PR #32, opened 2014-09-06
refactor "_sanitize" for Python < 2.7

patgmiller:
I have a python2.6 project which I'd like to use this on, so it does not handle dictionary comprehensions. I believe this patch would make it compatible for Python 2.6+.

### DIFF
--- a/gitlab.py
+++ b/gitlab.py
@@ -468,7 +468,7 @@ def _sanitize(value):
 
 
 def _sanitize_dict(src):
-    return {k: _sanitize(v) for k, v in src.items()}
+    return dict((k, _sanitize(v)) for k, v in src.items())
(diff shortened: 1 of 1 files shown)


ilesser:
Hi I found some issues with python2.66 (I am stuck with Red Hat 6) .

File gitlab/v3/objects.py line 2094 and 2095 (https://github.com/python-gitlab/python-gitlab/blob/master/gitlab/v3/objects.py#L2094)

And file gitlab/base.py line 524 https://github.com/python-gitlab/python-gitlab/blob/master/gitlab/base.py#L524

I am not familiar with the pull request full or even github itself so I am not very sure how to proceed.

Should I create a pull request ?

PR #50, opened 2014-12-05
Add Debian packaging.

aegiap:
Signed-off-by: Nicolas Chipaux aegiap@gandi.net


fpytloun:
This looks fine.
It could also build documentation to have html docs (in separate package, eg. python-gitlab-doc) and man pages.

```override_dh_auto_clean:
	dh_auto_clean
	rm -rf doc/build

override_dh_auto_build:
	dh_auto_build
	sphinx-build -b man doc/source doc/build/man
	sphinx-build -b html doc/source doc/build/html
```

avdyk:
There is a package for another gitlab wrapper for python in Debian: https://packages.debian.org/stretch/python-gitlab But it's not maintained "upstream" so maybe someone can contact the Debian maintainer and ask to package this project instead...

gpocentek:
Hi @aegiap 

I'm closing this PR because I don't want to handle the packaging in the code. I'm still convince that pushing the package to debian is a better solution, although the conflict with the existing package must be handled, and might be complicated.

PR #55, opened 2015-02-13
Add support for searching groups

dserodio:
It can probably be improved, but works for my needs

### DIFF
--- a/gitlab.py
+++ b/gitlab.py
@@ -513,7 +513,7 @@ class Gitlab(object):
         return l
 
     def search_groups(self, query, **kwargs):
-        """Searches groups by name.
+        """Get all groups that match your query string in their name or path.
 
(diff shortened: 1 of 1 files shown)


Asher256:
I agree with you Gauvain. It would be nice if these kind of 'search' functions are implemented directly in the classes (like Group or project).

What do you think about this:

A special class to do queries (we will have the possibility to choose between id=None, to list everything, or GitlabQuery(), to add some rules):
    projects = git.Project(GitlabQuery(name="projectname"))
    projects = git.Project(GitlabQuery(is_admin=True))

    groups = git.Group(GitlabQuery(name="groupname"))

gpocentek:
The code has changed recently and it's been easier to implement this method on the group manager. I pushed the change in commit 5513d0f. Thank you!

Issue #63, opened 2015-07-28
Iteration of results


jantman:
I'm using GitLab CE 7.12.2, and trying to list users via the API. I have 162 users. After connecting to GitLab using python-gitlab 0.9.2 from PyPi (my connection is ``self.conn``), ``self.conn.User()`` returns a list with 20 items, a fraction of my users.

The [GitLab API documentation](https://github.com/gitlabhq/gitlabhq/tree/master/doc/api#pagination) says that it paginates responses with a default of 20 results per response.

It appears that python-gitlab is ignoring the pagination. The only use of the "page" response attribute that I see in the source code is on [line 341](https://github.com/gpocentek/python-gitlab/blob/master/gitlab/__init__.py#L341) where it's deleted before passing to the constructor.

Could you please either implement pagination of responses so python-gitlab doesn't silently ignore results, or at least mention in the docs that this is happening?

Thanks.

gpocentek:
You can use the ``page`` and ``per_page`` arguments.

From the README:

`````python
for g in gl.Group(page=1, per_page=10):
    print (g)
`````

jantman:
@gpocentek Ok, I'd missed that, sorry. But doesn't that just add those as query parameters and build a URL based on them? The [API Docs for Pagination](https://github.com/gitlabhq/gitlabhq/tree/master/doc/api#pagination) specifically say,

> Link headers are send back with each response. These have rel prev/next/first/last and contain the relevant URL. Please use these instead of generating your own URLs.

dserodio:
The "automatic" pagination proposed by @jantman would be a great addition. There's no way to find out the max number of pages for a given resource

gpocentek:
Closing as this as been implemented in #64 . Thanks.

PR #64, opened 2015-07-28
python-gitlab Issue #63 - implement pagination for list()

jantman:
This implements iteration/pagination of list() results for #63

### DIFF
--- a/gitlab/__init__.py
+++ b/gitlab/__init__.py
@@ -26,9 +26,6 @@ import warnings
 import requests
 import six
 
-import logging
-logger = logging.getLogger(__name__)
-
 __title__ = 'python-gitlab'
 __version__ = '0.9.1'
 __author__ = 'Gauvain Pocentek'
@@ -352,7 +349,6 @@ class Gitlab(object):
             if 'next' in r.links and 'url' in r.links['next']:
                 args = kwargs.copy()
                 args['next_url'] = r.links['next']['url']
-                logger.debug("Iterating results 'next' link: %s", args['next_url'])
                 results.extend(self.list(obj_class, **args))
             return results
         else:
--- a/gitlab/__init__.py
+++ b/gitlab/__init__.py
@@ -26,6 +26,9 @@ import warnings
 import requests
 import six
 
+import logging
+logger = logging.getLogger(__name__)
+
 __title__ = 'python-gitlab'
 __version__ = '0.9.1'
 __author__ = 'Gauvain Pocentek'
@@ -189,6 +192,8 @@ class Gitlab(object):
         self._url = '%s/api/v3' % url
 
     def _construct_url(self, id_, obj, parameters):
+        if 'next_url' in parameters:
+            return parameters['next_url']
         args = _sanitize_dict(parameters)
         url = obj._url % args
         if id_ is not None:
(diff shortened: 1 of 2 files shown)


gitlab/__init__.py
```diff
@@ -26,6 +26,9 @@
 import requests
 import six
 
+import logging
+logger = logging.getLogger(__name__)
```
gpocentek: Could you remove the logging code for this change? I'm not against adding some logging but I'd rather do it in another patch.

gitlab/__init__.py
```diff
@@ -189,6 +192,8 @@ def set_url(self, url):
         self._url = '%s/api/v3' % url
 
     def _construct_url(self, id_, obj, parameters):
+        if 'next_url' in parameters:
```
gpocentek: Could you handle prev, first and last as well?

gitlab/__init__.py
```diff
@@ -342,8 +347,14 @@ def list(self, obj_class, **kwargs):
...
 
-            return [cls(self, item, **cls_kwargs) for item in r.json()
-                    if item is not None]
+            results = [cls(self, item, **cls_kwargs) for item in r.json()
+                       if item is not None]
+            if 'next' in r.links and 'url' in r.links['next']:
+                args = kwargs.copy()
+                args['next_url'] = r.links['next']['url']
+                logger.debug("Iterating results 'next' link: %s", args['next_url'])
+                results.extend(self.list(obj_class, **args))
```
gpocentek: This doesn't seem to give a choice to the user, this will always return the full list and do multiple API calls. If a user has hundreds of items this is a problem. It should be optional.

dserodio: IMHO, the perfect implementation would use Python generators instead of a list, and only fetch the next page when needed.

gpocentek: Sounds like a good plan. I'm a bit short on time to work on this these days but I'd be happy to merge this solution. Thanks!

gitlab/__init__.py
jantman: Done.

gitlab/__init__.py
jantman: @gpocentek I'm not sure I understand the purpose of that, given how this function is called. Even if I don't use the generator idea, from @dserodio, it's your project but I'd rather not add in additional unused code paths...

gitlab/__init__.py
jantman: I'm not sure I see this as "a problem" - the whole reason why I opened this PR is because I have about 200 users, and will soon have about 300 repositories. I've seen some latency when fetching all that, but not much.

I like @dserodio's idea of using a generator for this. I'm going to have a try at implementing it, though I haven't done much with generators before, and perhaps worse, I've never really worked with ``unittest`` or ``httmock``.

jantman:
I've removed the logging code as requested, and added unit tests for the _current_ implementation of just the ``next`` link. However, I think I've hit a bit of a roadblock...

I have nothing against giving a choice to the user, but I'd much rather prefer the iteration to be the default, with disabling it an option. I guess this is personal preference, but I find it very annoying when an API client presents a method like ``Users()`` that only retrieves partial results without making that extremely clear.

I looked into turning this into a generator, but as far as I can tell, I'd need to drastically change how arguments and return values are passed around to get away from the current recursive pattern.

gitlab/__init__.py
gpocentek: My comment was just plain stupid. Sorry.

gpocentek:
Thanks Jason.

I've merged your patch as is but I'll probably restore the default behavior to avoid surprises on updates (people will not expect to suddenly retrieve all the items).

jantman:
Ok, thanks! I'll keep an eye out for it, and update my code to use the pagination once it's released.

PR #66, opened 2015-08-12
Fix error when fetching single MergeRequests

stefanklug:
Sorry, I hit enter too early.
Fetching a single Merge Request didn't work because of the wrong url. I was puzzled by this gitlab inconsistency, but it is like that. See 
https://github.com/gitlabhq/gitlabhq/blob/master/doc/api/merge_requests.md

I am working with gitlab 7.13.4

Regards Stefan


gpocentek:
Thanks!

PR #72, opened 2015-09-23
Can bypassing confirm when creating new user now

pa4373:
GitLab add 'confirm' field support to its API since 7.9.0. However, it takes `'false'` string only when the content type of request is set to 'application/json'. So I just add `confirm` to `optionalCreateAttrs` list and override `_data_for_gitlab` to make sure `confirm` is converted to boolean string.

It has been tested on GitLab HQ 7.14.3 and API is v3.

Additional info: <https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/1281>


gpocentek:
Looks good, thanks!

PR #74, opened 2015-10-20
Add fork project support

fgouteroux:
- Forks a project into the user namespace of the authenticated user.
- Allows modification of the forked relationship between existing projects (for admins)

### DIFF
--- a/gitlab/__init__.py
+++ b/gitlab/__init__.py
@@ -522,12 +522,12 @@ class Gitlab(object):
         """
         return Team._get_list_or_object(self, id, **kwargs)
 
-    def ForkProject(self, id=None, **kwargs):
+    def ProjectFork(self, id=None, **kwargs):
         """Fork a project for a user.
 
         id must be a dict.
         """
-        return ForkProject._get_list_or_object(self, id, **kwargs)
+        return ProjectFork._get_list_or_object(self, id, **kwargs)
 
 
 def _get_display_encoding():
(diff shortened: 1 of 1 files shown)


gitlab/__init__.py
```diff
@@ -614,6 +614,13 @@ def Team(self, id=None, **kwargs):
         """
         return Team._get_list_or_object(self, id, **kwargs)
 
+    def ForkProject(self, id=None, **kwargs):
```
gpocentek: The class name should be ProjectFork.

gitlab/__init__.py
```diff
@@ -1376,3 +1404,12 @@ def Project(self, id=None, **kwargs):
...
                                                **kwargs)
+
+class ForkProject(GitlabObject):
+    _url = '/projects/fork/%(project_id)s'
+    canUpdate = False
+    canDelete = False
+    canList = False
+    canGet = False
+    requiredUrlAttrs = ['project_id']
+    requiredCreateAttrs = ['project_id']
```
gpocentek: requiredCreateAttrs should not be defined. If it exists the CLI will not work (the project-id argument would be defined twice).

gitlab/__init__.py
```diff
@@ -1376,3 +1404,12 @@ def Project(self, id=None, **kwargs):
         return TeamProject._get_list_or_object(self.gitlab, id,
                                                team_id=self.id,
                                                **kwargs)
+
+class ForkProject(GitlabObject):
```
gpocentek: The function name should be ProjectFork.

gpocentek:
Looks good but could you address the comments? Thanks.

fgouteroux:
Thanks for your comments, I have updated the code. I also have re-synchronized the branch due to the updates between the two commits.

gpocentek:
I've changed a lot of things last week and it looks like you merge didn't take that into account. could you have a look? Sorry for making your work harder...

fgouteroux:
It will be more simple to create a new PR with the last updates.

PR #78, opened 2015-12-23
Use name as sort key to fix Python 3 TypeError

cdbennett:
Sort types explicitly by name to fix unorderable types TypeError in
Python 3.

The call to sort() on cli.py line 259 produced the error:

    TypeError: unorderable types: type() < type()


gpocentek:
Thanks!

PR #79, opened 2015-12-23
Check for action arg before use (fix Python 3 error)

cdbennett:
Check that the parsed `action` attribute exists before using it.
Previously, whenever the command-line tool was invoked on Python 3.4
without a bare-word argument, which is expected to be the action, the
program aborts with an AttributeError in cli.py on the line assigning
`arg.action` to `action`:

    AttributeError: 'Namespace' object has no attribute 'action'

### DIFF
--- a/gitlab/cli.py
+++ b/gitlab/cli.py
@@ -271,7 +271,14 @@ def main():
     config_files = arg.config_file
     gitlab_id = arg.gitlab
     verbose = arg.verbose
-    action = arg.action
+
+    if hasattr(arg, "action"):
+        action = arg.action
+    else:
+        sys.stderr.write("error: an action is required\n\n")
+        parser.print_help()
+        sys.exit(1)
+
     what = arg.what
 
     # Remove CLI behavior-related args


gpocentek:
Could you give an example of a command that would generate this error? Thanks.

cdbennett:
This issue occurs whenever there is no "bareword" argument given to the program.  For instance, with no arguments at all `gitlab` will hit it:

    $ gitlab
      File "C:\Python34\Scripts\gitlab-script.py", line 9, in <module>
        load_entry_point('python-gitlab==0.9.1', 'console_scripts', 'gitlab')()
      File "C:\Python34\lib\site-packages\python_gitlab-0.9.1-py3.4.egg\gitlab\cli.py", line 275, in main
    AttributeError: 'Namespace' object has no attribute 'action'

It will also occur if you provide option arguments, but still don't provide a "bareword" argument, one that is not an option flag, as in `gitlab -v`:

    $ gitlab -v
      File "C:\Python34\Scripts\gitlab-script.py", line 9, in <module>
        load_entry_point('python-gitlab==0.9.1', 'console_scripts', 'gitlab')()
      File "C:\Python34\lib\site-packages\python_gitlab-0.9.1-py3.4.egg\gitlab\cli.py", line 275, in main
    AttributeError: 'Namespace' object has no attribute 'action'

This test was performed with the #78 (`fix_python3_sort_types`) fix applied -- otherwise Python 3 will fail earlier in the program before this point is reached.

gpocentek:
I just push commit 7c38ef6f which should fix for this bug. It uses the argparse features instead of providing a custom test. Could you test it to make sure it behaves as you would expect? Thanks.

cdbennett:
Verified fixed with commit 7c38ef6.  Thanks.

PR #80, opened 2016-01-06
add fork project support (#Refs 74)

fgouteroux:
- Forks a project into the user namespace of the authenticated user.
- Allows modification of the forked relationship between existing projects (for admins)


gpocentek:
Thanks for your patch.

PR #82, opened 2016-01-07
Support setting commit status

cdbennett:
Support commit status updates.  Commit status can be set by a POST to
the appropriate commit URL.  The status can be updated by a subsequent
POST to the same URL with the same `name` and `ref`, but different
values for `state`, `description`, etc.

### DIFF
--- a/gitlab/objects.py
+++ b/gitlab/objects.py
@@ -681,6 +681,22 @@ class ProjectCommitManager(BaseManager):
     obj_cls = ProjectCommit
 
 
+class ProjectCommitStatus(GitlabObject):
+    _url = '/projects/%(project_id)s/statuses/%(commit_id)s'
+    canUpdate = False
+    canDelete = False
+    requiredUrlAttrs = ['project_id', 'commit_id']
+    requiredCreateAttrs = ['state']
+    optionalCreateAttrs = ['description', 'name', 'ref', 'target_url']
+    requiredGetAttrs = []
+    requiredUpdateAttrs = []
+    requiredDeleteAttrs = []
+
+
+class ProjectCommitStatusManager(BaseManager):
+    obj_cls = ProjectCommitStatus
+
+
 class ProjectKey(GitlabObject):
     _url = '/projects/%(project_id)s/keys'
     canUpdate = False
@@ -961,6 +977,7 @@ class Project(GitlabObject):
     managers = [
         ('branches', ProjectBranchManager, [('project_id', 'id')]),
         ('commits', ProjectCommitManager, [('project_id', 'id')]),
+        ('commitstatuses', ProjectCommitStatusManager, [('project_id', 'id')]),
         ('events', ProjectEventManager, [('project_id', 'id')]),
         ('files', ProjectFileManager, [('project_id', 'id')]),
         ('forks', ProjectForkManager, [('project_id', 'id')]),


gitlab/objects.py
```diff
@@ -787,6 +802,7 @@ class Project(GitlabObject):
     managers = [
         ('branches', ProjectBranchManager, [('project_id', 'id')]),
         ('commits', ProjectCommitManager, [('project_id', 'id')]),
+        ('statuses', ProjectCommitStatusManager, [('project_id', 'id')]),
```
gpocentek: For consistency I'd call the attribute 'commitstatuses' instead of 'statuses' (to match the class name).

gitlab/objects.py
```diff
@@ -520,6 +520,21 @@ class ProjectCommitManager(BaseManager):
...
+    _url = '/projects/%(project_id)s/statuses/%(commit_id)s'
+    canUpdate = False
+    canDelete = False
+    requiredUrlAttrs = ['project_id', 'commit_id']
+    requiredCreateAttrs = ['state']
+    optionalCreateAttrs = ['description', 'name', 'ref', 'target_url']
+    requiredGetAttrs = []
+    requiredUpdateAttrs = []
+    requiredDeleteAttrs = []
+
```
gpocentek: Add a blank line to make the pep8 tests pass (you can run `tox -epep8` to check)

gitlab/__init__.py
```diff
@@ -521,3 +521,21 @@ def Team(self, id=None, **kwargs):
         to write it on the server.
         """
         return Team._get_list_or_object(self, id, **kwargs)
+
+    def Status(self, id=None, **kwargs):
```
gpocentek: I plan to deprecate the similar functions in the Gitlab class (it's already done in the other classes), so I'm not sure this is useful to add this one. Could you also add a manager attribute in the constructor?

gpocentek:
Looks good, but could you address my comments? Thanks!

cdbennett:
Sure, thanks for the review and suggestions.  I update based on your recommendations.

gitlab/__init__.py
cdbennett: @gpocentek Do you recommend I simply delete this function?  I only created it because it seemed consistent with the other functions -- I was not sure if it was necessary.

gpocentek: Yes, you can remove this function.

cdbennett:
OK, I addressed your comments by doing the following and rebased the topic branch fresh on master:

1. Rename attribute to commitstatuses
2. Add blank line per PEP8
3. Remove deprecated function that was being added to Gitlab class
4. And also: Remove usage example from the deleted `README.md` file

The usage example seems helpful to give a quick summary of the expectations of the multiple GitLab attributes for this operation:

    gitlab project-commit-status create --project-id 2 --commit-id a43290c \
        --state success --name ci/jenkins --target-url http://server/build/123 \
        --description "Jenkins build succeeded"

And the docstring that was in the Gitlab class function was a useful quick reference:

        This is often used for automated software build or test status.

        Args:
            id: Commit reference.
            kwargs: Arbitrary keyword arguments for the status:
                Required:
                - state: pending, running, success, failed, canceled
                Optional:
                - description: short description of this status
                - name:        label for this status e.g., name=ci/jenkins
                - ref:         branch reference e.g., ref=master
                - target_url:  hyperlink target for this status

At some point it would be nice to add this kind of info to the docs for the project-commit-status creation operation.

gpocentek:
Thanks for the patch. I'll add the documentation bits

PR #84, opened 2016-01-23
Fix User.json()

rhansen:
A couple of bugs are preventing `User.json()` from working:
  * the `Gitlab` class has not been imported into `objects.py`
  * the `jsonEncoder` class does not skip over `BaseManager` attributes, causing `json.dumps()` to raise an exception

These commits fix the above issues.


rhansen:
Thank you!

gpocentek:
Thanks for your patch :)

PR #85, opened 2016-01-24
Test script cleanups

rhansen:
Here are a bunch of commits to fix and clean up some minor issues in the `tools/*.sh` scripts.

### DIFF
--- a/tools/build_test_env.sh
+++ b/tools/build_test_env.sh
@@ -55,10 +55,15 @@ CONFIG=/tmp/python-gitlab.cfg
 
 cleanup() {
     rm -f "${CONFIG}"
+    log "Killing gitlab-test docker container..."
     docker kill gitlab-test >/dev/null 2>&1
+    log "Removing gitlab-test docker container..."
     docker rm gitlab-test >/dev/null 2>&1
+    log "Deactivating Python virtualenv..."
     command -v deactivate >/dev/null 2>&1 && deactivate || true
+    log "Deleting python virtualenv..."
     rm -rf "$VENV"
+    log "Done."
 }
 [ -z "${BUILD_TEST_ENV_AUTO_CLEANUP+set}" ] || {
     trap cleanup EXIT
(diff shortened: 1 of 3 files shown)


gpocentek:
Thanks @rhansen !

This PR looks good but it changes the behavior of `build_test_env.sh`. I use this script to set up a dev environment. With your changes the environment is destroyed right after it's been created if you execute the script. Could you avoid calling the `cleanup` function if the script is executed directly?

gpocentek:
I'm also getting an error on the file creation test.

rhansen:
> This PR looks good but it changes the behavior of `build_test_env.sh`. I use this script to set up a dev environment. With your changes the environment is destroyed right after it's been created if you execute the script.

Ah, apologies—I didn't realize that script was used that way.  I thought it was just a shared library of sorts.

> Could you avoid calling the `cleanup` function if the script is executed directly?

Yes, I'll fix it soon.

> I'm also getting an error on the file creation test.

Can you paste the error message?

I did see random nginx "502 bad gateway" failures when testing this branch.  Those don't always happen, and they currently happen on master, so I don't believe any of the commits on this branch made anything worse.  I may be wrong.

gpocentek:
On 2016-01-24 17:28, Richard Hansen wrote:
>> This PR looks good but it changes the behavior of build_test_env.sh.
>> I use this script to set up a dev environment. With your changes the
>> environment is destroyed right after it's been created if you
>> execute the script.
> 
> Ah, apologies—I didn't realize that script was used that way. I
> thought it was just a shared library of sorts.
> 
>> Could you avoid calling the cleanup function if the script is
>> executed directly?
> 
> Yes, I'll fix it soon.

Thanks.

>> I'm also getting an error on the file creation test.
> 
> Can you paste the error message?
> 
> I did see random nginx "502 bad gateway" failures when testing this
> branch. Those don't always happen, and they currently happen on
> master, so I don't believe any of the commits on this branch made
> anything worse. I may be wrong.

I couldn't reproduce exactly this error, but I've also seen those 502 
failures after an upgrade to 8.4.0. I agree this doesn't seem related to 
your patch.

Thank you!

rhansen:
I made a few changes and rebased:

  * when run on its own, `build_test_env.sh` no longer runs `cleanup()` on exit (this preserves the behavior it has on `master`)
  * use the gentler `docker stop` instead of the aggressive `docker kill`
  * wait for the docker container to stop before removing it
  * don't suppress docker's stderr
  * add more log messages
  * move the `sleep 20` to `build_test_env.sh`

gpocentek:
Thank you!

rhansen:
Thank you!

PR #90, opened 2016-02-05
fix GitlabObject creation in _custom_list

ms-boom:
call GitlabObject.__init__(self, gl, ...)
with self.gitlab(gitlab.Gitlab) instread self(BaseManager)


gpocentek:
Nice catch, thank you!

PR #94, opened 2016-02-12
Add support for searching for users and getting a user by username

rhansen:
Add support for `/users?search=foo` and `/users?username=foo`.

Also, in order to make it easier to write tests for these new features, add:
  * `GitlabObject.as_dict()`, which is the same as calling `json.loads()` on the output of `GitlabObject.json()`
  * `GitlabObject.__eq__()` and `GitlabObject.__ne__()`, which checks the dictionaries returned by `GitlabObject.as_dict()` for equivalence

### DIFF
--- a/gitlab/objects.py
+++ b/gitlab/objects.py
@@ -565,6 +565,14 @@ class User(GitlabObject):
 class UserManager(BaseManager):
     obj_cls = User
 
+    def search(self, query, **kwargs):
+        """Search users.
+
+        Returns a list of matching users.
+        """
+        url = self.obj_cls._url + '?search=' + query
+        return self._custom_list(url, self.obj_cls, **kwargs)
+
 
 class CurrentUserKey(GitlabObject):
     _url = '/user/keys'
--- a/gitlab/objects.py
+++ b/gitlab/objects.py
@@ -491,6 +491,14 @@ class GitlabObject(object):
         return {k: v for k, v in six.iteritems(self.__dict__)
                 if (not isinstance(v, BaseManager) and not k[0] == '_')}
 
+    def __eq__(self, other):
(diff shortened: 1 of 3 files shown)


gitlab/objects.py
```diff
@@ -541,10 +552,41 @@ def unblock(self, **kwargs):
         raise_error_from_response(r, GitlabUnblockError)
         self.state = 'active'
 
+    def __eq__(self, other):
+        if type(other) is type(self):
+            selfdict = self.as_dict()
+            otherdict = self.as_dict()
```
gpocentek: Do you mean `otherdict = other.as_dict()`?

gitlab/objects.py
```diff
@@ -541,10 +552,41 @@ def unblock(self, **kwargs):
...
+        Returns a User object or None if the named user does not
+        exist.
+        """
+        url = self.obj_cls._url + '?username=' + username
+        results = self._custom_list(url, self.obj_cls, **kwargs)
+        assert len(results) in (0, 1)
+        try:
+            return results[0]
+        except IndexError:
+            return None
```
gpocentek: Other methods raise a `GilabGetError` exception when the resource cannot be found. Could you do the same here, instead of returning `None`?

gpocentek:
Thanks for this patch! Could you address the comments, and I'll merge.

gitlab/objects.py
rhansen: Oops!

rhansen:
Thank you for reviewing!  I pushed some rebased commits that I believe address your comments.

gpocentek:
Thank you :)

rhansen:
Thanks for merging!

PR #98, opened 2016-03-02
Fix the 'invalid syntax' on Python 3.2, because of u'password'

Asher256:
More informations regarding this issue:

Operating system: Debian Wheezy, with Python 3.2 and the last
version of python-gitlab.

The gitlab module raised this exception, because of the 'u' (Unicode):

  File "push_settings.py", line 14, in <module>
    from helper import ROOT_EMAIL, ADMINS, git, old_git
  File "/opt/scripts/gitlab/helpers/helper.py", line 25, in <module>
    from gitlab import Gitlab
  File "/opt/scripts/gitlab/helpers/gitlab/__init__.py", line 32, in <module>
    from gitlab.objects import *  # noqa
  File "/opt/scripts/gitlab/helpers/gitlab/objects.py", line 546
    selfdict.pop(u'password', None)
			   ^
SyntaxError: invalid syntax
It is a recent change:
01802c0 (Richard Hansen 2016-02-11 22:43:25 -0500 546) selfdict.pop(u'password', None)
01802c0 (Richard Hansen 2016-02-11 22:43:25 -0500 547) otherdict.pop(u'password', None)

To solve the issue, 'u' was removed.

### DIFF
--- a/gitlab/objects.py
+++ b/gitlab/objects.py
@@ -543,8 +543,8 @@ class User(GitlabObject):
         if type(other) is type(self):
             selfdict = self.as_dict()
             otherdict = other.as_dict()
-            selfdict.pop(u'password', None)
-            otherdict.pop(u'password', None)
+            selfdict.pop('password', None)
+            otherdict.pop('password', None)
             return selfdict == otherdict
         return False
 


Asher256:
To be backware compatible, have you ever thought about unicode_literals?

1. Adding "from __future__ import unicode_literals" at the top of your Python modules – it’s best to put it in each and every module, otherwise you’ll keep checking the top of your files to see which mode is in effect;
2. Removing the u prefix before unicode strings;
3. Adding a b prefix before bytestrings.

gpocentek:
@Asher256 : Thanks! I'll take the time to learn a bit more about unicode_literals. Thank you for dealing with this problem.

PR #101, opened 2016-03-04
Wrong create attributes check in update.

s-m-b:
There are new requiredUpdateAttrs in GitlabObject, I guess you want to use it in Gitlab.update instead of requiredCreateAttrs.
Right now with requiredCreateAttrs for user ['email', 'username', 'name', 'password'], one cannot update user without updating password.

### DIFF
--- a/gitlab/__init__.py
+++ b/gitlab/__init__.py
@@ -568,7 +568,7 @@ class Gitlab(object):
         params.update(kwargs)
         missing = []
         for k in itertools.chain(obj.requiredUrlAttrs,
-                                 obj.requiredCreateAttrs):
+                                 obj.requiredUpdateAttrs):
             if k not in params:
                 missing.append(k)
         if missing:


gitlab/__init__.py
```diff
@@ -568,7 +568,7 @@ def update(self, obj, **kwargs):
         params.update(kwargs)
         missing = []
         for k in itertools.chain(obj.requiredUrlAttrs,
-                                 obj.requiredCreateAttrs):
+                                 obj.requiredUpdateAttrs):
```
gpocentek: You need to deal with GitlabObjects that don't have requiredUpdateAttrs defined. For these classes requiredCreateAttrs needs to be used.

This might be why the unit tests are failing.

s-m-b: Thank you for explanation. As I see it now that change also requires to specify correct requiredUpdateAttrs for all GitlabObjects. Please correct me if that is not in sync with changes since 9.0.2 to address fixme on User object.

gpocentek: On most objects update and create attributes are the same, so I'd prefer to keep the possibility to use requiredCreateAttrs if requiredUpdateAttrs is not defined.

Your change is at the right place, but you need to add a test to make sure you use the proper attribute.

I hope this explanation is clear enough.

gpocentek:
See commit f8528cce8d79b13e90e1f87b56c8cdbe30b2067e for the update. Thanks.

PR #108, opened 2016-04-02
Adding a Session instance for all HTTP requests

guyzmo:
The session instance will make it easier for setting up once headers, including
the authentication payload, and it's also making it easier to override HTTP queries
handling, when using [betamax](https://github.com/sigmavirus24/betamax) for recording
and replaying the test suites.


gpocentek:
Thanks!

PR #110, opened 2016-04-21
Remove 'next_url' from kwargs before passing it to the cls constructor.

chrwen-omicron:
I had troubles iterating through all projects and changing some settings. The save method failed and I figured out that the 'next_url' attribute is still set and used in the _construct_url method.

This is an attempt to fix this, I might have missed something ...


gpocentek:
Looks good, thank you!

chrwen-omicron:
You're welcome. Keep up the good work!

PR #111, opened 2016-04-27
List projects in a group

missionrulz:
Hello,

I needed the ability to list projects under a group. I believe this is where the code should belong, please correct/advise me if there are any changes that need to be made.

This will allow me to list some/all of the projects that belong to a group.

`p = gl.groups.list_projects(37, all=True)`

Thanks.

### DIFF
--- a/gitlab/__init__.py
+++ b/gitlab/__init__.py
@@ -65,6 +65,7 @@ class Gitlab(object):
     Attributes:
         user_keys (UserKeyManager): Manager for GitLab users' SSH keys.
         users (UserManager): Manager for GitLab users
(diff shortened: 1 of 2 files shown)


gpocentek:
Thank you for your patch.

To keep the same logic used with other object, you should create a GroupProject class that only allows listing operation, and the related manager.

missionrulz:
@gpocentek, I have added in the changes you recommended. Please let me know if there is anything else.

gpocentek:
Thank you!

This was not exactly what I had in mind so I made an extra change more for consistency with the rest of the API.

PR #118, opened 2016-05-20
Add support for subscribe and unsubscribe in issues

IvicaArsov:
Add methods subscribe and unsubscribe in ProjectIssue to let the user control subscription in issues.


gpocentek:
Thanks!

PR #120, opened 2016-05-30
Added support for HTTP basic authentication

PeterMosmans:
Support for HTTP basic authentication

The username and password can be specified in the configuration file
```
http_username
http_password
```


gpocentek:
The patch looks good, thank you (will test ASAP).

I'm a bit curious about how you use this though. Do you use both HTTP auth and token auth? Is there any reason why?

PeterMosmans:
Hi @gpocentek , yes, I use both: Certain gitlab servers that I use are protected behind a HTTP basic authentication interface. This protects them from potential issues in the gitlab API (think defense-in-depth). After that is the 'usual' gitlab authentication (either tokens or username/password).

gpocentek:
Thanks!

PR #121, opened 2016-05-31
Added HTTPauth support for even more methods :)

PeterMosmans:
Hi @gpocentek , thanks for merging the previous method (and adding the documentation, I didn't thought about that). This pull request adds HTTPauth support for the remaining methods, which should cover pretty much all.

Thanks in advance,

Peter


gpocentek:
Looks good, thank you.

PR #124, opened 2016-06-11
Fix: --title is required argument, when in reality optional

Condla:
I add one line that includes "title" to be an optional command line argument.


gpocentek:
Thanks for the patch, it looks good. Could you just shorten your git message (``git commit --amend`` and ``git push --force`` are OK with me)?

Condla:
should be fine now...

gpocentek:
Thank you!

PR #128, opened 2016-07-11
add `note_events` to project hooks attributes

### LINKED ISSUE #127: Add 'note_events' or 'comments' to ProjectHook
Would it be possible to add attributes for note_events (aka comments) and also Wiki Page events in webhooks?

Scanning through the code I think it should be added here: https://github.com/gpocentek/python-gitlab/blob/master/gitlab/objects.py#L1025

I would like to know whether that's all or if there's more to it. I would gladly try the MR myself in a couple of days.

rafaeleyng:
closes #127


gpocentek:
Looks good, thank you!

PR #131, opened 2016-07-25
Added a new project attribute to enable the container registry.

chrwen-omicron:
Do you handle incompatible changes between GitLab versions somehow?


gpocentek:
I'd like to do something about incompatible changes but I don't know of a way to automate things, and I don't have the time to keep track of the changes manually, at all. Seeing how https://github.com/pyapi-gitlab/pyapi-gitlab struggles I don't think it would be wise to follow this track.

gpocentek:
Thanks for the change!

chrwen-omicron:
You're probably right, maybe it's enough to mention the supported GitLab version on each python-gitlab release.

You're welcome, I really appreciate your work here.

PR #135, opened 2016-08-03
Implement slack integration service.

marcoblos:
Implement method to _put_ a new integration with Slack.

`project.slack_service('https://hooks.slack.com/services/...', 'optional-username', '#optional-channel')`

If not inform username and channel assumes the configured the slack service.

### DIFF
--- a/gitlab/exceptions.py
+++ b/gitlab/exceptions.py
@@ -110,6 +110,7 @@ class GitlabMRClosedError(GitlabOperationError):
 class GitlabMROnBuildSuccessError(GitlabOperationError):
     pass
 
+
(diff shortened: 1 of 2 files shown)


gpocentek:
Thanks for this patch, but I'm afraid it breaks the CRUD logic that is used with other objects. I'd prefer to see a new ProjectSlackService class (or similar, that's just an idea) to manage the other methods (create/update/get/delete).

gpocentek:
Support for services has been added. Thanks.

marcoblos:
@gpocentek Thank's for add this feature and sorry for don't response you, i am very busy in last days.
I see your code and it is very nice! Awesome!

PR #138, opened 2016-08-29
mention per_page in users.py

timfeirg:
I can't seem to find any reference on the `all` and `per_page` argument, thus it took me a long time to finally know how to list all gitlab users using `gitlab.users.list`

however I'm not sure this is the right way to solve this, perhaps add docs to `gitlab.users.list`? or change the default value for `per_page`? tell me and I'll be happy to contribute.

### DIFF
--- a/docs/gl_objects/users.py
+++ b/docs/gl_objects/users.py
@@ -1,5 +1,5 @@
 # list
-users = gl.users.list()
+users = gl.users.list(all=True)
 # end list
 
 # search


gpocentek:
Hi,

Thank you for taking the time to tackle this. There is a section about pagination in the Getting Started section of the documentation: http://python-gitlab.readthedocs.io/en/stable/api-usage.html#pagination. I've improved it a little in 29e2efe.

The problem I see with using per_page in the users example is that you would probably have to use it everywhere. It's a real pain to maintain.

ww110052181:
@timfeirg  hi,timfeirg   
  users_list = gl.users.list(all=True) 
        print(users_list)
the result is 
[<User id:2>, <User id:1>]
but I cannot find the username ,or should I get  the id:2's username
thank you

ww110052181:
[<User id:2>, <User id:1>]

PR #142, opened 2016-09-07
Fix JIRA service - add api_url to mandatory attributes

galet:
The `api_url` attribute is mandatory at least since GitLab 8.11. Otherwise server returns `gitlab.exceptions.GitlabUpdateError: 400: 400 (Bad request) "api_url" not given`.


gpocentek:
Thank you for the patch. I'd rather define this parameter as optional, to keep compatibility with older versions of Gitlab. I know it's not optimal but I don't have a solution to handle multiple APIs in multiple Gitlab versions at the moment.

galet:
@gpocentek sounds good to me. Closing this pull request. Thanks!

gpocentek:
@galet you can update this pull request if you still want to submit the change.

galet:
Created a new clean pull request #146.

PR #146, opened 2016-09-13
JIRA service - add api_url to optional attributes

galet:
The api_url attribute is mandatory at least since GitLab 8.11. Otherwise
server returns gitlab.exceptions.GitlabUpdateError: 400: 400 (Bad request) "api_url" not given.

Keeps the parameter as optional to maintain backward compatibility with older GitLab
versions.


gpocentek:
Looks good, thanks!

PR #147, opened 2016-09-13
Missing coma concatenates array values

derek-austin:
'enabled_git_access_protocolgravatar_enabled' were two distinct values in ApplicationSettings.optionalUpdateAttrs.


gpocentek:
Nice catch, thanks.

PR #150, opened 2016-09-15
Add branch protection notes

vilhelmen:
You can pass developers_can_push and developers_can_merge to the protect function. Handy!


gpocentek:
Nice addition, thank you!

PR #151, opened 2016-09-16
Add initial support for manipulation of LDAP links

costela:
This PR add support for group.ldaplinks.create() and group.ldaplinks.delete().

It's unclear to me what the best way to implement .list() would be. Since the API returns the LDAP information embedded in the group, I see two options (maybe I'm missing something obvious):
- hide the ldap_group_links attribute behind a manager and add a list() method that fetches the group again, ignoring all info but the LDAP links
- keep the .ldap_group_links and the .ldaplinks manager separate, but somehow update the .ldap_group_links list when we perform changes (currently the list remains out-dated whenever we perform changes, which obviously sucks)

Not sure how any of these could be implemented... :)

### DIFF
--- a/docs/gl_objects/groups.py
+++ b/docs/gl_objects/groups.py
@@ -64,3 +64,24 @@ group.members.delete(member_id)
 # or
 member.delete()
 # end member delete
+
+# ldap sync list
+group.ldap_group_links
+# end ldap sync list
+
+# ldap sync create
+gl.group_ldap_links.create({'cn': 'some_ldap_group_name',
+                             'provider': 'ldapmain', 'group_access': 50},
+                            group_id=1)
+# or
+group.ldaplinks.create({'cn': 'some_ldap_group_name',
+                             'provider': 'ldapmain', 'group_access': 50})
+# end ldap sync create
+
+# ldap sync delete
+gl.group_ldap_links.delete('some_ldap_group_name', group_id=1)
+# or
+group.ldaplinks.delete('some_ldap_group_name')
+# or
+ldaplink.delete()
+# end ldap sync delete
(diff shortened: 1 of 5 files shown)


gpocentek:
Thank you for your work on this patch.

It looks OK apart from a few syntax changes required. But I'm not comfortable merging it because this is a gitlab EE feature that I can't test. I'm not sure how I could automate testing and make sure the code works.

costela:
Oh, I totally forgot this annoying detail :/
I believe the enterprise edition is actually installable without a license (it's user based), so I could ask the Gitlab guys if they could issue a sort of "for automated testing purposes" blanket license. Than you could test against EE. Would you be OK with this?

gpocentek:
It's worth a try. I'd be happy to add EE features to the lib. Thanks!

costela:
Just added a commit that changes the testing to use a fork of your ```test-python-gitlab``` docker image using gitlab EE, but I keep getting timeouts. Strangely enough my local docker image seems to work...
Also, the timeouts don't seem to signal a build-failure since the whole functional testing is done inside ```after_success```. Was this intentional?

jouve:
I have this working (create/delete) on 1.4.0:

``` python

from gitlab.base import RESTObject, RESTManager
from gitlab.mixins import CreateMixin, DeleteMixin
import gitlab.v4.objects

class GroupLdapGroupLink(RESTObject):
    pass


class GroupLdapGroupLinkManager(CreateMixin, DeleteMixin, RESTManager):
    _path = '/groups/%(group_id)s/ldap_group_links'
    _obj_cls = GroupLdapGroupLink
    _from_parent_attrs = {'group_id': 'id'}
    _create_attrs = (('cn', 'group_access', 'provider'), tuple())

gitlab.v4.objects.GroupLdapGroupLinkManager = GroupLdapGroupLinkManager

gitlab.v4.objects.Group._managers = tuple(
    gitlab.v4.objects.Group._managers + (('ldapgrouplinks', 'GroupLdapGroupLinkManager'), )
)

```

gpocentek:
The LDAP group link API endpoint being quite different from the other endpoints I implemented support for this feature as Group methods instead of new classes. See commit d6a61af.

PR #154, opened 2016-09-21
Create a project in a group

derek-austin:
Just a sketch, feel free to toss it and do it in the right way.


gpocentek:
Let's merge it, at least we got the info. Thank you!

PR #155, opened 2016-09-22
add only_allow_merge_if_build_succeeds option to project objects

rafaeleyng:


gpocentek:
The patch looks good, thank you. Please fix the pep8 failure and I'll merge.

rafaeleyng:
It didn't work =(

I tried:

```
    project.only_allow_merge_if_build_succeeds = True
    project.save()
```

and 

```
    project.only_allow_merge_if_build_succeeds = 1
    project.save()
```

but none worked.

Am I doing something wrong?

gpocentek:
I tested here and everything seems to work, using a boolean value. Are you sure that your gitlab version support this attribute? It's recent.

rafaeleyng:
I can set this value using the GUI, but I'm not sure about the API support.

Since I'm no Python developer, I may have done something wrong. Here is what I did:

```
$ sudo pip uninstall python-gitlab
$ git clone https://github.com/gpocentek/python-gitlab.git
$ cd python-gitlab
$ python setup.py install
```

Does this suffices to use the newest version containing the change introduced in this MR?

gpocentek:
Yes you should be able to use your change. Do you know the version of the gitlab server?

gpocentek:
From what I see in the git log you need gitlab 8.12 to get this feature in the API.

rafaeleyng:
We are using GitLab Community Edition 8.9.0.

I'll try to update and give you some input again.

PR #159, opened 2016-10-02
Add ProjectBuild.erase()

### LINKED ISSUE #158: Erasing builds is not supported
The [documentation](https://python-gitlab.readthedocs.io/en/stable/gl_objects/builds.html#id3) indicates that you can erase a build:

```python
build.delete()
```

But when I try to do this, I get a `NotImplementedError`:

```
  File "/usr/lib/python2.7/site-packages/gitlab/objects.py", line 326, in delete
    raise NotImplementedError
NotImplementedError
```

This happens because [`canDelete = False`](https://github.com/gpocentek/python-gitlab/blob/90ad2dec7a0b158b2e77ae0cb403b01f2e1498d4/gitlab/objects.py#L942-L948) in the `ProjectBuild` class.

For whatever reason, the GitLab team decided to make erasing a build a  [`POST /projects/:id/builds/:build_id/erase`](https://docs.gitlab.com/ce/api/builds.html#erase-a-build), instead of a `DELETE`, like many of their other APIs.

Regardless, it looks like we will need to special-case this operation to POST.  Which option would you prefer?

1. Override the `delete` method of `ProjectBuild`
2. Add an `erase` method to `ProjectBuild`

Seeing as GitLab might make the deletion more fine-grained in the future (e.g. I hope they eventually allow erasing only artifacts, keeping the trace), I would prefer option 2.

---
GitLab version:  8.12.3
python-gitlab version: 0.15.1

JonathonReinhart:
This adds `ProjectBuild.erase()`.

We can't use the existing delete() functionality, because GitLab uses
`POST /projects/:id/builds/:build_id/erase` to erase a build. Instead of
overriding delete(), we add a separate erase() method to keep the naming
consistent, and allow potentially more fine-grained operations in the
future.

- https://docs.gitlab.com/ce/api/builds.html#erase-a-build

This fixes #158

### DIFF
--- a/docs/gl_objects/builds.py
+++ b/docs/gl_objects/builds.py
@@ -103,10 +103,6 @@ build.cancel()
 build.retry()
 # end retry
 
-# delete
-gl.project_builds.delete(build_id, project_id=1)
-# or
-project.builds.delete(build_id)
-# or
-build.delete()
-# end delete
+# erase
+build.erase()
+# end erase
--- a/docs/gl_objects/builds.rst
+++ b/docs/gl_objects/builds.rst
@@ -150,8 +150,8 @@ Cancel/retry a build:
    :start-after: # retry
    :end-before: # end retry
 
-Erase a build:
+Erase a build (artifacts and trace):
 
 .. literalinclude:: builds.py
-   :start-after: # delete
-   :end-before: # end delete
+   :start-after: # erase
+   :end-before: # end erase
--- a/gitlab/exceptions.py
+++ b/gitlab/exceptions.py
@@ -95,6 +95,10 @@ class GitlabBuildRetryError(GitlabRetryError):
     pass
 
 
+class GitlabBuildEraseError(GitlabRetryError):
+    pass
+
+
 class GitlabPipelineRetryError(GitlabRetryError):
     pass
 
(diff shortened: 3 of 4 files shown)


JonathonReinhart:
There doesn't appear to be anything in place for testing APIs related to builds, so I haven't added any automated tests.  However, I have successfully tested this using a script I'm writing (will be released soon), and it works as expected.

gpocentek:
Looks good to me, thanks!

JonathonReinhart:
@gpocentek Thanks for merging! Any chance this will make it into an official release any time soon?  My tool JonathonReinhart/gitlab-artifact-cleanup requires it, and I'm looking to put it on PyPi soon, so it would be great to be able to set a proper minimum version requirement. Currently I'm [monkey-patching](https://github.com/JonathonReinhart/gitlab-artifact-cleanup/blob/a0365500e2f0a9b4038d55fbf5546d848780b3af/gitlab-artifact-cleanup#L121-L142) this functionality in if I find that it is missing.

Also FYI: you're missing a `0.15.1` tag.

gpocentek:
@JonathonReinhart 0.15.1 is tagged and I'm preparing the 0.16 release.

PR #165, opened 2016-10-17
Fix ProjectBuild.play raising error even on success

hakkeroid:
Running play on ProjectBuild raises a GitlabBuildPlayError although the request was successful. It looks like gitlab returns a 200-OK instead of 201-CREATED response and as such always raises an exception.


gpocentek:
Thank you!

PR #168, opened 2016-10-20
Pass kwargs to object factory

hakkeroid:
### The Issue
When you fetch tags from a project and try to access the builds through a tag commit object like in the following snippet:
```
client = gitlab.Gitlab(..)
project = client.projects.get('some-id')
tag = project.tags.list()[0]
tag.commit.builds()
```
… then python-gitlab raises an AttributeError in line 1128 in objects.py. This happens because it tries to construct the request url and cannot find `self.project_id`.
```
url = '/projects/%s/repository/commits/%s/builds' % (self.project_id, self.id)
```
### What Causes the Issue?
When you fetch tags python-gitlab constructs Tag-objects which call `_set_from_dict` internally. Then this method checks if it contains data for subobjects and calls subsequently `_get_object` . At this point the subobjects will be created from the gitlab response data but do not receive any "kwargs" information and as such are missing the project-id.

### Possible Solution
I simply passed down the kwargs to `_get_object`. It solves the specific issue although I am not exactly sure if this is the best way to go. Additionally I am not a huge fan of passing down parameters for too many levels as it feels conceptually wrong.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ MANIFEST
 .idea/
 docs/_build
 .testrepository/
+.tox
--- a/gitlab/objects.py
+++ b/gitlab/objects.py
@@ -280,13 +280,14 @@ class GitlabObject(object):
 
             raise GitlabGetError("Object not found")
 
-    def _get_object(self, k, v):
+    def _get_object(self, k, v, **kwargs):
         if self._constructorTypes and k in self._constructorTypes:
-            return globals()[self._constructorTypes[k]](self.gitlab, v)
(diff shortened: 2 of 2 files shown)


hakkeroid:
Besides I purposely did not "fix" all locations where `_set_from_dict` is used. Mainly because I am not sure if you want it to be solved that way. :-)

gpocentek:
Thank you for taking care of this!

The other ``_set_from_dict`` calls update existing objects so they are probably fine as they are.

I made the choice to have something completely dynamic so having several levels of ``**kwargs`` is fine with me. Even if it is conceptually wrong ;)

Issue #160, opened 2016-10-03
Finding closes_issues() for merge_request give 404


kvdveer:
My example code:

```python
x = project.mergerequests.get(iid=184)
# "GET /api/v3/projects/68/merge_requests?iid=184&project_id=68 HTTP/1.1" 200 998

print(x.closes_issues())
# "GET /api/v3/projects/68/merge_requests/None/closes_issues?project_id=68 HTTP/1.1" 404 27

print(x.as_dict() )
#{'gitlab': <gitlab.Gitlab at 0x7f383c31da90>,
# 'id': None,
# 'iid': 184,
# 'project_id': 68}
```

As you can see, the url is constructed incorrectly (None instead of the merge request iid). This is probably because the id of the MR is not filled in. 

Oddly enough, this works fine when the MR is retrieved using the list method:

```python
 project.mergerequests.list(iid=184)[0].as_dict()
# {'assignee': None,
#  'author': <gitlab.objects.User at 0x7f3836f757b8>,
#  'created_at': '2016-10-03T03:55:09.473+02:00',
#  'description': '<redacted>',
#  'downvotes': 0,
#  'force_remove_source_branch': False,
#  'gitlab': <gitlab.Gitlab at 0x7f383c31da90>,
#  'id': 186,
#  'iid': 184,
#  'labels': [],
#  'merge_commit_sha': None,
#  'merge_status': 'can_be_merged',
#  'merge_when_build_succeeds': False,
#  'milestone': None,
#  'project_id': 68,
#  'sha': '51ceba81898eef854d2929805771cdcb61e1e132',
#  'should_remove_source_branch': False,
#  'source_branch': 'show_version',
#  'source_project_id': 68,
#  'state': 'opened',
#  'subscribed': True,
#  'target_branch': 'master',
#  'target_project_id': 68,
#  'title': '<redacted>',
#  'updated_at': '2016-10-03T04:30:46.671+02:00',
#  'upvotes': 0,
#  'user_notes_count': 0,
#  'web_url': '<redacted>/merge_requests/184',
#  'work_in_progress': False}
```

gpocentek:
Thank you for your detailed report. This certainly looks like a bug in python-gitlab.

hakkeroid:
[...]

PR #169, opened 2016-10-22
Convert response list to single data source for iid requests

hakkeroid:
Solves #160 

It is questionable though if the iid check should be located here as iids are specific and not available with all gitlab resources.

### DIFF
--- a/gitlab/objects.py
+++ b/gitlab/objects.py
@@ -378,6 +378,22 @@ class GitlabObject(object):
             data = self.gitlab.get(self.__class__, data, **kwargs)
             self._from_api = True
 
+            # the API returned a list because custom kwargs where used
+            # instead of the id to request an object. Usually parameters
+            # other than an id return ambiguous results. However in the
+            # gitlab universe iids together with a project_id are
+            # unambiguous for merge requests and issues, too.
+            # So if there is only one element we can use it as our data
+            # source.
+            if 'iid' in kwargs and isinstance(data, list):
+                if len(data) < 1:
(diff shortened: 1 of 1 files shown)


gpocentek:
Moving the checks elsewhere will lead to extra complexity. Still I'm not entirely convinced that this is the place for this test, but can't seem to find a better place to handle it.

Could you rebase your patch and I think I'll merge it as is since it clearly fixes the bugs and add support for a nice feature. Thank you.

hakkeroid:
No problem. I rebased the branch.

I thought about that. What I could think of was another class attribute like "identifiers" which is a list of possible id's to check. This could be a simple `['id']` on the parent class GitlabObject as a sane default and a `['id', 'iid']` on the respective subclasses. 
Then any id relevant tests can expect a list of identifiers to check against. Although when going into that direction a simple list might not be enough because iid itself isn't clearly distinctive. Only in combination with project_id.

Another possibility could be to have a dedicated class for id handling. Usually
that seems to be too extreme but gitlab's different concepts of "identity"
might justify that.

Anyway I am happy to see you are merging it.

hakkeroid:
Sorry I just saw that there is indeed `GitlabObject.idAttr`. That might be an opportunity but as you said; definitely more complex.

gpocentek:
Let's merge your patch, it's simple enough and fixes a corner case related to gitlab's special behavior. If we need to deal with more special cases in the future we'll consider a more solid solution.

Thank you for tackling this :)

PR #174, opened 2016-10-26
fix oauth token access

deanrock:
authorization header doesn't have a colon in it's value, requests' auth parameter overrides custom authorization header

(and there was a PEP8 error for line 441)

### DIFF
--- a/gitlab/__init__.py
+++ b/gitlab/__init__.py
@@ -358,7 +358,7 @@ class Gitlab(object):
 
         if oauth_token:
             self.headers.pop("PRIVATE-TOKEN", None)
-            self.headers["Authorization"] = "Bearer: %s" % oauth_token
+            self.headers["Authorization"] = "Bearer %s" % oauth_token
         elif token:
             self.headers.pop("Authorization", None)
             self.headers["PRIVATE-TOKEN"] = token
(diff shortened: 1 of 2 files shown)


dzervas:
I think that this will have problems with HTTP requests apart from GET (have you tested PUT/PUSH/DELETE?) 'cause the authorization header is overwritten by basic auth.

I fixed all these just yesterday (I wish I could see your request earlier...). Have a look #176

deanrock:
@dzervas yes, you are correct, I didn't test with other requests :/

can you please fix example: https://github.com/gpocentek/python-gitlab/pull/174/commits/3164a0b1a96d35b2260732ea09bab59be8dfee7c#diff-8a23d682236897c1a2f5bafede722cb3L28 ?

dzervas:
I think that the example should not be fixed, as the private key is not static (and you should give that "feel"  in the documentation). It's not doing the whole auth thing (you have to talk with the server to gain the key).

gpocentek:
@deanrock @dzervas thank you for your patches and comments, really appreciated. I should be able to review and test the changes next week.

gitlab/__init__.py
```diff
@@ -397,16 +397,22 @@ def _raw_get(self, path_, content_type=None, streamed=False, **kwargs):
...
 
         headers = self._create_headers(content_type)
+
+        if 'Authorization' in self.headers:
+            auth = None
+        else:
+            auth = requests.auth.HTTPBasicAuth(
+                self.http_username,
+                self.http_password)
+
```
gpocentek: There is one thing bothering me here: with oauth token we explicitly define headers, with basic HTTP auth we don't and use a python-requests feature instead. I'd prefer to have the same method for both authentication methods.

Updating the ``_create_headers`` method might be the easiest way to achieve consistency. This would also avoid code duplication in the ``_raw_*`` methods.

What do you think about this?

gpocentek:
Hi @deanrock 

This MR has not been updated in almost a year so I'm closing it. #357 is a new proposal to support oauth.

Thank you for your time.

PR #176, opened 2016-10-27
Fix oauth

dzervas:
Removed colon from the authorization header
Fixed oauth header been replaced by basic auth


gpocentek:
Hi @dzervas 

This MR has not been updated in ove a year so I'm closing it. #357 is a new proposal to support oauth.

Thank you for your time.

PR #178, opened 2016-10-28
fix bug when retrieving changes for merge request

### LINKED ISSUE #177: ProjectMergeRequest.changes() returns commit instead of changes
Hi,

I think there is a typo in this line which leads to a bug:

https://github.com/gpocentek/python-gitlab/blob/master/gitlab/objects.py#L1608

When querying a merge request for the changes, it returns the associated commits.

Cheers

cgumpert:
Erroneously a merge request would return its commit when being queried for its changes. Fixes #177.

### DIFF
--- a/gitlab/objects.py
+++ b/gitlab/objects.py
@@ -1605,7 +1605,7 @@ class ProjectMergeRequest(GitlabObject):
             GitlabConnectionError: If the server cannot be reached.
             GitlabListError: If the server fails to perform the request.
         """
-        url = ('/projects/%s/merge_requests/%s/commits' %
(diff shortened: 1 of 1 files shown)


gpocentek:
Nice catch, thank you!

cgumpert:
Just out of curiosity: is there already a plan when the next release will be available? We need to have the fix from the merge request for our production system. Of course we would like to get it from the official repository rather than from my local fork as soon as feasible.

Thanks

gpocentek:
I plan to do a release soon, hopefully by the end of the week.

PR #180, opened 2016-10-29
Feature: enable / disable the deploy key in a project

Asher256:
To solve the issue https://github.com/gpocentek/python-gitlab/issues/179 I added 2 new features to the class ProjectKeyManager: enable_deploy_key() and disable_deploy_key() .

### DIFF
--- a/docs/gl_objects/deploy_keys.py
+++ b/docs/gl_objects/deploy_keys.py
@@ -34,3 +34,15 @@ key = project.keys.list(key_id)
 # or
 key.delete()
 # end delete
+
+# enable
+key = gl.project_keys.enable(key_id, project_id=1)
+# or
+key = project.keys.enable(key_id)
+# end enable
+
+# disable
+key = gl.project_keys.disable(key_id, project_id=1)
+# or
+key = project.keys.disable(key_id)
+# end disable
--- a/docs/gl_objects/deploy_keys.py
+++ b/docs/gl_objects/deploy_keys.py
@@ -36,13 +36,9 @@ key.delete()
 # end delete
 
 # enable
-key = gl.project_keys.enable(key_id, project_id=1)
-# or
 key = project.keys.enable(key_id)
 # end enable
 
 # disable
-key = gl.project_keys.disable(key_id, project_id=1)
-# or
 key = project.keys.disable(key_id)
 # end disable
--- a/gitlab/objects.py
+++ b/gitlab/objects.py
@@ -1249,21 +1249,21 @@ class ProjectKey(GitlabObject):
     requiredUrlAttrs = ['project_id']
     requiredCreateAttrs = ['title', 'key']
 
-    def enable(self, key_id):
-        """Enable a deploy key for a project."""
-        url = '/projects/%s/deploy_keys/%s/enable' % (self.project_id, key_id)
-        r = self.gitlab._raw_post(url)
-        raise_error_from_response(r, GitlabProjectDeployKeyError, 200)
-
-    def disable(self, key_id):
-        """Disable a deploy key for a project."""
-        url = '/projects/%s/deploy_keys/%s/disable' % (self.project_id, key_id)
-        r = self.gitlab._raw_delete(url)
-        raise_error_from_response(r, GitlabProjectDeployKeyError, 200)
-
 
 class ProjectKeyManager(BaseManager):
     obj_cls = ProjectKey
+
+    def enable(self, key_id):
+        """Enable a deploy key for a project."""
+        url = '/projects/%s/deploy_keys/%s/enable' % (self.parent.id, key_id)
+        r = self.gitlab._raw_post(url)
+        raise_error_from_response(r, GitlabProjectDeployKeyError, 201)
+
+    def disable(self, key_id):
+        """Disable a deploy key for a project."""
+        url = '/projects/%s/deploy_keys/%s/disable' % (self.parent.id, key_id)
+        r = self.gitlab._raw_delete(url)
+        raise_error_from_response(r, GitlabProjectDeployKeyError, 201)
 
 
 class ProjectEvent(GitlabObject):
--- a/gitlab/objects.py
+++ b/gitlab/objects.py
@@ -1249,21 +1249,21 @@ class ProjectKey(GitlabObject):
     requiredUrlAttrs = ['project_id']
     requiredCreateAttrs = ['title', 'key']
 
-
-class ProjectKeyManager(BaseManager):
-    obj_cls = ProjectKey
-
     def enable_deploy_key(self, project_id, key_id):
-        """Enable a deploy key in a project."""
+        """Enable a deploy key for a project."""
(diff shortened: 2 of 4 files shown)


gitlab/objects.py
```diff
@@ -1253,6 +1253,18 @@ class ProjectKey(GitlabObject):
 class ProjectKeyManager(BaseManager):
     obj_cls = ProjectKey
 
+    def enable_deploy_key(self, project_id, key_id):
+        """Enable a deploy key in a project."""
+        url = '/projects/%s/deploy_keys/%s/enable' % (project_id, key_id)
+        r = self.gitlab._raw_post(url)
+        raise_error_from_response(r, GitlabBuildRetryError, 201)
```
gpocentek: The ``GitlabBuildRetryError`` exception doesn't seem appropriate here. Maybe define new exceptions for these 2 methods.

gpocentek: Also I'd rather add ``enable`` and ``disable`` methods on the ``ProjectKey`` class instead of adding these methods on the manager, to keep the existing logic.

gitlab/objects.py
```diff
@@ -1253,6 +1253,18 @@ class ProjectKey(GitlabObject):
...
+
+    def disable_deploy_key(self, project_id, key_id):
+        """Disable a deploy key in a project."""
+        url = '/projects/%s/deploy_keys/%s/disable' % (project_id, key_id)
+        r = self.gitlab._raw_post(url)
```
gpocentek: Looking at the doc the method is DELETE, not POST: https://github.com/gitlabhq/gitlabhq/blob/master/doc/api/deploy_keys.md

gitlab/objects.py
Asher256: Thanks for the suggestions.

Question: if I move the functions enable/disable to ProjectKey, should I add an a new enable/disable functions to BaseManager?

gitlab/objects.py
```diff
@@ -1249,6 +1249,18 @@ class ProjectKey(GitlabObject):
     requiredUrlAttrs = ['project_id']
     requiredCreateAttrs = ['title', 'key']
 
+    def enable(self, key_id, project_id):
```
gpocentek: self should be the only argument, project and key id should be retrieved for the object attributes:
`````
url = '...' % (self.project_id, self.id)
`````

File: gitlab/objects.py [hunk 4]
```diff
@@ -1249,6 +1249,18 @@ class ProjectKey(GitlabObject):
...
+        """Enable a deploy key for a project."""
+        url = '/projects/%s/deploy_keys/%s/enable' % (project_id, key_id)
+        r = self.gitlab._raw_post(url)
+        raise_error_from_response(r, GitlabProjectDeployKeyError, 201)
+
+    def disable(self, key_id, project_id):
```
gpocentek: Same here.

gitlab/objects.py
```diff
@@ -1249,6 +1249,18 @@ class ProjectKey(GitlabObject):
...
+        """Disable a deploy key for a project."""
+        url = '/projects/%s/deploy_keys/%s/disable' % (project_id, key_id)
+        r = self.gitlab._raw_delete(url)
+        raise_error_from_response(r, GitlabProjectDeployKeyError, 201)
```
gpocentek: The expected return code is 200, not 201.

docs/gl_objects/deploy_keys.py
```diff
@@ -34,3 +34,15 @@
 # or
 key.delete()
 # end delete
+
+# enable
+key = gl.project_keys.enable(key_id, project_id=1)
```
gpocentek: Since the new methods are defined for the ProjectKey class you should only have one example:

````` python
key.enable()
`````

File: gitlab/objects.py (hunk 5 above)
Comment (Asher256) [inline gitlab/objects.py]: No, the expected code is 201. Check:
https://docs.gitlab.com/ce/api/deploy_keys.html

```
Enable a deploy key

Enables a deploy key for a project so this can be used. Returns the enabled key, with a status code 201 when successful.

curl --request POST --header "PRIVATE-TOKEN: 9koXpg98eAheJpvBs5tK" https://gitlab.example.com/api/v3/projects/5/deploy_keys/13/enable
Attribute	Type	Required	Description
id	integer	yes	The ID of the project
key_id	integer	yes	The ID of the deploy key
Example response:

{
   "key" : "ssh-rsa AAAA...",
   "id" : 12,
   "title" : "My deploy key",
   "created_at" : "2015-08-29T12:44:31.550Z"
}
```

File: docs/gl_objects/deploy_keys.py (hunk 6 above)
Comment (Asher256) [inline docs/gl_objects/deploy_keys.py]: Done.

**Comment by Asher256** [comment]
I put back enable() and disable() functions to ProjectKeyManager() because:
```
>>> type(project.keys)
gitlab.objects.ProjectKeyManager
```

File: gitlab/objects.py (hunk 3 above)
Comment (Asher256) [inline gitlab/objects.py]: Done.

But I read from self.parent.id in ProjectKeyManaged. Check below...

(if you have another suggestion related to that, tell me)

**Comment by Asher256** [comment]
Please check again Gauvain.

**Comment by gpocentek** [comment]
@Asher256 To make things more consistent I moved the methods to the ``ProjectKey`` class.

I'm considering adding methods to the managers to avoid multiple API calls in such cases but this is something that I'd rather change for all the objects/managers at the same time.

Thank you for the patch :)

**Comment by Asher256** [comment]
You are welcome!

I'll try your the master branch on monday.

**Comment by zhujinhe** [comment]
How to enable a public key to a specific  project? Public key(created by admin) seems do not belong to any project.
Assume that public key id is 13, project id is 34.
```
project = gl.projects.get(34)
project.keys.enable(13)
```
Error Message:
AttributeError: 'ProjectKeyManager' object has no attribute 'enable'

**Comment by jnancel** [comment]
@Asher256 @gpocentek 

Did you update the documentation on that issue? Cause I can't find a way to do that following what's in readthedocs : if I try to create my ProjectKey object with the existing project id and key id ( knowing that the key id is global and does not exist yet in the project ), I get a 404 so I cannot apply enable : 
```
>>> gl.project_keys.get(project_id=74, id=63)
  File "/usr/lib/python2.7/site-packages/gitlab/objects.py", line 102, in get
    return self.obj_cls.get(self.gitlab, id, **args)
  File "/usr/lib/python2.7/site-packages/gitlab/objects.py", line 276, in get
    return cls(gl, id, **kwargs)
  File "/usr/lib/python2.7/site-packages/gitlab/objects.py", line 380, in __init__
    data = self.gitlab.get(self.__class__, data, **kwargs)
  File "/usr/lib/python2.7/site-packages/gitlab/__init__.py", line 425, in get
    raise_error_from_response(r, GitlabGetError)
  File "/usr/lib/python2.7/site-packages/gitlab/exceptions.py", line 180, in raise_error_from_response
    response_body=response.content)
gitlab.exceptions.GitlabGetError: 404: 404 Not found

```

Thanks

PR #185, opened 2016-11-21
Allow blank token in config

so0k:
Using a config with `http_username/http_password` only (*no token*) did not work...

The changes in this pull request allow usage of gitlab-cli with the following config:
```
[global]
default = example
ssl_verify = true
timeout = 5

[example]
url = https://gitlab.example.com
http_username = root
http_password = mypassword
```

### DIFF
--- a/gitlab/__init__.py
+++ b/gitlab/__init__.py
@@ -168,9 +168,12 @@ class Gitlab(object):
     def credentials_auth(self):
         """Performs an authentication using email/password."""
         if not self.email or not self.password:
-            raise GitlabAuthenticationError("Missing email/password")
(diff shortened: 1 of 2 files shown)


so0k:
should probably test if `http_username` & `http_password` are blank and throw auth error

gpocentek:
Thank you for this patch.

I didn't know it was possible to disabled token authentication. Could you tell me where this should be configured on the gitlab side?

Also the travis test fails, this probably means something unexpected now happens with token authentication. Could you check?

gpocentek:
I'm closing this PR since it didn't get an answer or an update in months. Feel free to provide a new one.

so0k:
It's ok, works in my fork

PR #186, opened 2016-11-21
Fix `should_remove_source_branch`

petebrowne:
I'm getting the following 400 error when using `merge`:

```
  File "/usr/local/lib/python2.7/site-packages/gitlab/objects.py", line 1533, in merge
    raise_error_from_response(r, errors)
  File "/usr/local/lib/python2.7/site-packages/gitlab/exceptions.py", line 168, in raise_error_from_response
    response_body=response.content)
gitlab.exceptions.GitlabOperationError: 400: {"error":"should_remove_source_branch is invalid"}
```

It seems like Gitlab has changed their validations - using `True` seems to work.

### DIFF
--- a/gitlab/objects.py
+++ b/gitlab/objects.py
@@ -1732,9 +1732,9 @@ class ProjectMergeRequest(GitlabObject):
         if merge_commit_message:
             data['merge_commit_message'] = merge_commit_message
         if should_remove_source_branch:
-            data['should_remove_source_branch'] = 'should_remove_source_branch'
+            data['should_remove_source_branch'] = True
         if merged_when_build_succeeds:
-            data['merged_when_build_succeeds'] = 'merged_when_build_succeeds'
+            data['merged_when_build_succeeds'] = True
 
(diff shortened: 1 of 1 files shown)


gpocentek:
@petebrowne Thank you for your patch!

The problem I see here is that this code will fail for users dealing with older gitlab installations. That type of change shouldn't append on the gitlab side but we'll have to deal with it.

I don't have a solution to offer. Maybe check the version before deciding what to do with this attribute?

petebrowne:
How do we check the Gitlab version? Is that an API call we need to make? Or is that configuration for the end user of this client?

Taking a step back - why was this set to `"should_remove_source_branch"` in the first place? It seems very odd that the API wouldn't accept `true`. Which versions of the Gitlab API required this?

gpocentek:
I honestly don't remember if I had to use this value or if any value would have worked. I'll test on an older gitlab version to make sure switching to `true` works.

gpocentek:
Current code doesn't seem correct. Let's merge, thank you!

petebrowne:
Thanks!

PR #187, opened 2016-11-27
Added support for Snippets

guyzmo:
new API, that should be coming in Gitlab 8.15

cf [Gitlab-CE MR !6373](https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/6373)

### DIFF
--- a/gitlab/__init__.py
+++ b/gitlab/__init__.py
@@ -103,6 +103,7 @@ class Gitlab(object):
         self.runners = RunnerManager(self)
         self.settings = ApplicationSettingsManager(self)
         self.sidekiq = SidekiqManager(self)
+        self.snippets = SnippetManager(self)
         self.users = UserManager(self)
         self.teams = TeamManager(self)
         self.todos = TodoManager(self)
@@ -469,7 +470,8 @@ class Gitlab(object):
             params.pop(obj.idAttr)
 
         r = self._raw_delete(url, **params)
-        raise_error_from_response(r, GitlabDeleteError)
+        raise_error_from_response(r, GitlabDeleteError,
+                                  expected_code=[200, 204])
         return True
 
     def create(self, obj, **kwargs):
--- a/gitlab/objects.py
+++ b/gitlab/objects.py
@@ -1018,6 +1018,67 @@ class LicenseManager(BaseManager):
     obj_cls = License
 
 
+class Snippet(GitlabObject):
+    _url = '/snippets'
+    _constructorTypes = {'author': 'User'}
+    requiredCreateAttrs = ['title', 'file_name', 'content']
+    optionalCreateAttrs = ['lifetime', 'visibility_level']
+    optionalUpdateAttrs = ['title', 'file_name', 'content', 'visibility_level']
+    shortPrintAttr = 'title'
+
+    def content(self, streamed=False, action=None, chunk_size=1024, **kwargs):
+        """Return the raw content of a snippet.
+
+        Args:
+            streamed (bool): If True the data will be processed by chunks of
+                `chunk_size` and each chunk is passed to `action` for
+                treatment.
+            action (callable): Callable responsible of dealing with chunk of
+                data.
+            chunk_size (int): Size of each chunk.
+
+        Returns:
+            str: The snippet content
+
+        Raises:
+            GitlabConnectionError: If the server cannot be reached.
+            GitlabGetError: If the server fails to perform the request.
+        """
+        url = ("/snippets/%(snippet_id)s/raw" %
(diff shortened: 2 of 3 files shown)


guyzmo:
the code is still untested — and will fail up until the !6373 is accepted, but it's a start to have the new API implemented as soon as !6373 is merged and 8.15 released.

gpocentek:
Thank you! I did a quick review and the code looks OK to me. Let's wait for the feature to land in gitlab and I'll test properly.

The unit tests fail because of a typo, shouldn't be too hard to fix :)

guyzmo:
ok, so far so good, all tests are 👌

BTW, about APIs, you might want to have a look at upcoming changes that they're planning to do for the next major release:

* [API Changes for 9.0](https://gitlab.com/gitlab-org/gitlab-ce/issues/20070)
* [Use of swagger](https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/2397)

guyzmo:
It'd be also nice if we could throw a nice error if the API is missing (using the gitlab server's version, for example?)

gpocentek:
I have some ideas about how we could better deal with versionning, but it would require important changes. This is getting clearer in my head by I'm afraid it will not implemented soon (lack of time basically).

I'd be happy to share if you're interested, and I'm open to suggestions :)

guyzmo:
as an update, the API is now merged in master, and due to release in 8.15, which is supposed to be out in 3 days. 🤞

guyzmo:
hurra 🙌 gitlab 8.15 has been released yesterday, and it's featuring [the `/snippets` endpoint](https://about.gitlab.com/2016/12/22/gitlab-8-15-released/#api-changes) through the patch I implemented.

So now we can test this PR and merge it if it's actually working fine, which it should, as I wrote the other side of the API as well 😛

gpocentek:
@guyzmo thank you for the update and the other side of the API.

I should be able to have a look at the patches in the coming days.

gitlab/__init__.py
```diff
@@ -484,7 +484,7 @@ def delete(self, obj, id=None, **kwargs):
             params.pop(obj.idAttr)
 
         r = self._raw_delete(url, **params)
-        raise_error_from_response(r, GitlabDeleteError)
+        raise_error_from_response(r, GitlabDeleteError, expected_code=[200, 204])
```
guyzmo: cf [delete return values](https://docs.gitlab.com/ce/api/snippets.html#delete-snippet), new API spec returns 204 for deletes and return no contents.

guyzmo:
I made a quick check with my git-repo implementation (with python-gitlab pointing to my branch) and it's working fine. I updated the tests as well, but they're failing as for some reasons [my tests are not using my fork of python-gitlab](https://travis-ci.org/guyzmo/git-repo/jobs/186391777#L395).

gpocentek:
The test failure is related to issue #199.

gpocentek:
Thank you!

gpocentek:
Question: is the ``owned`` method really needed? ``list()`` will only list the current user snippets.

And I think ``all()`` should be renamed ``public()`` to avoid any confusion. What do you think?

gpocentek:
Also the ``content()`` method is a problem when updating an object, it conflicts with the optional attribute.

guyzmo:
> Also the content() method is a problem when updating an object, it conflicts with the optional attribute.

ah! good catch, I did not notice that… I guess `content()` should be renamed `raw()` then.

> And I think all() should be renamed public() to avoid any confusion. What do you think?

sure!

> Question: is the owned method really needed? list() will only list the current user snippets.

good catch, and yes! ☺

did you handled those when you did the merge?

guyzmo:
(BTW, you can ping me on IRC if you use it, I'm zmo on freenode)

and happy holiday ☺

gpocentek:
I've made the changes this morning, nothing really complicated. Thank you for you feedback.

FYI the `visibility_level` attribute doesn't seem to be exposed by the server API. Is this intended?

guyzmo:
This is not intended and I actually did not notice, and neither did the reviewers /o\

Though I guess it's kind of ok… Someone else will have to fix that 😀

https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/6373/diffs#cc752c66542cbeeb21ca7a57c430836b865580c9_203_203

guyzmo:
Otherwise, if you've got an idea of when you think you'll make a new release, I'd be happy to know so I can be prepared to release the new version of git-repo using it ☺

And thank you 😀

gpocentek:
I'll probably make a new release this week to fix a few problems with gitlab 8.15.

PR #192, opened 2016-12-12
Fix JIRA service editing for GitLab 8.14+

galet:
GitLab simplified the configuration for JIRA service and renamed most of
the fields. To maintain backward compatibility all mandatory fields were
moved to optional section. 

See: https://docs.gitlab.com/ce/api/services.html#jira

**Example of exception it fixes**:
`gitlab.exceptions.GitlabUpdateError: 400: 400 (Bad request) "url" not given`


gpocentek:
Looks good, thanks!

PR #208, opened 2017-01-23
minor thing

nutztherookie:
it's just confusing that it would say "pip" again :)


gpocentek:
Thanks!

I merged your change but updated the commit message, hope you don't mind. Nice catch :)

PR #210, opened 2017-01-26
Add builds-email and pipelines-email services

comel:
Add builds emails and pipelines emails services.


gpocentek:
Looks good, thank you!

PR #220, opened 2017-02-09
Added pipeline_events to ProjectHook attrs

### LINKED ISSUE #219: Missing pipeline_events in hooks
https://github.com/gpocentek/python-gitlab/blob/master/gitlab/objects.py#L1418

pipeline_events should be in here, and it's missing. 

This results in creating a hook with no pipeline_events, but all the other events.

If you update a hook, it will update everything except for this. 

It just uses the default value of False.

alexwidener:
Ran tests, all passed.

This will fix #219


alexwidener:
Doesn't look like the failure is an issue with anything that I pushed:

```
**gitlab.exceptions.GitlabUpdateError: 400: {"error":"domain_whitelist is invalid"}**
```

gpocentek:
The patch looks good. The test failure is due to another problem. I'm working on it.

Issue #235, opened 2017-03-15
Implement API for cherry picking commits


cgumpert:
Hi,

the Gitlab API provides a functionality for [cherry-picking](https://docs.gitlab.com/ce/api/commits.html#cherry-pick-a-commit) individual commits. This is not yet implemented in the nice python wrapper. Could this functionality be added?

Thanks,
Christian

gpocentek:
#236 has been merge, thanks :+1:

cgumpert:
Hi @gpocentek, do you have an idea when the next release containing this enhancement will be available?

PR #236, opened 2017-03-15
Provide API wrapper for cherry picking commits

cgumpert:
This is a possible implementation for addressing #235. It allows to cherry pick individual commits through the Gitlab API both from the python code as well as from the command line. The documentation has been updated.

### DIFF
--- a/docs/gl_objects/commits.rst
+++ b/docs/gl_objects/commits.rst
@@ -43,7 +43,7 @@ Get the diff for a commit:
    :start-after: # diff
    :end-before: # end diff
 
-Cherry-pick a commit into another branch
+Cherry-pick a commit into another branch:
 
 .. literalinclude:: commits.py
    :start-after: # cherry
--- a/docs/gl_objects/commits.rst
+++ b/docs/gl_objects/commits.rst
@@ -43,6 +43,12 @@ Get the diff for a commit:
    :start-after: # diff
    :end-before: # end diff
 
+Cherry-pick a commit into another branch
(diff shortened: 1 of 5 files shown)


docs/gl_objects/commits.rst
```diff
@@ -43,6 +43,12 @@ Get the diff for a commit:
    :start-after: # diff
    :end-before: # end diff
 
+Cherry-pick a commit into another branch
```
gpocentek: Please end the line with a colon (``:``) to introduce the example.

gitlab/objects.py
```diff
@@ -1353,6 +1353,23 @@ def builds(self, **kwargs):
...
+        """Cherry-pick a commit into a branch.
+
+        Args:
+            branch (str): Name of target branch.
+
+        Raises:
+            GitlabCherryPickError: If the cherry pick could not be applied.
+        """
+        url = '/projects/{0:s}/repository/commits/{1:s}/cherry_pick'.format(
+            self.project_id, self.id)
```
gpocentek: For the sake of consistency could you use the old ``%()s`` format? Yes, this is nitpicking :)

cgumpert: I had problems with keeping the line length below the limit of 79 characters. I worked around it now by introducing another local variable with a short name. If you have any other suggestion on how to break this string across two lines, I'd be happy to hear.

gpocentek:
Thank you for this patch!

PR #241, opened 2017-03-20
add 'delete source branch' option when creating MR

cgumpert:
The Gitlab API provides the possibility to check the option "delete source branch when merge is accepted" [(here)](https://docs.gitlab.com/ce/api/merge_requests.html#create-mr). This PR allows to make use of this API feature.


gpocentek:
Thank you :)

PR #245, opened 2017-03-29
Add time_stats to ProjectMergeRequest

TimNN:
The code was more or less copied from the equivalent function on


gpocentek:
Looks good, thank you.

PR #257, opened 2017-05-10
Add support for nested groups

matejzero:
Add `parent_id` attribute support when creating groups, so we can create nested groups.

I'm not sure if this is the only change needed, but it is working on my local copy and I have successfully created nested groups while migrating SVN to GitLab.

### DIFF
--- a/gitlab/objects.py
+++ b/gitlab/objects.py
@@ -2702,8 +2702,10 @@ class GroupProjectManager(BaseManager):
 class Group(GitlabObject):
     _url = '/groups'
     requiredCreateAttrs = ['name', 'path']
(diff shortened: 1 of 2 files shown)


matejzero:
I've added all group attributes found on GitLab API page.

Tests currently only for `parent_id`.

gpocentek:
Thank you for this change, I actually needed it myself :+1: 

I took the liberty to fix the minor problem which made the tests fail.

matejzero:
Thanks for merging. I wanted to fix the failing tests, but was busy with gitlab migration for the last 2 weeks, so thanks for fixing that as well:)

PR #265, opened 2017-05-20
Fixed repository_tree and repository_blob path encoding

yosyos04:
This is done to enable support for paths with spaces and other special characters


gpocentek:
Looks good.

I wonder if we could make the change more generic but this is a first step and it will do for now.

Thanks!

PR #268, opened 2017-05-28
import urlencode() from six.moves.urllib.parse instead of from urllib

anthrotype:
Fixes AttributeError on Python 3, as `urlencode` function has been moved from `urllib` to `urllib.parse` module.

`six.moves.urllib.parse.urlencode()` is an py2.py3 compatible alias of `urllib.parse.urlencode()` on Python 3, and of `urllib.urlencode()` on Python 2.

### DIFF
--- a/gitlab/v3/objects.py
+++ b/gitlab/v3/objects.py
@@ -23,7 +23,7 @@ import json
 import warnings
 
 import six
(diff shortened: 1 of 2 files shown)


anthrotype:
ping @gpocentek

gpocentek:
Thanks!

PR #272, opened 2017-06-05
Add new event types to ProjectHook

astronouth7303:
These are being returned in the live API, but can't set them.


gpocentek:
Looks good. Thanks!

PR #276, opened 2017-06-16
Missing expires_at in GroupMembers update

elisarver:
optionalCreateAttrs was set twice in GroupMember due to possible copy-paste error. This corrects the second call from optionalCreateAttrs to optionUpdateAttrs.


gpocentek:
Nice catch, thank you!

PR #278, opened 2017-06-23
Docs 📖: Add link to gitlab docs on obtaining a token

asfaltboy:
I find these sort of links very user friendly 😅

### DIFF
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -80,12 +80,15 @@ section.
      - URL for the GitLab server
    * - ``private_token``
      - Your user token. Login/password is not supported.
+       Refer `the official documentation`__ to learn how to obtain a token.
    * - ``api_version``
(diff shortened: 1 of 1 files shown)


docs/cli.rst
```diff
@@ -80,13 +80,16 @@ section.
      - URL for the GitLab server
    * - ``private_token``
      - Your user token. Login/password is not supported.
+       Please refer `the official documentation`__ to learn how to obtain a token.
```
gpocentek: You can remove `Please` here as you're not asking for permission to the reader ;)

I'm not a native english speaker but I believe "Refer to the official doc..." is better.

asfaltboy:
Thanks @gpocentek , removed "Please" and amended the commit.

gpocentek:
Thanks! The test failures are not related to this patch so let's merge!

PR #282, opened 2017-06-26
Fixed repository_compare examples

velvetz7:


docs/gl_objects/projects.py
```diff
@@ -178,11 +178,11 @@
 result = project.repository_compare('master', 'branch1')
 
 # get the commits
-for i in commit:
-    print(result.commits)
+for commit in result.commits:
```
gpocentek: `result` is a dict, so you need `result['commits']` here.

docs/gl_objects/projects.py
```diff
@@ -178,11 +178,11 @@
...
+    print(commit)
 
 # get the diffs
-for file_diff in commit.diffs:
+for file_diff in result.diffs:
```
gpocentek: Same here for diffs.

velvetz7:
You are right, I fixed them.

gpocentek:
Looks good, thank you for the quick update!

PR #287, opened 2017-07-08
Added dependency injection support for Session


### LINKED ISSUE #280: methods deprecation
I've had a report for deprecation warnings in git-repo when using gitlab. I then looked at the release notes, and found out that a bunch of the `Gitlab` object's API is getting deprecated.

Though, I actually need [those](https://github.com/python-gitlab/python-gitlab/issues/280#issuecomment-310841660) in [git-repo](https://github.com/guyzmo/git-repo/blob/devel/git_repo/services/ext/gitlab.py#L34,L36) where I initialise objects in two steps, first I initialise the instances (and importantly request's session) then I load the configuration, and finally the connect method configures the gitlab object.

So either having a method to configure the object after it has been instanciated (with an API similar to the constructor), or keeping the setters really matter to me! Or I'll end up forced to monkey patched internals of `python-gitlab` and that can end up ugly 😖

cf [this issue](https://github.com/guyzmo/git-repo/issues/166)

Thank you for your lib ♥

guyzmo:
fixes #280

### DIFF
--- a/gitlab/__init__.py
+++ b/gitlab/__init__.py
@@ -71,7 +71,7 @@ class Gitlab(object):
 
     def __init__(self, url, private_token=None, email=None, password=None,
                  ssl_verify=True, http_username=None, http_password=None,
-                 timeout=None, api_version='3'):
+                 timeout=None, api_version='3', session=None):
 
         self._api_version = str(api_version)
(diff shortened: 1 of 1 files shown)


gpocentek:
Looks good, thank you :+1:

Issue #270, opened 2017-06-02
Gitlab.from_config doesn't allow string ssl_verify


mdhausman:
For many users who are working with self-signed certs, ssl verification can be troublesome. I found that while I could use python-gitlab with our internal certs, I can't do it from a configuration file.

For instance, this works:
```
gitlab_dao = gitlab.Gitlab('https://our.gitlab.location', private_token='somethingprivate', ssl_verify='/our/companys/internal/certs/cert.crt')
```
But when I tried to set up a config file like
```
[global]
default = https://our.gitlab.location
ssl_verify = /our/companys/internal/certs/cert.crt
timeout = 5
api_version = 3

[somewhere]
url = https://our.gitlab.location
private_token = somethingprivate
```
And call it with `gitlab_dao = gitlab.Gitlab.from_config('somewhere', ['my-config.cfg'])`, it would break on `gitlab_dao.auth()`.

Some debugging led me to the reason: GitlabConfigParser [line 66](https://github.com/python-gitlab/python-gitlab/blob/master/gitlab/config.py#L67): `self.ssl_verify = self._config.getboolean(self.gitlab_id, 'ssl_verify')`

Because it's only looking for a boolean, it throws an exception here and sets ssl_verify to `True`. This kills the internal cert.

mdhausman:
I've got a hacky workaround in the meantime:

```
gitlab_dao = gitlab.Gitlab.from_config('somewhere', ['my-config.cfg'])
gitlab_dao.ssl_verify = '/our/companys/internal/certs/cert.crt'
gitlab_dao.auth()
```

mdhausman:
I can probably make a pull request to handle this within the next few weeks. Shouldn't be hard.

I'm thinking leave line 66 as-is, but on exception it should instead try to `self._config.get(self.gitlab_id, 'ssl_verify')` instead. If it's a falsy return (or yet another exception), default back to `True`. Then a little update to test_config.py and we're good.

Issue #204, opened 2017-01-13
Specify CA certificate


cmr:
In `requests`, one can set `verify` to the path of a CA certificate used to verify the the connection. This is really useful for CE instances when orgs run their own CA (ie, "self-signed").  This feature would allow me to not have to use ssl_verify=False.

gpocentek:
Definitely an interesting feature to implement. Thanks for the suggestion :)

mdhausman:
It looks like this has been implemented? When I set `ssl_verify='path/to/my/signed/cert.pem'`, it uses that cert appropriately, with no InsecureRequestWarnings. Exception being when I try to set ssl_verify from a config file as documented in #270

mdhausman:
Finally getting around to looking into this.

Yeah, it looks like under the hood this is already [exactly what requests does](https://github.com/requests/requests/blob/master/requests/sessions.py#L468) and that we should be able to use it that way. Furthermore, this can be done safely as this behavior appears [even in the earliest version of requests supported by python-gitlab](https://github.com/requests/requests/blob/b30b09e84c5709149e796af5cb07671db823af53/requests/sessions.py#L231).

gpocentek:
Fixed by @mdhausman in #294.

PR #294, opened 2017-07-31
Support SSL verification via internal CA bundle

mdhausman:
- Also updates documentation
- See issues #204 and #270

### DIFF
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -61,9 +61,10 @@ parameters. You can override the values in each GitLab server section.
      - Possible values
      - Description
    * - ``ssl_verify``
-     - ``True`` or ``False``
-     - Verify the SSL certificate. Set to ``False`` if your SSL certificate is
-       auto-signed.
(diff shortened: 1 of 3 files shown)


mdhausman:
Hm. It looks like all the failing checks are also failing on the current master branch. So the good news is, it's not by code breaking things. The bad news is, something is definitely broken.

gpocentek:
The python3.6 failure is definitely not related to your patch. The functional tests failure is probably a temporary error. Let's see how it goes with a new run.

gpocentek:
The patch looks good BTW, thanks!

mdhausman:
Glad it looks good. Wish the tests were working, though 😅

gpocentek:
Could you rebase your change on the current master? The tests should be OK now.

mdhausman:
Success! 💃 🎉 🎊 🍾

gpocentek:
Thank you!

PR #307, opened 2017-09-11
Fix tag api

RobberPhex:
* Tag can get by id
* on_http_error bring `error_message`, `response_code`, `response_body`

### DIFF
--- a/gitlab/v4/objects.py
+++ b/gitlab/v4/objects.py
@@ -1035,8 +1035,7 @@ class ProjectTag(ObjectDeleteMixin, RESTObject):
         self.release = server_data
 
 
-class ProjectTagManager(GetMixin, CreateMixin, DeleteMixin,
-                        RESTManager):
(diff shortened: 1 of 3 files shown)


gitlab/v4/objects.py
```diff
@@ -1035,7 +1035,7 @@ def set_release_description(self, description, **kwargs):
         self.release = server_data
 
 
-class ProjectTagManager(GetFromListMixin, CreateMixin, DeleteMixin,
+class ProjectTagManager(GetMixin, CreateMixin, DeleteMixin,
```
gpocentek: You could use the `NoUpdateMixin` instead of the 4 other mixins.

gpocentek:
Thanks!

The changes look good but you need to re-add the support for listing tags.

gpocentek:
Thanks for the quick update!

RobberPhex:
please release new version. currently, v1.0.0 have this bug.

PR #321, opened 2017-09-22
Add doc to get issue from iid

BancarelValentin:


gpocentek:
@BancarelValentin Sorry for the late comment.

In v4 issues are retrieved using iids, so the updated doc is only valid for v3. Could you add this information in the change? Thank you.

gpocentek:
Thanks!

PR #342, opened 2017-10-13
Add mattermost service support

matejzero:
Add support for mattermost service.

I hope this is all I need to do:)

### DIFF
--- a/gitlab/v3/objects.py
+++ b/gitlab/v3/objects.py
@@ -1675,6 +1675,7 @@ class ProjectService(GitlabObject):
 
                  # Optional fields
                  'username', 'password', 'jira_issue_transition_id')),
+        'mattermost': (('webhook',), ('username', 'channel')),
         'pivotaltracker': (('token', ), tuple()),
         'pushover': (('api_key', 'user_key', 'priority'), ('device', 'sound')),
         'redmine': (('new_issue_url', 'project_url', 'issues_url'),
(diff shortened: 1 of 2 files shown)


gitlab/v3/objects.py
```diff
@@ -1675,6 +1675,7 @@ class ProjectService(GitlabObject):
 
                  # Optional fields
                  'username', 'password', 'jira_issue_transition_id')),
+        'mattermost': (('webhook',), ('username', 'channel')),
```
gpocentek: I can't find references to `username` and `channel` in the gitlab documentation and code. Is this really supported?

matejzero: Looking at [services API](https://docs.gitlab.com/ee/api/services.html#mattermost-notifications), it looks like `username` and `channel` are optional fields.

gpocentek: You're right. I guess I missed it.

matejzero:
It seems like this is not working, but the problem is in GitLab. At the moment, even if we send username / channel parameters to GitLab, they are not set. Strange, since it's documented in docs.

I opened an issue on [gitlab](https://gitlab.com/gitlab-org/gitlab-ce/issues/39895).

matejzero:
Issue is fixed in GitLab.

gpocentek:
Thanks for the feedback @matejzero

Issue #163, opened 2016-10-09
Access token support


Moezen:
I use oauth2 to authorize gitlab and get the access token which use to call gitlab apis. But gitlab.Gitlab only support for email/password and private token.

gpocentek:
Hi,

I pushed https://github.com/gpocentek/python-gitlab/tree/oauth with a possibly working oauth2 support. I don't have a server set up to properly test this. Could you give it a try?

Thank you!

deanrock:
Hi,

I submitted a pull request #174 that correctly works with oauth2 tokens.

boer0924:
@gpocentek  why the master branch not oauth2, why not merage oauth into master?

gpocentek:
@boer0924 because last time I checked it didn't work. There are some PR to fix the branch but they need to be updated.

This is not a priority for me at the moment, but I'd be happy to review patches.

Thanks!

boer0924:
@gpocentek  thanks replay!

Yestoday, when I clone the oauth branch and use it, It's not work, error 401.

My gitlab version 9.4.2, help me!

PR #357, opened 2017-11-02
Oauth token support

nathschmidt:
Addresses #163 

Small change that adds support for Oauth tokens.  Follows the same flow as the standard private token - instead of setting the`PRIVATE-TOKEN` header sets `Authorization`.  Deprecated methods have not been updated/changed.  Existing open PR's for this issue haven't been used as they're over a year old and the codebase seems to have changed rather dramatically. 

There's possibly a slight issue with providing both `private_token` and `oath_token`, as `private_token` will silently take precedence maybe a warning or error would be clearer. I think the chance of a user actually doing that is pretty slim though.

I've verified against a v0.10 gitlab (v4 API). Shouldn't be any break to the existing interface.

Let me know if you've got any questions.

### DIFF
--- a/gitlab/__init__.py
+++ b/gitlab/__init__.py
@@ -70,7 +70,7 @@ class Gitlab(object):
         api_version (str): Gitlab API version to use (3 or 4)
     """
 
-    def __init__(self, url, private_token=None, email=None, password=None,
+    def __init__(self, url, private_token=None, oauth_token=None, email=None, password=None,
                  ssl_verify=True, http_username=None, http_password=None,
                  timeout=None, api_version='3', session=None):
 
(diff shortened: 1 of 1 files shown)


nathschmidt:
Happy to squash the linter issues & fixes.

gpocentek:
@nathschmidt Thank you for this patch! It looks good but I'm not sure how I can test it, I don't really have experience with oauth.

I'm wondering if I can test on gitlab.com with auth from github. Any feedback on this?

nathschmidt:
Hmmm, I'm not sure if you can chain the OAuth that way - worth a try.  I've tested it against an in house gitlab-ee (v0.10) - with great success.

nathschmidt:
This is great, thanks!  When do you think the next release will be?

PR #358, opened 2017-11-02
Update documentation for getting files with API v4

thejinx0r:


thejinx0r:
I realize now that I was using an much older version.

PR #378, opened 2017-11-17
add "list pipeline jobs" handler

akaWolf:
as described here:
https://docs.gitlab.com/ee/api/jobs.html#list-pipeline-jobs

example:
`jobs = pipeline.jobs.list()`

### DIFF
--- a/gitlab/v4/objects.py
+++ b/gitlab/v4/objects.py
@@ -1707,6 +1707,10 @@ class ProjectFileManager(GetMixin, CreateMixin, UpdateMixin, DeleteMixin,
 
 
 class ProjectPipeline(RESTObject):
+    _managers = (
+        ('jobs', 'PipelineJobManager'),
+    )
+
     @cli.register_custom_action('ProjectPipeline')
     @exc.on_http_error(exc.GitlabPipelineCancelError)
     def cancel(self, **kwargs):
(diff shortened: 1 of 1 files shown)


gitlab/v4/objects.py
```diff
@@ -1764,6 +1768,12 @@ def create(self, data, **kwargs):
         return CreateMixin.create(self, data, path=path, **kwargs)
 
 
+class PipelineJobManager(ListMixin, RESTManager):
```
gpocentek: Since you can list and get jobs, you should use a `RetrieveMixin` here.

gitlab/v4/objects.py
```diff
@@ -1764,6 +1768,12 @@ def create(self, data, **kwargs):
...
+    _path = '/projects/%(project_id)s/pipelines/%(pipeline_id)s/jobs'
+    _obj_cls = ProjectJob
```
gpocentek: Although this seems to be a good idea to use a ProjectJob, things get a bit more complicated than expected when using the CLI. To avoid bad surprises I try to stick to the upstream API logic: we are requesting a different URL so we get a different object class.

Could you create and use a `PipelineJob` object class instead?

ahsobhi:
i was just going to create a pull request for such handler, and it is very useful as well. Thanks to @akaWolf and @gpocentek

gpocentek:
I'm closing this MR since there's been no update, and #413 implements the feature.

PR #382, opened 2017-12-03
Expected HTTP response for subscribe is 201

ptomato:
It seems that the GitLab API gives HTTP response code 201 ("created") when
successfully subscribing to an object, not 200.


gpocentek:
Nice catch, thank you!

PR #389, opened 2017-12-06
mixins.py: Avoid sending empty update data to issue.save

csoriano1618:
When saving an issue we send the updated data only. 

However the server expect at least one parameter to be provided,
otherwise it fails with an exception.

Check whether we have some update to send, otherwise skip
the update altogether.

### DIFF
--- a/gitlab/mixins.py
+++ b/gitlab/mixins.py
@@ -299,7 +299,7 @@ class SaveMixin(object):
         """
         updated_data = self._get_updated_data()
         # Nothing to update. Server fails if sent an empty dict.
-        if len(updated_data.keys()) == 0:
+        if not updated_data.keys():
             return
 
         # call the manager
--- a/gitlab/mixins.py
+++ b/gitlab/mixins.py
@@ -298,8 +298,8 @@ class SaveMixin(object):
             GitlabUpdateError: If the server cannot perform the request
         """
         updated_data = self._get_updated_data()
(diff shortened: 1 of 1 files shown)


gitlab/mixins.py
```diff
@@ -298,6 +298,9 @@ def save(self, **kwargs):
             GitlabUpdateError: If the server cannot perform the request
         """
         updated_data = self._get_updated_data()
+        # Nothing to update. Servers fails if sent an empty dict.
+        if len(updated_data.keys()) == 9:
```
gpocentek: Could you explain the logic here? Where does the 9 come from?

Thanks.

csoriano1618: Whops, typo when creating the patch. It should be a 0.

gitlab/mixins.py
```diff
@@ -298,6 +298,9 @@ def save(self, **kwargs):
             GitlabUpdateError: If the server cannot perform the request
         """
         updated_data = self._get_updated_data()
+        # Nothing to update. Server fails if sent an empty dict.
+        if len(updated_data.keys()) == 0:
```
gpocentek: Thanks for the update!

A more pythonic way to write this would be:
```python
if not update_data:
    return
```

Could you update the code? Thanks again.

csoriano1618:
Thanks for the review, I didn't know en empty dictionary works with the "not".

gpocentek:
Thanks @csoriano1618 :+1:

PR #398, opened 2017-12-13
 Feature/pipeline schedules

mlq:
Hi,

I've started to implement support for pipeline schedules and wanted to ask if this is wanted to be picked up and reviewed. It allows to add pipeline schedules as well as variables. Any feedback is welcome as I am not so familiar with the code base.

Best regards

### DIFF
--- a/gitlab/v3/objects.py
+++ b/gitlab/v3/objects.py
@@ -1496,6 +1496,18 @@ class ProjectFileManager(BaseManager):
     obj_cls = ProjectFile
 
 
+class ProjectPipelineSchedule(GitlabObject):
+    _url = '/projects/%(project_id)s/pipeline_schedules'
+    _create_url = '/projects/%(project_id)s/pipeline_schedules'
+
+    requiredUrlAttrs = ['project_id']
+    requiredCreateAttrs = ['description', 'ref', 'cron']
+
+
+class ProjectPipelineSchedulesManager(BaseManager):
+    obj_cls = ProjectPipelineSchedule
+
+
 class ProjectPipeline(GitlabObject):
     _url = '/projects/%(project_id)s/pipelines'
     _create_url = '/projects/%(project_id)s/pipeline'
@@ -1803,6 +1815,7 @@ class Project(GitlabObject):
         ('notificationsettings', 'ProjectNotificationSettingsManager',
          [('project_id', 'id')]),
         ('pipelines', 'ProjectPipelineManager', [('project_id', 'id')]),
+        ('pipeline_schedules', 'ProjectPipelineSchedulesManager', [('project_id', 'id')]),
         ('runners', 'ProjectRunnerManager', [('project_id', 'id')]),
         ('services', 'ProjectServiceManager', [('project_id', 'id')]),
         ('snippets', 'ProjectSnippetManager', [('project_id', 'id')]),
--- a/gitlab/v4/objects.py
+++ b/gitlab/v4/objects.py
@@ -1706,7 +1706,23 @@ class ProjectFileManager(GetMixin, CreateMixin, UpdateMixin, DeleteMixin,
         return utils.response_content(result, streamed, action, chunk_size)
 
 
+class ProjectPipelineJob(ProjectJob):
+    pass
+
+
+class ProjectPipelineJobsManager(ListMixin, RESTManager):
+    _path = '/projects/%(project_id)s/pipelines/%(pipeline_id)s/jobs'
+    _obj_cls = ProjectPipelineJob
+    _from_parent_attrs = {'project_id': 'project_id',
+                          'pipeline_id' : 'id'}
+    _list_filters = ('scope',)
+
+
 class ProjectPipeline(RESTObject):
+    _managers = (
+        ('jobs', 'ProjectPipelineJobsManager'),
+    )
+
     @cli.register_custom_action('ProjectPipeline')
     @exc.on_http_error(exc.GitlabPipelineCancelError)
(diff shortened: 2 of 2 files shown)


gitlab/v3/objects.py
```diff
@@ -1496,6 +1496,18 @@ class ProjectFileManager(BaseManager):
     obj_cls = ProjectFile
 
 
+class ProjectPipelineSchedule(GitlabObject):
```
gpocentek: I don't think pipelines schedule are implemented in the v3 API, so this file should probably not be changed.

gitlab/v4/objects.py
```diff
@@ -1764,6 +1780,65 @@ def create(self, data, **kwargs):
...
+class ProjectPipelineScheduleVariable(SaveMixin, ObjectDeleteMixin, RESTObject):
+    _id_attr = 'key'
+
+
+class ProjectPipelineScheduleVariableManager(CRUDMixin, RESTManager):
+    _path = '/projects/%(project_id)s/pipeline_schedules/%(pipeline_schedule_id)s/variables'
+    _obj_cls = ProjectPipelineScheduleVariable
+    _from_parent_attrs = {'project_id': 'project_id',
+                          'pipeline_schedule_id' : 'id'}
+    _create_attrs = (('pipeline_schedule_id', 'key', 'value'), tuple())
```
gpocentek: You can remove this line since you redefine _create_attrs on the next line.

gitlab/v4/objects.py
```diff
@@ -1764,6 +1780,65 @@ def create(self, data, **kwargs):
...
+    _create_attrs = (('key', 'value'), tuple())
+
+    def list(self):
```
gpocentek: I'm not sure it's a good idea to implement this method. I'd like to avoid adding features that are not directly coming from the GitLab API. I'd rather remove this method.

gitlab/v4/objects.py
```diff
@@ -2035,6 +2110,7 @@ class Project(SaveMixin, ObjectDeleteMixin, RESTObject):
         ('notificationsettings', 'ProjectNotificationSettingsManager'),
         ('pipelines', 'ProjectPipelineManager'),
         ('protectedbranches', 'ProjectProtectedBranchManager'),
+        ('pipeline_schedules', 'ProjectPipelineSchedulesManager'),
```
gpocentek: Could you use `pipelineschedules` for consistency with the rest of the attribute names?

gitlab/v4/objects.py
```diff
@@ -1764,6 +1780,65 @@ def create(self, data, **kwargs):
...
+        else:
+            obj = self._parent.manager.get(self._parent.id)
+            for variable in obj._attrs['variables']:
+                schedule_variable = self._obj_cls(self, variable)
+                array.append(schedule_variable)
+
+        return array
+
+
+class ProjectPipelineSchedule(RESTObject):
```
gpocentek: GitLab supports updating and deleting these resources, so you need to add the `SaveMixin` and `ObjectDeleteMixin` here.

gitlab/v4/objects.py
```diff
@@ -1764,6 +1780,65 @@ def create(self, data, **kwargs):
...
+    _managers = (
+        ('variables', 'ProjectPipelineScheduleVariableManager'),
+    )
+
+
+class ProjectPipelineSchedulesManager(RetrieveMixin, CreateMixin, RESTManager):
```
gpocentek: You can use the `CRUDMixin` since all the methods are supported.

gitlab/v4/objects.py
```diff
@@ -1764,6 +1780,65 @@ def create(self, data, **kwargs):
...
+    _path = '/projects/%(project_id)s/pipeline_schedules'
+    _obj_cls = ProjectPipelineSchedule
+    _from_parent_attrs = {'project_id': 'id'}
+    _create_attrs = (('description', 'ref', 'cron'),
+                     ('cron_timezone', 'active'))
+
```
gpocentek: Could you define the `_update_attrs` attributes as well?

gitlab/v4/objects.py
```diff
@@ -1764,6 +1780,65 @@ def create(self, data, **kwargs):
...
+    def create(self, data, **kwargs):
```
gpocentek: I don't think you need this method (you're just calling the mixin method).

gitlab/v4/objects.py
gpocentek: And the class should be called `ProjectPipelineScheduleManager` (no plural).

gitlab/v4/objects.py
```diff
@@ -1706,7 +1706,23 @@ def raw(self, file_path, ref, streamed=False, action=None, chunk_size=1024,
         return utils.response_content(result, streamed, action, chunk_size)
 
 
+class ProjectPipelineJob(ProjectJob):
```
gpocentek: Pipeline jobs support has already be implemented, could you remove these items?

gpocentek:
@mlq if you don't mind I'll update your patch to fix the problems listed in the review.

mlq:
Hi!

Sorry that I haven't replied yet but I haven't found the time to introduce your suggested changes. If you can do that, it would be great! I just tried to build something that was working for me without knowing everything about the implementation!

Thank you!

gpocentek:
Thanks for your fast answer! I'll update your patch.

gpocentek:
Thanks for the initial work @mlq!

mlq:
Awesome, thank you for this project!

Issue #361, opened 2017-11-04
Add support for award emojis


gpocentek:
https://docs.gitlab.com/ce/api/award_emoji.html

PR #399, opened 2017-12-15
Introduce Award Emojis

zeddicus76:
Intoduce award_emoji access. Can be used similar to other merge request list items.

`award_emojis = mr.award_emojis.list()`

Should resolve https://github.com/python-gitlab/python-gitlab/issues/361


gpocentek:
@zeddicus76 thanks a lot! Unfortunately I have a patch almost ready which covers more objects (snippets and issues). I intended to finish and push it today.

PR #404, opened 2017-12-25
introduce a way to get all user's projects

### LINKED ISSUE #403: Support for getting list of user projects
API v4 has an endpoint `GET /users/:id/projects`, listing all of user's projects, however it is currently not supported.

akhmerov:
This is a proposal for accessing `GET /users/:id/projects`. Note that it isn't strictly backwards compatible because I renamed `gitlab.user.projects` into `gitlab.user.project` and used `gitlab.user.projects` for the new endpoint.

Closes #403

### DIFF
--- a/gitlab/v4/objects.py
+++ b/gitlab/v4/objects.py
@@ -181,6 +181,12 @@ class UserProject(RESTObject):
     pass
 
 
+class UserProjectsManager(GetFromListMixin, RESTManager):
+    _path = '/users/%(user_id)s/projects'
+    _obj_cls = UserProject
+    _from_parent_attrs = {'user_id': 'id'}
+
+
 class UserProjectManager(CreateMixin, RESTManager):
     _path = '/projects/user/%(user_id)s'
     _obj_cls = UserProject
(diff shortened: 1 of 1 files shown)


gpocentek:
Thanks for tackling this @akhmerov 

I don't think this is the best way to implement this feature though. I believe that you could use the existing ``UserProject`` and ``UserProjectManager`` classes to support a ``create`` method. This would require to write a custom method instead of using the mixin, because of different URLs in the GitLab API.

Do you want to have a look at this?

akhmerov:
I agree, my PR is mostly to illustrate what is missing, a custom class is a better solution.

Unfortunately after looking the codebase and not quite understanding what goes where, I believe it will take me too long time to implement a custom class adhering to the API, sorry. That is unless you have some specific pointers that I could straightforwardly adapt and adjust.

gpocentek:
@akhmerov thank you for your answer

No problem, I will take care of the implementation of the listing. I didn't want to take other this change before making sure it was OK with you.

Thanks again!

llincolnatal:
Hi, has this feature ever been implemented? Thanks

max-wittig:
Yes this has been there since a few years. I guess, we're missing documentation:

```python
gl.users.get(<id>).projects.list()
```

max-wittig:
I guess, we have some documentation in the projects api section about it. There is an example:

```python
user_project = alice.projects.create({'name': 'project'})
user_projects = alice.projects.list()
```

llincolnatal:
Thank you very much Max-wittig

matusnovak:
This is incorrect. Creating a project via `alice.projects.create({})` always fails with 403 Forbidden. To correctly create a new project at user level you must use: `POST /projects/user/%(user_id)` and not `/users/%(user_id)s/projects`. See https://docs.gitlab.com/ee/api/projects.html#create-project-for-user

PR #406, opened 2017-12-26
Allow per_page to be used with generators.

### LINKED ISSUE #405: Pagination generator with non-default per_page
Currently if you request `xxx.list(as_list=False, per_page=100) you end up getting a list instead of the generator.

I think the logic in in [http_list](https://github.com/python-gitlab/python-gitlab/blob/7efbc30b9d8cf8ea856b68ab85b9cd2340121358/gitlab/__init__.py#L715) needs some rework.
Perhaps just removing the `'per_page' in kwargs` condition [from here](https://github.com/python-gitlab/python-gitlab/blob/7efbc30b9d8cf8ea856b68ab85b9cd2340121358/gitlab/__init__.py#L746)

I should be able to specify something like...

```python
my_generator = gl.projects.list(per_page=100, as_list=False)
```

... and still get a generator.  Then I can hook it up to a progress bar like:

```python
my_generator = gl.projects.list(per_page=100, as_list=False, membership=True)
with click.progressbar(my_generator, length=my_generator.total, label='Projects') as bar:
    for project in bar:
        pass
```

The above code generates the following error:

```
AttributeError: 'list' object has no attribute 'total'
```

Today if I want to use the generator function I have to leave the `per_page` off and get them only 20 at a time; this makes it much slower.

ericfrederich:
Fixes #405

### DIFF
--- a/gitlab/__init__.py
+++ b/gitlab/__init__.py
@@ -743,7 +743,7 @@ class Gitlab(object):
         if get_all is True:
             return list(GitlabList(self, url, query_data, **kwargs))
 
-        if 'page' in kwargs or 'per_page' in kwargs or as_list is True:
+        if 'page' in kwargs or as_list is True:
             # pagination requested, we return a list
             return list(GitlabList(self, url, query_data, get_next=False,
                                    **kwargs))


gpocentek:
@ericfrederich thanks!

PR #408, opened 2018-01-05
Adding the supported version badge

### LINKED ISSUE #402: Document supported Python versions in README, and PyPi
When I am considering using a new Python library, perhaps the most important information I need is to be able to tell what versions of Python the library supports.

Typically I look for:

1. I look for PyPi badges at the top of the README.MD file. ([for an example, see pyhocon](https://github.com/chimpler/pyhocon))
2. A list of supported version in the README.MD more generally.
3. Version categories from the setup.py displayed in the PyPi page. ([for an example, see pyhocon](https://pypi.python.org/pypi/pyhocon/))

python-gitlab doesn't seem to have documentation describing this.

If you came up with a list, I would be willing to do the legwork to create a PR for this documentation fix.

(I realize that it mentions it in [the docs](https://python-gitlab.readthedocs.io/en/stable/install.html), but I might not even make it there if it's not clear.)

movermeyer:
This is probably enough. Thanks for your help. Closes #402

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -7,6 +7,9 @@
 .. image:: https://readthedocs.org/projects/python-gitlab/badge/?version=latest
    :target: https://python-gitlab.readthedocs.org/en/latest/?badge=latest
 
+.. image:: https://img.shields.io/pypi/pyversions/python-gitlab.svg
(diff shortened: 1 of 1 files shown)


gpocentek:
Thanks for both README updates!

PR #413, opened 2018-01-10
Add manager for jobs within a pipeline.

kw217:
This PR adds a manager for jobs within a pipeline, so you can say `gl.projects.get(prid).pipelines.get(plid).jobs.list()` to list all of the jobs within a single pipeline (in GitLab v4).

Please let me know if there are any tests I should add. The existing UTs still pass.

### DIFF
--- a/gitlab/v4/objects.py
+++ b/gitlab/v4/objects.py
@@ -1942,9 +1942,13 @@ class ProjectPipelineManager(RetrieveMixin, CreateMixin, RESTManager):
         return CreateMixin.create(self, data, path=path, **kwargs)
 
 
-class ProjectPipelineJobManager(RetrieveMixin, RESTManager):
+class ProjectPipelineJob(ProjectJob):
+    pass
+
+
+class ProjectPipelineJobManager(GetFromListMixin, RESTManager):
     _path = '/projects/%(project_id)s/pipelines/%(pipeline_id)s/jobs'
-    _obj_cls = ProjectJob
+    _obj_cls = ProjectPipelineJob
     _from_parent_attrs = {'project_id': 'project_id', 'pipeline_id': 'id'}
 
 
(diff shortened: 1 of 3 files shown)


gitlab/v4/objects.py
```diff
@@ -1940,6 +1942,12 @@ def create(self, data, **kwargs):
         return CreateMixin.create(self, data, path=path, **kwargs)
 
 
+class ProjectPipelineJobManager(RetrieveMixin, RESTManager):
```
gpocentek: Could you use `GetWithoutIdMixin` instead of `RetrieveMixin`?

It looks like getting a single job is not possible with the GitLab API, but we can simulate this feature by looping through the list.

gitlab/v4/objects.py
```diff
@@ -1940,6 +1942,12 @@ def create(self, data, **kwargs):
...
+    _path = '/projects/%(project_id)s/pipelines/%(pipeline_id)s/jobs'
+    _obj_cls = ProjectJob
```
gpocentek: I'd rather use a new `ProjectPipelineJob` instead of `ProjectJob`. This is choice I made for other objects and managers, so I'd rather stick to this logic.

kw217:
Thanks @gpocentek - I've added some docs, distinguished the classes as you requested (I just made an `pass` subclass of the existing class - I hope that's what you wanted), and I've added `get` support via `GetFromListMixin`.

In principle we could implement `get` more efficiently by just calling `ProjectJobManager.get` and checking the result is in the same pipeline, but there wasn't a mixin to do this already and I wasn't sure how to get hold of the `ProjectJobManager` so I followed your suggestion instead.

Thanks.

kw217:
Not sure why the Python 2.7 build failed, but it looks like a connection error pulling one of the libraries, not related to my change :-(

gpocentek:
@kw217 Thanks!

kw217:
Great, thank you!

kw217:
Hi @gpocentek - when are you likely to spin a new release containing this enhancement? I'd like to use it in a tool I'm writing, and it will be simpler to point people at PyPI than getting them to download master. Thanks.

gpocentek:
Hi @kw217 

I'd like to fix a few bugs and merge #398 before doing a release. I hope I'll have this ready in a couple weeks.

kw217:
Super - that would be great. Thanks!

PR #419, opened 2018-01-29
Simplify the example for streamed artifacts

tardyp:
Going through an object adds a lot of complication.
Adding example for unzipping on the fly

### DIFF
--- a/docs/gl_objects/builds.py
+++ b/docs/gl_objects/builds.py
@@ -72,7 +72,7 @@ project.jobs.get(job_id)  # v4
 build_or_job.artifacts()
 # end artifacts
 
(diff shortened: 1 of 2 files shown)


tardyp:
restarting ci

gpocentek:
Thanks again @tardyp  :)

Your example is clearly simpler, but I'd rather keep the class-based example as well, as a way to illustrate a bit more complex setups. Could you update your patch?

tardyp:
@gpocentek here you go

gpocentek:
@tardyp Thanks for the fast update! I'll let the CI do its job then merge.

PR #420, opened 2018-01-29
make trigger_pipeline return the pipeline

tardyp:
Trigger_pipeline returns nothing, which makes it difficult to track the pipeline being trigger.

Next PR will be about updating a pipeline object to get latest status (not sure yet the best way to do it)


gpocentek:
Thanks @tardyp 

The change looks good, could you fix fix the pep8 issue? The other failing test should pass.

PR #426, opened 2018-02-05
introduce RefreshMixin

tardyp:
RefreshMixin allows to update a REST object so that you can poll on it.
This is mostly useful for pipelines and jobs, but could be set on most of other objects, with unknown usecases.

### DIFF
--- a/docs/gl_objects/builds.py
+++ b/docs/gl_objects/builds.py
@@ -44,6 +44,22 @@ project.triggers.delete(trigger_token)
 trigger.delete()
 # end trigger delete
 
+# pipeline trigger
+def get_or_create_trigger(project):
+    trigger_decription = 'my_trigger_id'
+    for t in project.triggers.list():
+        if t.description == trigger_decription:
+            return t
+    return project.triggers.create({'description': trigger_decription})
+
+trigger = get_or_create_trigger(project)
+pipeline = project.trigger_pipeline('master', trigger.token, variables={"DEPLOY_ZONE": "us-west1"})
+while pipeline.finished_at is None:
+    pipeline.refresh()
+    os.sleep(1)
+
+# end pipeline trigger
+
 # list
 builds = project.builds.list()  # v3
 jobs = project.jobs.list()  # v4
--- a/docs/gl_objects/builds.rst
+++ b/docs/gl_objects/builds.rst
@@ -102,6 +102,12 @@ Remove a trigger:
    :start-after: # trigger delete
    :end-before: # end trigger delete
 
+Full example with wait for finish:
+
+.. literalinclude:: builds.py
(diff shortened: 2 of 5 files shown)


docs/gl_objects/projects.py
```diff
@@ -333,6 +333,22 @@
 pipeline = project.pipelines.create({'ref': 'master'})
 # end pipeline create
 
+# pipeline trigger
+def get_or_create_trigger(project):
```
tardyp: I think this could be  a nice improvement to add this get_or_create to trigger manager

docs/gl_objects/projects.py
```diff
@@ -333,6 +333,22 @@
...
+    trigger_decription = 'my_trigger_id'
+    for t in project.triggers.list():
+        if t.description == trigger_decription:
+            return t
+    return project.triggers.create({'description': trigger_decription})
+
+trigger = get_or_create_trigger(project)
+pipeline = project.trigger_pipeline('master', trigger.token, variables={"DEPLOY_ZONE": "us-west1"})
+while pipeline.finished_at is None:
```
tardyp: also, we could add a helper like ```pipeline.wait_for_finish(poll_freq=1)```

tardyp:
@gpocentek could you please give some feedback?

gpocentek:
Hi @tardyp 

Sorry for my late answer.

If I understand correctly the goal of the ReadMixin is to refresh the attributes an existing object. I like the idea but I wonder if adding a mixin is really the best option. What do you think of adding a `read` (or `refresh`?) method the the existing GetMixin? This way all the objects that you can GET could have the possibility of being refreshed.

Does that sound good to you?

Thanks for your work :)

tardyp:
@gpocentek I like the refresh name method better than read. Thanks for the suggestion.

I think that the ReadMixin is only for managers. Here we want to add this mixin to restobjects.
We can add this method directly to RESTObject, without need for a new mixin.

Let me know your though, I'll update the PR after we get an agreement.

gpocentek:
@tardyp you're right, using GetMixin doesn't make sense.

I don't think it will be possible to refresh all the objects, so I guess providing a new RefreshMixin with a refresh method is the way to go.

Thank you for the update.

tardyp:
Hi @gpocentek I updated my code. I believe the e2e tests are still a little bit flaky.

gpocentek:
Hi @tardyp 

The tests have been a problem for a little while now, but I didn't have time to try to understand why yet.

Do you mind removing the change on the test script? I don't think it will really help, and I'd prefer to have this kind of change in a separate path.

Apart from that the patch looks good, thanks!

tardyp:
@gpocentek done

gpocentek:
Thanks!

PR #450, opened 2018-03-05
wait for gitlab to finish starting up

larsks:
this modifies build_test_env.sh so that it will wait for a successful
response from the /version API endpoint before exiting.

this implementation will wait indefinitely; if you would prefer, I can add a
retry counter and exit with an error after some number of retries.

### DIFF
--- a/tools/build_test_env.sh
+++ b/tools/build_test_env.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 # Copyright (C) 2016 Gauvain Pocentek <gauvain@pocentek.net>
 #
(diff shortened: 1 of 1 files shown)


larsks:
Incidentally, this explicitly modifies the initial `/bin/sh` to `/bin/bash` since (a) this solution is using a bash-ism and (b) the rest of this script extensively uses bash-isms.  It would fail anyplace that `/bin/sh` is not actually bash (I think there are still distributions like that?).

gpocentek:
I think setting a retry counter to fail at some point would be great, something like 120 seconds (no real reason for this number).

Switching to bash does make a lot of sense. Thanks!

PR #484, opened 2018-04-13
Change method for getting content of snippet

Matusf:
There is an [example](http://python-gitlab.readthedocs.io/en/stable/gl_objects/snippets.html#examples) in snippet section of the docs (get content) that raises AttributeError. The `raw` method was supported in APIv3. Maybe the example should use APIv4 method.


gpocentek:
Thanks for this PR @Matusf 

Could you update your change to have both v3 and v4 examples? v3 is still used so the documentation should still exist.

PR #488, opened 2018-04-18
feat: obey the rate limit

max-wittig:
This PR enables python-gitlab to wait if the GitLab rate-limit is hit.
It waits as long as GitLab specifies via the [Retry-After header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After)

GitLab lacks documentation about this feature, but rate-limits can be set in the Admin UI (see screenshot)
and via the settings API.
image

```json
{
 "throttle_authenticated_api_enabled": false,
 "throttle_authenticated_api_requests_per_period": 2,
 "throttle_authenticated_api_period_in_seconds": 10,
}
```

It can be disabled using the `obey_rate_limit=False` parameter

### DIFF
--- a/docs/api-usage.rst
+++ b/docs/api-usage.rst
@@ -326,3 +326,26 @@ The following sample illustrates how to use a client-side certificate:
 
 Reference:
 http://docs.python-requests.org/en/master/user/advanced/#client-side-certificates
+
+Rate limits
+-----------
+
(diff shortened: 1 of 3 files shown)


gpocentek:
Hi @max-wittig 

Thanks for this PR, it looks good and provides a very nice feature!

PR #505, opened 2018-05-21
add per_page config option

### LINKED ISSUE #503: Allow to set a global per_page in the configuration
This is a feature request.

I would like to be able to set per_page globally in the configuration file instead of setting it for each request.

## Expected Behavior
Setting per_page in http://python-gitlab.readthedocs.io/en/stable/cli.html#configuration would set per page for all request with the configured gitlab.Gitlab instance

## Actual Behavior
per_page must be given everytime in ListMixin.list.

## Specifications

The setting could then be defined in:
  - the configuration file
  - gitlab.Gitlab constructor
  - ListMixin.list

  - python-gitlab version: python-gitlab==1.4.0
  - Gitlab server version (or gitlab.com): 9.4.5-ee

jouve:
Close #503 .

### DIFF
--- a/gitlab/tests/test_config.py
+++ b/gitlab/tests/test_config.py
@@ -45,6 +45,7 @@ timeout = 10
 url = https://three.url
 private_token = MNOPQR
 ssl_verify = /path/to/CA/bundle.crt
+per_page = 50
 
 [four]
 url = https://four.url
@@ -66,6 +67,11 @@ private_token = ABCDEF
 
 [three]
 meh = hem
+
+[four]
+url = http://four.url
+private_token = ABCDEF
+per_page = 200
 """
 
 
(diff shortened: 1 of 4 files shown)


gitlab/tests/test_config.py
```diff
@@ -87,13 +93,19 @@ def test_invalid_id(self, m_open):
     @mock.patch('six.moves.builtins.open')
     def test_invalid_data(self, m_open):
         fd = six.StringIO(missing_attr_config)
-        fd.close = mock.Mock(return_value=None)
+        fd.close = mock.Mock(return_value=None,
+                             side_effect=lambda: fd.seek(0))
         m_open.return_value = fd
         config.GitlabConfigParser('one')
+        config.GitlabConfigParser('one')
```
gpocentek: I don't think this duplicated line is useful :) Could you remove it?

gpocentek:
LGTM, can you just drop the extra line (see comment). Thanks!

gitlab/tests/test_config.py
jouve: I added this to test the change line#97.

Without the fd.seek(0), all calls to config.GitlabConfigParser fail except the 1st one:

``` python
# fd.tell() = 0
config.GitlabConfigParser('one')  # reads fd
# fd.tell() = end of file
config.GitlabConfigParser('one')  # reads nothing because fd is at end of file, so it fail
```

gpocentek: OK, thanks for the explanation.

PR #507, opened 2018-05-23
Add support for Project badges

Miouge1:
Add support for project badges https://docs.gitlab.com/ee/api/project_badges.html

### DIFF
--- a/gitlab/v4/objects.py
+++ b/gitlab/v4/objects.py
@@ -1250,6 +1250,18 @@ class ProjectKeyManager(CRUDMixin, RESTManager):
         self.gitlab.http_post(path, **kwargs)
 
 
+class ProjectBadge(SaveMixin, ObjectDeleteMixin, RESTObject):
+    pass
+
+
+class ProjectBadgeManager(CRUDMixin, RESTManager):
+    _path = '/projects/%(project_id)s/badges'
+    _obj_cls = ProjectBadge
+    _from_parent_attrs = {'project_id': 'id'}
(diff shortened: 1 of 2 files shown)


gitlab/v4/objects.py
```diff
@@ -1250,6 +1250,17 @@ def enable(self, key_id, **kwargs):
...
 
+class ProjectBadge(SaveMixin, ObjectDeleteMixin, RESTObject):
+    pass
+
+
+class ProjectBadgeManager(CRUDMixin, RESTManager):
+    _path = '/projects/%(project_id)s/badges'
+    _obj_cls = ProjectBadge
+    _from_parent_attrs = {'project_id': 'id'}
+    _create_attrs = (('link_url', 'image_url'), tuple())
```
gpocentek: It is also possible to update the badge objects, so could you define the `_update_attrs` variable as well?

gitlab/v4/objects.py
```diff
@@ -2505,6 +2516,7 @@ class Project(SaveMixin, ObjectDeleteMixin, RESTObject):
         ('triggers', 'ProjectTriggerManager'),
         ('variables', 'ProjectVariableManager'),
         ('wikis', 'ProjectWikiManager'),
+        ('badges', 'ProjectBadgeManager'),
```
gpocentek: Could you move this line to keep the list sorted?

Miouge1:
@gpocentek thanks for the feedback. Can you review the latest commit, it should add tests, the update_attrs etc... ?

gpocentek:
Thanks!

PR #514, opened 2018-06-06
make as_list=False work for all=True queries

jouve:


gpocentek:
I've not done a proper review/test yet, but I have one small request: could you move the stderr-related code in tests to another merge request?

Thanks!

gpocentek:
Looks good, and thanks for the tests improvements!

PR #520, opened 2018-06-07
 Add project push rules configuration 

esabouraud:
This PR introduces support for [project push rules](https://docs.gitlab.com/ee/api/projects.html#push-rules) configuration.

### DIFF
--- a/docs/gl_objects/projects.rst
+++ b/docs/gl_objects/projects.rst
@@ -636,11 +636,19 @@ Reference
 Examples
 ---------
 
-Get project push rules::
+Create project push rules (at least one rule is necessary)::
+
+    project.pushrules.create({'deny_delete_tag': True})
+
+Get project push rules (returns None is there are no push rules)::
 
     pr = project.pushrules.get()
 
 Edit project push rules::
 
     pr.branch_name_regex = '^(master|develop|support-\d+|release-\d+\..+|hotfix-.+|feature-.+)$'
-	pr.save()
+    pr.save()
+
+Delete project push rules::
+
+	pr.delete()
--- a/docs/gl_objects/projects.rst
+++ b/docs/gl_objects/projects.rst
@@ -618,3 +618,29 @@ markdown to reference the uploaded file::
     issue.notes.create({
         "body": "See the [attached file]({})".format(uploaded_file["url"])
     })
+
+Project push rules
+==================
+
+Reference
+---------
+
+* v4 API:
+
+  + :class:`gitlab.v4.objects.ProjectPushRules`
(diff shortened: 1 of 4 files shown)


gpocentek:
Hi @esabouraud 

I don't really understand why you removed the CreateMixin, it looks like creation is possible (from the docs and the code). The only thing that doesn't seem possible is to get a single rule. Did I miss something?

esabouraud:
Hi @gpocentek 

At first glance the `/projects/:id/push_rule` API endpoint looked like a perfect match for CRUDMixin. However this approach [broke the tests](https://travis-ci.org/esabouraud/python-gitlab/jobs/389357816). My understanding is that CRUDMixin and GetWithoutIdMixin (with `_id_attr = None`) do not play nice with each other. I could not find this exact pairing anywhere else in the code and did not quite know how to fix it.
```
======================================================================
FAIL: gitlab.tests.test_cli.TestV4CLI.test_parser
tags: worker-1
----------------------------------------------------------------------
  File "/home/travis/build/esabouraud/python-gitlab/gitlab/tests/test_cli.py", line 96, in test_parser
    parser = cli._get_parser(gitlab.v4.cli)
  File "/home/travis/build/esabouraud/python-gitlab/gitlab/cli.py", line 115, in _get_parser
    return cli_module.extend_parser(parser)
  File "/home/travis/build/esabouraud/python-gitlab/gitlab/v4/cli.py", line 254, in extend_parser
    _populate_sub_parser_by_class(cls, object_subparsers)
  File "/home/travis/build/esabouraud/python-gitlab/gitlab/v4/cli.py", line 146, in _populate_sub_parser_by_class
    id_attr = cls._id_attr.replace('_', '-')
AttributeError: 'NoneType' object has no attribute 'replace'
```
As for the API itself, it's a bit unusual in that push rules are not referenced by id. Rather they are optional attributes that can be edited. The documentation itself is inconsistent as it sometimes refers to "_a_ push rule", and sometimes to "push rules".

gpocentek:
This endpoint is really weird, and indeed the mixins don't really work well together. The only solution I see to implement all the features, including create and delete, is to write class-specific methods instead of using the mixins. Mixins make most objects very simple to define, but sometimes duplicating a little bit of code is necessary. Could you try that?

Thanks for your work!

esabouraud:
OK I've implemented the create and delete methods of the endpoint, while keeping things fairly nice and tidy (I think). It was mostly a matter of allowing delete without an object id.

gpocentek:
Thanks @esabouraud !

esabouraud:
Thank you @gpocentek for your diligent work on this module.

PR #522, opened 2018-06-08
fix #521 change post_data default value to None

### LINKED ISSUE #521: http_request method of Gitlab class should accept None as the default value for param post_data
http_request method of Gitlab class (in the file \_\_init__.py) should accept None as the default value for param post_data.

For some caching servers, they reject HTTP GET request with body. If post_data's default value is {}, the request body of a GET request will get content b'{}' with Content-length header, which some caching server will just respond with an error. To maximize the support, I feel it suitable to change post_data's default value to None. Also it is not best practice in Python to use a mutable default value as an argument. 

I didn't have time to read all the code. So let me know if this change is appropriate.

beyondliu:


gitlab/__init__.py
```diff
@@ -362,7 +362,7 @@ def _build_url(self, path):
         else:
             return '%s%s' % (self._url, path)
 
-    def http_request(self, verb, path, query_data={}, post_data={},
+    def http_request(self, verb, path, query_data={}, post_data=None,
```
gpocentek: The default value of `post_data` for `http_post` and `http_put` are also an empty dict, I guess you should change the value to `None` for those methods as well (`http_request` is never called directly).

beyondliu: I didn't read the latest code carefully. The version I used is 1.3. But what was causing problem is GET request with a body (by default will be b'{}'), which get rejected by caching servers. PUT and POST requests with a non empty body should be alright. There might be better ways or better places to make the change.

gpocentek: Ah yes, sorry. So this looks good. Thanks!

PR #561, opened 2018-07-23
Add support for project transfers from the projects interface.

wrouesnel:
Adds missing transfer_project function to the Project object, which provides for an ergonomic way to move projects around.

See https://docs.gitlab.com/ee/api/projects.html#transfer-a-project-to-a-new-namespace


gpocentek:
Thanks @wrouesnel 

Could you fix the PEP8 test? Otherwise the change looks OK.

wrouesnel:
Fixed.

gpocentek:
Thanks!

PR #569, opened 2018-08-14
Minor typo "ou" vs. "or"

mattthias:
This change fixes a minor type in the table of possible values for options in the global section.


gpocentek:
Thank you!

PR #581, opened 2018-09-07
Add project protected tags management

esabouraud:
This PR introduces support for [protected tags](https://docs.gitlab.com/ce/api/protected_tags.html) management.

### DIFF
--- a/docs/gl_objects/projects.rst
+++ b/docs/gl_objects/projects.rst
@@ -657,3 +657,36 @@ Edit project push rules::
 Delete project push rules::
 
     pr.delete()
+
+Project protected tags
+==================
+
+Reference
+---------
+
+* v4 API:
+
+  + :class:`gitlab.v4.objects.ProjectProtectedTag`
+  + :class:`gitlab.v4.objects.ProjectProtectedTagManager`
+  + :attr:`gitlab.v4.objects.Project.protectedtags`
+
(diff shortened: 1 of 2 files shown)


esabouraud:
Hi @gpocentek, is there anything wrong with this PR ? Gitlab 11.3 is now released so this new API is available for general use.

gpocentek:
Hi @esabouraud

Sorry for the delay but I don't have a lot of time for reviews these days.

@max-wittig @hans-d do you think you could have a look?

docs/gl_objects/projects.rst
```diff
@@ -657,3 +657,36 @@ Edit project push rules::
...
+Examples
+---------
+
+Get a list of protected tags from a project::
+
+    protected_tags = project.protectedtags.list()
+
+Get a single protected tag or wildcard protected tag::
+
+    protected_tag = project.protectedtags.get('v*')
```
max-wittig: This actually doesn't work for me. Just throws 404, even though all my protected tags are called v3.xx and v2.xx

max-wittig: Actually it works. GitLab just implemented the protected tag feature in an unexpected way. It actually remembers the way (including wildcards how the tags where protected)

max-wittig:
@hans-d Do you want to take another look or can we merge?

hans-d:
@max-wittig if it's LGTM no need to specifically wait for me unless you have doubts...

PR #584, opened 2018-09-10
Exclude tests from package installation.

miesorozec:
Modify setup.py to exclude tests.py files from package installation.
At this moment after package installation you can see tests directory with many redundant tesy_*.py files.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -20,7 +20,7 @@ setup(name='python-gitlab',
       author_email='gauvain@pocentek.net',
       license='LGPLv3',
       url='https://github.com/python-gitlab/python-gitlab',
-      packages=find_packages(),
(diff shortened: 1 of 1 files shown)


miesorozec:
But is there any reasonable reason to install tests with package? I can't see any benefits here because tests are part of developing process and it should remain in that phase.  Users don't use tests. Also it uses disk space, it may be inrelevant in your case but let imagine tests with fixed huge json/xml files. If we can save space with not installing files which will be never used by the user, lets do it. And the fact that everyone is doing it, it does not mean that they are doing it right.

gpocentek:
I don't see any good reason to not have the tests available on the system either, especially when they are very small.

PR #599, opened 2018-09-29
add group members all

### LINKED ISSUE #589: /groups/:id/members/all not yet supported?

I want to use the api endpoint ```GET /groups/:id/members/all```, but that does not seem supported. ```GET /groups/:id/members/``` is (difference: /all includes all inheritance)

## Expected Behavior

Being able to use it, eg ```gitlab group-member-all list --group_id $id```

## Actual Behavior

not supported

## Specifications

  - python-gitlab version: 1.6.0

see also: https://docs.gitlab.com/ee/api/members.html#list-all-members-of-a-group-or-project-including-inherited-members

hans-d:
fixes #589: `gitlab group-member-all list --group-id $id`

### DIFF
--- a/gitlab/v4/objects.py
+++ b/gitlab/v4/objects.py
@@ -709,10 +709,12 @@ class GroupMemberManager(CRUDMixin, RESTManager):
 class GroupMemberAll(RESTObject):
     pass
 
+
 class GroupMemberAllManager(ListMixin, RESTManager):
     _path = '/groups/%(group_id)s/members/all'
     _obj_cls = GroupMemberAll
     _from_parent_attrs = {'group_id': 'id'}
+
 
 class GroupMergeRequest(RESTObject):
     pass
--- a/gitlab/v4/objects.py
+++ b/gitlab/v4/objects.py
@@ -706,6 +706,14 @@ class GroupMemberManager(CRUDMixin, RESTManager):
     _update_attrs = (('access_level', ), ('expires_at', ))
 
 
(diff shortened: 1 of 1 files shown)


max-wittig:
@hans-d Thanks for your contribution! Could you please also add some tests and add this endpoint to the documentation?

Thanks :smiley:

gitlab/v4/objects.py
```diff
@@ -706,6 +706,16 @@ class GroupMemberManager(CRUDMixin, RESTManager):
     _update_attrs = (('access_level', ), ('expires_at', ))
 
 
+class GroupMemberAll(RESTObject):
+    pass
+
+
+class GroupMemberAllManager(ListMixin, RESTManager):
```
gpocentek: I understand your goal here, but I think it would make more sense to add an `all()` method to the `GroupMemberManager` class. The `RunnerManager` class has the same logic. From a REST point of view `all` is an action applied to the resource at path `/groups/:id:/members` (GroupMember).

Does that make sense?

hans-d: Makes sense, hadn't seen that variation yet. Will try to change it like that soonish

gpocentek:
Hi @hans-d 

Is it OK with you if I take over this MR?

Thanks

hans-d:
Fine by me

TobiX:
This should probably be implemented in the same way for project members...

gpocentek:
I pushed #642 which will add support for the method for groups and projects.

Closing this one.

PR #605, opened 2018-10-05
fix(docker): use docker image with current sources

### LINKED ISSUE #602: Build docker image according to current source tree
Currently in the docker image, the package just gets installed from PyPI, which isn't the usual way to provide docker files in a source repository.

## Expected Behavior
Dockerfile get's build into wheel, which get's installed into multi-layer build

## Actual Behavior
Package installed from PyPI

max-wittig:
Fixes #602


Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+
+WORKDIR /opt/python-gitlab
+COPY . .
+RUN python setup.py bdist_wheel
+
+FROM python:alpine
+
+WORKDIR /opt/python-gitlab
+COPY --from=build /opt/python-gitlab/dist dist/
+RUN pip install $(find dist -name *.whl)
```
zoidbergwill: We could also delete the wheel here, and build the docker image with [`--squash`](https://docs.docker.com/engine/reference/commandline/build/#squash-an-images-layers---squash-experimental), to save a tiny bit of extra space.

Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM python:alpine AS build
```
zoidbergwill: Should we pin to a Python version?

README.rst
```diff
@@ -41,6 +41,27 @@ Install with pip
...
+
+How to build
+------------
+
+``docker build -t python-gitlab:TAG .``
+
+How to use
+----------
+
+``docker run -it --rm  -e GITLAB_PRIVATE_TOKEN=<your token> =/python-gitlab.cfg python-gitlab <command> ...``
```
zoidbergwill: Ghost ` =/python-gitlab.cfg` ?

Dockerfile
max-wittig: `--squash` still seems to be experimental, but otherwise good idea!

README.rst
max-wittig: I copied that from contrib/docker/README.rst and just glanced over it. Nice find.

Dockerfile
max-wittig: Might be better. Maybe 3.8 will break something. I don't think so, but best practice.

max-wittig:
@gpocentek Feel free to merge, if you approve :smiley:

PR #651, opened 2018-12-04
fix: docker entry point argument passing

roozbehf:
Fixes the problem of passing spaces in the arguments to the docker entrypoint.

Before this fix, there was virtually no way to pass spaces in arguments such as task description.


max-wittig:
@roozbehf Thanks for your contribution!

PR #669, opened 2018-12-21
Modify list method of UserProjectManager 

carleski:
...to use kwargs['user_id'] and self._parent.id

When running via CLI, as with `python-gitlab user-projects list --user-id 1`, the `self._parent.id` variable does not exist. However, the user id is passed through to the function as `kwargs['user_id']`. This changes adds an if/else block to check to see if `self._parent.id` exists, and then use the appropriate variable.

### DIFF
--- a/gitlab/v4/objects.py
+++ b/gitlab/v4/objects.py
@@ -236,8 +236,11 @@ class UserProjectManager(ListMixin, CreateMixin, RESTManager):
             GitlabAuthenticationError: If authentication is not correct
             GitlabListError: If the server cannot perform the request
         """
+        if self._parent is not None:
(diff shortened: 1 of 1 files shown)


gpocentek:
Hi @carleski 

Thanks for working on this!

You patch works but I think it doesn't address the root problem. The user_id attribute should be discovered automatically.

max-wittig:
@gpocentek What did you mean by automatically?

evgkrsk:
Works for me: this patch let me list user's projects using symbolic user name (as well as numeric user's id).

Why it doesnt merged yet ?

max-wittig:
Merged in https://github.com/python-gitlab/python-gitlab/pull/886

Issue #632, opened 2018-11-13
Retry-After header may be missing: need default timeout


wvxvw:
## Description of the problem, including code/CLI snippet
The `retry-after` header is not always present, in which case, the client crashes.

## Expected Behavior
When there isn't a `retry-after` header, the client uses a default retry timeout.

## Actual Behavior
When there isn't a `retry-after` header, the client crashes

## Specifications

  - python-gitlab version: 1.6.0
  - API version you are using (v3/v4): Whatever they use on gitlab.com
  - Gitlab server version (or gitlab.com): gitlab.com

max-wittig:
@wvxvw You can temporary disable the `obey_rate_limit` feature in the meantime (which actually then causes a 429 crash, not sure if helpful)

I will fix the feature, but I would vote against implementing a default timeout, as no one knows what the right value should be there. Could be different for every server.

Instead the feature should just be disabled, if there is no retry-after header and cause the 429 error, which also can be caught.

wvxvw:
@max-wittig Default may come from configuration, or as an argument to the client etc.  It doesn't have to be non-configurable.

Another approach is to do some sort of "smart" back-off, where you increase the timeout every time you get "too many requests".

There are many ways to make this more useful. Raising an error in this case will only delegate the responsibility for implementation to the library user, but will not make any progress towards solving it.

max-wittig:
@wvxvw good idea and you're right. The library user shouldn't have to deal with this stuff. It should just be possible to disable it. Thanks for your input.

PR #678, opened 2019-01-09
Fix missing "Retry-After" header and fix integration tests

srikanthchelluri:
This PR does two things: a) implements an exponential backoff in the case that we are throttled and "Retry-After" isn't in the response header (this currently causes a `KeyError`), b) fix an integration test because we were decoding something that was already a string.

Closes https://github.com/python-gitlab/python-gitlab/issues/632.

### DIFF
--- a/docs/api-usage.rst
+++ b/docs/api-usage.rst
@@ -299,7 +299,9 @@ Rate limits
 
 python-gitlab obeys the rate limit of the GitLab server by default.  On
 receiving a 429 response (Too Many Requests), python-gitlab sleeps for the
-amount of time in the Retry-After header that GitLab sends back.
+amount of time in the Retry-After header that GitLab sends back.  If GitLab
+does not return a response with the Retry-After header, python-gitlab will
+perform an exponential backoff.
 
 If you don't want to wait, you can disable the rate-limiting feature, by
 supplying the ``obey_rate_limit`` argument.
@@ -312,6 +314,18 @@ supplying the ``obey_rate_limit`` argument.
    gl = gitlab.gitlab(url, token, api_version=4)
    gl.projects.list(all=True, obey_rate_limit=False)
 
+If you do not disable the rate-limiting feature, you can supply a custom value
+for ``max_retries``; by default, this is set to 10. To retry without bound when
+throttled, you can set this parameter to -1. This parameter is ignored if
+``obey_rate_limit`` is set to ``False``.
+
+.. code-block:: python
+
+   import gitlab
+   import requests
+
+   gl = gitlab.gitlab(url, token, api_version=4)
+   gl.projects.list(all=True, max_retries=12)
 
 .. warning::
 
--- a/gitlab/__init__.py
+++ b/gitlab/__init__.py
@@ -477,6 +477,10 @@ class Gitlab(object):
         # obey the rate limit by default
         obey_rate_limit = kwargs.get("obey_rate_limit", True)
 
+        # set max_retries to 10 by default, disable by setting it to -1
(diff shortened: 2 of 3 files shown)


max-wittig:
Thanks for this contribution! Looks pretty good. Any way to easily test this? (against a public GitLab instance without `Retry-After` header)

srikanthchelluri:
Hm, this is a bit difficult to test. But here's how I validated the code path - let me know what you think.

1. Created a virtual environment and built the library with the following code:
```Python
while True:
    # result = self.session.send(prepped, timeout=timeout, **settings)
    result = requests.get('https://httpstat.us/429')

    self._check_redirects(result)

        return result

    if 429 == result.status_code and obey_rate_limit:
        if max_retries == -1 or cur_retries < max_retries:
            wait_time = 2 ** cur_retries * 0.1
            if "Retry-AfterXYZ" in result.headers:
                wait_time = int(result.headers["Retry-After"])
            cur_retries += 1
            print('DEBUG: backing off', str(wait_time))
            time.sleep(wait_time)
            continue
```

Note that https://httpstat.us returns a response with the `Retry-After` header set to `3600`. I first validated that we backed off 3600 seconds; then I changed the header we were checking for (`Retry-AfterXYZ` as shown above) and validated that we were backing off exponentially (see below).

2. Wrote this small script:
```
from getpass import getpass
import gitlab

private_token = getpass('GitLab API Token: ')
gl = gitlab.Gitlab('https://gitlab.com/', private_token=private_token)
project = gl.projects.get(<some-project-id-we-own>)

project.members.list(all=True)
```

3. Validated the exponential backoff:
```
(env) ➜  python-gitlab python throttle.py 
GitLab API Token: 
DEBUG: backing off 0.1
DEBUG: backing off 0.2
DEBUG: backing off 0.4
DEBUG: backing off 0.8
DEBUG: backing off 1.6
DEBUG: backing off 3.2
DEBUG: backing off 6.4
DEBUG: backing off 12.8
DEBUG: backing off 25.6
DEBUG: backing off 51.2
  File "/Users/srikanth.chelluri/repo/python-gitlab/gitlab/exceptions.py", line 251, in wrapped_f
    return f(*args, **kwargs)
  File "/Users/srikanth.chelluri/repo/python-gitlab/gitlab/mixins.py", line 49, in get
    server_data = self.gitlab.http_get(path, **kwargs)
  File "/Users/srikanth.chelluri/repo/python-gitlab/gitlab/__init__.py", line 542, in http_get
    streamed=streamed, **kwargs)
  File "/Users/srikanth.chelluri/repo/python-gitlab/gitlab/__init__.py", line 520, in http_request
    response_body=result.content)
gitlab.exceptions.GitlabHttpError: 429: 429 Too Many Requests

During handling of the above exception, another exception occurred:

  File "throttle.py", line 10, in <module>
    project = gl.projects.get(8244461)
  File "/Users/srikanth.chelluri/repo/python-gitlab/gitlab/exceptions.py", line 253, in wrapped_f
    raise error(e.error_message, e.response_code, e.response_body)
gitlab.exceptions.GitlabGetError: 429: 429 Too Many Requests
```

Thoughts/feedback on this approach? Didn't try spinning up and configuring a GitLab instance without the header since we don't really need it to test the code path.

max-wittig:
@srikanthchelluri Works as expected. Many thanks for the contribution 👍 

Just a hint for future contributions: It's best practice to use a separate branch in your fork. Makes it easier for us to checkout the changes and easier for you to maintain your next contribution. Otherwise your master in your fork is polluted by another commit etc.

max-wittig:
ah really. Okay weird. Maybe I miss typed something and didn't see it. Not sure 😄

PR #701, opened 2019-02-18
Fix all kwarg behaviour

jpiron:
`all` kwarg is used to manage GitlabList generator behaviour.
However, as it is not pop'ed from kwargs, it is sent to Gitlab API.
Some endpoints such as [the project commits](https://docs.gitlab.com/ee/api/commits.html#list-repository-commits) one,
support a `all` attribute.
This means a call like `project.commits.list(all=True, ref_name='master')`
won't return all the master commits as one might expect but all the
repository's commits.
To prevent confusion, the same kwarg shouldn't be used for 2 distinct
purposes.
Moreover according to [the documentation](https://python-gitlab.readthedocs.io/en/stable/gl_objects/commits.html#examples),
the `all` project commits API endpoint attribute doesn't seem supported.

### DIFF
--- a/gitlab/__init__.py
+++ b/gitlab/__init__.py
@@ -590,7 +590,7 @@ class Gitlab(object):
         # In case we want to change the default behavior at some point
         as_list = True if as_list is None else as_list
 
-        get_all = kwargs.get('all', False)
+        get_all = kwargs.pop('all', False)
         url = self._build_url(path)
 
         if get_all is True:


max-wittig:
Makes sense. To use the `all` endpoint, you can supply it with `query_parameters` instead.

gitlab/__init__.py
```diff
@@ -590,7 +590,7 @@ def http_list(self, path, query_data={}, as_list=None, **kwargs):
         # In case we want to change the default behavior at some point
         as_list = True if as_list is None else as_list
 
-        get_all = kwargs.get('all', False)
+        get_all = kwargs.pop('all', False)
```
max-wittig: won't this cause an `KeyError`, if `all` is not used?

max-wittig:
@gpocentek We may need to update the documentation that `all` is not send to the GitLab API anymore by default. What do you think?

gpocentek:
@max-wittig yes, we need to add that info in the release notes

max-wittig:
@gpocentek Could you disable the travis requirement temporary to merge this?

PR #702, opened 2019-02-18
Implement __eq__ and __hash__ methods

jpiron:
To ease lists and sets manipulations.

### DIFF
--- a/gitlab/base.py
+++ b/gitlab/base.py
@@ -96,6 +96,16 @@ class RESTObject(object):
         else:
             return '<%s>' % self.__class__.__name__
 
+    def __eq__(self, other):
+        if self.get_id() and other.get_id():
+            return self.get_id() == other.get_id()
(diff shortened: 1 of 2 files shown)


max-wittig:
Thanks for your contribution. What use-case do you have for this? How can I test it?

jpiron:
My use case is to compute statistics. One of them being the number of commits not part of a merge request. 
So for a project, I retrieve all of its commits and then all the commits linked to an MR and do something like : 
```python
merge_requets_commits = XXX
commits_not_linked_to_mr = [commit for commit in project.commits.list(all=True, ref_name='master') if commit not in merge_requets_commits]
```
The `not in` part requires `__eq__` to be defined on  `RESTObject`.

For the `__hash__`, I often transform commits lists into sets to ensure uniqueness.

Sorry I forgot about the tests, I'll add some of them to test the changes.

Issue #717, opened 2019-02-27
CLI event list fails on None


xarx00:
## Description of the problem, including code/CLI snippet
`gitlab event list` fails in v4/cli.py#l.38 with `AttributeError: 'NoneType' object has no attribute 'lower'`.

## Specifications
  - python-gitlab version: 1.8.0
  - API version you are using (v3/v4): na
  - Gitlab server version (or gitlab.com): na

## Cause
The error is caused by the EventManager class in objects.py, because one of `_list_filters` is called "action". This name collides with the args attribute `action`, and one of them is probably overwritten. When renaming "action" in `_list_filters` to something else, `gitlab event list` continues further (and fails with another similar error).

xarx00:
The subsequent error mentioned above is 
```
  File "c:\program files\python27\lib\site-packages\gitlab\v4\cli.py", line 364, in display
    value = value.replace('\r', '').replace('\n', ' ')
AttributeError: 'NoneType' object has no attribute 'replace'
```
It is caused by that some events on my gilab server do not have "target_title" filled in.

### Fix
Fix of the subsequent (NOT the original error):
```
--- cli.py.orig	2019-02-24 17:37:53.000000000 +0100
+++ cli.py	2019-02-27 19:17:56.045767100 +0100
@@ -357,7 +358,7 @@
                 id = getattr(obj, obj._id_attr)
                 print('%s: %s' % (obj._id_attr.replace('_', '-'), id))
             if hasattr(obj, '_short_print_attr'):
-                value = getattr(obj, obj._short_print_attr)
+                value = getattr(obj, obj._short_print_attr) or 'None'
                 value = value.replace('\r', '').replace('\n', ' ')
                 # If the attribute is a note (ProjectCommitComment) then we do
                 # some modifications to fit everything on one line
```

xarx00:
### Fix
In order to fix the primary error of this issue, I had to rename the "action" attribute:
[fix.diff.zip](https://github.com/python-gitlab/python-gitlab/files/2912006/fix.diff.zip)

gpocentek:
Hi @xarx00 

Could you create a pull request with this change? Thanks.

PR #727, opened 2019-03-04
Fix for issue #717

xarx00:
I propose this fix for isuue #717. It fixes `event list`, which didn't worked when invoked from CLI (and probably neither when used from gitlab API).

### DIFF
--- a/gitlab/v4/cli.py
+++ b/gitlab/v4/cli.py
@@ -273,7 +273,7 @@ def extend_parser(parser):
 
         object_subparsers = object_group.add_subparsers(
             title='action',
-            dest='action', help="Action to execute.")
+            dest='whaction', help="Action to execute.")
         _populate_sub_parser_by_class(cls, object_subparsers)
         object_subparsers.required = True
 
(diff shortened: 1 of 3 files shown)


gitlab/v4/cli.py
```diff
@@ -357,7 +357,7 @@ def display_dict(d, padding):
                 id = getattr(obj, obj._id_attr)
                 print('%s: %s' % (obj._id_attr.replace('_', '-'), id))
             if hasattr(obj, '_short_print_attr'):
-                value = getattr(obj, obj._short_print_attr)
+                value = getattr(obj, obj._short_print_attr) or 'None'
```
max-wittig: shouldn't `None` events just be skipped in the CLI?

```
target_title: None

target_title: None

target_title: None

target_title: None

target_title: None

target_title: None

target_title: None

target_title: None

target_title: None
```

xarx00: Actually, I don't know what these events are. It seems to me better to display them then to hode them. Because I consider github as a tool that helps me to show precisely what's going on. I it hides the results, it may hide an important clue to a problem I'm solving.

But I don't know how other parts of github behave. It's up to you to decide what you want it to do.

gpocentek: The problem I see with this issue is that there is no way to have a usable output without displaying multiple attributes. For events that would be target_type and action_name for exemple

What we could do (to be discussed):

1. make the `display_dict` check for a `repr` method for the object. This method would return a key and a value to be displayed. We could use `__repr__` ou `__unicode__` for this maybe
2. if the method doesn't exist, fallback to the `_short_print_attr` attribute
3. if this attribute is None raise an error so we can fix the code and add the method from 1 (not sure about this step, it's not nice for end users).

Other idea: `_short_print_attr` could become a tuple, and we display multiple lines when needed

gpocentek:
@xarx00 could you make the commit messages a bit more meaningful? Something like:

```
fix(cli): Don't fail when the short print attr value is None

Fixes #717 
```

Thanks for your help on fixing the CLI :+1:

xarx00:
@gpocentek OK, I have thought that link to the defect is enough.

PR #728, opened 2019-03-05
fix(types): return _value when _value is not iterable

Drinkey:
- In gitlab/types.py, the `get_for_api` in ListAttribute will trigger an exception if self._value is None. The error like below:
`TypeError: can only join an iterable`
- Catch the exception and return self._value if there is TypeError.

### DIFF
--- a/gitlab/types.py
+++ b/gitlab/types.py
@@ -43,6 +43,7 @@ class ListAttribute(GitlabAttribute):
         except TypeError:
             return self._value
 
+
 class LowercaseStringAttribute(GitlabAttribute):
     def get_for_api(self):
         return str(self._value).lower()
--- a/gitlab/types.py
+++ b/gitlab/types.py
@@ -38,8 +38,10 @@ class ListAttribute(GitlabAttribute):
             self._value = [item.strip() for item in cli_value.split(',')]
 
     def get_for_api(self):
-        return ",".join(self._value)
-
+        try:
+            return ",".join(self._value)
+        except TypeError:
+            return self._value
 
 class LowercaseStringAttribute(GitlabAttribute):
     def get_for_api(self):


max-wittig:
@Drinkey Thanks for your contribution. In which case does the error occur currently? Could you provide an example code snippet?

Drinkey:
@max-wittig 

Here is an example

```python
def get_ms_issues(**kwargs):
    gl = gitlab.Gitlab(GITLAB_URL, GITLAB_TOKEN)
    group = gl.groups.get(8)
    return group.issues.list(**kwargs)

default_args = dict(milestone='3.16', labels=None)

print(get_ms_issues(**default_args))
```
When `labels` is `None`, `','.join(self._value)` will trigger `TypeError`.

gpocentek:
`labels` is supposed to be a list, so you shouldn't define it as `None`. Using an empty list should work.

I'd prefer to raise an exception in the method if the type is not correct, rather than returning a value on which we have no control.

gitlab/types.py
```diff
@@ -38,7 +38,10 @@ def set_from_cli(self, cli_value):
             self._value = [item.strip() for item in cli_value.split(',')]
 
     def get_for_api(self):
-        return ",".join(self._value)
+        try:
+            return ",".join(self._value)
+        except TypeError:
+            return self._value
```
gpocentek: If `self._value` is not a list nor None, then we end up returning unusable data. I'd prefer to have an exception raised is `self._value` is not a list/set.

max-wittig: I agree with @gpocentek

Drinkey: @gpocentek @max-wittig Thanks, guys. I agree with you, too. I should convert the labels to an empty list when it's `None`.

I close this one, and thanks for your time.

Issue #716, opened 2019-02-27
TypeError: %d format: a number is required, not str


xarx00:
## Description of the problem, including code/CLI snippet
For instance, `gitlab group search --id=31 --scope projects --search XYZ` ends with the error `TypeError: %d format: a number is required, not str`. Several other commands end up the same way.

## Expected Behavior
The command succeeds.

## Actual Behavior
The code fails in objects.py at line `path = '/groups/%d/search' % self.get_id()`

## Specifications

  - python-gitlab version:  1.8.0
  - API version you are using (v3/v4): na
  - Gitlab server version (or gitlab.com): na

## Fix:
Replace `%d` everywhere in objects.py with `%s`. Diff file attached:
[fix.diff.zip](https://github.com/python-gitlab/python-gitlab/files/2910834/fix.diff.zip)

nejch:
Closed in #729 last year. Just double-checked and there's no more `%d` in the repo 😁

PR #729, opened 2019-03-05
Fix for #716: %d replaced by %s

xarx00:
This is a proposed fix for #716. While the id-s are integers when gitlab is used as a library, they are strings when used from the CLI. 
Replacement of %d by %s should be safe.


max-wittig:
Thanks alot for your many contributions @xarx00. I will test and review them tomorrow.

Issue #718, opened 2019-02-27
CLI project repository-tree doesn't allow --recursive


xarx00:
## Description of the problem, including code/CLI snippet
`gitlab project repository-tree` doesn't allow the `--recursive` parameter. 
This issue is a follow up for #452 for CLI, where the problem was fixed for LIB.

## Specifications

  - python-gitlab version: 1.8.0

## Fix
The fix implements --recursive as a boolean parameter:
[fix.diff.zip](https://github.com/python-gitlab/python-gitlab/files/2912556/fix.diff.zip)

PR #731, opened 2019-03-05
fix for issue #718

xarx00:
Fix for issue #718.

### DIFF
--- a/gitlab/v4/cli.py
+++ b/gitlab/v4/cli.py
@@ -230,8 +230,7 @@ def _populate_sub_parser_by_class(cls, sub_parser):
              for x in required if x != cls._id_attr]
             [sub_parser_action.add_argument("--%s" % x.replace('_', '-'),
                                             required=False)
-             for x in optional if x != cls._id_attr 
-                                            and x not in ['recursive']]
+             for x in optional if x != cls._id_attr and x not in ['recursive']]
             [sub_parser_action.add_argument("--%s" % x, action='store_true',
                                             required=False)
              for x in optional if x in ['recursive']]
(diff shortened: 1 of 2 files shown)


gitlab/v4/cli.py
```diff
@@ -230,7 +230,10 @@ def _populate_sub_parser_by_class(cls, sub_parser):
              for x in required if x != cls._id_attr]
             [sub_parser_action.add_argument("--%s" % x.replace('_', '-'),
                                             required=False)
-             for x in optional if x != cls._id_attr]
+             for x in optional if x != cls._id_attr and x not in ['recursive']]
```
gpocentek: I see a couple of problems here:

* What happens if another object gets a `recursive` attribute that requires a value?
* You are dealing with a specific attribute. If we start adding exceptions it will become a nightmare to maintain very quickly.

I think the method should not be changed. All other booleans need an explicit value on the CLI.

xarx00: In `gitlab.v4.cli._populate_sub_parser_by_class`, there are handled various specific commands and arguments. Thie procedure expects that actions like `list, get, create, update, delete` have the same meaning for every gitlab resource, and similarly that arguments like `--page, --per-page, --all` have the same meaning everywhere. Hence:

- It seems to me safe to expect that `--recursive` is boolean everywhere. Actually, I think that all boolean attributes should be treated this way. However, I didn't want to do that in this bugfix.
- As I mentioned in the previous point, rules for handling specific attributes and actions are already handled here. Not only that, in `gitlab.mixins` there are handled specific classes from `gitlab.v4.objects`!

Nevertheless, This piece of code (that you pointed out) is not critical for the bugfix, it can be omitted, if you prefer that.

xarx00: Actually, for now it would be better to remove this piece of code, as there are many other attributes that are boolean but require an argument ('true').

However, if #733 (replacement of argparse with click) will be realized, I vote for changing these boolean attributes to regular attributes without parameter.

xarx00: Removed changes from cli.py, left changes in objects.py only.

PR #732, opened 2019-03-05
Re-enable command specific help messages

hakanf:
This makes sure that the global help message wont be printed instead of the command specific one unless we fail to read the configuration file

Currently if you try, for example:

  gitlab project list --help

you only get the global argument help message

### DIFF
--- a/gitlab/cli.py
+++ b/gitlab/cli.py
@@ -136,9 +136,6 @@ def main():
         exit(0)
 
     parser = _get_base_parser(add_help=False)
-    if "--help" in sys.argv or "-h" in sys.argv:
-        parser.print_help()
-        exit(0)
 
     # This first parsing step is used to find the gitlab config to use, and
     # load the propermodule (v3 or v4) accordingly. At that point we don't have
@@ -150,6 +147,9 @@ def main():
             options.config_file
         )
     except gitlab.config.ConfigError as e:
+        if "--help" in sys.argv or "-h" in sys.argv:
+            parser.print_help()
+            exit(0)
         sys.exit(e)
     cli_module = importlib.import_module('gitlab.v%s.cli' % config.api_version)
 
--- a/gitlab/cli.py
+++ b/gitlab/cli.py
@@ -133,7 +133,7 @@ def _parse_value(v):
 def main():
     if "--version" in sys.argv:
         print(gitlab.__version__)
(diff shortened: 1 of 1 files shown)


xarx00:
Good, it seems to work (within argparse limits).

This certainly fixes the main problem of the CLI help - that you weren't able to obtain a proper advice on available parameters. But when you use wrong arguments (e.g. `gitlab group list -x`), you still won't obtain a proper help (for `gitlab group list`) parameters, but only for `gitlab` parameters. And the help itself is very ugly and confusing.

Unfortunately, I think these problems are built into `argparse`. I was playing with `argparse` a little, but I was not able to make it work satisfactorily. I think that `argparse` is not suitable for handling two levels of subparsers properly, and neither such an number of subparsers (27 **rows** of subparsers listed in the usage).

I suggest to replace it with Click. I was playing with click too, and I was able to make it work very good for a script similar to gitlab.

hakanf:
Sure, I can see your point. And yes, Click seems like a nice fit...
Makes it a bit of a bigger project though. Is your intention that we move to Click instead of this quick patch?

hakanf:
What I meant to say is that I could help out with moving to Click if you want, just that perhaps we separate that out into another patch, so that we dont delay this one...

xarx00:
I am not a maintainer of the project, so I cannot decide about the move to Click. I have created issue #733 for this purpose.

max-wittig:
@xarx00 @hakanf What do you think, if we merge this one after testing and move to click later. Is someone willing to contribute that?

hakanf:
@max-wittig , sure, I can work with @xarx00 to move to click.

gitlab/cli.py
```diff
@@ -150,6 +147,9 @@ def main():
             options.config_file
         )
     except gitlab.config.ConfigError as e:
+        if "--help" in sys.argv or "-h" in sys.argv:
+            parser.print_help()
+            exit(0)
```
max-wittig: please use `sys.exit(0)` like below

xarx00:
> 
> 
> @xarx00 @hakanf What do you think, if we merge this one after testing and move to click later. Is someone willing to contribute that?

I definitly agree.

hakanf:
OK, done... (also fixed another location I found using exit instead of sys.exit)

Issue #734, opened 2019-03-06
Subgroups do not return Group objects


xarx00:
## Description of the problem, including code/CLI snippet
`gl.groups.list()` returns a list of `gitlab.v4.objects.Group` class instances. On the other hand,
`group.subgroups.list()` returns a list of `gitlab.v4.objects.GroupSubgroup` instances. They do not provide methods and properties of `Group` objects.

Similarly, `group.projects.list()` does not return instances of `gitlab.v4.objects.Project`; neither `group.mergerequests.list()` nor `project.mergerequests.list()` return instances of `gitlab.v4.objects.MergeRequest`, etc.

If this is not intentional (I don't see why it should be), I suggest fixing the classes of the returned objects.

## Specifications
  - python-gitlab version: 1.8.0

## Fix
As a fix, it would suffice to change the `_obj_cls` parameter in the appropriate Manager classes to the correct instance classes. For example, in `GroupProjectManager`, it would suffice to change
`_obj_cls = GroupProject` to `_obj_cls = Project`. Unfortunately, this solution breaks the gitlab CLI help for these commands.

Hence, it is currently better to leave `_obj_cls` untouched, and inherit the current instance class from the correct instance class. For example, change `class GroupProject(RESTObject):` to `class GroupProject(Project):`.

max-wittig:
I think this behavior should be expected, as we just follow the GitLab API specification. Thanks for working on this, but I don't think we should change this.

To get a group object from a subgroup, you simply have to request it.
```python
subgroups = gl.groups.get(1).subgroups.list()
group = gl.groups.get(subgroups[0].id)
```

PR #735, opened 2019-03-07
Fix for defect #734

xarx00:
In order to fix the defect, I had to:

- Change the base class for GroupProject, GroupSubgroup, GroupMergeRequest and ProjectMergeRequest classes.
- Rearrange the classes in order to satisfy the dependencies.
- Modify the path variable in MergeRequest methods, in order they work in various contexts, not only from projects. Also, I had to change the calls-name in calls to `@cli.register_custom_action`.

However, the same kind of refactoring should be done, in my opinion, for many more classes, like CodeSnippets, CustomAttributes, Events, Runners etc. Simply for all duplicate classes that differ only by the way they were obtained.
For list of others, see https://github.com/gitlabhq/gitlabhq/tree/master/doc/api, in particular the paragraph "Standalone resources".

My question is: should I continue in refactoring the other classes too?

### DIFF
--- a/gitlab/base.py
+++ b/gitlab/base.py
@@ -29,6 +29,7 @@ class RESTObject(object):
     without ID in the url.
     """
     _id_attr = 'id'
+    _path = None
 
     def __init__(self, manager, attrs):
         self.__dict__.update({
@@ -130,6 +131,10 @@ class RESTObject(object):
         if self._id_attr is None or not hasattr(self, self._id_attr):
             return None
         return getattr(self, self._id_attr)
+
+    def get_path(self):
+        """Returns the path of the resource."""
+        return '%s/%s' % (self._path or self.manager.path, self.get_id())
 
     @property
     def attributes(self):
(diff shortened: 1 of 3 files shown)


gpocentek:
We had the same of logic while ago, and this ended up being full of buggy side effects. This needs to be well tested before considering it. Integration tests are needed.

This also changes the existing API, and we lose the exact match with the GitLab API, where group MR are NOT the same objects as Project MR.

xarx00:
> @gpocentek: group MR are NOT the same objects as Project MR.

Is group MR really different from project MR?? According to the docs https://github.com/gitlabhq/gitlabhq/blob/master/doc/api/merge_requests.md it seems to me that they are the same. 

- `GET /groups/:id/merge_requests`  - Get all merge requests for this group and its subgroups.
- `GET /projects/:id/merge_requests` - Get all merge requests for this project.

And the returned merge requests look identical too.

> @gpocentek: This also changes the existing API

It seems to me that the API will be extended, not restricted. Can you point me to an example where the API will be changed?

> @gpocentek: and this ended up being full of buggy side effects

You're right, it turned out to be more complicated than I expected. However, I think I'm almost finished with the changes. I'm trying not to change the API, as mentioned above.

max-wittig:
I'm also not so sure about this patch. I think people are used to the fact that Groups and subgroups are a different object. That's the same way GitLab does it with their API.

Issue #722, opened 2019-03-03
Implement project releases API


gpocentek:
New API in Gitlab 11.7: https://docs.gitlab.com/ce/api/releases/index.html

max-wittig:
This was merged

PR #746, opened 2019-03-27
add project releases api

therealgambo:
This PR address #722 

Not entirely sure how to implement the update functionality, as it requires an additional argument in the http request on the end so I've left that out for now.


max-wittig:
Thanks for your contribution!

PR #768, opened 2019-05-01
docs: Add an example of trigger token usage

### LINKED ISSUE #752: Make Project.trigger_pipeline useable without general authentication

Gitlab offers [Trigger tokens](https://docs.gitlab.com/ee/ci/triggers/#trigger-token) to allow $things to authenticate against the API to only trigger a pipeline. This is currently not possible (or at least I don't see how) with python-gitlab because we first need to authenticate against the Gitlab server.

## Expected Behavior

```python
pipeline = gitlab.Project(id).trigger_pipeline('master', token, variables={"DEPLOY": "true"})
```

## Actual Behavior

```python
gl = gitlab.Gitlab('https://gitlab.example-org', private_token=private_token)
project = gl.projects.get(id)
project.trigger_pipeline('master', trigger_token, variables={"DEPLOY": 'true'})
```

The problem is, that I need to have a valid user account/API token with API privileges to get the project. After that I can use the trigger_token to trigger the pipeline which is kind of counterproductive.

## Specifications

  - python-gitlab version: 1.8.0
  - Gitlab server version (or gitlab.com): 11.8.6-ee

gpocentek:
Closes #752

### DIFF
--- a/docs/gl_objects/builds.rst
+++ b/docs/gl_objects/builds.rst
@@ -94,6 +94,16 @@ Full example with wait for finish::
         pipeline.refresh()
         time.sleep(1)
 
+You can trigger a pipeline using token authentication instead of user
+authentication. To do so create an anonymous Gitlab instance and use lazy
+objects to get the associated project::
+
+    gl = gitlab.Gitlab(URL)  # no authentication
+    project = gl.projects.get(project_id, lazy=True)  # no API call
+    project.trigger_pipeline('master', trigger_token)
+
+Reference: https://docs.gitlab.com/ee/ci/triggers/#trigger-token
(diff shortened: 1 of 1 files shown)


docs/gl_objects/builds.rst
```diff
@@ -94,6 +94,16 @@ Full example with wait for finish::
         pipeline.refresh()
         time.sleep(1)
 
+You can trigger a pipeline using a token authentication instead of user
```
max-wittig:
```diff
- You can trigger a pipeline using a token authentication instead of user
+ You can trigger a pipeline using token authentication instead of user
```

PR #773, opened 2019-05-09
chore(ci): use reliable ci system

max-wittig:
We currently still need to rely on Travis for the docker tests, but I will look into changing that.

Job succeeded on GitLab.com: https://gitlab.com/python-gitlab/python-gitlab/pipelines/60606601

I'm using the GitLab feature to report back to GitHub for my own open source project and it works pretty well: https://github.com/max-wittig/gitlab-languages/commits/master (certainly better than travis)

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,0 +1,96 @@
+image: python:3.7
+
+stages:
+  - lint
+  - build-test-image
+  - test
+  - deploy
+
+commitlint:
+  image: node:12
+  stage: lint
+  before_script:
+    - npm install -g @commitlint/cli @commitlint/config-conventional
+    - 'echo "module.exports = {extends: [\"@commitlint/config-conventional\"]}" > commitlint.config.js'
+  script:
+    - npx commitlint --from=origin/master
+  except:
+    - master
+
+#build_test_image:  # Currently hangs forever, because of GitLab Runner infrastructure issues
+#  stage: build-test-image
+#  image:
+#    name: gcr.io/kaniko-project/executor:debug
+#    entrypoint: [""]
+#  script:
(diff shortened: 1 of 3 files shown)


.gitlab-ci.yml
```diff
@@ -0,0 +1,96 @@
...
+  stage: lint
+  before_script:
+    - npm install -g @commitlint/cli @commitlint/config-conventional
+    - 'echo "module.exports = {extends: [\"@commitlint/config-conventional\"]}" > commitlint.config.js'
+  script:
+    - npx commitlint --from=origin/master
+  except:
+    - master
+
+#build_test_image:  # Currently hangs forever, because of GitLab Runner infrastructure issues
```
gpocentek: I'm wondering if the image build should really be done here. It takes a lot of time and disk space, why not use the image from the docker hub?

max-wittig: The image from the dockerhub is pretty old and manually pushed. We could just push the latest changes, when the test folder changes and the build shouldn't take that long. Disk space is free on GitLa and it's just the test image.

Normally it works pretty well with kaniko, but seems like the GitLab guys have some problems currently. Not sure: https://gitlab.com/gitlab-org/gitlab-runner/issues/4223

max-wittig: But I could also remove the comment for now

gpocentek: I just think that this should not be done for all the builds. Maybe move the job to another repo so we can use it when GitLab has fixed their docker issue?

max-wittig: It's not done for all builds. It's only done for master, but only if the `tools` folder changes (so when the actual docker file or some build script changes)

```yaml
only:
  refs:
     - master
   changes:
     - tools/
```

See: 
https://docs.gitlab.com/ee/ci/yaml/#onlychangesexceptchanges

gpocentek: Ah right, that sounds good then :)

Issue #93, opened 2016-02-10
projects.all() returns only 20 projects


godmar:
In issue #53 I had asked how to list all projects. The answer was pagination.
I am now trying the same with the current version of python-gitlab.  This time, I am using gl.projects.all() as recommended in the deprecation warning I received with my old code.

However, it seems the code isn't doing any pagination (is it no longer necessary? In Issue #53, you said there's no way to avoid it?

As a result, I'm only getting 20 projects returned.

What's the way to iterate over all projects using python-gitlab?

gpocentek:
There's some inconsistency in the listing methods. The managers' `list()`method can return all the items with `all=True`, so this might do what you're looking for:

````` python
all_projects = gl.projects.list(all=True)
`````

This is not supported on other listing method (`all`, `search`, `owned`) for the moment.

godmar:
gl.projects.list(all=True) returns 10 projects only. gl.projects.all() returns 20 projects.

If you look at the code: projects.all() calls custom_list, which makes a single API request.

projects.list call gl.list(), which has some code here: https://github.com/gpocentek/python-gitlab/blob/master/gitlab/__init__.py#L372 but that stops after 10 results. The returned response is:

    {'last': {'url': 'https://git.cs.vt.edu/api/v3/projects?page=1&per_page=0', 'rel': 'last'}, 'first': {'url':
    'https://git.cs.vt.edu/api/v3/projects?page=1&per_page=0', 'rel': 'first'}}

which does not have 'next' property.


How do I iterate to return all projects in the new code base?

godmar:
ps: gl.projects.all() receives this API response:

    {'first': {'url': 'https://git.cs.vt.edu/api/v3/projects/all?page=1&per_page=0', 'rel': 'first'}, 'last': {'url': 'https://git.cs.vt.edu/api/v3/projects/all?page=38&per_page=0', 'rel': 'last'}, 'next': {'url': 'https://git.cs.vt.edu/api/v3/projects/all?page=2&per_page=0', 'rel': 'next'}}

but then [fails to process the 'next' field](https://github.com/gpocentek/python-gitlab/blob/master/gitlab/objects.py#L160) as far as I can see.

How is this supposed to work?

godmar:
Oh, and I know why. projects.all() issues a request for /projects/all, but then doesn't process the 'next' field in the r.links.  projects.list(all=True) issues a request for /projects, in which case the API response doesn't have a 'next' field.

[...]

PR #784, opened 2019-05-28
Update projects.rst

jkogut:
Having exactly 20 internal and 5 private projects in the group
spent some time debugging this issue.

Hopefully that helped: https://github.com/python-gitlab/python-gitlab/issues/93

Imho should be definitely mention about `all=True` parameter.

### DIFF
--- a/docs/gl_objects/projects.rst
+++ b/docs/gl_objects/projects.rst
@@ -39,6 +39,8 @@ Results can also be sorted using the following parameters:
 
 ::
 
+    # List all projects (default 20 )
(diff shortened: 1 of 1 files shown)


docs/gl_objects/projects.rst
```diff
@@ -39,6 +39,8 @@ Results can also be sorted using the following parameters:
 
 ::
 
+    # List all projects (default 20 )
```
gpocentek: Please remove the extra space before the parenthesis?

PR #789, opened 2019-06-08
docs: remove v3 support

gpocentek:


docs/switching-to-v4.rst
```diff
@@ -8,24 +8,24 @@ GitLab provides a new API version (v4) since its 9.0 release. ``python-gitlab``
...
 Using the v4 API
 ================
 
-python-gitlab uses the v4 API by default since the 1.3.0 release. To use the
-old v3 API, explicitly define ``api_version`` in the ``Gitlab`` constructor:
+python-gitlab uses the v4 API by default since the 1.3.0 release. If you are
+migrating from an older release, make sure that you remove the ``api_version``
+definition in you constructors and configuration file:
+
+The following examples are **not valid** anymore:
```
max-wittig: could we remove these examples instead?

gpocentek: I'd like to keep something to ease the migration from v3 to v4, but I don't really know how to show what not to do. I'll try to improve that.

PR #797, opened 2019-06-18
feat: bump version to 1.9.0

max-wittig:


gpocentek:
I don't see modifications in ChangeLog.rst, is that expected?

max-wittig:
Ahh no that's not expected. Forgot about that. I will fix it.

max-wittig:
@gpocentek Also added the changelog now :smiley:

PR #827, opened 2019-07-17
docs: re-order api examples

ahaynssen:
`Pipelines and Jobs` and `Protected Branches` are out of order in contents and sometimes hard to find when looking for examples.


max-wittig:
@ahaynssen Thanks for the contribution!

PR #846, opened 2019-07-25
fix auth() for cookie based authentication

buster3:
It is possible to do a cookie based authentication. This change enable the use of the GitLab user element.

https://python-gitlab.readthedocs.io/en/stable/api-usage.html
https://gist.github.com/gpocentek/bd4c3fbf8a6ce226ebddc4aad6b46c0a


max-wittig:
@buster3 Thanks for your contribution. But wasn't this possible already, like this?

```python
import gitlab

gl = gitlab.Gitlab(username="yourname", password="yourpassword")
gl.auth()
```

PR #861, opened 2019-08-14
Fix avatar uploading for projects

ravanscafi:
The `avatar` attribute was already set for projects but it wasn't working because the type was missing.


ravanscafi:
@max-wittig could you please take a look?
It's a really simple change, I'm using it and can ensure that it works.

Thanks!

PR #865, opened 2019-08-16
feat: add methods to retrieve an individual project environment

orf:
Gitlab supports this: https://docs.gitlab.com/ce/api/environments.html#get-a-specific-environment

I'm using the same `ProjectEnvironment` object, even though retrieving an environment has wildly more attributes (including nested ones). Please let me know if I should add anything else to this.

### DIFF
--- a/docs/gl_objects/environments.rst
+++ b/docs/gl_objects/environments.rst
@@ -24,6 +24,10 @@ Create an environment for a project::
 
     environment = project.environments.create({'name': 'production'})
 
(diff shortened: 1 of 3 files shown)


LaurentDumont:
What is the status on this? I'd like to get the Pipeline status for the env in question and the default list doesnt have the info :(

max-wittig:
@orf Doesn't this disable the listing of environments?

orf:
I've added a test @max-wittig. Let me know what you think!

PR #867, opened 2019-08-20
Add environment_scope to optional ProjectVariable attributes

orf:
Right now you cannot update a scoped variable in a project!


orf:
Seems to be an issue with the Gitlab API when updating scoped variables :/

Issue #871, opened 2019-09-04
Add "submodules" API


krase:
## Description of the problem, including code/CLI snippet
Since GitLab 11.5 there is an API for handling Git submodules.
https://docs.gitlab.com/ee/api/repository_submodules.html

I would love to help here, but i don't get how to add new API elements. python-gitlab looks like having a nice pluggable structure, but currently i do not understand what i need to do exactly. A short introduction for developers in the documentation would be nice.

## Expected Behavior
Having submodules API available in python-gitlab

## Actual Behavior
Submodules API not implemented

## Specifications

  - python-gitlab version: 1.11
  - API version you are using (v3/v4): v4
  - Gitlab server version (or gitlab.com): 12.2.4

PR #873, opened 2019-09-04
feat: impement update_submodule

krase:
This is an implementation for #871 
Not sure if this is as intended ;)


max-wittig:
Thanks for your contribution! Could you use black to format your code and add some tests?

PR #876, opened 2019-09-05
Add support for job token

sathieu:
See https://docs.gitlab.com/ee/api/jobs.html#get-job-artifacts for usage

### DIFF
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -83,9 +83,9 @@ You must define the ``url`` in each GitLab server section.
    If the GitLab server you are using redirects requests from http to https,
    make sure to use the ``https://`` protocol in the ``url`` definition.
 
-Only one of ``private_token`` or ``oauth_token`` should be defined. If neither
-are defined an anonymous request will be sent to the Gitlab server, with very
-limited permissions.
+Only one of ``private_token``, ``oauth_token`` or ``job_token`` should be
+defined. If neither are defined an anonymous request will be sent to the Gitlab
+server, with very limited permissions.
 
 .. list-table:: GitLab server options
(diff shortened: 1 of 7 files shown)


docs/cli.rst
```diff
@@ -108,7 +110,8 @@ limited permissions.
    * - ``http_password``
      - Password for optional HTTP authentication
 
-__ https://docs.gitlab.com/ce/user/profile/personal_access_tokens.html
+.. _pat: https://docs.gitlab.com/ce/user/profile/personal_access_tokens.html
+.. _job-token: https://docs.gitlab.com/ee/api/jobs.html#get-job-artifacts
```
max-wittig: let's keep links with `CE` here to be consistent

sathieu:
@max-wittig All done. Please merge.

gitlab/__init__.py
```diff
@@ -346,9 +350,17 @@ def _construct_url(self, id_, obj, parameters, action=None):
...
     def _set_auth_info(self):
-        if self.private_token and self.oauth_token:
+        if (
+            self.private_token
+            and self.oauth_token
+            or self.private_token
+            and self.job_token
+            or self.oauth_token
+            and self.job_token
+        ):
```
max-wittig: This seems confusing. 

What about a true XOR?

```python
if sum(bool(arg) for arg in [self.private_token, self.oauth_token, self.job_token]) == 1
```

sathieu: Done (it's `!= 1`).

max-wittig: ah, you're right!

sathieu:
Thanks

PR #885, opened 2019-09-23
Don't mangle label name on update

sathieu:


bufferoverflow:
@sathieu commit message should be in conventional changlog style, see https://github.com/python-gitlab/python-gitlab#contributing

sathieu:
@max-wittig All done. Please review again!

max-wittig:
Thanks for the PR!

PR #896, opened 2019-10-06
improve functionnal tests

jouve:
* add a new flag -k to reuse the container
  * adapt python_test_v4 to be run multiple times
  * add reset_gitlab.py to undo changes made by python_test_v4
* use gitlab container from gitlab
* new check when gitlab is started (instead of long sleep)
* simplify generate_token.py with requests-html

### DIFF
--- a/tools/build_test_env.sh
+++ b/tools/build_test_env.sh
@@ -25,11 +25,13 @@ error() { log "ERROR: $@" >&2; }
 fatal() { error "$@"; exit 1; }
 try() { "$@" || fatal "'$@' failed"; }
 
+REUSE_CONTAINER=
 NOVENV=
 PY_VER=3
 API_VER=4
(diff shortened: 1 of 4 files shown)


max-wittig:
Thank you very much @jouve. I've always wanted to improve the tests and use the GitLab container.

You may need to wait a while longer or use the [healthcheck-endpoint](https://docs.gitlab.com/ce/user/admin_area/monitoring/health_check.html) so you can be sure that GitLab started completely.

jouve:
I tried the healthcheck, unfortunatelly it works on my computer by not with travis :( so I had to put the sleep 200 again

max-wittig:
Yeah travis is frustrating :(

max-wittig:
@jouve Thanks again!

PR #900, opened 2019-10-07
Fix constructing Gitlab object without authentication

jalr:
unauthenticated access is broken since #876 was merged

fixes getting the error `Only one of private_token, oauth_token or job_token should be defined` when using no authentication method at all.

### DIFF
--- a/gitlab/__init__.py
+++ b/gitlab/__init__.py
@@ -355,7 +355,7 @@ class Gitlab(object):
                 bool(arg)
                 for arg in [self.private_token, self.oauth_token, self.job_token]
             )
-            != 1
+            > 1
         ):
             raise ValueError(
                 "Only one of private_token, oauth_token or job_token should "


max-wittig:
Totally missed that. Thanks for the PR. Would you mind to use the commit conventions? So change to commit message to something like: `fix: allow Gitlab object with auth`

Thank you!

jalr:
No worries, sometimes it happens that you overlook something. I just noticed the issue today because my chat bot wasn't working anymore :smile: 
Just learned about the commit linter. Good idea to implement that! I hope that it passes now.

gitlab/__init__.py
```diff
@@ -355,7 +355,7 @@ def _set_auth_info(self):
                 bool(arg)
                 for arg in [self.private_token, self.oauth_token, self.job_token]
             )
-            != 1
+            > 1
```
max-wittig: But this does exactly the reverse. It only works, if we have multiple parameters.

We should only check this condition, if the items are larger than 1

PR #904, opened 2019-10-12
remove deprecated session auth

jouve:
session endpoint was removed in gitlab api v4 and api v3 support was removed in python-gitlab 1.5.0


max-wittig:
Thanks for this MR!

PR #908, opened 2019-10-12
Remove warning about open files from test_todo()

derekschrock:
When running unittests python warns that the json file from test_todo()
was still open. Use with to open, read, and create encoded json data
that is used by resp_get_todo().


max-wittig:
@derekschrock Thanks for the MR!

PR #914, opened 2019-10-18
Remove inaccurate projects.get() example

### LINKED ISSUE #913: projects.get() raises error when project is specified as "userspace/name" rather than ID

Attempt to make a query using the project's path, per the [example in the documentation](https://github.com/python-gitlab/python-gitlab/blame/v1.12.1/docs/gl_objects/projects.rst#L68).

```python
gl.projects.get('mygroup/myname')
```

## Expected Behavior

python-gitlab resolves the group and project name, returning a `Project` object for that project.

## Actual Behavior

Exception is raised with the following error:

```
Error encountered triggering pipeline: 404: <!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>404 Not Found</title>
</head><body>
<h1>Not Found</h1>
<p>The requested URL /api/v4/projects/mygroup/myname was not found on this server.</p>
</body></html>
```

From the URL in the error, it appears that the project is not being resolved at all and that python-gitlab is assuming the value passed in was the project ID.

## Specifications

  - python-gitlab version: 1.12.1
  - Gitlab server version (or gitlab.com): 11.11.2

terminalmage:
Also, add an example value for project_id to the other `projects.get()` example.

Resolves #913

### DIFF
--- a/docs/gl_objects/projects.rst
+++ b/docs/gl_objects/projects.rst
@@ -63,9 +63,8 @@ Results can also be sorted using the following parameters:
 Get a single project::
 
     # Get a project by ID
+    project_id = 851
     project = gl.projects.get(project_id)
-    # Get a project by userspace/name
-    project = gl.projects.get('myteam/myproject')
 
 Create a project::
 


bufferoverflow:
Thanks for the contribution @terminalmage !

Could you please adopt the commit message according to https://github.com/python-gitlab/python-gitlab/blob/master/README.rst#contributing to e.g. `docs: projects get requires id`, this is the reason why CI fails.

terminalmage:
CI passing now, thanks.

PR #919, opened 2019-10-23
fix(projects): support `approval_rules` endpoint for projects

srikanthchelluri:
The `approvers` API endpoint is deprecated [1]. GitLab instead uses
the `approval_rules` API endpoint to modify approval settings for
merge requests. This adds the functionality for project-level
merge request approval settings.

Note that there does not exist an endpoint to 'get' a single
approval rule at this moment - only 'list'.

[1] https://docs.gitlab.com/ee/api/merge_request_approvals.html

### DIFF
--- a/docs/gl_objects/mr_approvals.rst
+++ b/docs/gl_objects/mr_approvals.rst
@@ -12,6 +12,8 @@ References
 
   + :class:`gitlab.v4.objects.ProjectApproval`
   + :class:`gitlab.v4.objects.ProjectApprovalManager`
+  + :class:`gitlab.v4.objects.ProjectApprovalRule`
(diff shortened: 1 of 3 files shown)


max-wittig:
@srikanthchelluri Thanks for this MR! Would you mind to also add some tests for this feature?

srikanthchelluri:
@max-wittig Force pushed to fix merge conflicts.

It seems like merge request approvals aren't available in the community edition (which is what we're spinning up as part of the integration tests): https://gitlab.com/gitlab-org/gitlab/issues/20696

I added a few assertions to `ee-test.py`, but I don't have a private EE instance to test on. Any suggestions?

max-wittig:
@srikanthchelluri ee-test.py seems like the place to put this, yes. I also don't have an EE version to test.

PR #925, opened 2019-10-30
use default template for Docker container pipeline

CybotTM:
simplify/unify pipeline setup;
enable Docker image build for all tags and branches incl. master;

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,41 +1,29 @@
-image: python:3.7
-
-stages:
-  - deploy
-  - deploy-latest
-
-deploy:
-  stage: deploy
+# This file is a template, and might need editing before it works on your project.
+build-master:
+  # Official docker image.
+  image: docker:latest
+  stage: build
+  services:
+    - docker:dind
+  before_script:
+    - docker login -u "$CI_REGISTRY_USER" -p "$CI_REGISTRY_PASSWORD" $CI_REGISTRY
   script:
-    - pip install -U setuptools wheel twine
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -1,41 +1,29 @@
...
-
-deploy:
-  stage: deploy
+# This file is a template, and might need editing before it works on your project.
+build-master:
+  # Official docker image.
+  image: docker:latest
+  stage: build
+  services:
+    - docker:dind
```
max-wittig: we don't need `dind`, if we can build the image using `kaniko`

I don't see the reason to go a step back here.

CybotTM: But you are adding another layer of complexity without using any of the benefits of kaniko or without any problem which needs to be solved by using kaniko.

So maybe i was just confused, and missed the "latest" image in your build process and image registry, and this "deploy" was also irritating, because ... i don't know, i missed something, i expected this pipeline just builds the image for every tag and branch and pushes it into the registry. But if there is more, than, well.

max-wittig: The latest image will be added automatically as soon as a new tag is created.

I don't think we need an image for every branch.

max-wittig: See: https://github.com/python-gitlab/python-gitlab/commit/01cbc7ad04a875bea93a08c0ce563ab5b4fe896b

max-wittig:
I don't see a reason for this MR. `latest` docker images for every tag will be build, starting with the next `python-gitlab` release.. See: 01cbc7a

PR #926, opened 2019-10-30
Housekeeping

CybotTM:
just some technical tweaks to incorporate some best practices and require no insecure modules

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,16 +1,25 @@
-FROM python:3.7-alpine AS build
+FROM python:3.7-alpine AS base
+
+# ensure latest updates are installed (security)
+RUN apk upgrade --no-cache
+RUN pip list --outdated --format=freeze | grep -v '^\-e' | cut -d = -f 1  | xargs -n1 pip install --no-cache-dir -U
 
 WORKDIR /opt/python-gitlab
+
+
+FROM base AS build
+
 COPY . .
 RUN python setup.py bdist_wheel
 
-FROM python:3.7-alpine
 
-WORKDIR /opt/python-gitlab
-COPY --from=build /opt/python-gitlab/dist dist/
-RUN pip install PyYaml
-RUN pip install $(find dist -name *.whl) && \
-    rm -rf dist/
+FROM base
+
+COPY --from=build /opt/python-gitlab/dist/ /opt/python-gitlab/dist/
+
+RUN pip install --no-cache-dir PyYaml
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -1,16 +1,25 @@
-FROM python:3.7-alpine AS build
+FROM python:3.7-alpine AS base
+
+# ensure latest updates are installed (security)
+RUN apk upgrade --no-cache
```
max-wittig: This is against the docker best practices. Never upgrade packages in the image itself. The base should be upgraded 

> Avoid RUN apt-get upgrade and dist-upgrade, as many of the “essential” packages from the parent images cannot upgrade inside an unprivileged container

https://docs.docker.com/develop/develop-images/dockerfile_best-practices/

Dockerfile
```diff
@@ -1,16 +1,25 @@
-FROM python:3.7-alpine AS build
+FROM python:3.7-alpine AS base
```
max-wittig: let's use 3.8, while we're at it

max-wittig:
Thanks for the MR! I'm not sure about upgrading of base packages.

Could you please format your commit messages according to our contributing guidelines?
https://github.com/python-gitlab/python-gitlab#contributing

Dockerfile
CybotTM: Python-request does not support 3.8 officially.

max-wittig: yes. It does. https://github.com/python-gitlab/python-gitlab/commit/6048175ef2c21fda298754e9b07515b0a56d66bd

Dockerfile
CybotTM: But "I" have no control over the base image, I have no control when they will update their packages. To incorporate a security update docker:alpine needs to be updated, after this docker:python needs to be update, and finally docker:python-gitlab has to be updated and maybe any custom company specific image based on python-gitlab has to be updated too.

But I want this fix for the pecl lib now which is used by some version-parser-module to extract version info from tags which holds some malicious formatted content. (Just an example)

So, i do not agree to this best-practice. :-)

max-wittig: I do agree to follow best-practice and we shouldn't maintain the base image of python ourselves.

> But "I" have no control over the base image

These changes should rather go here: https://github.com/docker-library/python

max-wittig:
Closing this, as base image changes should rather go in the base-image at https://github.com/docker-library/python

PR #937, opened 2019-11-14
fix(labels): ensure label.save() works

sathieu:
Otherwise, we get:
```
   File "gitlabracadabra/mixins/labels.py", line 67, in _process_labels
     current_label.save()
   File "gitlab/exceptions.py", line 267, in wrapped_f
     return f(*args, **kwargs)
   File "gitlab/v4/objects.py", line 896, in save
     self._update_attrs(server_data)
   File "gitlab/base.py", line 131, in _update_attrs
     self.__dict__["_attrs"].update(new_attrs)
 TypeError: 'NoneType' object is not iterable
```

Because `server_data` is `None`.


max-wittig:
Thanks for your MR!

Issue #939, opened 2019-11-14
Proper usage of exceptions under Python3


ericfrederich:
## Description of the problem, including code/CLI snippet

Python3 has support to `raise from` to distinguish between the following circumstances:
* an exception while handling another exception
* an exception which is the direct cause of another

For instance when a project doesn't exist as in the code below, you'll see two exceptions:
* gitlab.exceptions.GitlabHttpError
* gitlab.exceptions.GitlabGetError

```python
import os
import gitlab

gl = gitlab.Gitlab(
    url=os.environ['GITLAB_URL'],
    private_token=os.environ['GITLAB_TOKEN'],
    api_version="4",
)
gl.auth()

gl.projects.get('non-existent-project')
```

On Python3 between the two exceptions you'll see the statement: `During handling of the above exception, another exception occurred:`

This is clearly not the case.

## Expected Behavior

You see in the console between the two exceptions:
```
The above exception was the direct cause of the following exception:
```

## Actual Behavior

You see in the console between the two exceptions:
```
During handling of the above exception, another exception occurred:
```

## Specifications

  - python-gitlab version: 1.13.0
  - API version you are using (v3/v4): v4
  - Gitlab server version (or gitlab.com): 12.4.1

PR #940, opened 2019-11-14
Use six.raise_from in GitLabError's wrapper

ericfrederich:
Fixes Issue #939

### DIFF
--- a/gitlab/exceptions.py
+++ b/gitlab/exceptions.py
@@ -17,6 +17,8 @@
 
 import functools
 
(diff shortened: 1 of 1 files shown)


max-wittig:
Thanks for your contribution. Would you mind to format your code with black and use the commit conventions that are described in the README?

bufferoverflow:
I think this is not worth to implement, as 2.7 and six will be gone very soon, see https://pythonclock.org/

PR #941, opened 2019-11-16
feat: add autocompletion support

### LINKED ISSUE #637: Feature idea: autocompletion for CLI mode

Feature/improve suggestion: in many of my cases, I would like to have an autocompletion to have less time to spend at docs and source code to find out all possible combination of commands/object/args.

## Expected Behavior

Would be great to have autocompletion for bash or zsh.

## Actual Behavior

No completion right now.

## Specifications

  - python-gitlab version: 1.6.0
  - API version you are using (v3/v4): 4

mchlumsky:
Fixes #637

### DIFF
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -271,3 +271,61 @@ command line. This is handy for values containing new lines for instance:
    It is obviously the best project around
    EOF
    $ gitlab project create --name SuperProject --description @/tmp/description
+
+Enabling shell autocompletion
+============================
+
+To get autocompletion, you'll need to install the package with the extra
+"autocompletion":
+
+.. code-block:: console
+
+    pip install python_gitlab[autocompletion]
+
+
+Add the appropriate command below to your shell's config file so that it is run on
+startup. You will likely have to restart or re-login for the autocompletion to
+start working.
+
+Bash
+----
+
+.. code-block:: console
+
+   eval "$(register-python-argcomplete gitlab)"
+
+tcsh
+----
+
+.. code-block:: console
+
+   eval `register-python-argcomplete --shell tcsh gitlab`
+
+fish
+----
+
+.. code-block:: console
+
+   register-python-argcomplete --shell fish gitlab | .
+
+Zsh
+---
+
+.. warning::
+
+    Zsh autocompletion support is broken right now in the argcomplete python
+    package. Perhaps it will be fixed in a future release of argcomplete at
+    which point the following instructions will enable autocompletion in zsh.
+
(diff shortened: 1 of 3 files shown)


max-wittig:
Thanks for this feature. Doesn't work for me with the provided instructions and zsh.

mchlumsky:
Hi!

Indeed it doesn't work. I thought it was just my setup that didn't work. I don't really use zsh so my testing of it was light. The other shells do work though. I use this patch in bash all the time with no problems at all.

Argcomplete sounded like a quick win. It's used by pipx [1] and others.
It doesn't sound like the author of argcomplete cares much about zsh [2].

Should I remove any mention of zsh from the docs? Or should we mention zsh support as experiemental/broken with a reference to the issue [2]?
Or is this a show stopper?

[1]: https://github.com/pipxproject/pipx/blob/master/src/pipx/main.py#L587
[2]: https://github.com/kislyuk/argcomplete/issues/10

max-wittig:
@mchlumsky I guess we should add a notice that it doesn't work with zsh.

Can we make `argcomplete` optional somehow? I think most people don't use the CLI.

mchlumsky:
I kept the commits separate for clarity but I can squash them into one once they pass peer-review.

max-wittig:
@mchlumsky Looks good to me now! Could you just squash those commits and then we can merge it :smiley:

mchlumsky:
Commits squashed! However travis-ci testing is failing during some rate-limit testing in py_func_v4.
I tried rebasing against latest master but I still get the failure.
I don't think the failure is related to this merge request.

max-wittig:
@mchlumsky Sadly the build pipeline is not as stable as I want it to be. But I can't figure out why as everything works fine locally, everytime.

mchlumsky:
@max-wittig The test that failed in my run was [this one](https://github.com/python-gitlab/python-gitlab/blob/2a01326e8e02bbf418b3f4c49ffa60c735b107dc/tools/python_test_v4.py#L902). It's a rate limit test, which can be almost described as a performance test. The test fails because it can't trigger the rate limiting feature of gitlab. This kind of test is hard to achieve on a public shared service like travis-ci where the VMs running your tests can have very different sizes and/or workloads.
I am pretty impressed with the functional testing you've achieved (spinning up a gitlab instance and testing against it). I didn't know you could pull that off with travis-ci! Kudos for that! ;)

max-wittig:
@mchlumsky Thank you, but credit goes to the original creator of this lib, not to me. @gpocentek

max-wittig:
But I guess we should remove this test.

Issue #942, opened 2019-11-20
404 for project-fork list


choyrim:
## Description of the problem, including code/CLI snippet
```
$ gitlab project-fork list --project-id 15368194
Impossible to list objects (404: 404 Project Not Found)
```

but browsing to https://gitlab.com/api/v4/projects/15368194/forks directly works (logged in)

## Expected Behavior
should show a list of forks.

## Actual Behavior
404 response.

## Specifications

  - python-gitlab version: 1.13.0
  - API version you are using (v3/v4): v4
  - Gitlab server version (or gitlab.com): gitlab.com

PR #943, opened 2019-11-20
#942: fix up path computation for project-fork list

choyrim:
fix for #942 - path computation wasn't working.

### DIFF
--- a/gitlab/v4/objects.py
+++ b/gitlab/v4/objects.py
@@ -2049,7 +2049,7 @@ class ProjectFork(RESTObject):
 
 
 class ProjectForkManager(CreateMixin, ListMixin, RESTManager):
-    _path = "/projects/%(project_id)s/fork"
+    _path = "/projects/%(project_id)s/forks"
     _obj_cls = ProjectFork
     _from_parent_attrs = {"project_id": "id"}
     _list_filters = (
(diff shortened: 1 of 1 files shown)


max-wittig:
@choyrim But don't you delete the `list` method here and replace it with the `create` method?

choyrim:
> @choyrim But don't you delete the `list` method here and replace it with the `create` method?

@max-wittig yes that is the end result. I originally just tried to fix the list method but that caused the create to fail. So I noticed that the `ProjectPipelineManager` had resolved a similar path plurality (pipeline vs pipelines) issue and imitated the approach.

gitlab/v4/objects.py
```diff
@@ -2069,27 +2069,24 @@ class ProjectForkManager(CreateMixin, ListMixin, RESTManager):
...
-        """
+            GitlabCreateError: If the server cannot perform the request
 
-        path = self._compute_path("/projects/%(project_id)s/forks")
-        return ListMixin.list(self, path=path, **kwargs)
+        Returns:
+            RESTObject: A new instance of the managed object class build with
+                the data sent by the server
+        """
+        path = self.path[:-1]  # drop the 's'
```
max-wittig: Maybe we could use the real path instead to avoid confusion.

```
/projects/%(project_id)s/fork
```

max-wittig:
Ah I see. Makes sense! Thank you!

gitlab/v4/objects.py
choyrim: I am just trusting that the person who did the pipeline manager knew what he/she was doing. I suspect that if I replace the path as above, it won't substitute in for `%(project_id)s` which was the cause for the 404 in the earlier `ProjectForkManager.list` method implementation.

max-wittig: ah you're right. Sorry.

Issue #551, opened 2018-07-11
feature idea : Kubernetes configuration


jpic:
Thanks a lot for maintaining this library !

The feature idea : Manage k8s environments like environments, to edit clusters in the "Operations" -> "Kubernetes" project menu inside gitlab.

 gitlab.v4.objects.ProjectKubernetesManager & friends :joy:

gpocentek:
Hello @jpic 

Thank you :)

Nice idea but this needs to be implemented in GitLab first. python-gitlab only manages what exists in the GitLab API. You should open an issue on their bugtracker to suggest this idea: https://gitlab.com/gitlab-org/gitlab-ce.

I'll be happy to add support for this feature as soon as it lands in Gitlab.

jpic:
Thanks! [It's on](https://gitlab.com/gitlab-org/gitlab-ce/issues/49145) :joy:

jpic:
FTR: it was [already on](https://gitlab.com/gitlab-org/gitlab-ce/issues/40473)

J4lf:
The Gitlab API is now able to manage project level clusters and group level clusters. Would it be possible to add the support of these apis?
Thanks!

max-wittig:
Sure. I just don't have any way to try this out, but I would be happy to review a MR regarding this topic.

nejch:
I think this is done, with #946 and #947 merged :)

PR #946, opened 2019-11-24
feat: add project and group clusters

lundbird:
Resolve issue #551. @max.wittig@siemens.com

This is the only gitlab api type that has nested json. You may have trouble using this feature from the command line as the nested-json parameter will be interpreted as a string which will cause the request to fail.


max-wittig:
Awesome addition! Thank you!

max-wittig:
@lundbird Just noticed that we got no docs for this. Would you mind to add another MR with the docs? Looks like maybe you just forgot `git add`?

Issue #952, opened 2019-11-28
Missing imports api 


weaming:
## Description of the problem, including code/CLI snippet

missing https://docs.gitlab.com/ee/api/import.html

## Expected Behavior

gl.imports.github ...

## Actual Behavior

not found in this lib.

## Specifications

  - python-gitlab version: 1.13.0
  - API version you are using (v3/v4): v4
  - Gitlab server version (or gitlab.com): whatever

max-wittig:
This was added by @andrew-littlebits in the last release. Thanks again!

PR #959, opened 2019-12-03
feat: add support for /import/github

andrew-littlebits:
Addresses #952

This adds a method to the `ProjectManager` called `import_github`, which maps to the `/import/github` API endpoint. Calling `import_github` will trigger an import operation from <repo_id> into <target_namespace>, using <personal_access_token> to authenticate against github. In practice a gitlab server may take many 10's of seconds to respond to this API call, so we also take the liberty of increasing the default timeout (only for this method invocation).

Unfortunately since `import` is a protected keyword in python, I was unable to follow the endpoint structure with the manager namespace (eg: `gitlab.import.github` is not allowed`). I'm open to suggestions on a more sensible interface.

I'm successfully using this addition to batch-import hundreds of github repositories into gitlab. I didn't see any low-hanging-fruit way to add a unit test for this - please advise if this is acceptable.

### DIFF
--- a/gitlab/v4/objects.py
+++ b/gitlab/v4/objects.py
@@ -4744,6 +4744,69 @@ class ProjectManager(CRUDMixin, RESTManager):
             "/projects/import", post_data=data, files=files, **kwargs
         )
 
+    def import_github(
+        self,
+        personal_access_token,
+        repo_id,
+        target_namespace,
+        new_name=None,
+        timeout_override=60.0,
+        **kwargs
+    ):
+        """Import a project from Github to Gitlab (schedule the import)
+
+        This method will return when an import operation has been safely queued,
+        or an error has occurred. After triggering an import, check the
+        `import_status` of the newly created project to detect when the import
+        operation has completed.
+
+        NOTE: this request may take longer than most other API requests.
+        So this method will override the session timeout with <timeout_override>,
+        which defaults to 60 seconds.
+
+        Args:
+            personal_access_token (str): GitHub personal access token
+            repo_id (int): Github repository ID
+            target_namespace (str): Namespace to import repo into
+            new_name (str): New repo name (Optional)
+            timeout_override (int or float): Timeout to use for this request
+            **kwargs: Extra options to send to the server (e.g. sudo)
+
+        Raises:
+            GitlabAuthenticationError: If authentication is not correct
+            GitlabListError: If the server failed to perform the request
+
+        Returns:
+            dict: A representation of the import status.
+
+        Example:
+        ```
+            gl = gitlab.Gitlab_from_config()
+            print "Triggering import"
+            result = gl.projects.import_github(ACCESS_TOKEN,
+                                               123456,
+                                               "my-group/my-subgroup")
+            project = gl.projects.get(ret['id'])
+            print "Waiting for import to complete"
+            while project.import_status == u'started':
+                time.sleep(1.0)
+                project = gl.projects.get(project.id)
+            print "Github import complete"
+        ```
+        """
+        data = {
+            "personal_access_token": personal_access_token,
+            "repo_id": repo_id,
+            "target_namespace": target_namespace,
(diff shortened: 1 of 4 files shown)


max-wittig:
Thanks for the MR. I think we should follow the GitLab API as closely as possible, without adding helpers to keep it as thin of a layer as possible. Then we could also remove the dependency to another API wrapper library. 

It should be implemented like descriped here: https://docs.gitlab.com/ee/api/import.html#import-repository-from-github.

API requests can be faked pretty easily as test, like done here: https://github.com/python-gitlab/python-gitlab/blob/master/gitlab/tests/test_gitlab.py#L757

andrew-littlebits:
Ok, no problem, I can clean out the `github` dependency and add a test. I'll try to get on this shortly.

andrew-littlebits:
@max-wittig i'm not quite familiar with this error (see failed pipeline build), and it doesn't look like its anywhere near code that i've touched. Do you have a moment to take a look? I can dig deeper too, but I wanted to check first if this is a known issue.

gitlab/v4/objects.py
```diff
@@ -4712,6 +4712,74 @@ def import_project(
...
+        operation has completed.
+
+        NOTE: this request may take longer than most other API requests.
+        So this method will override the session timeout with <timeout_override>,
+        which defaults to 60 seconds.
+
+        If repo_id is specified as an integer, it will be used directly.
+        If repo_id is a string, we will attempt to look it up using PyGithub.
+        The preferred string format is <org>/<repository>. However the python
+        package `PyGithub` must be installed to perform this lookup.
```
max-wittig: The docs will need some updates

gitlab/v4/objects.py
```diff
@@ -4712,6 +4712,74 @@ def import_project(
...
+        """
+        data = {
+            "personal_access_token": personal_access_token,
+            "repo_id": repo_id,
+            "target_namespace": target_namespace,
+        }
+        if new_name:
+            data["new_name"] = new_name
+        prev_timeout = self.gitlab.timeout
+        self.gitlab.timeout = timeout_override
```
max-wittig: This seems hacky. Can't you just provide timeout in the post method? (haven't looked into the code too much)

```python
kwargs["timeout"] = 60
result = self.gitlab.http_post("/import/github", post_data=data, **kwargs)
```

max-wittig:
Left a few comments.

gitlab/v4/objects.py
andrew-littlebits: I'm afraid not. It looks like on each invocation of `http_post` (or `http_request`) the timeout is freshly retrieved from `_get_session_opts()` and therefore pulled from the base gitlab object:

https://github.com/python-gitlab/python-gitlab/blob/master/gitlab/__init__.py#L490-L493

If I were to amend this behavior to also accept a timeout from `kwargs`, would this MR be an appropriate enough place to do it?

gitlab/v4/objects.py
andrew-littlebits: Whoops! Resolved in 89d8f2b253803923c9fd633a11df66d409709cda.

gitlab/v4/objects.py
max-wittig: ah yes. Feel free to create a new commit with this.

andrew-littlebits:
As requested in an outdated discussion: i've lightly modified the base `gitlab` object to allow `timeout` to be specified in `kwargs`. This helps to avoid hacky code: instead of needing to change the global timeout parameter for the `/import/github` API call, I can set it just for that request.

max-wittig:
@andrew-littlebits Sorry for letting you wait so long. LGTM. Could you just rebase your branch to the latest master? Thanks!

andrew-littlebits:
@max-wittig no worries, thank you for walking me through the process! I've rebased and lightly squashed.

PR #968, opened 2019-12-10
Stats

mitar:


mitar:
I have no idea if this is right, but it works. :-)

gitlab/v4/objects.py
```diff
@@ -4042,6 +4062,8 @@ class Project(SaveMixin, ObjectDeleteMixin, RESTObject):
         ("variables", "ProjectVariableManager"),
         ("wikis", "ProjectWikiManager"),
         ("clusters", "ProjectClusterManager"),
+        ("additionalstatistics", "ProjectAdditionalStatisticsManager"),
```
max-wittig: Why not just `statistics`, like GitLab does it? 

I see that this might be introduce confusion with `project.attributes["statistics"]`. What do you think?

gitlab/v4/objects.py
```diff
@@ -4042,6 +4062,8 @@ class Project(SaveMixin, ObjectDeleteMixin, RESTObject):
...
+        ("issuesstatistics", "ProjectIssuesStatisticsManager"),
```
max-wittig: I think `issue_statistics` would sound better.

max-wittig:
@mitar Thanks for the MR. Would you mind to use the conventional changelog for your commits, add docs and tests?

gitlab/v4/objects.py
mitar: Yes, it overrides that. This is why I had to rename it.

gitlab/v4/objects.py
mitar: But above you have things like `protectedbranches`, not `protected_branches`?

mitar: Also, GitLab API names it [issue**s** statistics](https://docs.gitlab.com/ee/api/issues_statistics.html#get-project-issues-statistics).

max-wittig:
@mitar Nearly complete 👍 Would you mind to just add some tests to this?

mitar:
Do you have an existing example to look at? I could not find an example for such a simple endpoint like this.

mitar:
Added tests.

max-wittig:
@mitar Woah. You're fast. Just format your changes using black and you're good to go after the pipeline passes.

mitar:
Done.

mitar:
Tests are passing.

max-wittig:
@mitar Awesome thanks!

mitar:
When would this be released?

max-wittig:
@mitar I guess I could do another release for you on Monday.

max-wittig:
@mitar You got a christmas present! :christmas_tree: :gift: 

https://github.com/python-gitlab/python-gitlab/releases/tag/v1.15.0

Thanks for your contributions!

mitar:
Thank you!

PR #971, opened 2019-12-12
feat: add more options for project/group ci variables manipulation

jooola:
Add more flags for CI variables Create and Update operations.


max-wittig:
Thanks for your MR!

PR #980, opened 2019-12-18
Refactor/cleanup upgrade

max-wittig:


bufferoverflow:
very nice, goodby six

PR #982, opened 2019-12-18
chore: enforce python version requirements

max-wittig:


bufferoverflow:
@max-wittig very nice! do you like wait until python 2 is gone?

max-wittig:
Yes we will. I will craft a new release next year.

max-wittig:
@bufferoverflow We already dropped support for 2 with the previous MR merged.

max-wittig:
This is actually important, so that python2 users can still use the old version.

PR #984, opened 2019-12-22
chore: build_sphinx needs sphinx >= 1.7.6

derekschrock:
Stepping thru Sphinx versions from 1.6.5 to 1.7.5 build_sphinx fails.  Once Sphinx == 1.7.6 build_sphinx finished.


max-wittig:
Thanks for the fix!

PR #986, opened 2020-01-07
fix snippet get example

jgroom33:
This fixes the snippet example in the documentation


max-wittig:
Thanks for the MR :+1:  Could you please change the commit message to comply with the conventions?

PR #996, opened 2020-01-21
feat: add appearance API

max-wittig:


bufferoverflow:
awesome, this helps to reduce our efforts!

PR #997, opened 2020-01-21
Refactor/structure objects

max-wittig:


max-wittig:
@nejch Already did a better job here!

PR #998, opened 2020-01-22
WIP: migrate to asyncio and httpx

### LINKED ISSUE #945: migrate to asyncio

all the requests done in this module are syncronous. to use it more efficiently all requests should be converted to https://aiohttp.readthedocs.io/en/stable/ to make them concurrent

## Expected Behavior

posibility for concurrent requests

## Actual Behavior

only syncronus requests

## Specifications

  - python-gitlab version:

L0ric0:
starting to migrate to async while staying compatible to the syncronous versions.

TODO:
* [x] rewrite GitlabList __init__ to be async compatible
* [ ] find all functions that need a async_ version 

which funktions are suposed to be called by the user? if possible i would like to keep all internal funktions async and run the functions that have to be sync with `run_until_complete()`

would it be better to have a centrel event_loop instead of getting it with `get_event_loop` every time and if so where would i put it?

fix #945

### DIFF
--- a/gitlab/__init__.py
+++ b/gitlab/__init__.py
@@ -650,14 +650,18 @@ class Gitlab(object):
         url = self._build_url(path)
 
         if get_all is True and as_list is True:
-            return list(GitlabList(self, url, query_data, **kwargs))
+            return list(await GitlabList(self, url, query_data, **kwargs).__setup__())
 
         if "page" in kwargs or as_list is True:
             # pagination requested, we return a list
-            return list(GitlabList(self, url, query_data, get_next=False, **kwargs))
+            return list(
+                await GitlabList(
+                    self, url, query_data, get_next=False, **kwargs
+                ).__setup__()
+            )
 
         # No pagination, generator requested
-        return GitlabList(self, url, query_data, **kwargs)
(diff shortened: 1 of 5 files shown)


L0ric0:
at the moment the httpx module dose not support `auth` in the `httpx.Request` object and doesn't supply a `client.prepare_request` method so that the `http_request` method in `__init__.py` doesn't work. Does somebody know of an alternative?

max-wittig:
Thanks for all the efforts on this. Maybe we should wait until the 1.0 release of https in April?

> A 1.0 release is expected to be issued sometime on or before April 2020.`

L0ric0:
yeah probably hopefully they fix the two issues that are blocking this pr

Tadaboody:
Hello! I'm willing to help with getting this PR done to learn some more asyncio :) 

One question - why httpx? how does it compare to libraries like [aiohttp](https://docs.aiohttp.org/en/stable/)?

darkdragon-001:
Can you check if [httpx version 1.0.0.beta0](https://github.com/encode/httpx/releases/tag/1.0.0.beta0) fixes the remaining issues?

JohnVillalovos:
No activity in almost two years from the author. Closing.

PR #1000, opened 2020-01-25
Update auth docs

matusf:
Hi, I removed the docs regarding `email/password` auth since it's no longer supported by `python-gitlab`.


max-wittig:
Thanks for your MR. That's helpful. Just a small comment

.gitignore
```diff
@@ -9,3 +9,6 @@ docs/_build
 .testrepository/
 .tox
 venv/
+.mypy_cache
+.vscode
+
```
max-wittig: These are local files, which should be managed by a global gitconfig instead.

matusf: Thanks, fixed

max-wittig:
Thanks for the MR!

PR #1003, opened 2020-01-27
feat: use keyset pagination by default for `all=True`

max-wittig:


max-wittig:
PTAL @mitar

docs/api-usage.rst
```diff
@@ -204,6 +204,11 @@ listing methods support the ``page`` and ``per_page`` parameters:
 By default GitLab does not return the complete list of items. Use the ``all``
 parameter to get all the items when using listing methods:
 
+.. warning::
+
+   The all=True option uses keyset pagination by default, if order_by="id"
+   or if order_by is not supplied.
```
mitar: This is slightly hard to read. Maybe:

```suggestion
   The all=True option uses keyset pagination by default if order_by is not supplied,
   or if order_by="id".
```

PR #1004, opened 2020-01-29
[WIP] Switch project import to requests-toolbelt

zapp42:
Maybe this is of use for anybody. It works for me but I have not tested it thoroughly.
The Gitlab project import has lots of open issues, so your mileage may vary. Also I had to adapt the ProxyTimeout (Apache reverse proxy) and the unicorn['worker_timeout'] values.

### DIFF
--- a/gitlab/__init__.py
+++ b/gitlab/__init__.py
@@ -23,6 +23,7 @@ import time
 import warnings
 
 import requests
+from requests_toolbelt.multipart.encoder import MultipartEncoder
 
 import gitlab.config
 from gitlab.const import *  # noqa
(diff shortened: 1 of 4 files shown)


max-wittig:
@zapp42 Thanks for this MR! Looks good. Could you just format your code with black and change your commit message to something like:

feat(project-import): use multipart upload

max-wittig:
@zapp42 I think this is gonna be solved by the switch to `httpx` as it supports multi-part uploads in #1036

max-wittig:
I guess, as we're not switching to httpx in the near future, we should consider this.

PR #1008, opened 2020-02-05
Add capability to control GitLab features per project or group

filipowm:
Currently GitLab Features API supports configuring features not only globally or per user, but also per group or project. `python-gitlab` is currently missing this option, thus this PR.

Additionally fixed issue when setting up feature with `None `(null) values, e.g. none `feature_group` or `user`. It caused GitLab to respond with HTTP 500. Thus now we're filtering out `None` values from data.

### DIFF
--- a/gitlab/v4/objects.py
+++ b/gitlab/v4/objects.py
@@ -760,6 +760,7 @@ class FeatureManager(ListMixin, DeleteMixin, RESTManager):
             "group": group,
             "project": project,
         }
+        data = utils.remove_none_from_dict(data)
         server_data = self.gitlab.http_post(path, post_data=data, **kwargs)
         return self._obj_cls(self, server_data)
 
--- a/gitlab/v4/objects.py
+++ b/gitlab/v4/objects.py
@@ -724,7 +724,16 @@ class FeatureManager(ListMixin, DeleteMixin, RESTManager):
     _obj_cls = Feature
 
     @exc.on_http_error(exc.GitlabSetError)
-    def set(self, name, value, feature_group=None, user=None, **kwargs):
+    def set(
+        self,
+        name,
+        value,
+        feature_group=None,
+        user=None,
+        group=None,
+        project=None,
+        **kwargs
+    ):
         """Create or update the object.
 
         Args:
@@ -732,6 +741,8 @@ class FeatureManager(ListMixin, DeleteMixin, RESTManager):
             value (bool/int): The value to set for the object
             feature_group (str): A feature group name
             user (str): A GitLab username
+            group (str): A GitLab group
+            project (str): A GitLab project in form group/project
             **kwargs: Extra options to send to the server (e.g. sudo)
 
         Raises:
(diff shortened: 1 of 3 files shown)


filipowm:
@max-wittig Could you please take a look?
This one is currently blocking PR in gitlab configuration-as-code tool: https://github.com/Roche/gitlab-configuration-as-code/pull/11

gitlab/v4/objects.py
```diff
@@ -742,7 +753,14 @@ def set(self, name, value, feature_group=None, user=None, **kwargs):
...
         path = "%s/%s" % (self.path, name.replace("/", "%2F"))
-        data = {"value": value, "feature_group": feature_group, "user": user}
+        data = {
+            "value": value,
+            "feature_group": feature_group,
+            "user": user,
+            "group": group,
+            "project": project,
+        }
+        data = utils.remove_none_from_dict(data)
```
max-wittig: Why is this required?

max-wittig:
@filipowm Thanks for the MR. Could you please drop the duplicate commits from master and rebase instead?

gitlab/v4/objects.py
filipowm: So, in API there are also `group` and `project` properties: https://docs.gitlab.com/ee/api/features.html

Regarding removing empty values from dict, see the checks below:
```
# 1
curl --header "PRIVATE-TOKEN: <token>" --data "value=true&user=" https://mygitlab.com/api/v4/features/test 
# 404 Not Found

# 2
curl --header "PRIVATE-TOKEN:<token>" --data "value=true&user=myuser" https://mygitlab.com/api/v4/features/test 
# OK: {"name":"test","state":"conditional","gates":[{"key":"boolean","value":false},{"key":"actors","value":["User:23"]}]}

# 3
curl --header "PRIVATE-TOKEN: <token>" --data "value=true&project=&group=&feature_group=&user=" https://mygitlab.com/api/v4/features/test
# 404 Not Found

# 4
curl --header "PRIVATE-TOKEN: <token>" --data "value=true&project=&group=&feature_group=" https://mygitlab.com/api/v4/features/test
# NOT OK (not expected result): {"name":"test","state":"conditional","gates":[{"key":"boolean","value":false},{"key":"groups","value":[""]}]}

# 5
curl --header "PRIVATE-TOKEN: <token>" --data "value=true" https://mygitlab.com/api/v4/features/test
# OK (expected result, should be same as the above one): {"name":"test","state":"on","gates":[{"key":"boolean","value":true}]}
```

Thus if empty values are sent (what happens), the value is set inappropriately. Examples:
1. `gitlab.features.set(name, value)` - I would expect that such feature will be set globally, but instead now it fails with 404 Not Found, because it's unable to find an empty user. This scenario #3 above.

2. `gitlab.features.set(name, value, group=existing_group)` - it will fail with 404 because empty user is sent.

3. `gitlab.features.set(name, value, user=existing_user)` - it should set a feature for given user (if user not exist, ofc it will return 404). But instead it sets it for user, empty group (like in case #4 above).

Based on above I came with idea to clean all `None` values from dict. Cause this solves the problem. If you have any better idea how to do this, I'd be happy to rewrite it :)

filipowm:
> @filipowm Thanks for the MR. Could you please drop the duplicate commits from master and rebase instead?

Done. I missed to notice that :)

max-wittig:
@filipowm Thanks for your efforts!

PR #1014, opened 2020-02-14
Add MR pipeline trigger function

nbtavis:
There are two types of pipelines:

1. Branch
2. Merge request

Trigger_pipeline only triggers branch pipelines.
This function would create and trigger an MR pipeline.


max-wittig:
@nbtavis Sorry for the late response and thanks for this MR! Would you mind to change the following things, before we can merge:
* update the docs
* format your code with black
* squash your commits
* use conventional changelog for your commit message: `feat(project): add mr trigger`

PR #1022, opened 2020-02-19
chore: ensure developers use same gitlab image as CI

nejch:
I had really weird behavior (tests failing, omnibus config wouldn't load properly) until I realized I had a super old gitlab image locally that wasn't being pulled for tests.

This does an explicit docker pull instead of `--pull` since that's only been added [recently](https://github.com/docker/cli/pull/1498) and might not work for everyone, and it should only be a tiny penalty anyway.

I added a CLI flag so this should make it easy to build a CI matrix with tox that tests against several GitLab versions (e.g. latest + 3 last minor releases etc) or to explicitly test against a version reported in an issue. (Although I see currently Travis is excruciatingly slow so not sure if that's desired :D)

### DIFF
--- a/tools/build_test_env.sh
+++ b/tools/build_test_env.sh
@@ -29,12 +29,15 @@ REUSE_CONTAINER=
 NOVENV=
 PY_VER=3
 API_VER=4
+GITLAB_IMAGE="gitlab/gitlab-ce"
+GITLAB_TAG="latest"
 while getopts :knp:a: opt "$@"; do
     case $opt in
         k) REUSE_CONTAINER=1;;
         n) NOVENV=1;;
         p) PY_VER=$OPTARG;;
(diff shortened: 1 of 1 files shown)


max-wittig:
@nejch Thanks for the MR! I like the idea that we always pull the latest image, but I don't think we should test older version. I'm sadly the only maintainer of this library and I don't think it's worth the time to spend on older GitLab versions.

nejch:
Ok sure! If you think the flag for explicit tags is helpful I can leave it, e.g. to manually run tests to check an issue is only with an old gitlab version or so (it doesn't do anything by default); otherwise I can remove it.

max-wittig:
You're right. It does make sense that it's configurable and the default is the latest! Thanks!

PR #1028, opened 2020-02-23
chore: use pytest for unit tests and coverage

nejch:


nejch:
I'll add this to #1024.

PR #1034, opened 2020-03-01
feat(api): add support for GitLab OAuth Applications using Applications API

filipowm:
GitLab can be used as an OAuth authentication provider. GitLab API allows creating Applications which could benefit from this feature. Currently `python-gitlab` does not support managing OAuth Apps, so I added this here.

Reference: https://docs.gitlab.com/ce/api/applications.html

### DIFF
--- a/docs/api-objects.rst
+++ b/docs/api-objects.rst
@@ -7,6 +7,7 @@ API examples
 
    gl_objects/access_requests
    gl_objects/appearance
+   gl_objects/applications
    gl_objects/emojis
    gl_objects/badges
    gl_objects/branches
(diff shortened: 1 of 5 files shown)


max-wittig:
Thank you for the MR!

I guess the commit message is a bit too long for commitlint. Would you mind to shorten it?

You can put the rest of the information in the body of the commit message.

filipowm:
@max-wittig Finally, all passes :-) There were some gotchas that were difficult to find why tests are failing, but it works now! Would be nice to have some basic "developing guide", but it's truly nice to have..

max-wittig:
@filipowm I know, we kinda need some guidance, but first we need to have a nicer test setup.

Issue #1025, opened 2020-02-22
async support & httpx-based python-gitlab


vishes-shell:
I've completely reworked python-gitlab to be async-compatible: https://github.com/vishes-shell/python-gitlab : 

- code it fully async compatible via awesome [`httpx`](https://github.com/encode/httpx)
- async tests are rewritten with [`pytest`](https://github.com/pytest-dev/pytest), [`pytest-asyncio`](https://github.com/pytest-dev/pytest-asyncio) and [`respx`](https://github.com/lundberg/respx)
- functional tests are also rewritten to be async

You can see changes in pull request https://github.com/vishes-shell/python-gitlab/pull/1 , i've tried to stick to project rules, but without `pytest` it would be harder...

I've seen some work that's done in https://github.com/python-gitlab/python-gitlab/pull/998 , and at the beginning i started on top of it, but found it hard to continue, due to some incomplete changes that are done, and trying to do sync stuff via `run_until_complete` would not work.

I would like to release it under `python-gitlab-async` module.

How i see it:

- i would hear your thoughts on this
- make some changes that you require (maybe drop `cli` support since `python-gitlab` do it good)
- update documentation
- transfer my fork under your `python-gitlab` namespace as `python-gitlab-async`, become a member of it
- release it in pypi.org
- take care of it

I think it's not that hard to keep up with changes from `master` branch and try to release as `python-gitlab` with same versions.

It was a pleasure to work with this brilliantly written library, and the same time you have a lot of `def`s to become `async def`s.:)

nejch:
Why not try to have the changes merged into this project? :) The maintainer has already signaled in #998 that he was willing to merge httpx-based work, they were just waiting with the other contributor for a stable httpx 1.0 to be released in the next month or so.

I totally agree with migrating to pytest though. All existing unit tests can already be run via pytest (with 2 lines of code changed); it would make proper testcases possible in functional tests both for the API and CLI, and both could then reuse fixtures, run entire suites independently of individual failures etc.

vishes-shell:
@nejch as far as I know it's really bad practice to implement every methods' sync functions through

```
loop = asyncio.get_event_loop()
loop.run_until_complete(async_foo())
```

So without sync methods that would break any sync projects that rely on this. Maybe I'm not aware of a good way to make async/sync compatible library.

nejch:
Ah sorry, the async/sync mess :/ I'm not sure either. It seems like some people are trying though, if this helps with future discussion:
[...]

PR #1036, opened 2020-03-04
Async and sync compatible wrapper

vishes-shell:
I've rebuild wrapper to support async and sync interface with support of `httpx`.

There is two gitlab clients: `Gitlab` and `AsyncGitlab`, initerfaces are the same, except most of the methods would return awaitable object in AsyncGitlab.

Related #1025


vishes-shell:
Here is some sort of notable changes:

1. [`response_content`](https://github.com/python-gitlab/python-gitlab/pull/1036/files#diff-c546ef8ace575bce9f52186c784ed721R29-R54) no longer support `chunk_size` since `httpx` does not support this. (https://github.com/encode/httpx/issues/394)
2. gitlab clients are moved to `client.py`
3. 3 implementations of gitlab clients: [`gitlab.BaseGitlab`](https://github.com/python-gitlab/python-gitlab/pull/1036/files#diff-dfcce258283be5679b4b4e318027ee43R48-R573) abstract implementation that share common methods and store all docsrings, [`gitlab.Gitlab`](https://github.com/python-gitlab/python-gitlab/pull/1036/files#diff-dfcce258283be5679b4b4e318027ee43R577-R783) is sync implementation with `httpx.Client`, [`gitlab.AsyncGitlab`](https://github.com/python-gitlab/python-gitlab/pull/1036/files#diff-dfcce258283be5679b4b4e318027ee43R787-R999) is async implementation with `httpx.AsyncClient`.
4. [`on_http_error`](https://github.com/python-gitlab/python-gitlab/pull/1036/files#diff-753797cb124e131242fc0f7a76115d34R265-R283) make nested since we can get coroutine and need to catch same error in awaiting response
5. [`RESTObject`](https://github.com/python-gitlab/python-gitlab/pull/1036/files#diff-88b7b76c9248a07e02a7119df9c96dcfR24-R60) must be created (with `create()`), not initialized, since arguments could be coroutines, and need to be awaited, and `__init__` does not support that interface
6. Everything should be returned, since it can be couroutin
7. every postprocess of response should be a function and be decorated with `gitlab.utils.awaitable_postprocess`, as it's done with `_update_attrs` and other edge cases
8. unit tests are run by `pytest` and run against _async_ and _sync_ client implementations
9. since unit tests are run with async gitlab, tests are required to be async, but fetching result is done through fixture that knows which client is used and `return value` if that's _sync_ version or `return await value` if that's `async` version.
10. `GitlabList` support two versions of iteration
11. `requests` are completely removed

nejch:
>     1. `response_content` no longer support `chunk_size` since `httpx` does not support
>         this. ([encode/httpx#394](https://github.com/encode/httpx/issues/394))

I just noticed something: `chunk_size` is a configurable argument in several public methods (for blobs/snippet content, export download, etc) so it's possible people are using it and this would be a breaking change, right? Since you're replacing it with `aiter_bytes` and using chunks there, is there any way to preserve backward compatibility (I haven't checked the implementation)? Or wait for the upstream issue to resolve?

>     3. 3 implementations of gitlab clients: `gitlab.BaseGitlab` abstract implementation that share common methods and store all docsrings, `gitlab.Gitlab` is sync implementation with `httpx.Client`, `gitlab.AsyncGitlab` is async implementation with `httpx.AsyncClient`.
>     9. since unit tests are run with async gitlab, tests are required to be async, but fetching result is done through fixture that knows which client is used and `return value` if that's _sync_ version or `return await value` if that's `async` version.
That's really cool :)

max-wittig:
@vishes-shell Thanks for all the work that you're doing!

People could have been using chunk_size, by providing a different `requests` instance, right?
(Which is also not supported anymore then?)

But we have nothing about this officially in the docs, so I would say we can solve this by mentioning it in the release notes. We anyway need to bump the version to 3.X.

nejch:
@max-wittig I tried earlier and at least in some cases it can even be just:

```python
project = gl.projects.get(1)
export = project.exports.create({})
dl = export.download(chunk_size=512)
```
From what I see in the current solution by @vishes-shell this would still work and wouldn't break anything, the argument would just be ignored by `response_content()`. But if anyone relies on `chunk_size` for some specific reason, it might break their behavior. I'm not sure of all the use cases but it seems useful especially for project exports: https://stackoverflow.com/questions/46205586/why-to-use-iter-content-and-chunk-size-in-python-requests/46205745

test-requirements.txt
```diff
@@ -7,3 +7,7 @@ jinja2
 mock
 sphinx>=1.3
 sphinx_rtd_theme
+requests>=2.22.0
+respx>=0.12.1,<0.13
+pytest
+pytest-asyncio
```
graingert, line 13: can you use the anyio pytest plugin so you test on both trio and asyncio?

darkdragon-001:
What's the current state? Could you publish wheels with your Github releases such that one can install it via `requirements.txt`? Would be quite convenient for your scripts. Breaking changes are not problem since they can easily be adjusted.

darkdragon-001:
When using your library, I get the following warning:
> httpx-0.18.2-py3.9.egg/httpx/_client.py:1978: UserWarning: Unclosed <httpx.AsyncClient object at 0x7ffabf61e280>. See https://www.python-httpx.org/async/#opening-and-closing-clients for details.

darkdragon-001:
Btw: `chunk_size` is [supported](https://github.com/encode/httpx/pull/1277) by httpx now.

graingert:
> When using your library, I get the following warning:
> 
> > httpx-0.18.2-py3.9.egg/httpx/_client.py:1978: UserWarning: Unclosed <httpx.AsyncClient object at 0x7ffabf61e280>. See https://www.python-httpx.org/async/#opening-and-closing-clients for details.
> 
> Apart from that, it works like a charm 😍

You'll need to use the client with `async with AsyncGitlab() as gitlab: ...`

darkdragon-001:
I am using it since a few months without any problems. `chunk_size` is implemented by now as well. Any other blockers? Or would it be accepted when updated to implement the latest features in the main branch?

lmilbaum:
As part of onboarding myself into this project, I stumbled into this PR. Could someone provide a use case when an async wrapper is being used?

darkdragon-001:
@lmilbaum We are evaluating the jobs from the last day via script such that our QA team has a good overview about possible regressions or problems with our hardware text boxes (e.g. when the same code runs on one runner but not the other one). As network is slow, waiting for the answer of the previous request before submitting the next one takes quite a lot of time. If we can submit all requests asynchronously and just wait for all of them to finish in parallel, we have reduced the run time of our script by a multitude. Unfortunately, this MR is outdated and newer features are not available yet...

nejch:
The wrapper here https://github.com/pan-net-security/aio-gitlab/ has a few points similar to what @darkdragon-001 described above. But true, this PR would need quite a bit of rework, or be used as starting point for a new one, since the library has changed a bit since then. I personally have not yet had enough need for this to work on it.

Another option would be to first switch from requests to httpx for the sync client (with a breaking change, because we're currently allowing people to bring their own requests sessions) and smooth out any kinks, then add async later, to make the transition less painful.

lmilbaum:
@darkdragon-001 and @nejch thanks for your feedback. Now, I can understand better the need for an async capability. 
Would it make sense to move this discussion to [Discussions](https://github.com/python-gitlab/python-gitlab/discussions) while closing this PR? 
The first discussion point might be the migration from `requests` to `httpx` as suggested by @nejch.

nejch:
> @darkdragon-001 and @nejch thanks for your feedback. Now, I can understand better the need for an async capability. Would it make sense to move this discussion to [Discussions](https://github.com/python-gitlab/python-gitlab/discussions) while closing this PR? The first discussion point might be the migration from `requests` to `httpx` as suggested by @nejch.

@lmilbaum thanks for looking into this! I'd keep this PR open even if as draft, just so it stays on our radar as there's a lot of useful discussion and work done here. And just so we remember to credit @vishes-shell as author whether they want to still work on this or not.

I'd maybe open a new issue to plan the switch to httpx as that would be for v4.0.0. We could then get rid of `types-requests` and `requests-toolbelt` I assume, as httpx is fully typed and does multipart encoding well IIRC.

PR #1038, opened 2020-03-07
Fix: do not require empty data dict for create()

nejch:
There's a quirk in the `CreateMixin` when checking for required attributes in [`_check_missing_create_attrs`](https://github.com/python-gitlab/python-gitlab/blob/e5afb554bf4bcc28555bde4030f50558f175a53b/gitlab/mixins.py#L153), which requires a data dict. But some create methods don't require additional data, e.g. if the only required attribute is already provided by their parent manager id (e.g. for project export, see https://docs.gitlab.com/ee/api/project_import_export.html).

This means the docs are sometimes asking people to pass an empty dict which I think is confusing for newcomers (and maybe unpythonic?).

This small fix allows using `create()` without passing a dict. The `UpdateMixin` is doing something similar with [`new_data = new_data or {}`](https://github.com/python-gitlab/python-gitlab/blob/e5afb554bf4bcc28555bde4030f50558f175a53b/gitlab/mixins.py#L250-L265) but I thought this is a bit more explicit with None.

I didn't add any checks as the data is already validated in `_check_missing_create_attrs` based on each Manager's `_create_attrs` and is already tested (see below), so I hope I covered everything :)  https://github.com/python-gitlab/python-gitlab/blob/e5afb554bf4bcc28555bde4030f50558f175a53b/gitlab/tests/test_mixins.py#L265-L276.


max-wittig:
You're right. This was kinda unpythonic before. Thanks for the change!

PR #1011, opened 2020-02-12
fix(objects): Added approval rule name to data

charlesfayal:
The new gitlab api needs a name for the approval rules, added a default one to maintain the python-gitlab api.

The prior pull request wasn't ready for merge, needed this fix.

PR #1039, opened 2020-03-08
Fix/set approvers

max-wittig:


max-wittig:
Closes https://github.com/python-gitlab/python-gitlab/pull/1011

Issue #1037, opened 2020-03-06
Support for group import/export API


nejch:
I was checking exports for work and realized 12.8 introduced (experimental) support for Group import/export. Cool! :)

https://docs.gitlab.com/ce/api/group_import_export.html

PR #1040, opened 2020-03-08
test: update tests and params for project export/import

nejch:
This will need to be changed for #1025 but I'd like to have some tests before refactoring Export code for #1037.

### DIFF
--- a/gitlab/tests/objects/test_projects.py
+++ b/gitlab/tests/objects/test_projects.py
@@ -15,6 +15,108 @@ headers = {"content-type": "application/json"}
 binary_content = b"binary content"
 
 
+@urlmatch(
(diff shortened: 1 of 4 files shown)


nejch:
I updated this PR as I found some missing parameters for imports while adding tests, @max-wittig could you take a look?

Currently the new mocks/tests are a bit inconsistent with the existing approach but I liked that it made the tests very readable (I first saw it [here](https://github.com/ansible/ansible/tree/devel/test/units/modules/source_control/gitlab)). I can change this back to `with HTTMock` if it's more consistent, or change the other mocks in this module, but I guess in any case with the switch to httpx this will become kind of irrelevant :P

max-wittig:
This does look a lot more readable! Thanks a lot!

This is gonna be a good addition, until we go for httpx!

PR #1042, opened 2020-03-09
fix(docs): additional project statistics example

khuedoan98:
Using `get()['fetches']['total']` in the example will raise an error:

```
TypeError: 'ProjectAdditionalStatistics' object is not subscriptable
```

Updated to use new method with `get().fetches['total']`


max-wittig:
@khuedoan98 Thanks for the fix!

PR #1045, opened 2020-03-09
Revert "feat: use keyset pagination by default for `all=True`"

max-wittig:
Reverts python-gitlab/python-gitlab#1003


bufferoverflow:
autsch

PR #1049, opened 2020-03-18
docs: fix comment of prev_page()

donhui:
It shoud be `prev`, not `next`.

### DIFF
--- a/gitlab/__init__.py
+++ b/gitlab/__init__.py
@@ -800,9 +800,9 @@ class GitlabList(object):
 
     @property
     def prev_page(self):
-        """The next page number.
+        """The prev page number.
 
-        If None, the current page is the last.
+        If None, the current page is the first.
         """
         return int(self._prev_page) if self._prev_page else None
 
--- a/gitlab/__init__.py
+++ b/gitlab/__init__.py
@@ -800,7 +800,7 @@ class GitlabList(object):
 
     @property
     def prev_page(self):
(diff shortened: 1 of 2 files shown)


gitlab/__init__.py
```diff
@@ -800,9 +800,9 @@ def current_page(self):
 
     @property
     def prev_page(self):
-        """The next page number.
+        """The prev page number.
```
nejch:
```suggestion
        """The previous page number.
```

gitlab/base.py
```diff
@@ -194,9 +194,9 @@ def current_page(self):
 
     @property
     def prev_page(self):
-        """The next page number.
+        """The prev page number.
```
nejch:
```suggestion
        """The previous page number.
```

donhui:
> Thank you @donhui, nice catch! And thanks for the commitlint fix :)
> 
> I just added 2 small suggestions (sorry for the nitpick). I think we can have the full word in the docs; as is done in https://docs.gitlab.com/ee/api/#other-pagination-headers.

Thank you, I have applied your suggestions.

nejch:
That was quick, thanks! I've approved but Travis is currently down for maintenance, so we just have to wait for that to go green.

If you find any more inconsistencies, feel free to open a new PR :)

donhui:
OK, thanks for your review.

PR #1052, opened 2020-03-19
feat(api): add support for Gitlab Deploy Token API

machine424:
There is a conflict between the required parameters of `POST /projects/:id/deploy_tokens` (and `POST /groups/:id/deploy_tokens`) as documented here : `https://docs.gitlab.com/ce/api/deploy_tokens.html#create-a-project-deploy-token` and required parameters that the API waits for: `gitlab.com` that is uses 12.9.0-pre 4b94245907d (As I write these lines) and the docker image `gitlab/gitlab-ce:nightly` require four parameters: `name`, `expires_at`, `username` and `scopes`.

Creating a deploy token from `gitlab.com` UI agrees with the Deploy Token API documentation: only the `name` and the `scopes` are required.

The managers I added require the four parameters to make tests pass with `gitlab/gitlab-ce:nightly`.

the documentation on the other hand reflects the provided Deploy Token API documentation.

I've created an issue to make this clear: https://gitlab.com/gitlab-org/gitlab/-/issues/211878

### DIFF
--- a/docs/gl_objects/deploy_tokens.rst
+++ b/docs/gl_objects/deploy_tokens.rst
@@ -1,0 +1,137 @@
+#######
+Deploy tokens
+#######
+
+Deploy tokens allow read-only access to your repository and registry images
+without having a user and a password.
+
+Deploy tokens
+=============
+
+This endpoint requires admin access.
+
+Reference
+---------
+
+* v4 API:
+
+  + :class:`gitlab.v4.objects.DeployToken`
+  + :class:`gitlab.v4.objects.DeployTokenManager`
+  + :attr:`gitlab.Gitlab.deploytokens`
+
+* GitLab API: https://docs.gitlab.com/ce/api/deploy_tokens.html
+
+Examples
+--------
+
+Use the ``list()`` method to list all deploy tokens across the GitLab instance.
+
+::
+
+    # List deploy tokens
+    deploy_tokens = gl.deploytokens.list()
+
+Project deploy tokens
+=====================
+
+This endpoint requires project maintainer access or higher.
+
+Reference
+---------
+
+* v4 API:
+
+  + :class:`gitlab.v4.objects.ProjectDeployToken`
+  + :class:`gitlab.v4.objects.ProjectDeployTokenManager`
+  + :attr:`gitlab.v4.objects.Project.deploytokens`
+
+* GitLab API: https://docs.gitlab.com/ce/api/deploy_tokens.html#project-deploy-tokens
+
+Examples
+--------
+
+List the deploy tokens for a project::
+
+    deploy_tokens = project.deploytokens.list()
+
+Create a new deploy token to access registry images of a project:
+
+In addition to required parameters ``name`` and ``scopes``, this method accepts
+the following parameters:
+
+* ``expires_at`` Expiration date of the deploy token. Does not expire if no value is provided.
+* ``username`` Username for deploy token. Default is ``gitlab+deploy-token-{n}``
+
+
+::
+
+    deploy_token = project.deploytokens.create({'name': 'token1', 'scopes': ['read_registry'], 'username':'', 'expires_at':''})
+    # show its id
+    print(deploy_token.id)
+    # show the token value. Make sure you save it, you won't be able to access it again.
+    print(deploy_token.token)
+
+.. warning::
+
+   With GitLab 12.9, even though ``username`` and ``expires_at`` are not required, they always have to be passed to the API.
+   You can set them to empty strings, see: https://gitlab.com/gitlab-org/gitlab/-/issues/211878.
+   Also, the ``username``'s value is ignored by the API and will be overriden with ``gitlab+deploy-token-{n}``,
+   see: https://gitlab.com/gitlab-org/gitlab/-/issues/211963
+   These issues were fixed in GitLab 12.10.
+
+Remove a deploy token from the project::
+
+    deploy_token.delete()
+    # or
+    project.deploytokens.delete(deploy_token.id)
+
+
+Group deploy tokens
+===================
+
+Reference
+---------
+
+* v4 API:
+
+  + :class:`gitlab.v4.objects.GroupDeployToken`
+  + :class:`gitlab.v4.objects.GroupDeployTokenManager`
+  + :attr:`gitlab.v4.objects.Group.deploytokens`
+
+* GitLab API: https://docs.gitlab.com/ce/api/deploy_tokens.html#group-deploy-tokens
+
+Examples
+--------
+
+List the deploy tokens for a group::
+
+    deploy_tokens = group.deploytokens.list()
+
+Create a new deploy token to access all repositories of all projects in a group:
+
+In addition to required parameters ``name`` and ``scopes``, this method accepts
+the following parameters:
+
+* ``expires_at`` Expiration date of the deploy token. Does not expire if no value is provided.
+* ``username`` Username for deploy token. Default is ``gitlab+deploy-token-{n}``
+
+::
+
+    deploy_token = group.deploytokens.create({'name': 'token1', 'scopes': ['read_repository'], 'username':'', 'expires_at':''})
+    # show its id
+    print(deploy_token.id)
+
+.. warning::
+
+   With GitLab 12.9, even though ``username`` and ``expires_at`` are not required, they always have to be passed to the API.
+   You can set them to empty strings, see: https://gitlab.com/gitlab-org/gitlab/-/issues/211878.
+   Also, the ``username``'s value is ignored by the API and will be overriden with ``gitlab+deploy-token-{n}``,
+   see: https://gitlab.com/gitlab-org/gitlab/-/issues/211963
+   These issues were fixed in GitLab 12.10.
+
+Remove a deploy token from the group::
+
+    deploy_token.delete()
+    # or
+    group.deploytokens.delete(deploy_token.id)
+
--- a/gitlab/v4/objects.py
+++ b/gitlab/v4/objects.py
@@ -695,6 +695,43 @@ class DeployKeyManager(ListMixin, RESTManager):
     _obj_cls = DeployKey
 
 
+class DeployToken(ObjectDeleteMixin, RESTObject):
+    pass
+
+
+class DeployTokenManager(ListMixin, RESTManager):
+    _path = "/deploy_tokens"
+    _obj_cls = DeployToken
+
+
+class ProjectDeployToken(ObjectDeleteMixin, RESTObject):
+    pass
+
+
+class ProjectDeployTokenManager(ListMixin, CreateMixin, DeleteMixin, RESTManager):
+    _path = "/projects/%(project_id)s/deploy_tokens"
+    _from_parent_attrs = {"project_id": "id"}
+    _obj_cls = ProjectDeployToken
+    _create_attrs = (
+        ("name", "scopes",),
+        ("expires_at", "username",),
+    )
+
+
+class GroupDeployToken(ObjectDeleteMixin, RESTObject):
+    pass
+
+
+class GroupDeployTokenManager(ListMixin, CreateMixin, DeleteMixin, RESTManager):
+    _path = "/groups/%(group_id)s/deploy_tokens"
+    _from_parent_attrs = {"group_id": "id"}
+    _obj_cls = GroupDeployToken
+    _create_attrs = (
+        ("name", "scopes",),
+        ("expires_at", "username",),
+    )
+
+
 class NotificationSettings(SaveMixin, RESTObject):
     _id_attr = None
 
@@ -1301,6 +1338,7 @@ class Group(SaveMixin, ObjectDeleteMixin, RESTObject):
         ("subgroups", "GroupSubgroupManager"),
         ("variables", "GroupVariableManager"),
         ("clusters", "GroupClusterManager"),
+        ("deploytokens", "GroupDeployTokenManager"),
     )
 
     @cli.register_custom_action("Group", ("to_project_id",))
@@ -4212,6 +4250,7 @@ class Project(SaveMixin, ObjectDeleteMixin, RESTObject):
         ("clusters", "ProjectClusterManager"),
         ("additionalstatistics", "ProjectAdditionalStatisticsManager"),
         ("issuesstatistics", "ProjectIssuesStatisticsManager"),
+        ("deploytokens", "ProjectDeployTokenManager"),
     )
 
     @cli.register_custom_action("Project", ("submodule", "branch", "commit_sha"))
--- a/tools/cli_test_v4.sh
+++ b/tools/cli_test_v4.sh
@@ -195,14 +195,6 @@ testcase "project upload" '
         --filename '$(basename $0)' --filepath '$0' >/dev/null 2>&1
 '
 
-testcase "project deletion" '
-    GITLAB project delete --id "$PROJECT_ID"
-'
-
-testcase "group deletion" '
-    OUTPUT=$(try GITLAB group delete --id $GROUP_ID)
-'
-
 testcase "application settings get" '
     GITLAB application-settings get >/dev/null 2>&1
 '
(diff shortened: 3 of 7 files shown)


docs/gl_objects/deploy_tokens.rst
```diff
@@ -0,0 +1,120 @@
+#######
+Deploy tokens
+#######
+
+Deploy tokens allow read-only access to your repository and registry images
+without having a user and a password.
+
+Deploy tokens can be managed by maintainers only.
```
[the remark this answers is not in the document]
max-wittig: I would say we remove this comment and add the required permissions the sections below.

docs/gl_objects/deploy_tokens.rst
```diff
@@ -0,0 +1,120 @@
...
+
+Deploy tokens
+=============
+
```
max-wittig: This endpoint requires admin access

docs/gl_objects/deploy_tokens.rst
```diff
@@ -0,0 +1,120 @@
...
+This requires admin access.
+
+::
+
+    # List deploy tokens
+    deploy_tokens = gl.deploytokens.list()
+
+Project deploy tokens
+=====================
+
```
max-wittig: This endpoint requires maintainer project access or higher.

docs/gl_objects/deploy_tokens.rst
```diff
@@ -0,0 +1,120 @@
...
+Examples
+--------
+
+List the deploy tokens for a project::
+
+    deploy_tokens = project.deploytokens.list()
+
+Create a new deploy token to access registry images of a project:
+
+This method accepts the following none required parameters:
```
max-wittig:
```suggestion
This method accepts the following parameters:
```

docs/gl_objects/deploy_tokens.rst
```diff
@@ -0,0 +1,120 @@
...
+
+* ``name`` New deploy token’s name
+* ``expires_at`` Expiration date for the deploy token. Does not expire if no value is provided.
```
max-wittig:
```suggestion
* ``expires_at`` Expiration date of the deploy token. Does not expire if no value is provided.
```

docs/gl_objects/deploy_tokens.rst
```diff
@@ -0,0 +1,120 @@
...
+Reference
+---------
+
+* v4 API:
+
+  + :class:`gitlab.v4.objects.GroupDeployToken`
+  + :class:`gitlab.v4.objects.GroupDeployTokenManager`
+  + :attr:`gitlab.v4.objects.Group.deploytokens`
+
+* GitLab API: https://docs.gitlab.com/ce/api/deploy_tokens.html
```
max-wittig:
```suggestion
* GitLab API: https://docs.gitlab.com/ce/api/deploy_tokens.html#group-deploy-tokens
```

docs/gl_objects/deploy_tokens.rst
```diff
@@ -0,0 +1,120 @@
...
+Reference
+---------
+
+* v4 API:
+
+  + :class:`gitlab.v4.objects.ProjectDeployToken`
+  + :class:`gitlab.v4.objects.ProjectDeployTokenManager`
+  + :attr:`gitlab.v4.objects.Project.deploytokens`
+
+* GitLab API: https://docs.gitlab.com/ce/api/deploy_tokens.html
```
max-wittig:
```suggestion
* GitLab API: https://docs.gitlab.com/ce/api/deploy_tokens.html#project-deploy-tokens
```

docs/gl_objects/deploy_tokens.rst
```diff
@@ -0,0 +1,120 @@
...
+Examples
+--------
+
+List the deploy tokens for a group::
+
+    deploy_tokens = group.deploytokens.list()
+
+Create a new deploy token to access all repositories of all projects in a group:
+
+This method accepts the following none required parameters:
```
max-wittig:
```suggestion
This method accepts the following parameters:
```

gitlab/v4/objects.py
```diff
@@ -695,6 +695,43 @@ class DeployKeyManager(ListMixin, RESTManager):
...
+class ProjectDeployToken(ObjectDeleteMixin, RESTObject):
+    pass
+
+
+class ProjectDeployTokenManager(ListMixin, CreateMixin, DeleteMixin, RESTManager):
+    _path = "/projects/%(project_id)s/deploy_tokens"
+    _from_parent_attrs = {"project_id": "id"}
+    _obj_cls = ProjectDeployToken
+    _create_attrs = (
+        ("name", "expires_at", "username", "scopes",),
```
max-wittig: Not all of these are required, right?

Only `name` and `scopes` should be required. The rest should be in the currently empty tuple

```suggestion
        ("name", "scopes",),
```

gitlab/v4/objects.py
```diff
@@ -695,6 +695,43 @@ class DeployKeyManager(ListMixin, RESTManager):
...
+        tuple(),
```
max-wittig:
```suggestion
        ("expires_at", "username",),
```

gitlab/v4/objects.py
```diff
@@ -695,6 +695,43 @@ class DeployKeyManager(ListMixin, RESTManager):
...
+class GroupDeployToken(ObjectDeleteMixin, RESTObject):
+    pass
+
+
+class GroupDeployTokenManager(ListMixin, CreateMixin, DeleteMixin, RESTManager):
+    _path = "/groups/%(group_id)s/deploy_tokens"
+    _from_parent_attrs = {"group_id": "id"}
+    _obj_cls = GroupDeployToken
+    _create_attrs = (
+        ("name", "expires_at", "username", "scopes",),
```
max-wittig:
```suggestion
        ("name", "scopes",),
```

gitlab/v4/objects.py
```diff
@@ -695,6 +695,43 @@ class DeployKeyManager(ListMixin, RESTManager):
...
+        tuple(),
```
max-wittig:
```suggestion
        ("expires_at", "username",),
```

tools/build_test_env.sh
```diff
@@ -30,7 +30,7 @@ NOVENV=
 PY_VER=3
 API_VER=4
 GITLAB_IMAGE="gitlab/gitlab-ce"
-GITLAB_TAG="latest"
+GITLAB_TAG="nightly"
```
max-wittig: I guess this could hurt, if GitLab changes one of their APIs around a couple times.

Let's stay at stable and wait for the 22nd

docs/gl_objects/deploy_tokens.rst
```diff
@@ -0,0 +1,120 @@
...
+* ``expires_at`` Expiration date for the deploy token. Does not expire if no value is provided.
+* ``username`` Username for deploy token. Default is ``gitlab+deploy-token-{n}``
+
+::
+
+    deploy_token = group.deploytokens.create({'name': 'token1', 'scopes': ['read_repository']})
+    # show its id
+    print(deploy_token.id)
+
+Revoke (delete) the deploy token::
```
max-wittig: It's not revoked, it's just deleted in GitLab

```suggestion
Removes a deploy token from the group::
```

docs/gl_objects/deploy_tokens.rst
```diff
@@ -0,0 +1,120 @@
...
+
+::
+
+    deploy_token = project.deploytokens.create({'name': 'token1', 'scopes': ['read_registry']})
+    # show its id
+    print(deploy_token.id)
+    # show the token value. Make sure you save it, you won't be able to access it again.
+    print(deploy_token.token)
+
+Revoke (delete) the deploy token::
```
max-wittig:
```suggestion
Removes a deploy token from the project::
```

docs/gl_objects/deploy_tokens.rst
```diff
@@ -0,0 +1,120 @@
...
+  + :attr:`gitlab.Gitlab.deploytokens`
+
+* GitLab API: https://docs.gitlab.com/ce/api/deploy_tokens.html
+
+Examples
+--------
+
+Use the ``list()`` method to list all deploy tokens across the GitLab instance.
+
+This requires admin access.
```
max-wittig: Better place this line above

```suggestion

```

gitlab/v4/objects.py
max-wittig: Ah I guess one should read the description 😆

docs/gl_objects/deploy_tokens.rst
machine424: `for` the token to be created.

gitlab/v4/objects.py
machine424: As I explained is the description:
```
The managers I added require the four parameters to make tests pass with gitlab/gitlab-ce:nightly.
```
I re-defined them as not required.
It turned out that the API doesn't comply with the spec (the doc), waiting for https://gitlab.com/gitlab-org/gitlab/-/issues/211878 to be fixed.

gitlab/v4/objects.py
machine424: See above.

tools/build_test_env.sh
machine424: It was temporary to test on Travis, `latest` doesn't ship the API yet.

machine424: I hope the new release will ship the feature; there're some bugs to be fixed: https://gitlab.com/gitlab-org/gitlab/-/issues/211963, https://gitlab.com/gitlab-org/gitlab/-/issues/211878

nejch: I started #1054 as a first step to try and improve on this problem, so it can still be tested. We will need proper skippable test cases (based on gitlab versions) to make this smoother, hopefully very soon :P.

machine424:
Cool, yes of course, I’ll add them.
But first, we have some problems to clear up, I don’t think the description of the PR and my last comments were clear, so I’ll rephrase:
1- There is an incoherence (at least for me) between the Gitlab documentation and the API:
Here in the doc: https://docs.gitlab.com/ee/api/deploy_tokens.html#create-a-project-deploy-token they state that `expires_at` and `username` are not required to create a deploy token. BUT the API waits for them. I opened an Issue about this here: https://gitlab.com/gitlab-org/gitlab/-/issues/211878 and Gitlab folks recognised the bug and proposed a workaround: PROVIDE empty values for `expires_at` and `username` ==> should we mark `expires_at` and `username` as required in `_create_attrs` ? should we override `create` method for the two Managers to pass empty `expires_at` and `username` to the API if the user didn’t provide them ?

2- Another problem: the API doesn’t take provided `username` into consideration, it overrides it, see: https://gitlab.com/gitlab-org/gitlab/-/issues/211963 => it’s not of python-gitlab’s business, but we should warn the users, do something else ?


I added some tests here : https://github.com/python-gitlab/python-gitlab/pull/1052/commits/e9814b2d7d8f5c8375556ea771d74f7658230c55 to illustrate my sayings.

nejch:
> Cool, yes of course, I’ll add them.
> But first, we have some problems to clear up, I don’t think the description of the PR and my last comments were clear, so I’ll rephrase:

Nice, thanks! Sorry I wasn't fully following the previous round on this but I'll chime in.

> 1- There is an incoherence (at least for me) between the Gitlab documentation and the API:
> Here in the doc: https://docs.gitlab.com/ee/api/deploy_tokens.html#create-a-project-deploy-token they state that `expires_at` and `username` are not required to create a deploy token. BUT the API waits for them. I opened an Issue about this here: https://gitlab.com/gitlab-org/gitlab/-/issues/211878 and Gitlab folks recognised the bug and proposed a workaround: PROVIDE empty values for `expires_at` and `username` ==> should we mark `expires_at` and `username` as required in `_create_attrs` ? should we override `create` method for the two Managers to pass empty `expires_at` and `username` to the API if the user didn’t provide them ?
>
> 2- Another problem: the API doesn’t take provided `username` into consideration, it overrides it, see: https://gitlab.com/gitlab-org/gitlab/-/issues/211963 => it’s not of python-gitlab’s business, but we should warn the users, do something else ?
> 
> I added some tests here : [e9814b2](https://github.com/python-gitlab/python-gitlab/commit/e9814b2d7d8f5c8375556ea771d74f7658230c55) to illustrate my sayings.

I agree the empty value workaround won't work for libraries. How about just documenting this as a warning in the docs with links to upstream issues (at least that's what gitlab does for their open issues) - for both the optional attributes being required, and the username having no effect?

And once it's fixed upstream the exact GitLab versions affected can be defined. The tests will just need to provide the optional args as well for now. As you said it's the API behavior and not the wrapper. I see `go-gitlab` added this recently and aren't handling it on their side.

This way you don't have additional logic or breaking it for specific gitlab versions if a `create()` override is done and removed at some point later when the upstream fix is done.

Hm: Looking at it a bit more, seems like the upstream fix for the optional args is just changing required to optional ([here](https://gitlab.com/gitlab-org/gitlab/-/blob/master/lib/api/deploy_tokens.rb#L55-58) and [here](https://gitlab.com/gitlab-org/gitlab/-/blob/master/lib/api/deploy_tokens.rb#L116-119). I'll see if I can do a few lines of Ruby, maybe then it can already get into a patch release soonish :P

nejch:
I've started an upstream PR for the optional params, let's see if that gets merged into the next patch release, then we might not need to worry about it here. Not quite sure yet what's going on with `username` being ignored though..

machine424:
Great.
I've found another incoherence while adjusting some tests: https://gitlab.com/gitlab-org/gitlab/-/issues/212523
I have the impression that I'm testing the API.

tools/cli_test_v4.sh
```diff
@@ -215,3 +207,82 @@ testcase "values from files" '
...
+    echo $CREATE_PROJECT_DEPLOY_TOKEN_OUTPUT | grep "scopes: " | grep -q "read_registry"
+'
+# Uncomment once https://gitlab.com/gitlab-org/gitlab/-/issues/211963 is fixed
+#testcase "create project deploy token" '
+#    echo $CREATE_PROJECT_DEPLOY_TOKEN_OUTPUT | grep  "username: root"
+#'
+
+# Remove once https://gitlab.com/gitlab-org/gitlab/-/issues/211963 is fixed
+testcase "create project deploy token" '
+    echo $CREATE_PROJECT_DEPLOY_TOKEN_OUTPUT | grep "gitlab+deploy-token"
```
max-wittig:
```suggestion
    echo $CREATE_PROJECT_DEPLOY_TOKEN_OUTPUT | grep -q "gitlab+deploy-token"
```

tools/cli_test_v4.sh
```diff
@@ -215,3 +207,82 @@ testcase "values from files" '
...
+'
+testcase "create project deploy token" '
+    echo $CREATE_PROJECT_DEPLOY_TOKEN_OUTPUT | grep -q "expires-at: 2021-09-09T00:00:00.000Z"
+'
+testcase "create project deploy token" '
+    echo $CREATE_PROJECT_DEPLOY_TOKEN_OUTPUT | grep "scopes: " | grep -q "read_registry"
+'
+# Uncomment once https://gitlab.com/gitlab-org/gitlab/-/issues/211963 is fixed
+#testcase "create project deploy token" '
+#    echo $CREATE_PROJECT_DEPLOY_TOKEN_OUTPUT | grep  "username: root"
```
max-wittig:
```suggestion
#    echo $CREATE_PROJECT_DEPLOY_TOKEN_OUTPUT | grep -q  "username: root"
```

machine424:
@max-wittig , when the client serves the server ! I'll not add more tests, I need this feature, lool.
I followed @nejch and I submitted a MR to correct the username overriding: https://gitlab.com/gitlab-org/gitlab/-/merge_requests/28175, the third bug https://gitlab.com/gitlab-org/gitlab/-/issues/212523 will be fixed by a Gitlab folk, Steve. 
Once the 12.10 is released, I'll enable the tests that are commented.

docs/gl_objects/deploy_tokens.rst
```diff
@@ -0,0 +1,135 @@
...
+    print(deploy_token.id)
+    # show the token value. Make sure you save it, you won't be able to access it again.
+    print(deploy_token.token)
+
+.. warning::
+
+   Even though ``username`` and ``expires_at`` are not required, they always have to be passed to the API,
+   you can set them to empty strings, see: https://gitlab.com/gitlab-org/gitlab/-/issues/211878.
+   The ``username``'s value is ignored by the API and will be overriden with ``gitlab+deploy-token-{n}``,
+   see: https://gitlab.com/gitlab-org/gitlab/-/issues/211963
```
nejch:
```suggestion
   With GitLab 12.9, even though ``username`` and ``expires_at`` are not
   required, they always have to be passed to the API. You can set them to
   empty strings, see: https://gitlab.com/gitlab-org/gitlab/-/issues/211878.
   Also, the ``username``'s value is ignored by the API in GitLab 12.9 and
   will be overridden with ``gitlab+deploy-token-{n}``, see:
   https://gitlab.com/gitlab-org/gitlab/-/issues/211963
   
   These issues were fixed in GitLab 12.10.
```

docs/gl_objects/deploy_tokens.rst
```diff
@@ -0,0 +1,135 @@
...
+    deploy_token = group.deploytokens.create({'name': 'token1', 'scopes': ['read_repository'], 'username':'', 'expires_at':''})
+    # show its id
+    print(deploy_token.id)
+
+.. warning::
+
+   Even though ``username`` and ``expires_at`` are not required, they always have to be passed to the API,
+   you can set them to empty strings, see: https://gitlab.com/gitlab-org/gitlab/-/issues/211878.
+   The ``username``'s value is ignored by the API and will be overriden with ``gitlab+deploy-token-{n}``,
+   see: https://gitlab.com/gitlab-org/gitlab/-/issues/211963
```
nejch:
```suggestion
   With GitLab 12.9, even though ``username`` and ``expires_at`` are not
   required, they always have to be passed to the API. You can set them to
   empty strings, see: https://gitlab.com/gitlab-org/gitlab/-/issues/211878.
   Also, the ``username``'s value is ignored by the API in GitLab 12.9 and
   will be overridden with ``gitlab+deploy-token-{n}``, see:
   https://gitlab.com/gitlab-org/gitlab/-/issues/211963
   
   These issues were fixed in GitLab 12.10.
```

nejch:
> @max-wittig , when the client serves the server ! I'll not add more tests, I need this feature, lool.
> I followed @nejch and I submitted a MR to correct the username overriding: https://gitlab.com/gitlab-org/gitlab/-/merge_requests/28175, the third bug https://gitlab.com/gitlab-org/gitlab/-/issues/212523 will be fixed by a Gitlab folk, Steve.
> Once the 12.10 is released, I'll enable the tests that are commented.

Wow cool, I never got to that one, thanks for that :) I approved with the two suggestions as you mentioned, it seems both upstream MR's will be merged in time for 12.10.

Yeah, thanks for adding all the CLI/integration tests, that's really extensive :grin:

machine424:
> Wow cool, I never got to that one, thanks for that :) I approved with the two suggestions as you mentioned, it seems both upstream MR's will be merged in time for 12.10.
> 
> Yeah, thanks for adding all the CLI/integration tests, that's really extensive 😁

There're never that extensive, but I really don't want to open a new issue on Gitlab :wink:

PR #1053, opened 2020-03-20
feat: add create from template args to ProjectManager

lassimus:
This commit adds the v4 Create project attributes necessary to create a
project from a project, instance, or group level template as documented
in https://docs.gitlab.com/ee/api/projects.html#create-project

### DIFF
--- a/gitlab/v4/objects.py
+++ b/gitlab/v4/objects.py
@@ -4788,6 +4788,10 @@ class ProjectManager(CRUDMixin, RESTManager):
             "avatar",
             "printing_merge_request_link_enabled",
             "ci_config_path",
+            "template_name",
+            "template_project_id",
+            "use_custom_template",
+            "group_with_project_templates_id",
         ),
     )
     _update_attrs = (


lassimus:
`**kwargs` is useful when you're using the library, but if the optional attributes aren't added, they aren't usable in the cli.

I'm currently using the cli for chatops ci jobs, as it's much less maintenance than a purpose-built script that uses the library. I know you guys have talked about removing the cli completely, so I understand if you don't want to accept the PR for those reasons.

If you accept this PR, I'm certainly open to contributing more api updates.

nejch:
> 
> 
> `**kwargs` is useful when you're using the library, but if the optional attributes aren't added, they aren't usable in the cli.
> 
> I'm currently using the cli for chatops ci jobs, as it's much less maintenance than a purpose-built script that uses the library. I know you guys have talked about removing the cli completely, so I understand if you don't want to accept the PR for those reasons.
> 
> If you accept this PR, I'm certainly open to contributing more api updates.

Ah I might have worded that strangely, it was a general comment that we have some catching up do to across the entire library to match the current GitLab API attributes ;) Sorry if that came out wrong! 

That's a good point about CLI usage as well, thanks. We should probably review all the missing attributes and add more as a follow-up to this PR.

PR #1054, opened 2020-03-21
chore: improve test environment for upcoming features

nejch:
In #1052 and in some previous PRs, we've had to wait for a release to test new features on `latest`.

This PR allows contributors and CI (via either command-line flags or env variables) to define the image and tag to test against, and removes references to older python test environments.

Since `nightly` might be unstable I've added it to `allow_failures` so it's not blocking but at least provides some feedback on new features. This can later be combined with `@skipif` (or similar) to run tests only against versions equal or higher than a specific GitLab version to improve development for upcoming features.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,6 +40,22 @@ jobs:
         - pip3 install tox
         - tox -e py_func_v4
     - stage: test
+      name: cli_func_nightly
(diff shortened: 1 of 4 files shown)


max-wittig:
That's really cool!

PR #1062, opened 2020-04-02
chore(group): update group_manager attributes

max-wittig:
Synced with https://docs.gitlab.com/ee/api/groups.html


gitlab/v4/objects.py
```diff
@@ -1425,9 +1437,22 @@ class GroupManager(CRUDMixin, RESTManager):
             "name",
             "path",
             "description",
+            "membership_lock",
+            "share_with_group_lock",
             "visibility",
+            "request_access_enabled",
+            "mentions_disabled",
```
nejch, line 1444: Thanks for having me take a look :D sorry this took so long!
It seems like there are some duplicate entries (I saw some duplicates even in the GitLab docs) that argpase and the CLI do not like - this should make it go green:
```suggestion
```

max-wittig: Ah no problem. Sorry for assigning, if not even the pipeline is green 🙈

PR #1066, opened 2020-04-07
chore: use pytest to run unit tests and coverage

nejch:
I closed this one earlier thinking I'd include it in a later PR, but I guess it doesn't hurt to add it already, what do you think @max-wittig? :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,11 +1,14 @@
 *.pyc
 build/
 dist/
+htmlcov/
 MANIFEST
 .*.swp
 *.egg-info
 .idea/
+coverage.xml
 docs/_build
-.testrepository/
+.coverage
 .tox
+.venv/
 venv/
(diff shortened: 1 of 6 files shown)


max-wittig:
Doesn't hurt. That's true. Is there a way to show coverage changes in GitHub as well? I'm always missing features in Github, but that's where the community is still (and GitLab is missing a follow feature) 😄

nejch:
> 
> 
> Doesn't hurt. That's true. Is there a way to show coverage changes in GitHub as well? I'm always missing features in Github, but that's where the community is still (and GitLab is missing a follow feature) 😄

I've been think about this as well 😄 Badges and integrations with outside providers seems to be the only way, but probably no way of having it displayed in-line in diffs like with GitLab 12.9. Some integrations like Coveralls/SonarQube and other code quality tools might at least give us PR decorations or CI checks to display a summary, we'd have to see and decide on one.

Here's one example:
image

(from https://github.com/coveralls-clients/coveralls-python/pull/224)

We can open a follow-up issue after some research :P

PR #1067, opened 2020-04-07
fix(project): add missing project parameters

### LINKED ISSUE #765: How to create a job with init README？
I want to use `gitlab project create ...` to create a job and Initialize repository with a README,
but there is no options.
gitlab api https://docs.gitlab.com/ee/api/projects.html#create-project in create job has option `initialize_with_readme`.
in the last i use this command create README .`gitlab project-file create --project-id $(shell cat REPO_INFO | grep 'id' | sed 's/^id: //g') --file-path README.md --branch master --content '# makefile-init' --commit-message init-with-makefile`
if there has better choice?

## Specifications

  - python-gitlab version:

nejch:
As a follow-up to #1062, adding missing project attributes as well.

This is just a copy/paste with some regex replace from the markdown table in GitLab's API docs, so it includes EE attributes as well that we can't test..
It'd be cool to periodically parse those tables (if they were consistent) and auto-create issues/PR's against missing attributes :P

This closes #765. And probably some other CLI issues that require attributes to be explicitly added to the library.


max-wittig:
Thanks for keeping these updated! Regarding EE attributes. Many of them are gonna be core soon 😄 

https://about.gitlab.com/blog/2020/03/30/new-features-to-core/

PR #1072, opened 2020-04-14
feat(types): add __dir__ to RESTObject to expose attributes

spyoungtech:
This change adds the `__dir__` magic method to the `RESTObject` class in order to expose and allow introspection on the attributes that the object might have.

This is very useful to me when exploring the API in an interactive console. 

Normally I'll call `dir(obj)` (or `obj?` in iPython shell) to try to discover what methods and attributes exist for the object. However, the attributes in the `.attributes` property are not currently exposed, requiring the extra step of checking `.attributes` on each object I explore this way.

Hope this makes sense and that this change, if accepted, will help others.

### DIFF
--- a/gitlab/base.py
+++ b/gitlab/base.py
@@ -111,6 +111,9 @@ class RESTObject(object):
             return self.get_id() != other.get_id()
         return super(RESTObject, self) != other
 
+    def __dir__(self):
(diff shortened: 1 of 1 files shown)


max-wittig:
Didn't even know about `dir()` 😮 

This change does make sense! Thanks for your contribution!

PR #1073, opened 2020-04-14
docs(readme): update test docs

max-wittig:


nejch:
Nice, I guess we really could move away from custom helper scripts to a more native python/tox setup to make it simpler :)

PR #1077, opened 2020-04-16
feat(api): added support in the GroupManager to upload Group avatars

Flor1an-dev:
Similar to the `UserManager` and `ProjectManager` the `GroupManager` now supports uploading avatar images. The `GroupManager` object was missing the `ImageAttribute` type.

The feature to upload avatars to groups was introduced in Gitlab 12.9.
See also https://docs.gitlab.com/ce/api/groups.html#new-group (Attribute "avatar")

### DIFF
--- a/docs/gl_objects/groups.rst
+++ b/docs/gl_objects/groups.rst
@@ -61,6 +61,13 @@ Update a group::
     group.description = 'My awesome group'
     group.save()
 
+Set the avatar image for a group::
+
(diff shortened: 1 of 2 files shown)


Flor1an-dev:
@nejch I've added the documentation as well :)

nejch:
> @nejch I've added the documentation as well :)

Thanks again! :)

PR #1082, opened 2020-04-22
chore: bring commit signatures up to date with 12.10

nejch:
The commit signature test was failing against nightly after some changes to commit signatures (https://gitlab.com/gitlab-org/gitlab/-/merge_requests/27327). This fixes the test and some minor docs, since with 12.10 released we'll be getting failed builds without this (it was just a silly assert for the 404 :D).

I only left the unit test as is since the GPG response is one of the possible options.


nejch:
Hm, to get this to pass I had to also remove an outdated token test for the CLI.

PR #1085, opened 2020-04-25
chore(ci): add codecov integration to Travis

nejch:
We discussed adding some kind of coverage reporting, codecov seemed to be the easiest to add. After this is merged, codecov should also start reporting PR comments with a summary (on successful builds):
https://docs.codecov.io/docs/pull-request-comments

Edit: ah the summary is already here :) If this comment style is too intrusive, we can probably customize this to make it smaller with `codecov.yml`. Looks like there was an initial run a few years ago, since this reports against an old master commit. I'll look into adding this as a separate check as well. Seems like we can later split unit and functional tests into separate checks with flags.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -87,8 +87,11 @@ jobs:
       dist: bionic
       name: coverage
       python: 3.8
+      install:
+        - pip3 install tox codecov
       script:
-        - pip3 install tox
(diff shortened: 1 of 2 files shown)


max-wittig:
Cool feature!

PR #1108, opened 2020-06-01
Added NO_ACCESS const

stuartgunter:
This constant is useful for cases where no access is granted, e.g. when creating a protected branch.

The `NO_ACCESS` const corresponds to the definition in https://docs.gitlab.com/ee/api/protected_branches.html

### DIFF
--- a/gitlab/const.py
+++ b/gitlab/const.py
@@ -15,6 +15,7 @@
 # You should have received a copy of the GNU Lesser General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+NO_ACCESS = 0
 GUEST_ACCESS = 10
 REPORTER_ACCESS = 20
 DEVELOPER_ACCESS = 30


nejch:
Thanks for this @stuartgunter, that is useful and might pop up in other scenarios later. 

There have been some breaking changes in GitLab 13.0, which is why there have been some unrelated build failures recently. Could you rebase this on top of master - then this should go green? Thanks! :)

max-wittig:
kinda little nitpick on how the conventional changelog linter works. Could you begin your message in small case so that the pipeline goes green? `feat: add NO_ACCESS const`

stuartgunter:
Hi @max-wittig ... no problem. I've updated the commit message.

While I like the idea behind conventional changelog, it's quite an overhead for a small change like this from an infrequent contributor. I understand why it's been chosen but it has discouraged me from contributing more.

If only there was a consistent contribution guide for all/most OSS projects so we could reduce the barrier to entry :(

max-wittig:
@stuartgunter That's true, but otherwise the history is gonna be a mess and you will never find any reasonable messages (everyone would just use. `Update objects.py` or have to keep a changelog file. It's a difficult thing to align

max-wittig:
Thanks for the addition and sorry for the hassle

stuartgunter:
No problem at all. I do understand the rationale for using conventional changelog. My comments weren't intended as criticism, and I hope they were received well.

I wonder whether a pre-commit hook would help avoid having to amend commit messages and force push. That would certainly help new contributors get it right the first time and avoid too much Git-foo :)

max-wittig:
No I totally get your point and I'm annoyed by it myself that others cannot contribute that easily, but I'm not sure that having the wild west of commit messages be the better solution 😄  

A pre-commit hooks would actually be very nice. Thanks for the idea.

Issue #1120, opened 2020-06-19
ProjectMergeRequest.merge fails because of incorrect request behaviour


ferhat-aram:
## Problem

When I'm trying to merge a merge request via the ProjectMergeRequest.merge function and with "should_remove_source_branch" and/or "merge_when_pipeline_succeeds", I get the HTTP Code `405` with the message `Method not allowed` back. 

## Expected Behavior

I tried to do a API-Call as explained here https://docs.gitlab.com/ee/api/merge_requests.html#accept-mr and it works.

Using curl and sending a put request to `/projects/:id/merge_requests/:merge_request_iid/merge?remove_source_branch=True` worked fine. 

## Actual Behavior

The problem here is, that pythlon_gitlab merge API doesn't generate a query string, but send the paramters as post_data, which isn't accepted by the GitLab API.

I tried a curl request with `"Content-type: application/json" -d '{"merge_when_pipeline_succeeds": "True"}'` and `/projects/:id/merge_requests/:merge_request_iid/merge` and the request failed. 

It's very clear, that the request itself isn't done properly as expected (File :

gitlab/v4/objects.py
```
        path = "%s/%s/merge" % (self.manager.path, self.get_id())
        data = {}
        if merge_commit_message:
            data["merge_commit_message"] = merge_commit_message
        if should_remove_source_branch:
            data["should_remove_source_branch"] = True
        if merge_when_pipeline_succeeds:
            data["merge_when_pipeline_succeeds"] = True

        server_data = self.manager.gitlab.http_put(path, post_data=data, **kwargs)
        self._update_attrs(server_data)
```

The code handles those two parameters as post_data, while it should handle them as query_data:

gitlab/__init__.py
```
        query_data = query_data or {} #Expected: query_data["merge_when_pipeline_succeeds"]=True, Actual: None
        post_data = post_data or {} #Expected: None, Actual: post_data["merge_when_pipeline_succeeds"]=True
```

## Specifications

  - package version: 2.3.1
  - API version: v4
  - Gitlab server version (or gitlab.com): 12.9.3-ee

ferhat-aram:
I've already created a branch locally, but I have no rights to push a new branch, sadly. 
Could you please allow me access?

max-wittig:
@ferhat-aram Python-gitlab is just a wrapper around the GitLab API. I can't help you with your permission issues.

ferhat-aram:
[...]

PR #1121, opened 2020-06-19
fix(merge): parse arguments as query_data

ferhat-aram:
Refer to #1120 

Merge request requests aren't working, as the request parameters `should_remove_source_branch` and `remove_source_branch` are parsed as `post_data`, when they should be send as query_string, as described in the official GitLab API documentation itself [GitLab API Accept Merge](https://docs.gitlab.com/ee/api/merge_requests.html#accept-mr)


max-wittig:
Thanks for the MR!

Issue #1123, opened 2020-06-22
GitlabList only passing kwargs to first query call


tyates-indeed:
## Description of the problem, including code/CLI snippet
`GitlabList` accepts `**kwargs` in the constructor and passes those keyword args to the first call to `_query`. The keyword args are not saved though, so subsequent calls to `_query` in the `next` method do not pass the keyword args. This means that the first query will respect arguments like `retry_transient_errors` but subsequent calls will not.

## Expected Behavior
Keyword args passed to `GitlabList` apply to all query calls for that list.

## Actual Behavior
Keyword args passed to `GitlabList` apply only to the first query call.

## Specifications

  - python-gitlab version: 2.3.1
  - API version you are using (v3/v4): v4
  - Gitlab server version (or gitlab.com): irrelevant

PR #1124, opened 2020-06-22
Pass kwargs to subsequent queries in GitlabList (fixes: #1123)

tyates-indeed:
Pass kwargs to subsequent queries in GitlabList (fixes: #1123)

### DIFF
--- a/gitlab/__init__.py
+++ b/gitlab/__init__.py
@@ -780,6 +780,12 @@ class GitlabList(object):
         self._query(url, query_data, **kwargs)
         self._get_next = get_next
 
+        # Preserve kwargs for subsequent queries
+        if kwargs is None:
+            self._kwargs = {}
+        else:
+            self._kwargs = kwargs.copy()
+
     def _query(self, url, query_data=None, **kwargs):
         query_data = query_data or {}
         result = self._gl.http_request("get", url, query_data=query_data, **kwargs)
@@ -864,7 +870,7 @@ class GitlabList(object):
             pass
 
         if self._next_url and self._get_next is True:
-            self._query(self._next_url)
+            self._query(self._next_url, **self._kwargs)
             return self.next()
 
         raise StopIteration
--- a/gitlab/__init__.py
+++ b/gitlab/__init__.py
@@ -779,10 +779,7 @@ class GitlabList(object):
         self._gl = gl
 
         # Preserve kwargs for subsequent queries
-        if kwargs is None:
-            self._kwargs = {}
-        else:
-            self._kwargs = kwargs.copy()
+        self._kwargs = kwargs.copy()
 
         self._query(url, query_data, **self._kwargs)
         self._get_next = get_next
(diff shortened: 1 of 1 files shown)


gitlab/__init__.py
```diff
@@ -780,6 +780,12 @@ def __init__(self, gl, url, query_data, get_next=True, **kwargs):
         self._query(url, query_data, **kwargs)
         self._get_next = get_next
 
+        # Preserve kwargs for subsequent queries
```
rkpatel7, line 783: Should this happen before https://github.com/python-gitlab/python-gitlab/pull/1124/commits/1d011ac72aeb18b5f31d10e42ffb49cf703c3e3a#diff-15f1286708cbbf354eafe7b7f49a1622L780? And at that line do:

```python
self._query(url, query_data, **self._kwargs)
```

That way this class would be internally consistent with how it calls `self._query()` in both `__init__()` and `next()`.

gitlab/__init__.py
```diff
@@ -780,6 +780,12 @@ def __init__(self, gl, url, query_data, get_next=True, **kwargs):
...
+        if kwargs is None:
+            self._kwargs = {}
+        else:
+            self._kwargs = kwargs.copy()
+
     def _query(self, url, query_data=None, **kwargs):
```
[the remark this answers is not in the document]
tyates-indeed: Good call. Changed.

gitlab/__init__.py
tyates-indeed: Yeah, that was my thinking as well.

gitlab/__init__.py
```diff
@@ -777,7 +777,14 @@ class GitlabList(object):
 
     def __init__(self, gl, url, query_data, get_next=True, **kwargs):
         self._gl = gl
-        self._query(url, query_data, **kwargs)
+
+        # Preserve kwargs for subsequent queries
+        if kwargs is None:
+            self._kwargs = {}
+        else:
+            self._kwargs = kwargs.copy()
```
tron1point0, line 785: `kwargs` can't be `None`:

```python
>>> def foo(**kwargs):
...   print(kwargs)
... 
>>> foo()
{}
>>> foo(**None)
  File "<input>", line 1, in <module>
    foo(**None)
TypeError: foo() argument after ** must be a mapping, not NoneType
```

So this  can just be:

```suggestion
        self._kwargs = kwargs.copy()
```

rkpatel7:
@nejch, @max-wittig, or @bufferoverflow, any chance one of y'all can take a look at this pull request and merge it in?

max-wittig:
Hi @rkpatel7 Sorry for the delay! I will have a look.

max-wittig:
Thanks for the PR!

tyates-indeed:
@max-wittig is there a release planned soon to pull in this change?

max-wittig:
Sure makes sense

PR #1139, opened 2020-07-23
feat: add share/unshare the group with a group

sathieu:

PR #1140, opened 2020-07-23
feat: debug broken CI

sathieu:


sathieu:
@max-wittig It looks like Gitlab's behavior has changed: `group2/gr2_project` is now considered as owned by `root` (while there is a `group2.members.delete(gl.user.id)`. :astonished:

sathieu:
Fix integrated in #1139.

PR #1145, opened 2020-08-07
Allow pagination to be set in config globally and via cli

ericfrederich:
Also allow order_by to be set in globally in config

### DIFF
--- a/gitlab/config.py
+++ b/gitlab/config.py
@@ -167,13 +167,15 @@ class GitlabConfigParser(object):
             raise GitlabDataError("Unsupported per_page number: %s" % self.per_page)
 
         self.pagination = None
-        try:
-            self.pagination = self._config.get(self.gitlab_id, "pagination")
-        except Exception:
-            pass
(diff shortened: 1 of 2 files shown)


max-wittig:
Thanks for the MR! We had a similar change before. If pagination is set to `keyset` and you e.g. request merge requests, then GitLab will just throw 400 at you, so we would need a allow-list for this case to only use `keyset`, if the endpoint supports it.

max-wittig:
@ericfrederich See here: https://github.com/python-gitlab/python-gitlab/pull/1110/files

The problem with `keyset pagination` is that it breaks the workflow of some people, because GitLab will not send out pagination headers in this pagination mode. I've tried to re-add these upstream (https://gitlab.com/gitlab-org/gitlab/-/merge_requests/34273), but GitLab doesn't seem to want that ☹️

max-wittig:
@ericfrederich Any ideas how to continue here without breaking peoples workflow?

PR #1159, opened 2020-08-28
Feat: Project job artifacts for latest successful pipeline

nejch:
This adds support for an endpoint that was missing (artifacts from the latest pipeline per branch/tag):
https://docs.gitlab.com/ee/api/job_artifacts.html#download-the-artifacts-archive

The unrelated changes are there to keep black happy :) looks like there was a new release with different behavior


max-wittig (approved):
Very nice!

PR #1162, opened 2020-08-30
chore(env): add pre-commit and commit-msg hooks

### LINKED ISSUE #1128: Create pre-commit hook for commit message linting
Many people are having problems with conventional changelog. Maybe we should add a pre-commit hook to prevent this.

nejch:
Closes #1128. I was having this issue too sometimes when switching workstations and forgot to format :P

I went with `commitlint` to be consistent with CI but maybe `commitizen` makes sense in the future to keep dev dependencies Python-only.

### DIFF
--- a/.commitlintrc.json
+++ b/.commitlintrc.json
@@ -1,0 +1,3 @@
+{
+  "extends": ["@commitlint/config-conventional"]
+}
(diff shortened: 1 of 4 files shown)


max-wittig:
Looks good. Yes commitizen might make sense in the future!

PR #1163, opened 2020-08-30
Feat: add support for instance variables API

### LINKED ISSUE #1143: add support for instance-level CI/CD variables
Add support for instance-level CI/CD variables that became available in 13.1.

https://docs.gitlab.com/ee/api/instance_level_ci_variables.html

## Expected Behavior
Can interact with instance-level CI/CD variables the same as with groups and projects

## Actual Behavior
Not supported

nejch:
Closes #1143.

I turned objects into a subpackage here as it's getting more and more convoluted to find the right spot to shove them into `objects.py` :grin:. Would this work? If so I can try to split more in a separate PR.

Also, got some super repetitive tests here unfortunately as I wanted to also add missing project/group vars. Would be nice to parametrize fixtures (gl/group/project, and have just 1 set of tests), but I didn't find a clean way to do it as you would with params. This is the closest I found, any more ideas? https://stackoverflow.com/a/56871701


tools/functional/conftest.py
```diff
@@ -17,27 +22,7 @@ def random_id():
...
-        command = ["gitlab", "--config-file", CONFIG]
-
-        for subcommand in subcommands:
-            # ensure we get strings (e.g from IDs)
-            command.append(str(subcommand))
-
-        return script_runner.run(*command)
-
-    return _gitlab_cli
+    return os.path.join(TEMP_DIR, "python-gitlab.cfg")
```
max-wittig, line 25: Suggestion: Might as well use pathlib here, as we're Python 3.6+ 😄 

```python
from pathlib import Path
```

```suggestion
    return Path(TEMP_DIR) / "python-gitlab.cfg"
```

max-wittig (commented):
Amazing work as always, just a small suggestion. Let me know what you think!

max-wittig (commented):
Awesome work as always! 👍 

Just a small suggestion. Let me know what you think!

tools/functional/conftest.py
nejch: Thank you! Ah yes, definitely don't want to bring back python2 ghosts 😁 fixed.

max-wittig: Both are valid approaches still I guess, but Pathlib is kind of nicer and we can use it now 😄

PR #1165, opened 2020-08-31
Update objects.py

DylannCordel:
Hi !

Since version 10.3 (and later), param to not send (re)confirmation when updating an user is `skip_reconfirmation` (and not `skip_confirmation`).

See:

* https://gitlab.com/gitlab-org/gitlab-foss/-/merge_requests/15175?tab=
* https://docs.gitlab.com/11.11/ee/api/users.html#user-modification
* https://docs.gitlab.com/ee/api/users.html#user-modification


nejch:
Thanks you for this PR, nice catch! :)

PR #1174, opened 2020-09-04
 Add unit tests for resource label events API

Shkurupii:
Hi! I propose to add another few unit test cases. At this time is for resource label events.


nejch:
Thanks for helping out with tests! :)

PR #1178, opened 2020-09-05
test(env): replace custom scripts with pytest and docker-compose

nejch:
I like writing functional/integration tests because the GitLab API keeps changing so it's the only way to really check the responses are what is assumed in unit test fixtures. But the custom bash scripting in the environment was kind of unreliable with the REUSE_CONTAINER option, the custom venv creation, hardcoded temp dirs (I'm sometimes on a Windows machine 😢 ). It was also not a consistent experience compared to unit tests, so might be why people don't usually add them. 

I saw that gitbeaker was doing something similar so I'm hoping this is a stable approach.

The idea was to:
- have all virtualenvs fully managed by tox for consistency
- fully cross-platform I hope (even tested on Windows cmd 😁 )
- use native docker-compose, making it easy to add real integration tests with other services like gitlab-runner, LDAP etc.
- create a static token directly with the rails runner, so no need for `requests-html` dependency and `generate_token.py`
- remove all unneeded/wrapper bash scripts when it's all put together, use pytest libraries that can be more robust instead
- easier to rewrite `python_test_v4.py` into real test cases that run independently (though it will make the runs longer I assume)
- can use pytest's tmpdir fixtures later which might be more reliable than `/tmp`

Cons/TODOs before merge:
- performance - not sure if it's just me but seems like the container takes a bit longer to spin up.
- more pytest dependencies to install - I've added a separate `docker-requirements.txt` to avoid polluting other environments, does that make sense?
- still need to implement a "reuse existing container" option upstream on `pytest-docker`. I have something working locally with `pytest_addoption` and there is some work already upstream in https://github.com/avast/pytest-docker/pull/51

It's also possible this is a bit overkill 😁 but would love any feedback on this if anyone else was also having issues with the bash scripts before.


max-wittig (approved):
Very very nice!

tools/functional/conftest.py
```diff
@@ -1,40 +1,122 @@
...
 from random import randint
+from subprocess import check_output
 
 import pytest
 
 import gitlab
 
 
 TEMP_DIR = tempfile.gettempdir()
+TEST_DIR = Path(__file__).resolve().parent
```
max-wittig, line 14:
```suggestion
TEST_DIR = Path.cwd()
```

tools/functional/python_test_v4.py
```diff
@@ -56,11 +57,11 @@
 qG2ZdhHHmSK2LaQLFiSprUkikStNU9BqSQ==
 =5OGa
 -----END PGP PUBLIC KEY BLOCK-----"""
-AVATAR_PATH = os.path.join(os.path.dirname(__file__), "avatar.png")
-
+AVATAR_PATH = Path(__file__).resolve().parent / "fixtures" / "avatar.png"
```
max-wittig, line 60:
```suggestion
AVATAR_PATH = Path.cwd() / "fixtures" / "avatar.png"
```

max-wittig (changes requested):
I'm really a fan of this, just two small comments.

tools/functional/conftest.py
nejch: Thanks! I had this as I am always a bit wary of using `cwd()` since I can't be sure where it's executed from. So I did the "derive path relative from the current script" dance 😁 I know It's a bit uglier but 100% safe. Currently with `tox -e py_func_v4` executed from the root dir this would give me an error because cwd is higher up (please don't laugh at the windows paths 😄):
```
ERROR tools/functional/api/test_gitlab.py::test_api_v4 - Exception: Command docker-compose -f "C:\Projects\python-gitlab\fixtures\docker-compose.yml" -p "pyt
est20584" up --build -d returned 1: """.FileNotFoundError: [Errno 2] No such file or directory: 'C:\\Projects\\python-gitlab\\fixtures\\docker-compose.yml'
```
I'll check if pytest has some cleaner way for this.

max-wittig: I guess that's okay, but pathlib should have some equivalent

max-wittig:
@nejch Sadly some conflicts happened. Could you please rebase?

nejch:
> 
> 
> @nejch Sadly some conflicts happened. Could you please rebase?

I've rebased and refactored those paths with helper fixtures where possible so it's hopefully more readable :)

PR #1180, opened 2020-09-07
test: add unit tests for badges API

Shkurupii:
Hi! Take a look at the unit tests inside. At that time for project and group badges.


nejch (approved):
Thanks again for the thorough tests! :)

PR #1181, opened 2020-09-07
feat(api): add support for user identity provider deletion

### LINKED ISSUE #1070: How to delete identities with the API?

```
user.identities = []
user.save()
```

## Expected Behavior

Identities are deleted.

## Actual Behavior

Identities are unchanged.

## Specifications

  - python-gitlab version: 2.2.0
  - Gitlab server version (or gitlab.com): 12.9

nejch:
Closes #1070.

This one is weird because the endpoint is kind of special (only supports delete, plus the creation of identities is inconsistent with this endpoint).

https://docs.gitlab.com/ee/api/users.html#delete-authentication-identity-from-user

I couldn't do `user.identities` for the manager because that's the attribute returned in the nested JSON. Maybe this could be done by having a `__repr__()` that returns that list from the attribute of the parent manager? But it felt like I would just be adding to the inconsistency, not sure. I didn't find any other examples that solved it for nested json.

I opted for `user.identityproviders.delete()` since its explicit and the provider is a required param. I also considered `user.identity.delete()`, `user.delete_identity()` (custom method) and `user.providers.delete()`, I'm open to any other suggestions here :grin:

### DIFF
--- a/docs/gl_objects/users.rst
+++ b/docs/gl_objects/users.rst
@@ -80,6 +80,10 @@ Set an external identity for a user::
     user.extern_uid = '3'
     user.save()
 
+Delete an external identity by provider name::
+
+    user.identityproviders.delete('oauth2_generic')
+
 User custom attributes
 ======================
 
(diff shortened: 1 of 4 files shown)


max-wittig (approved):
I like it and it's also intuitiv this way. Thank you!

PR #1182, opened 2020-09-08
Added MR wip filter param

jlpospisil:
Added filter param to allow searching merge requests using `wip` query param

### DIFF
--- a/gitlab/v4/objects/__init__.py
+++ b/gitlab/v4/objects/__init__.py
@@ -1210,6 +1210,7 @@ class GroupMergeRequestManager(ListMixin, RESTManager):
         "source_branch",
         "target_branch",
         "search",
+        "wip",
     )
     _types = {"labels": types.ListAttribute}
 
@@ -1732,6 +1733,7 @@ class MergeRequestManager(ListMixin, RESTManager):
         "source_branch",
         "target_branch",
         "search",
+        "wip",
     )
     _types = {"labels": types.ListAttribute}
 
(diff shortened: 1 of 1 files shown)


gitlab/v4/objects/__init__.py
```diff
@@ -1210,6 +1210,7 @@ class GroupMergeRequestManager(ListMixin, RESTManager):
         "source_branch",
         "target_branch",
         "search",
+        "wip",
```
nejch, line 1213: Thanks for the addition @jlpospisil! One small note. You added this to the `MergeRequestManager` (global) and `GroupMergeRequestManager` (groups). I think you can also add it to `ProjectMergeRequestManager`  (projects) per the API docs.

It seems like `wip` is not documented as a param in group merge requests in https://docs.gitlab.com/ee/api/merge_requests.html#list-group-merge-requests, but it probably works. Did you already use it this way?

jlpospisil: I use it via the UI all the time which is what led me to try it via this library.  The possible values are `yes` or `no`.  I primarily use `wip=no` to get all non WIP MRs.  This will filter out anything that is labelled as `WIP` or where the description starts with `WIP`.  I verified that it worked as expected for my use case.  I specifically used the `GroupMergeRequestManger`.  I did add the filter to the group manager as well.

nejch: Great, thanks! I wasn't sure if it was covered by the Groups API since the GitLab docs didn't have it there. Seems to happen sometimes with their documentation :)

Issue #1113, opened 2020-06-08
Re-enable note tests and revert in 13.1


max-wittig:
See: 
https://github.com/python-gitlab/python-gitlab/commit/63ae77ac1d963e2c45bbed7948d18313caf2c016

Issue #1033, opened 2020-02-25
Functional tests that create/update/delete resources are flaky


nejch:
I realized the weird unrelated failure in #1020 (see https://travis-ci.org/python-gitlab/python-gitlab/jobs/652064210) was a flaky test for rate limits. I'm getting it locally sometimes but only rarely.

I haven't had the time to investigate but I'll try to have a look at some point. Maybe something in the environment slows down the requests sometimes and it does not trigger 429 or something.

max-wittig:
This is not the only flaky test in these functional tests. Many of them fail randomly, but just in CI, never locally.

I spend a lot of effort a while ago (even switched CI systems), but haven't figured it out, yet
I would be happy for every hint I get :smile: 

Or maybe we got to completely change this functional test setup, not sure.

nejch:
Ah I see :( I do get some failures locally, maybe my measly laptop is closer to Travis VM specs :P 
It's hard to mark things for re-run now because certain tests prepare state for others, so maybe a refactor will help, let's see :)

nejch:
I now *think* that a lot of these failures happen when GitLab hands off tasks to Sidekiq asynchronously, and getting a response from the server does not mean that something is actually done (like deleting a project etc). If the next step/assert relies on that, it fails.

I see this more often now as I try to convert asserts to test cases for #1024 (I will push that once it gets greener :P ). Especially as I try to delete a resource in teardown and re-create it for the next test. If I run tests with `pytest-random-order` it goes green whent those 2 tests are further apart.

For example with the rate limit, if I try to reuse `tools/reset_gitlab.py` for teardown and add some checks with `projects.list()`, it takes almost 10-20 seconds in my case to delete all the projects that the test creates (around 20-40 projects?).

Have you or anyone here had any similar experience? Maybe we could have some kind of `wait_resource_available()` and `wait_resource_deleted()` helpers, and have as much of that in fixtures as possible to avoid having sleeps in tests. It would probably slow down the tests quite a bit, but make them more robust. Anyway, splitting these into testcases is already some progress it seems :)

Issue #1105, opened 2020-05-25
Fix snippet integration tests for GitLab 13.0


nejch:
I was working on something unrelated and snippet tests kept failing.. turns out that when GitLab introduced versioned snippets in 13.0, along the way there's been some changes in the create/update API behavior that broke the functional tests (maybe this one https://gitlab.com/gitlab-org/gitlab/-/commit/ac3438f5c2a4da780e730660bccea5f4ac12483d).

max-wittig:
This does look like a GitLab error. Renaming the snippet via API deletes it's contents.

max-wittig:
I reproduced it with curl and create an upstream issue: https://gitlab.com/gitlab-org/gitlab/-/issues/219348

max-wittig:
This will be fixed in 13.1. by GitLab. It would say we leave it in our tests like this for now or should we add allow_fail? @nejch

nejch:
Looks like I forgot about this one, but the tests are included as of https://github.com/python-gitlab/python-gitlab/pull/1205 :)

Issue #1065, opened 2020-04-07
Uncomment tests in GitLab 12.10


max-wittig:
https://github.com/python-gitlab/python-gitlab/pull/1052/files#diff-4a8817e452ed84e4d3b53c2523dac89aR655

nejch:
These ones were also included during the split.

PR #1205, opened 2020-10-08
refactor(tests): split functional tests

nejch:
Glorious copy/paste time! This just splits functional tests in the hope we get more control over weird timing failures.
No attempts yet to add fixture factories which would be very useful here :P But for now, this provides a `wait_for_sidekiq` helper that should hopefully make it easier to solve if there are still issues.

This will be a pain to review, I wish GitHub/GitLab had code move detection like Review Board :( But the file sizes should be more reasonable now:
```
$ ls tools/functional/api/*.py | xargs wc -l | sort -h
   12 tools/functional/api/test_deploy_keys.py
   13 tools/functional/api/test_packages.py
   36 tools/functional/api/test_deploy_tokens.py
   42 tools/functional/api/test_current_user.py
   46 tools/functional/api/test_clusters.py
   48 tools/functional/api/test_variables.py
   61 tools/functional/api/test_import_export.py
   74 tools/functional/api/test_snippets.py
   89 tools/functional/api/test_issues.py
   96 tools/functional/api/test_merge_requests.py
  126 tools/functional/api/test_repository.py
  170 tools/functional/api/test_users.py
  183 tools/functional/api/test_gitlab.py
  190 tools/functional/api/test_groups.py
  298 tools/functional/api/test_projects.py
```

Closes https://github.com/python-gitlab/python-gitlab/issues/1113, https://github.com/python-gitlab/python-gitlab/issues/1105, https://github.com/python-gitlab/python-gitlab/issues/1065, hopefully closes https://github.com/python-gitlab/python-gitlab/issues/1033.

Also I realized when doing this that anything to do with group creation/deletion is *really* slow so maybe it'll help with https://github.com/python-gitlab/python-gitlab/pull/1194.


tools/functional/conftest.py
```diff
@@ -1,6 +1,7 @@
 import tempfile
 import time
 import uuid
+from contextlib import contextmanager
```
nejch, line 4: oops this one needs to go, will remove it asap.

tools/functional/api/test_repository.py
```diff
@@ -0,0 +1,126 @@
...
+    # object method
+    assert readme.decode().decode() == "Initial content"
+
+    blame = project.files.blame(file_path="README.rst", ref="master")
+    assert blame
+
+
+def test_repository_tree(project):
+    tree = project.repository_tree()
+    assert len(tree) != 0
```
max-wittig, line 44:
```suggestion
    assert tree
```

tools/functional/api/test_users.py
```diff
@@ -3,6 +3,158 @@
...
+
+    new_user.delete()
+    wait_for_sidekiq()
+
+    assert new_user.id not in [user.id for user in gl.users.list()]
+
+
+def test_user_projects_list(gl, user):
+    projects = user.projects.list()
+    assert len(projects) == 0
```
max-wittig, line 66:
```suggestion
    assert projects
```

tools/functional/api/test_users.py
```diff
@@ -3,6 +3,158 @@
...
+
+
+def test_user_events_list(gl, user):
+    events = user.events.list()
+    assert len(events) == 0
```
max-wittig, line 71:
```suggestion
    assert events
```

tools/functional/conftest.py
max-wittig:
```suggestion
```

max-wittig (approved):
Really awesome work!

max-wittig (changes requested):
Awesome work. Just a few suggestions!

tools/functional/api/test_import_export.py
```diff
@@ -0,0 +1,61 @@
...
+    time.sleep(10)
+    group_import = gl.groups.get(import_path)
+
+    assert group_import.path == import_path
+    assert group_import.name == import_name
+
+
+def test_project_import_export(gl, project, temp_dir):
+    export = project.exports.create()
+    export.refresh()
```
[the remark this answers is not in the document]
nejch: It was probably there originally because the create/POST request only gets a `"message": "202 Accepted"` response so there's no initial `export_status` for the loop below, I will check. Maybe a for loop with a break and only one `refresh` would be more elegant though :)

tools/functional/api/test_users.py
nejch: These `len()` asserts are all evil, especially with group objects :D but maybe we could keep them as a sanity check as in "it _does_ return a list (rather than None) but the list it is empty"? Since None would fail with `len()`.

Or would you prefer the more explicit:
```python
assert isinstance(something, list)
assert not something
```

tools/functional/api/test_import_export.py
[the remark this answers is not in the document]
nejch: It was needed but I placed an explicit get here instead, maybe nicer :)
```diff
 def test_project_import_export(gl, project, temp_dir):
     export = project.exports.create()
-    export.refresh()
+    assert export.message == "202 Accepted"
+
+    export = project.exports.get()
+    assert isinstance(export, gitlab.v4.objects.ProjectExport)

     count = 0
     while export.export_status != "finished":
```

nejch:
> 
> 
> Awesome work. Just a few suggestions!

Thanks for the quick review Max! I pushed https://github.com/python-gitlab/python-gitlab/pull/1205/commits/65ce02675d9c9580860df91b41c3cf5e6bb8d318 to address the suggestions :)

max-wittig:
Very very nice!

PR #1209, opened 2020-10-11
docs(cli): add auto-generated CLI reference

### LINKED ISSUE #1197: Trigger export using CLI
I would love to be able to trigger project exports using the CLI.

I'm fine with the export notification going to my e-mail and me downloading them from there if I can just automate half of it.

nejch:
This adds a generated page similar to the API reference, including all the possible commands. It's a _long_ list, but better than nothing. 😁 I wasn't able to group them into sections due to the way they are currently dynamically generated, but maybe there's a way.

I've noticed a lot of people still use the CLI even though it has some issues. I'm hoping this can help people/prevent opening issues that are simply due to missing docs (recent example: https://github.com/python-gitlab/python-gitlab/issues/1197), or at least we can more quickly close those by providing the link rather than providing examples every time.

Also means no more manual examples for the CLI are needed, hopefully.

Click also has a [drop-in replacement](https://sphinx-click.readthedocs.io) for the sphinx extension used for this PR, so switching at a later point should be easy.

Closes #1197.

### DIFF
--- a/docs/cli-objects.rst
+++ b/docs/cli-objects.rst
@@ -1,0 +1,17 @@
+##################################
+CLI reference (``gitlab`` command)
+##################################
+
+.. warning::
+
+    The following is a complete, auto-generated list of subcommands available
+    via the :command:`gitlab` command-line tool. Some of the actions may
+    currently not work as expected or lack functionality available via the API.
+
+    Please see the existing `list of CLI related issues`_, or open a new one if
+    it is not already listed there.
+
+.. _list of CLI related issues: https://github.com/python-gitlab/python-gitlab/issues?q=is%3Aopen+is%3Aissue+label%3Acli
+
+.. autoprogram:: gitlab.cli:docs()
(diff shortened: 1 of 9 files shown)


docs/cli-objects.rst
```diff
@@ -0,0 +1,17 @@
...
+    The following is a complete, auto-generated list of subcommands available
+    via the :command:`gitlab` command-line tool. Some of the actions may
+    currently not work as expected or lack functionality available via the API.
+
+    Please see the existing `list of CLI related issues`_, or open a new one if
+    it is not already listed there.
+
+.. _list of CLI related issues: https://github.com/python-gitlab/python-gitlab/issues?q=is%3Aopen+is%3Aissue+label%3Acli
+
+.. autoprogram:: gitlab.cli:docs()
```
max-wittig, line 16: nice didn't know about this!

max-wittig:
Very nice!

PR #1214, opened 2020-10-12
docs(readme): also add hint to delete gitlab-runner-test

max-wittig:
Otherwise the whole testsuite will refuse to run


nejch:
Oops, forgot that one! 👍

PR #1215, opened 2020-10-12
test(users): force 10 seconds sleep to improve reliability

max-wittig:
I'm welcome for better ideas

### DIFF
--- a/tools/functional/api/test_users.py
+++ b/tools/functional/api/test_users.py
@@ -54,7 +54,7 @@ def test_delete_user(gl, wait_for_sidekiq):
             "password": "delete-user-pass",
         }
     )
-
+    time.sleep(10)
     new_user.delete()
     wait_for_sidekiq()
 


tools/functional/api/test_users.py
```diff
@@ -54,7 +54,7 @@ def test_delete_user(gl, wait_for_sidekiq):
             "password": "delete-user-pass",
         }
     )
-
+    time.sleep(10)
     new_user.delete()
```
nejch, line 58: I think this should go after, because the delete itself is taking too long:

```suggestion
    new_user.delete()
    time.sleep(10)
```

Looks like we really need this. I thought my little sidekiq wait hack would work though 🤕  I based it off of [this](https://gitlab.com/gitlab-org/gitlab-foss/-/issues/31750#note_28757367), but I'm not sure if we can extract from the API when _this_ exact sidekiq deletion job is finished:
> We can't guarantee when the user will actually be deleted by Sidekiq; this could take a while if the user has a lot of projects/events/etc.

I'll keep looking into this later, maybe we could also just wrap these asserts in retries until the condition is met (with a timeout), since it seems to me like GitLab gets more bloated with new features and so things can take longer with new releases on these little Travis VMs?

They do perform better during the day/office hours though, just retried all the jobs and they're going green 🤣

max-wittig: Maybe use something like [flaky](https://github.com/box/flaky) for this?

JohnVillalovos:
Maybe we don't need this anymore.

Seems like https://github.com/python-gitlab/python-gitlab/pull/1316 might have resolved the issue.

Not 100% sure, but I haven't noticed this error since it got merged.

nejch:
> Maybe we don't need this anymore.
> 
> Seems like #1316 might have resolved the issue.
> 
> Not 100% sure, but I haven't noticed this error since it got merged.

True, I think your PR fixed it! We can probably close this now and re-enable the other flaky test as well.

PR #1230, opened 2020-11-08
cli-usage readme: fixed term

manuel-91:
see codeblock below, it's about issues, not snippets


max-wittig:
Sorry for ignoring you for so long. Thanks for the MR!

PR #1249, opened 2020-12-10
feat: add pipeline test report support

rmonat:
Gitlab (since v13.0) now supports [test reports](https://docs.gitlab.com/ee/ci/unit_test_reports.html#unit-test-reports). Here is [the api description](https://docs.gitlab.com/ee/api/pipelines.html#get-a-pipelines-test-report).

### DIFF
--- a/docs/gl_objects/pipelines_and_jobs.rst
+++ b/docs/gl_objects/pipelines_and_jobs.rst
@@ -326,3 +326,26 @@ Examples
 List bridges for the pipeline::
 
     bridges = pipeline.bridges.list()
+
+Pipeline test report
+====================
+
+Get a pipeline's complete test report.
+
+Reference
+---------
+
+* v4 API
+
+  + :class:`gitlab.v4.objects.ProjectPipelineTestReport`
+  + :class:`gitlab.v4.objects.ProjectPipelineTestReportManager`
+  + :attr:`gitlab.v4.objects.ProjectPipeline.test_report`
+
+* GitLab API: https://docs.gitlab.com/ee/api/pipelines.html#get-a-pipelines-test-report
+
+Examples
+--------
+
+Get the test report for a pipeline::
+
+    test_report = pipeline.test_report.get()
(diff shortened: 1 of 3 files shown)


gitlab/v4/objects/__init__.py
```diff
@@ -3880,11 +3880,22 @@ class ProjectPipelineVariableManager(ListMixin, RESTManager):
...
+    _obj_cls = ProjectPipelineTestReport
+    _from_parent_attrs = {"project_id": "project_id", "pipeline_id": "id"}
+
+
 class ProjectPipeline(RESTObject, RefreshMixin, ObjectDeleteMixin):
     _managers = (
         ("jobs", "ProjectPipelineJobManager"),
         ("bridges", "ProjectPipelineBridgeManager"),
         ("variables", "ProjectPipelineVariableManager"),
+        ("test_report", "ProjectPipelineTestReportManager")
```
nejch, line 3898: Hi @rmonat, thanks for this and sorry for the delay.

Could you please run `black` on your code so the linter is happy? See instructions in the readme: https://github.com/python-gitlab/python-gitlab#development-workflow

Then if you could just this to the docs maybe? Here's a recent example: https://github.com/python-gitlab/python-gitlab/commit/05cbdc224007e9dda10fc2f6f7d63c82cf36dec0.

bonus - If you'd like to also tackle adding unit tests, here is an example: https://github.com/python-gitlab/python-gitlab/blob/master/gitlab/tests/objects/test_pipeline_schedules.py.

nejch (changes requested):

nejch:
@rmonat just FYI - we've split the objects package into modules now so this will have to be rebased and moved into `gitlab/v4/objects/pipelines.py`, for example. And the manager also imported in `projects.py`. Take a look at #1304 for an example on how to add it.

I'll assign this to you for now but if you need help just let me know :)

nejch (approved):
@rmonat thanks again for this. Since the project is moving ahead I didn't want this PR to be blocked or conflict too much, so I took the liberty of rebasing and adding docs for this. Don't worry you'll still be credited for the commit/PR ;)

PR #1255, opened 2020-12-17
Use inline anonymous references for external links

bittner:
The links to "the official documentation" on tokens are currently broken in the [project documentation](https://python-gitlab.readthedocs.io/en/stable/cli.html).

There doesn't seem to be an obvious way to use an alias for identical text labels that link to different targets. With inline links we can work around this shortcoming. Until we know better.

### DIFF
--- a/docs/cli-usage.rst
+++ b/docs/cli-usage.rst
@@ -98,22 +98,23 @@ server, with very limited permissions.
    * - ``url``
      - URL for the GitLab server
    * - ``private_token``
-     - Your user token. Login/password is not supported. Refer to `the official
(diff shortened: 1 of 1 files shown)


max-wittig:
Thanks for the MR. Would you mind to change your commit message to comply with the conventional changelog?

```
docs(cli): use inline anonymous references for external links
```

bittner:
Done. I left the long comment in the second line. Is that fine?

max-wittig:
Totally fine. Thanks 👍

max-wittig:
Sorry for ignoring this for so long! Thanks for your MR!

PR #1258, opened 2020-12-22
feat: add retry_transient_errors global option

max-wittig:


nejch:
This was now done in https://github.com/python-gitlab/python-gitlab/pull/1565.

PR #1263, opened 2021-01-10
fix(api): add missing GetMixin to ProjectLabelManager

ePirat:
The ProjectLabelManager was missing the GetMixin, which made it impossible to get labels by name or ID, even though this is supported by the GitLab API.

### DIFF
--- a/gitlab/v4/objects/__init__.py
+++ b/gitlab/v4/objects/__init__.py
@@ -3601,7 +3601,7 @@ class ProjectLabel(SubscribableMixin, SaveMixin, ObjectDeleteMixin, RESTObject):
 
 
 class ProjectLabelManager(
-    ListMixin, CreateMixin, UpdateMixin, DeleteMixin, RESTManager
+    RetrieveMixin, CreateMixin, UpdateMixin, DeleteMixin, RESTManager
 ):
     _path = "/projects/%(project_id)s/labels"
     _obj_cls = ProjectLabel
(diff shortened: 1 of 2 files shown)


gitlab/v4/objects/__init__.py
```diff
@@ -3601,7 +3601,7 @@ def save(self, **kwargs):
 
 
 class ProjectLabelManager(
-    ListMixin, CreateMixin, UpdateMixin, DeleteMixin, RESTManager
+    GetMixin, ListMixin, CreateMixin, UpdateMixin, DeleteMixin, RESTManager
```
nejch, line 3604: Thanks for this and nice catch! There's a convenience `RetrieveMixin` for get+list that we could also use here:

```suggestion
    RetrieveMixin, CreateMixin, UpdateMixin, DeleteMixin, RESTManager
```


Don't worry about the functional test failure, the environment needs a bit of cleanup. You could also add a `get()` test to the functional test suite after this section if you'd like:
https://github.com/python-gitlab/python-gitlab/blob/265dbbdd37af88395574564aeb3fd0350288a18c/tools/functional/api/test_projects.py#L141-L143

nejch (changes requested):

gitlab/v4/objects/__init__.py
ePirat: Thanks! Changed.

nejch: Thanks and sorry for the delay!

PR #1252, opened 2020-12-16
feat: support multipart uploads

max-wittig:
Closes https://github.com/python-gitlab/python-gitlab/pull/1004
Closes https://github.com/python-gitlab/python-gitlab/issues/646

PR #1269, opened 2021-01-27
chore(ci): bring test environment back to life

nejch:
This does a few things to hopefully bring back a workable environment so people's PRs can go green :)

- at some point `docker-compose` started using python-dotenv which processes .env files differently (at least for paths to env files) and this broke everything in functional tests. This PR explicitly tracks a version in the tox environment so it works everywhere regardless of the system.
- replaces Travis with GHA. On .org the Travis job queues were getting into 3+ hours for jobs, the migration to .com seemed like a whole process, and it was getting less and less reliable.
- I like that Actions have test failure annotations (akin to inline coverage in GitLab) and it was basically zero-config to add it with https://pypi.org/project/pytest-github-actions-annotate-failures. Jobs get queued fast and builds are under 10 min including functional tests and they seem very stable :+1: (I was spamming it while writing the YAML and none of the functional tests failed). https://github.com/python-gitlab/python-gitlab/pull/1215 or similar will probably still be necessary but this looks promising with GitHub Actions IMO.

Edit: well I take that back about tests being stable haha


max-wittig:
Very very nice. I wanted to switch to Github actions for so long, but didn't do it. Thanks!

.github/workflows/lint.yml
```diff
@@ -0,0 +1,23 @@
...
+env:
+  PY_COLORS: 1
+
+jobs:
+  black:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+      - uses: psf/black@stable
```
max-wittig, line 14: things might break, if black decides to change their formatting. Is there a way to fix the version?

max-wittig (changes requested):
Looks very good. Just one question regarding black.

.github/workflows/lint.yml
nejch: Ah that's right, already happened on Travis! I think currently we could only pin it to commit hashes, since the Action is versioned in the black repo itself and they only do tags/releases a few times a year (and nothing since they've added the Action).

But if we wait for this (or add a PR upstream) we can start pinning to black versions: https://github.com/psf/black/issues/1939. I will check, I think it might just be a simple addition to their action.yml and Dockerfile.

max-wittig: Ahh if they don't have an option, then we can leave it like that for now.

max-wittig:
@nejch would this be ready to merge? I would love to rebase https://github.com/python-gitlab/python-gitlab/pull/1252 on it.

nejch:
> 
> 
> @nejch would this be ready to merge? I would love to rebase #1252 on it.

In the end I noticed some of the flakiness in functional tests is still there, but if we don't mind hitting retry on those jobs sometimes we can solve it in a follow-up! :) (e.g. https://github.com/python-gitlab/python-gitlab/pull/1215) In that case it's ready.

max-wittig:
I've uninstalled the travis app. I think it should work now.

nejch:
> 
> 
> I've uninstalled the travis app. I think it should work now.

The checks run really fast now with GHA. Nice! 🥳

max-wittig:
Soooo fast. Woah. And I had to disable branch protection rules for travis as well.

max-wittig:
The tests are flacky because they are not isolated and rely on order of execution, which is not always the same.

See: `test_merge_request.py`. We would need many more fixtures. E.g. to create random branches etc.

max-wittig:
and create new onces everytime @nejch

PR #1276, opened 2021-01-31
docs: switch from travis-ci.org to travis-ci.com

bufferoverflow:


README.rst
```diff
@@ -1,5 +1,5 @@
-.. image:: https://travis-ci.org/python-gitlab/python-gitlab.svg?branch=master
-   :target: https://travis-ci.org/python-gitlab/python-gitlab
+.. image:: https://travis-ci.com/python-gitlab/python-gitlab.svg?branch=master
+   :target: https://travis-ci.com/python-gitlab/python-gitlab
```
nejch, line 2: I think we could replace this with GitHub Actions now, I forgot that in https://github.com/python-gitlab/python-gitlab/pull/1269 oops!

But seems like GHA has badges per workflow only, not for entire "builds"?
E.g. one for for Test / Lint / Docs each:

https://github.com/python-gitlab/python-gitlab/workflows/Test/badge.svg
https://github.com/python-gitlab/python-gitlab/workflows/Lint/badge.svg
https://github.com/python-gitlab/python-gitlab/workflows/Docs/badge.svg

bufferoverflow: @max-wittig Which badges shall be added? Or just remove the travis badge that was leading me into the wrong direction?

bufferoverflow: I added the Test badge with link to GH actions

nejch: Agree that makes the most sense here :)

PR #1277, opened 2021-01-31
feat(api,cli): make user agent configurable

nejch:
I was thinking that some tools/libraries using python-gitlab (or people triggering the CLI in pipelines) might want to define their own user agent.

### DIFF
--- a/gitlab/const.py
+++ b/gitlab/const.py
@@ -15,6 +15,9 @@
 # You should have received a copy of the GNU Lesser General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+from gitlab.__version__ import __title__, __version__
+
+
 NO_ACCESS = 0
 MINIMAL_ACCESS = 5
 GUEST_ACCESS = 10
@@ -51,3 +54,5 @@ SEARCH_SCOPE_GLOBAL_SNIPPET_TITLES = "snippet_titles"
 
 # specific project scope
 SEARCH_SCOPE_PROJECT_NOTES = "notes"
+
+USER_AGENT = "{}/{}".format(__title__, __version__)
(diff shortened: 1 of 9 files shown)


nejch:
Should the "Require pull request reviews before merging" maybe also be added to required checks in branch protection? So I don't accidentally merge my own stuff :laughing:

gitlab/const.py
```diff
@@ -51,3 +54,5 @@
 
 # specific project scope
 SEARCH_SCOPE_PROJECT_NOTES = "notes"
+
+USER_AGENT = "{}/{}".format(__title__, __version__)
```
max-wittig, line 58: I guess we could already use f-strings here. Not a blocker though.

max-wittig (approved):

gitlab/const.py
max-wittig: @nejch Should I merge or do you want to change this still?

nejch: @max-wittig I was thinking about this as well but in the end decided against since, if I remember correctly I didn't see any f-strings in the production code (only tests), so I thought maybe some poor soul is still on 3.5 (even if unsupported officially) and a minor/patch version bump would be confusing if it started failing all of a sudden 😁 

Maybe a follow-up issue to also clear all old-style % operators as well and just turn everything to f-strings consistently?

max-wittig: Sure.

max-wittig:
Btw. there should be no 3.5 users, as pypi will just give you the old version, if you download with pip.

nejch:
> Btw. there should be no 3.5 users, as pypi will just give you the old version, if you download with pip.

Oh cool, I didn't even realize that :)

PR #1278, opened 2021-01-31
feat(api): default to gitlab.com if no URL given

nejch:
The github CLI (`gh`) defaults to github.com if no arguments are given. Something similar would make sense here IMO, just for convenience :)

### DIFF
--- a/docs/api-usage.rst
+++ b/docs/api-usage.rst
@@ -2,34 +2,38 @@
 Getting started with the API
 ############################
 
-python-gitlab only supports GitLab APIs v4.
+python-gitlab only supports GitLab API v4.
 
 ``gitlab.Gitlab`` class
 =======================
 
-To connect to a GitLab server, create a ``gitlab.Gitlab`` object:
+To connect to GitLab.com or another GitLab instance, create a ``gitlab.Gitlab`` object:
 
 .. code-block:: python
 
    import gitlab
 
-   # private token or personal token authentication
-   # Note that a 'url' that results in 301/302 redirects will cause an error
-   # (see below for more information).
+   # anonymous read-only access for public resources (GitLab.com)
+   gl = gitlab.Gitlab()
+
+   # anonymous read-only access for public resources (self-hosted GitLab instance)
+   gl = gitlab.Gitlab('https://gitlab.example.com')
+
+   # private token or personal token authentication (GitLab.com)
+   gl = gitlab.Gitlab(private_token='JVNSESs8EwWRx5yDxM5q')
+
+   # private token or personal token authentication (self-hosted GitLab instance)
    gl = gitlab.Gitlab(url='https://gitlab.example.com', private_token='JVNSESs8EwWRx5yDxM5q')
 
    # oauth token authentication
-   gl = gitlab.Gitlab('http://10.0.0.1', oauth_token='my_long_token_here')
+   gl = gitlab.Gitlab('https://gitlab.example.com', oauth_token='my_long_token_here')
 
    # job token authentication (to be used in CI)
    import os
-   gl = gitlab.Gitlab('http://10.0.0.1', job_token=os.environ['CI_JOB_TOKEN'])
-
-   # anonymous gitlab instance, read-only for public resources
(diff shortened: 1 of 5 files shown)


bufferoverflow (approved):
great idea!

gitlab/__init__.py
```diff
@@ -68,7 +68,7 @@ class Gitlab(object):
 
     def __init__(
         self,
-        url,
+        url=DEFAULT_URL,
```
nejch, line 71: @max-wittig can you think of this breaking anything? I couldn't think of a scenario :thinking: 

I was also thinking of adding the same for the CLI, but would like to add some cli args for the token for it to make sense, so I'll try in a follow-up.

max-wittig (approved):

gitlab/__init__.py
nejch: Oof, thanks for this, I found another bug based on your comment, as passing `None` explicitly as a positional arg with this would break the base url, so I pushed a new commit.

With this I think it covers all your scenarios as well (see new parametrized test in https://github.com/python-gitlab/python-gitlab/commit/77ac807f8c66d4debfde79ea450c4e35a17c506e) but I agree there may be other things. Should I mark this a draft and add it to a new 3.0.0 milestone? Is there any other breaking stuff maybe coming up other than the dream of httpx? :grin:

JohnVillalovos:
If you do decide to make a breaking change I wonder if requiring kwargs would be a good idea?

```
def __init__(
    self,
    *,
    url,
    private_token=None,
    oauth_token=None,
    job_token=None,
    ssl_verify=True,
    http_username=None,
    http_password=None,
    timeout=None,
    api_version="4",
    session=None,
    per_page=None,
    pagination=None,
    order_by=None,
    user_agent=gitlab.const.USER_AGENT,
):

```

max-wittig:
@JohnVillalovos Not sure what you mean by `requiring kwargs would be a good idea`. Could you elaborate?

JohnVillalovos:
> 
> 
> @JohnVillalovos Not sure what you mean by `requiring kwargs would be a good idea`. Could you elaborate?

Sure. So by using the `*` in the argument list it will require people to use keyword arguments.

Allowed:
```
gl = gitlab.Gitlab(url="gitlab.com")
```

Not allowed:
```
gl = gitlab.Gitlab("gitlab.com")

TypeError: __init__() takes 1 positional argument but 2 were given
```

max-wittig:
@JohnVillalovos I'm not so sure. I'm not really a fan of this change. This would really break everything.

JohnVillalovos:
> 
> 
> @JohnVillalovos I'm not so sure. I'm not really a fan of this change. This would really break everything.

No worries. As a note you can put the `*` where ever you like in the arguments list. So if you wanted to require keyword arguments for everything but `url` you could put the `*` after the `url` argument.

gitlab/__init__.py
nejch: I've added a breaking change trailer in the commit message to mark it for 3.0.0.

nejch:
> > @JohnVillalovos I'm not so sure. I'm not really a fan of this change. This would really break everything.
> 
> No worries. As a note you can put the `*` where ever you like in the arguments list. So if you wanted to require keyword arguments for everything but `url` you could put the `*` after the `url` argument.
> 
> ```
> def __init__(
>     self,
>     url,
>     *,
>     private_token=None,
>     oauth_token=None,
>     job_token=None,
>     ssl_verify=True,
>     http_username=None,
>     http_password=None,
>     timeout=None,
>     api_version="4",
>     session=None,
>     per_page=None,
>     pagination=None,
>     order_by=None,
>     user_agent=gitlab.const.USER_AGENT,
> ):
> ```

I agree with Max I've seen a lot of use of this in tiny scripts, and there people probably want to use quick one-liners without kwargs just to instantiate `Gitlab` with a private token, so I'd keep this as is for now.

nejch:
@JohnVillalovos while resolving conflicts I consolidated the URL redirect warnings as I saw they were in 3 places, in case you'd like to have 1 final look, but this should also be ready now.

JohnVillalovos (approved):

PR #1288, opened 2021-02-07
refactor(v4): split objects and managers per API resource

nejch:
This one is huge, so sorry :see_no_evil: It does what https://github.com/python-gitlab/python-gitlab/pull/997 was for (splitting the 6000 line monster), but with less restructuring (only splitting the resources into modules). I'm not sure if all the importing is the best approach, but it's quite explicit. 

I can imagine this is too much for review so I can try to split it, but the managers were quite intertwined so I was fixing things along the way. Here is the new import chain (mostly due to Project and Group including many other managers) - output of `pydeps --rmprefix='gitlab.v4.objects.' --only='gitlab.v4.objects' gitlab/v4/objects/__init__.py`:

image

Each module only imports the managers it needs, while `__init__.py` does a star import of all of them for backwards compatibility. Each module exports only its own classes for the star import explicitly via `__all__`.

In the end this needed to happen in my opinion as well, to keep sane file sizes. But not sure what's more painful, reviewing this big PR or doing it in stages. I'm sure there are better ways to do this so if anyone has ideas, while keeping backwards compatibility, I'd be open too! :)

If this gets merged I can rebase other people's PR's touching `objects` so it's not too demotivating for them.

This would now roughly imitate the structure of the GitLab API docs and gitlab's own `lib/api` directory:
- https://gitlab.com/gitlab-org/gitlab/-/tree/master/lib/api
- https://docs.gitlab.com/ee/api/api_resources.html


nejch:
Oh no :grin: 
image

max-wittig:
That's a nice MR 👍  Will indeed take some time to review 😄

nejch:
> That's a nice MR :+1:  Will indeed take some time to review :smile:

Understandable :smile: I found that with `git diff --color-moved master`, it should show with different colors what's moved and what's added/deleted (or `git diff --color-moved=plain master` since the default gives you a funky zebra pattern).

So it'd show it's mostly just cut/paste :) I couldn't really find a good GUI difftool that does it well for moves across files but I'm sure there's some out there, if it makes it a bit clearer.

max-wittig:
Thanks for all the refactoring done here 👍

bufferoverflow:
@nejch very nice work! @max-wittig a heavy review :smile:

JohnVillalovos:
@nejch Any plans on doing something similar for gitlab/__init__.py ?

nejch:
> @nejch Any plans on doing something similar for gitlab/**init**.py ?

I wasn't currently working on anything, but maybe it would make sense to take the class out into a `client` module to make it slim, and possibly split it a bit or if you have any other ideas. I think @max-wittig you also mentioned you wanted some cleanup there some time ago? :grinning: But maybe you can open a new issue (or PR) for that.

JohnVillalovos:
> 
> 
> > @nejch Any plans on doing something similar for gitlab/**init**.py ?
> 
> I wasn't currently working on anything, but maybe it would make sense to take the class out into something like a `client` module to make it slim, and possibly split it a bit or if you have any other ideas. I think @max-wittig you also mentioned you wanted some cleanup there some time ago? 😀 But maybe you can open a new issue (or PR) for that.

Thanks. I'll try to do a PR and move most of what is in `gitlab/__init__.py` to `gitlab/client.py`.

Issue #1289, opened 2021-02-09
Gitlab CLI typo in current-user-gpg-key


tback:
## Description of the problem, including code/CLI snippet
The command to list gpg keys of the current user contains a typo

```
gitlab current-user-gpg-keys
```

## Expected Behavior
```
[List of users gpg keys]
```

## Actual Behavior
```
gitlab current-user-gpg-keys
gitlab: error: argument what: invalid choice: 'current-user-gpg-keys'...

gitlab current-user-gp-gkeys
Traceback: ...

KeyError: CurrentUserGpGkey
```

## Specifications

  - python-gitlab version: 2.6.0
  - API version you are using (v3/v4): v4
  - Gitlab server version (or gitlab.com): 13.3.9

nejch:
Hi @tback, currently the CLI commands are dynamically converted from CamelCase classes (to create commands for the CLI) and vice versa (to parse them from the CLI). The library incorrectly assumes those are separate words because of the uppercase in UserGPGKey.

As stated in the docs these are all auto-generated, so some may not work as expected currently. This might need some hardcoded exception list in a few cases :thinking:

PR #1290, opened 2021-02-09
fix(cli): fix parsing CLI objects to classnames 

nejch:
Closes https://github.com/python-gitlab/python-gitlab/issues/1289. Not sure if relying on requests' internal `CaseInsensitiveDict` for this is the best idea here but I didn't want to reinvent the wheel. But it's needed since proper camelcase -> kebab-case conversions are not fully reversible (see https://stackoverflow.com/a/1176023)

Fixes the issue for:
CurrentUserGPGKey
LDAPGroup
UserGPGKey

### DIFF
--- a/gitlab/tests/test_cli.py
+++ b/gitlab/tests/test_cli.py
@@ -34,6 +34,9 @@ from gitlab import cli
         ("class", "Class"),
         ("test-class", "TestClass"),
         ("test-longer-class", "TestLongerClass"),
+        ("current-user-gpg-key", "CurrentUserGPGKey"),
+        ("user-gpg-key", "UserGPGKey"),
+        ("ldap-group", "LDAPGroup"),
     ],
 )
 def test_what_to_cls(what, expected_class):
@@ -53,6 +56,9 @@ def test_what_to_cls(what, expected_class):
         ("TestClass", "test-class"),
         ("TestUPPERCASEClass", "test-uppercase-class"),
         ("UPPERCASETestClass", "uppercase-test-class"),
+        ("CurrentUserGPGKey", "current-user-gpg-key"),
+        ("UserGPGKey", "user-gpg-key"),
+        ("LDAPGroup", "ldap-group"),
     ],
 )
 def test_cls_to_what(class_name, expected_what):
--- a/gitlab/tests/test_cli.py
+++ b/gitlab/tests/test_cli.py
@@ -28,20 +28,37 @@ import pytest
 from gitlab import cli
 
 
-def test_what_to_cls():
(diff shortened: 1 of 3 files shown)


JohnVillalovos:
What do you think about having an optional `_arg_name` attribute that could be put on classes. If that was present using that as the argument name, otherwise use the `what_to_cls` function.

gitlab/tests/test_cli.py
```diff
@@ -29,20 +29,37 @@
...
 
-    assert "test-class" == cli.cls_to_what(TestClass)
-    assert "class" == cli.cls_to_what(Class)
+@pytest.mark.parametrize(
+    "class_name,expected_what",
+    [
+        ("Class", "class"),
+        ("TestClass", "test-class"),
+        ("TestUPPERCASEClass", "test-uppercase-class"),
+        ("UPPERCASETestClass", "uppercase-test-class"),
```
JohnVillalovos, line 56: Maybe add a test for

```
UserGPGKey
user-gp-gkey
```

As that was the bug.

JohnVillalovos:
FYI: These classes seem like ones to fail before this patch.

```
CurrentUserGPGKey
current-user-gp-gkey

LDAPGroup
l-da-pgroup

UserGPGKey
user-gp-gkey
```

JohnVillalovos:
> 
> 
> What do you think about having an optional `_arg_name` attribute that could be put on classes. If that was present using that as the argument name, otherwise use the `what_to_cls` function.

Here is a Proof-of-Concept for this:
https://github.com/python-gitlab/python-gitlab/pull/1362

nejch:
> What do you think about having an optional `_arg_name` attribute that could be put on classes. If that was present using that as the argument name, otherwise use the `what_to_cls` function.

I actually did this whole namespace lookup dance so that we wouldn't need to manually maintain another attribute in case people forget about it when adding new features :sweat_smile: but maybe your approach is clearer. I haven't looked at whether the current approach is problematic for typing. Either way I'll rebase and add some more tests. :)

gitlab/tests/test_cli.py
nejch: Added all the previously buggy scenarios, thanks!

JohnVillalovos (approved):

nejch:
@JohnVillalovos I just rebased, could you take one last look and merge if it's ok? :)

Based on some other discussions on (not) manually maintaining too many object attributes I think #1362 can be closed if this is merged, WDYT?

JohnVillalovos (approved):

JohnVillalovos:
Looks good to me @nejch   Thanks.

Issue #960, opened 2019-12-07
cli command line arguments for url and token


rrauenza:
(enhancement request)

What do you think of adding command line arguments to pass the url and token on the command line?  This would make integration with gitlab ci easier,

`gitlab --url https://$CI_SERVER_HOST --token $CI_JOB_TOKEN ...`

Either that, or an option to read them from the environment (or default them from the environment?)

max-wittig:
Good idea.

chrisoldwood:
Is this a duplicate of #893?

nejch:
@rrauenza if you'd be OK with env variables only, this can maybe be closed as a duplicate as Chris suggests and I can try and work on #893.

Environment variables just make sense as CLI arguments would end up passing environment variables when used in CI systems anyway (which is probably what a lot of people do with this library, like your example).

There are many issues open for CLI so maybe args can be a topic for when that gets reworked.

rrauenza:
Yeah, that sounds reasonable.

bittner:
I would need this feature now. Where would we implement the API token? As in:

```console
curl --silent --header "Authorization: Bearer $GITLAB_API_TOKEN" ...
```

Aside, for clarification:
- Do we technically need the distinction between the [3 types of token](https://github.com/python-gitlab/python-gitlab/blob/master/gitlab/config.py#L122-L138)?
- There is now also a [project access token](https://docs.gitlab.com/ee/user/project/settings/project_access_tokens.html). Which one of the 3 tokens needs to be set in that case?

For simplicity, I'd prefer to have it like @rrauenza originally requested: A single `--token` option for everything, which overrides any value in configurations files. We probably also should cover `--username` and `--password` (with an optional value to allow interactive input), which as a side effect will make [the container entry point](https://github.com/python-gitlab/python-gitlab/blob/master/docker-entrypoint.sh) unnecessary, e.g.

```console
gitlab --url https://$CI_SERVER_HOST --token $GITLAB_API_TOKEN ...
gitlab --username $CI_DEPLOY_USER --password ...
```

I'll try to find out if we can read from the environment from within the `argparse` setup code via the defaults. To my knowledge, there is not direct way to make argparse consider environment variables (unlike [click](https://click.palletsprojects.com/en/7.x/arguments/#environment-variables)).

nejch:
Yeah, click would make a lot more sense for this, which is why this has kind of stalled.
To make it a global option you would probably want to add them to the base parser here:
[...]

Issue #893, opened 2019-10-04
CLI usage without configuration file


ghost:
Requiring a configuration file is not ideal for using the CLI on a CI server. It would be great if the required settings could also be passed as arguments or predefined environment variables.

bufferoverflow:
That's a good idea and I'm sure contributions on this are welcome.

atmishra:
Can I work on this?

max-wittig:
@atmishra Sure. Contributions are always welcome!

Issue #1195, opened 2020-09-23
Add Oauth2 login to cli


max-wittig:
## Description of the problem, including code/CLI snippet

Let's see if we can do something similar to the GitHub CLI

![image](https://user-images.githubusercontent.com/6639323/93970731-fe209700-fd6e-11ea-868f-30ec5126edba.png)

nejch:
I kind of looked into this a bit for fun, as a next step now that the CLI is a bit easier to use. I think it would be possible to do conveniently _only_ for GitLab.com, if we 
1. register an OAuth app on GitLab.com (e.g. a group-level oauth app from `python-gitlab/` with PKCE)
2. commit the client ID in the repo, and then
3. follow the [PKCE flow](https://docs.gitlab.com/ee/api/oauth2.html#authorization-code-with-proof-key-for-code-exchange-pkce). A nice description is actually available for this in the [pkce-flow](https://pypi.org/project/pkce-flow/) package.

But for self-managed users would probably need to create their own OAuth apps and paste in the ID and instance URL, so the convenience over just creating a PAT in the web UI is kind of lost. It would actually be easier then to just steal the browser cookies and create a PAT like that, which seems hacky to me.

If I'm not mistaken, it seems `gh` can get away with this because their CLI app is completely baked into the GitHub.com as well as GitHub EE servers themselves (just by reading things like https://github.com/cli/cli/pull/4523, but I can't tell as the rest is closed source).

See:
https://github.com/cli/cli/blob/a9b2c0dc82158a1d31fd9c3c2d94e5f2511be98a/internal/authflow/flow.go#L17-L24

and
https://github.com/cli/cli/blob/eaa64df80193f4f632d574f25729460d8b657819/internal/authflow/flow.go#L76-L78

But maybe I'm just a bit clueless here with OAuth. Open to ideas @max-wittig @JohnVillalovos :sweat_smile:

max-wittig:
Thanks for your investigation @nejch. As most people are using gitlab.com, we could think about adding this.

Otherwise it's really mostly a one-time thing for most users, maybe a message is really enough for now.

mstenz:
> As most people are using gitlab.com, we could think about adding this.

How did you come up with this? Probably this would be valid in the OpenSource world, but EE very likely uses self-installed variant for different reasons I have the feeling (and also speaking for myself)

nejch:
@mstenz yes I'd say that assumption is for open source.

[...]

Issue #715, opened 2019-02-26
Secure token storage


jaraco:
## Description of the problem, including code/CLI snippet

I haven't read the code, but from reading the documentation, it sounds like python-gitlab only has support for soliciting a private-token from a plain-text config file in the user's home directory. Better techniques exist for storing secrets more securely, including the [keyring library](https://pypi.org/project/keyring), which stores passwords in system-managed secure stores. It would be nice if the library would allow and even recommend that tokens be stored/queried there.

## Expected Behavior

Users would be incentivized not to store any secrets on the file system in plain text.

## Actual Behavior

Users are required to store secrets in the file system in plain text.

## Specifications

  - python-gitlab version: 1.8.0
  - API version you are using (v3/v4): n/a
  - Gitlab server version (or gitlab.com): n/a

jamesquilty:
It's a reasonable point, but I would note that passwordless SSH private keys, required for non-interactive scripted access to remote git repositories, are effectively the same as a token and are stored in the file system in plain text format. The important thing is that these files are readable only by the owner and not anyone else. That would mean that, for example, on macOS and Linux systems the ~/.python-gitlab.cfg file should have `-rw-------` permissions.

I'd suggest that this issue requesting "Secure token storage" should perhaps be for python-gitlab to raise a warning and/or enforce appropriate config file permissions.

devpie:
It just installed python-gitlab and wondered, why there is no keyring integration ;-)

@jamesquilty while you are probably right, i always choose a more secure way of storing secrets than simply not allowing others to read them.

For me this is a +1

nejch:
This is an old one, but closed now with https://github.com/python-gitlab/python-gitlab/pull/1359 thanks to @klorenz!

PR #1296, opened 2021-02-14
feat(cli): allow options from CLI args or environment variables

nejch:
Closes https://github.com/python-gitlab/python-gitlab/issues/960
Closes https://github.com/python-gitlab/python-gitlab/issues/893

Makes it easier to later introduce https://github.com/python-gitlab/python-gitlab/issues/715 hopefully, and maybe https://github.com/python-gitlab/python-gitlab/issues/1195.

BREAKING-CHANGE: The gitlab CLI will now accept CLI arguments
and environment variables for its global options in addition
to configuration file options. This may change behavior for
some workflows such as running inside GitLab CI and with
certain environment variables configured.

### DIFF
--- a/docs/cli-usage.rst
+++ b/docs/cli-usage.rst
@@ -3,17 +3,60 @@
 ####################
 
 ``python-gitlab`` provides a :command:`gitlab` command-line tool to interact
-with GitLab servers. It uses a configuration file to define how to connect to
-the servers. Without a configuration file, ``gitlab`` will default to
-https://gitlab.com and unauthenticated requests.
+with GitLab servers.
+
+This is especially convenient for running quick ad-hoc commands locally, easily
+interacting with the API inside GitLab CI, or with more advanced shell scripting
+when integrating with other tooling.
 
 .. _cli_configuration:
 
 Configuration
 =============
 
-Files
------
+``gitlab`` allows setting configuration options via command-line arguments,
+environment variables, and configuration files.
+
+For a complete list of global CLI options and their environment variable
+equivalents, see :doc:`/cli-objects`.
+
+With no configuration provided, ``gitlab`` will default to unauthenticated
+requests against `GitLab.com <https://gitlab.com>`__.
+
+With no configuration but running inside a GitLab CI job, it will default to
+authenticated requests using the current job token against the current instance
+(via ``CI_SERVER_URL`` and ``CI_JOB_TOKEN`` environment variables).
+
+.. warning::
+   Please note the job token has very limited permissions and can only be used
+   with certain endpoints. You may need to provide a personal access token instead.
+
+When you provide configuration, values are evaluated with the following precedence:
+
+1. Explicitly provided CLI arguments,
+2. Environment variables,
+3. Configuration files:
+
+   a. explicitly defined config files:
+
+      i. via the ``--config-file`` CLI argument,
+      ii. via the ``PYTHON_GITLAB_CFG`` environment variable,
+
+   b. user-specific config file,
+   c. system-level config file,
+
+4. Environment variables always present in CI (``CI_SERVER_URL``, ``CI_JOB_TOKEN``).
+
+Additionally, authentication will take the following precedence
+when multiple options or environment variables are present:
+
+1. Private token,
+2. OAuth token,
+3. CI job token.
+
+
+Configuration files
+-------------------
 
 ``gitlab`` looks up 3 configuration files by default:
 
@@ -35,8 +78,8 @@ You can use a different configuration file with the ``--config-file`` option.
     If the environment variable is defined and the target file cannot be accessed,
     ``gitlab`` will fail explicitly.
 
-Content
--------
+Configuration file format
+-------------------------
 
 The configuration file uses the ``INI`` format. It contains at least a
 ``[global]`` section, and a specific section for each GitLab server. For
--- a/gitlab/cli.py
+++ b/gitlab/cli.py
@@ -19,6 +19,7 @@
 
 import argparse
 import functools
+import os
 import re
 import sys
 from types import ModuleType
@@ -112,17 +113,25 @@ def _get_base_parser(add_help: bool = True) -> argparse.ArgumentParser:
         "-v",
         "--verbose",
         "--fancy",
-        help="Verbose mode (legacy format only)",
+        help="Verbose mode (legacy format only) [env var: GITLAB_VERBOSE]",
         action="store_true",
+        default=os.getenv("GITLAB_VERBOSE"),
     )
     parser.add_argument(
-        "-d", "--debug", help="Debug mode (display HTTP requests)", action="store_true"
+        "-d",
+        "--debug",
+        help="Debug mode (display HTTP requests) [env var: GITLAB_DEBUG]",
+        action="store_true",
+        default=os.getenv("GITLAB_DEBUG"),
     )
     parser.add_argument(
         "-c",
         "--config-file",
         action="append",
-        help="Configuration file to use. Can be used multiple times.",
+        help=(
+            "Configuration file to use. Can be used multiple times. "
+            "[env var: PYTHON_GITLAB_CFG]"
+        ),
     )
     parser.add_argument(
         "-g",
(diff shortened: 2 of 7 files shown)


gitlab/__init__.py
```diff
@@ -210,6 +211,55 @@ def from_config(cls, gitlab_id=None, config_files=None):
...
 
+    @classmethod
+    def merge_config(
+        cls,
+        options,
+        gitlab_id=None,
+        config_files=None,
+    ):
+        """Create a Gitlab connection by merging configuration in the following
+        order:
```
JohnVillalovos, line 222: s/order/priority/

I'm assuming that arguments via the CLI arguments have the highest priority and will not be over-written by other configs.

gitlab/cli.py
```diff
@@ -126,7 +132,86 @@ def _get_base_parser(add_help=True):
...
     )
-
+    # TODO: deduplicate everything by providing some data struct to loop over
+    parser.add_argument(
+        "--url",
+        help=("GitLab server URL [env var: GITLAB_URL]"),
+        required=False,
+        default=os.getenv("GITLAB_URL"),
+    )
+    parser.add_argument(
```
phillid, line 142: On many systems, command-line arguments passed to programs can be seen by all other users of the system. I'm not convinced it's a good approach to facilitate the passing of secret tokens using CLI arguments like this

phillid:
We found this PR after we tried including gitlab-cli in a GitLab CI/CD pipeline job. This PR looks like it will achieve what we wanted to do (drive the GitLab API without creating a config file in each job).

One outstanding wishlist item we had was if the tool could look for the same environment variables that are set inside a GitLab CI/CD job, i.e. CI_JOB_TOKEN, CI_SERVER_URL and so on [1]. If the tool shared these varaibles, it could authenticate with the GitLab instance - is that something that has already been considered?

[1]: https://docs.gitlab.com/ee/ci/variables/predefined_variables.html

gitlab/cli.py
nejch: This was mainly added for completeness and to be consistent with many other CLI tools that do offer it, including gitlab's own tooling https://gitlab.com/gitlab-org/release-cli/-/blob/master/docs/index.md#using-this-tool-with-a-private-token). Seems like even HashiCorp's vault allows login via CLI args.

But I agree otherwise and we already suggest using credential helpers for tokens in the docs. Maybe a compromise would be for `gitlab` to issue a warning if it finds that credentials were passed via CLI args + a mention in the docs advising against it.

nejch:
> 
> 
> We found this PR after we tried including gitlab-cli in a GitLab CI/CD pipeline job. This PR looks like it will achieve what we wanted to do (drive the GitLab API without creating a config file in each job) 🎉
> 
> One outstanding wishlist item we had was if the tool could look for the same environment variables that are set inside a GitLab CI/CD job, i.e. CI_JOB_TOKEN, CI_SERVER_URL and so on [1](https://docs.gitlab.com/ee/ci/variables/predefined_variables.html). If the tool shared these varaibles, it could authenticate with the GitLab instance without any extra code written in CI/CD jobs - is that something that has already been considered?

Yes, part of this was meant to implement zero-config for CI jobs. But I need to maybe tweak and add tests to really check all scenarios (for some endpoints, a PAT is needed, so we should allow overriding in a sensible precedence between cli args/env/config files/CI env). Currently the precedence planned is:

https://github.com/python-gitlab/python-gitlab/blob/cab343c1e9a75b52028b6f75f1a5c2cbeeb45749/gitlab/__init__.py#L224-L232

I'll try to get back to this PR in the coming days :)

phillid:
> Yes, part of this was meant to implement zero-config for CI jobs. But I need to maybe tweak and add tests to really check all scenarios (for some endpoints, a PAT is needed, so we should allow overriding in a sensible precedence between cli args/env/config files/CI env). Currently the precedence planned is:
> 
> https://github.com/python-gitlab/python-gitlab/blob/cab343c1e9a75b52028b6f75f1a5c2cbeeb45749/gitlab/__init__.py#L224-L232

Aha. I'm blind :eyes: 

That does indeed look like exactly the behaviour we wanted.

JohnVillalovos:
What do you think about using `argparse.ArgumentParser..add_mutually_exclusive_group()` ?

For example:

```
private_token_group = parser.add_mutually_exclusive_group()
private_token_group.add_argument(
        "--private-token",
        help=("GitLab private token"),
        required=False,
    )
private_token_group.add_argument(
        "--private-token-env-var",
        help=("Name of environment variable containing GitLab private token"),
        required=False,
    )
```

But now that I look at the code I'm not sure where to do the processing to load the environment variable 😟

I'll just leave this comment anyway and maybe it gives you an idea 🤷‍♂️

nejch:
> .add_mutually_exclusive_group()

Yes, I think it might be useful (not for arg/var exclusive I think but maybe for different types of tokens). But yes sometimes we want to validate those a bit later, because things like CI_JOB_TOKEN might always be present 🤦 

I think we just need to write a bunch of tests for expected behaviors in all the possible combinations, and write the code based on that.

nejch:
@JohnVillalovos took almost a year to get back to this, but ready for another look :sweat_smile:

docs/cli-usage.rst
```diff
@@ -3,17 +3,60 @@
...
+
+   a. explicitly defined config files:
+
+      i. via the ``--config-file`` CLI argument,
+      ii. via the ``PYTHON_GITLAB_CFG`` environment variable,
+
+   b. user-specific config file,
+   c. system-level config file,
+
+4. Environment variables always present in CI (CI_SERVER_URL, CI_JOB_TOKEN).
```
JohnVillalovos, line 48: Should it be `CI_SERVER_URL` and `CI_JOB_TOKEN` ?  Meaning use back-ticks on them.

gitlab/cli.py
```diff
@@ -151,7 +160,86 @@ def _get_base_parser(add_help: bool = True) -> argparse.ArgumentParser:
...
+        "--server-url",
+        help=("GitLab server URL [env var: GITLAB_URL]"),
+        required=False,
+        default=os.getenv("GITLAB_URL"),
+    )
+    parser.add_argument(
+        "--ssl-verify",
+        help=(
+            "Whether SSL certificates should be validated. [env var: GITLAB_SSL_VERIFY]"
```
JohnVillalovos, line 172: Are all of these environment variables documented somewhere besides here?

gitlab/client.py
```diff
@@ -16,7 +16,9 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """Wrapper for the GitLab API."""
 
+import os
 import time
+from argparse import Namespace
```
JohnVillalovos, line 21: nit: you know me I prefer `import argparse` as I like to read the code and see `argparse.Namespace` 😊

gitlab/client.py
```diff
@@ -256,6 +258,87 @@ def from_config(
...
+        job_token = options.job_token or config.job_token or os.getenv("CI_JOB_TOKEN")
+
+        if private_token:
+            return (private_token, None, None)
+        if oauth_token:
+            return (None, oauth_token, None)
+        if job_token:
+            return (None, None, job_token)
+
+        return (None, None, None)
```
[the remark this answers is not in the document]
JohnVillalovos: Makes sense. Thanks.

tests/unit/test_gitlab_auth.py
```diff
@@ -1,7 +1,10 @@
+from argparse import Namespace
```
JohnVillalovos, line 1: nit: ... 😊

gitlab/client.py
```diff
@@ -256,6 +258,87 @@ def from_config(
...
+            timeout=options.timeout or config.timeout,
+            api_version=options.api_version or config.api_version,
+            per_page=options.per_page or config.per_page,
+            pagination=options.pagination or config.pagination,
+            order_by=options.order_by or config.order_by,
+            user_agent=options.user_agent or config.user_agent,
+        )
+
+    @staticmethod
+    def _get_auth_from_env(
```
JohnVillalovos, line 317: nit: I'm not sure about this name. mainly the 'from_env' part of it.

tests/unit/test_gitlab_auth.py
```diff
@@ -83,3 +86,110 @@ def test_http_auth():
...
+    )
+    cp = GitlabConfigParser()
+    cp.private_token = config["private_token"]
+    cp.oauth_token = config["oauth_token"]
+    cp.job_token = config["job_token"]
+
+    private_token, oauth_token, job_token = Gitlab._get_auth_from_env(options, cp)
+    assert private_token == expected_private_token
+    assert oauth_token == expected_oauth_token
+    assert job_token == expected_job_token
```
JohnVillalovos, line 195: I'm asking a lot but would be nice to have a unit test which tests the ordering priority of the options.



1. Explicitly provided CLI arguments,
1. Environment variables,
1. Configuration files:
    -  explicitly defined config files:
    - via the --config-file CLI argument,
    - via the PYTHON_GITLAB_CFG environment variable,
    - user-specific config file,
    - system-level config file,
1. Environment variables always present in CI (CI_SERVER_URL, CI_JOB_TOKEN).

JohnVillalovos (approved):
A few nits. But overall looks good to me. Thanks!

gitlab/cli.py
nejch: They will get generated in the CLI reference documentation which I link to at the top of the CLI usage intro.

I agree it's not the nicest, but I also think the entire docs need a bit of a revamp to focus less on the configuration file (among other things).

I wasn't sure whether to add another manually maintained table that will start to go out of date :grin: just noticed the config file options are also incomplete. So maybe a single table with CLI/env/file config all in one would be best. With the current theme I think tables don't look so nice though, as the text doesn't really wrap. Should we open a follow-up?

gitlab/client.py
nejch: Thanks for this comment. After I looked at this, I realized I should probably decouple any `argparse` code from `client.py` since that's supposed to be independent of the CLI. Maybe we could just pass a dict of key-values to it, and do the conversion from `cli.py`.

This way people could even do something like this, if they want programmatically use config files but override something..
`gl = gitlab.Gitlab.merge_config({"private_token": "glpat-sometoken"}, gitlab_id="my-server", ["some-file.cfg"])`

Not sure, but some might find that useful :grinning: WDYT?

gitlab/client.py
nejch: True, especially with the comment above about removing `argparse` and using more generic dicts.

tests/unit/test_gitlab_auth.py
nejch: Hehe yeah, I thought the same. But it got quite tedious and not easily doable in unit tests, so I've added functional tests where it's a more realistic scenario.

I think large parts of this are covered in different tests already though, just not the whole chain. Is there a big gap you noticed here that should be covered in this PR already?

gitlab/cli.py
JohnVillalovos: I'm fine with a follow-up. I just looked at the generated docs and I think it would be very easy to not find that info so a separate table to me seems best.

gitlab/client.py
JohnVillalovos: That sounds reasonable to me. Makes sense that this file doesn't need to be worries about argument parsing. But not sure how much work that is for you...

tests/unit/test_gitlab_auth.py
JohnVillalovos: Yeah `functional` tests makes the most sense for me.

Not any big gaps that I can think of. I'll mark this as resolved.

nejch: Maybe doing a test for at least the server `url` with the whole chain would make sense, one with `CI_SERVER_URL` present and one without.

docs/cli-usage.rst
nejch: Absolutely! Done, thanks.

gitlab/client.py
nejch: Converted to a dict and using `vars`() in `cli.py` now.

Was even toying with the idea of using a `kwargs` dict and just let people do `gitlab.Gitlab.merge_config(private_token="glpat-bla")` which would be quite clean. Maybe I'll add that later.

gitlab/client.py
nejch: I used `_merge_auth()` now, let me know if it's still confusing!

tests/unit/test_gitlab_auth.py
nejch: Removed this part now as we only have `dict` for options.

gitlab/cli.py
nejch: Ok, I opened https://github.com/python-gitlab/python-gitlab/issues/1795 to track this.

JohnVillalovos (approved):
Thanks @nejch . Looks good.

PR #1298, opened 2021-02-15
Remove Python 2 code

JohnVillalovos:
httplib is a Python 2 library. It was renamed to http.client in Python
3.

https://docs.python.org/2.7/library/httplib.html


max-wittig:
Very nice! Thanks 👍

PR #1300, opened 2021-02-15
chore: remove unused ALLOWED_KEYSET_ENDPOINTS variable

JohnVillalovos:
The variable ALLOWED_KEYSET_ENDPOINTS was added in commit
f86ef3bbdb5bffa1348a802e62b281d3f31d33ad.

Then most of that commit was removed in commit
e71fe16b47835aa4db2834e98c7ffc6bdec36723, but ALLOWED_KEYSET_ENDPOINTS
was missed.

### DIFF
--- a/gitlab/__init__.py
+++ b/gitlab/__init__.py
@@ -45,8 +45,6 @@ REDIRECT_MSG = (
     "must update your GitLab URL to use https:// to avoid issues."
 )
 
-ALLOWED_KEYSET_ENDPOINTS = ["/projects"]
-
 
 class Gitlab(object):
     """Represents a GitLab server connection.


nejch:
I think this can go for now but for context, it was meant to follow endpoints where GitLab supports keyset pagination but it seems they're so far still stuck with only /projects (https://docs.gitlab.com/ee/api/README.html#keyset-based-pagination) and it broke some people's workflows initially due to missing headers.

@max-wittig maybe we can re-add this e.g.to `gitlab.const` when https://github.com/python-gitlab/python-gitlab/issues?q=is%3Aissue+is%3Aopen+keyset are revisited :smile:

max-wittig:
They clearly stated that their not gonna re-add X-Total-Items. Maybe we just have to document this, when you switch the default.

PR #1304, opened 2021-02-17
feat: add personal access token API

max-wittig:
See: https://docs.gitlab.com/ee/api/personal_access_tokens.html


nejch (approved):
👍 huh. sometimes these features just sneak into CE... need a script to scrape the API docs for new endpoints 😁

PR #949, opened 2019-11-26
feat: add audit endpoint

idanbensha:
Added audit events feature used for gitlab EE.
As requested in this issue:
https://github.com/python-gitlab/python-gitlab/issues/948

This code was tested and is working properly.

PR #1308, opened 2021-02-19
feat: add project audit endpoint

Sineaggi:
Similar to #949 but specific to [project audit events](https://docs.gitlab.com/ee/api/audit_events.html#project-audit-events).

No previous issues for this exist.

The code was tested to verify it works.

### DIFF
--- a/gitlab/tests/objects/test_audit_events.py
+++ b/gitlab/tests/objects/test_audit_events.py
@@ -1,0 +1,79 @@
+"""
+GitLab API:
+https://docs.gitlab.com/ee/api/audit_events.html#project-audit-events
+"""
+
+import re
+
+import pytest
+import responses
+
+from gitlab.v4.objects.audit_events import ProjectAudit
+
+id = 5
+
+audit_events_content = {
+    "id": 5,
+    "author_id": 1,
+    "entity_id": 7,
+    "entity_type": "Project",
+    "details": {
+        "change": "prevent merge request approval from reviewers",
+        "from": "",
+        "to": "true",
+        "author_name": "Administrator",
+        "target_id": 7,
+        "target_type": "Project",
+        "target_details": "twitter/typeahead-js",
+        "ip_address": "127.0.0.1",
+        "entity_path": "twitter/typeahead-js",
+    },
+    "created_at": "2020-05-26T22:55:04.230Z",
+}
+
+audit_events_url = re.compile(
+    r"http://localhost/api/v4/(((groups|projects)/1)|(admin/ci))/audit_events"
+)
+
+audit_events_url_id = re.compile(
+    rf"http://localhost/api/v4/(((groups|projects)/1)|(admin/ci))/audit_events/{id}"
+)
+
+
+@pytest.fixture
+def resp_list_audit_events():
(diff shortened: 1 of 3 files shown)


gitlab/v4/objects/projects.py
```diff
@@ -85,6 +87,17 @@ class GroupProjectManager(ListMixin, RESTManager):
     )
 
 
+class ProjectAudit(RESTObject):
+    _id_attr = "id"
+
+
+class ProjectAuditManager(ListMixin, RESTManager):
```
nejch, line 94: Based on the upstream docs, this endpoint supports both single + all events endpoint. We have a convenience `RetrieveMixin` for endpoitns that support both .get() and .list():

```suggestion
class ProjectAuditManager(RetrieveMixin, RESTManager):
```

nejch (changes requested):
Thanks a lot for this @Sineaggi! Just one small comment.

BTW - It might make sense to separate this into its own module, `audit_events.py`. Looking at the upstream docs, we also have instance and group audit events, so this could be expanded at a later point, and look something like the variables module:

https://github.com/python-gitlab/python-gitlab/blob/master/gitlab/v4/objects/variables.py

EDIT: I realize now that the current instance-level audit events were put into `events.py`, but they seem to have their own section in gitlab's api, so moving them together into `audit_events.py` might still make sense.

Then import the project-specific classes in `projects.py`.

If you feel like it, you could also add some unit tests for this as a sanity check, using the example responses from your API link as mocks. Have a look at the existing unit tests: https://github.com/python-gitlab/python-gitlab/tree/master/gitlab/tests/objects

Thanks again!

Sineaggi:
@nejch code moved to `audit_events.py`, moved to `RetrieveMixin`, added tests that execute both .get() and .list()

gitlab/tests/objects/test_audit_events.py
```diff
@@ -0,0 +1,79 @@
...
+        "target_type": "Project",
+        "target_details": "twitter/typeahead-js",
+        "ip_address": "127.0.0.1",
+        "entity_path": "twitter/typeahead-js",
+    },
+    "created_at": "2020-05-26T22:55:04.230Z",
+}
+
+audit_events_url = re.compile(
+    r"http://localhost/api/v4/(((groups|projects)/1)|(admin/ci))/audit_events"
```
nejch, line 35:
```suggestion
    r"http://localhost/api/v4/((groups|projects)/1/)audit_events"
```

gitlab/tests/objects/test_audit_events.py
```diff
@@ -0,0 +1,79 @@
...
+)
+
+audit_events_url_id = re.compile(
+    rf"http://localhost/api/v4/(((groups|projects)/1)|(admin/ci))/audit_events/{id}"
```
nejch, line 39:
```suggestion
    rf"http://localhost/api/v4/((groups|projects)/1/)audit_events/{id}"
```

nejch (commented):
Thanks again! Just found a little copy/paste leftover, I can push this.

nejch (approved):

PR #1309, opened 2021-02-21
WIP: chore: set default value for RESTManager._from_parent_attrs

JohnVillalovos:
Set a default value of {} for RESTManager._from_parent_attrs. This
will make using tools like mypy easier.


JohnVillalovos:
Not needed.

PR #1310, opened 2021-02-21
chore: explicitly import gitlab.v4.objects/cli

JohnVillalovos:
As we only support the v4 Gitlab API, explicitly import
gitlab.v4.objects instead of dynamically importing it depending on the
API version.

This has the added benefit of mypy being able to type check the Gitlab
__init__() function as currently it will fail if we enable type
checking of __init__() it will fail.

### DIFF
--- a/gitlab/cli.py
+++ b/gitlab/cli.py
@@ -19,7 +19,6 @@
 
 import argparse
 import functools
-import importlib
 import re
 import sys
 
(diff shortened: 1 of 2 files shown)


nejch (changes requested):
I just have 1 comment, currently we still dynamically import in the cli module:

https://github.com/python-gitlab/python-gitlab/blob/5cc60d5a8ac129652611d3dc12b350b5ca7262b9/gitlab/cli.py#L184

So IMO I'd change that too for consistency.

From a quick grep there's also this, might only be an issue later though, with objects:

https://github.com/python-gitlab/python-gitlab/blob/5cc60d5a8ac129652611d3dc12b350b5ca7262b9/gitlab/base.py#L62

Otherwise, this also helps by not confusing tools like pyinstaller/cx_freeze with dynamic imports so you don't need hooks for standalone executables. And according to https://docs.gitlab.com/ee/api/, 

> GraphQL co-exists with the current v4 REST API. If we have a v5 API, this should be a compatibility layer on top of GraphQL. 

So this makes sense to me, one if/elif block at some point in the future shouldn't hurt, people do it all the time for OS-dependent imports :smile: /cc @max-wittig

JohnVillalovos:
> I just have 1 comment, currently we still dynamically import in the cli module:
> 
> https://github.com/python-gitlab/python-gitlab/blob/5cc60d5a8ac129652611d3dc12b350b5ca7262b9/gitlab/cli.py#L184
> 
> So IMO I'd change that too for consistency.

Done.

> From a quick grep there's also this, might only be an issue later though, with objects:
> 
> https://github.com/python-gitlab/python-gitlab/blob/5cc60d5a8ac129652611d3dc12b350b5ca7262b9/gitlab/base.py#L62

I'm not quite sure what to do there 😟

PR #1311, opened 2021-02-21
fix: test_update_group() dependency on ordering

JohnVillalovos:
Since there are two groups we can't depend on the one we changed to
always be the first one returned.

Instead fetch the group we want and then test our assertion against
that group.

### DIFF
--- a/tools/functional/cli/test_cli_v4.py
+++ b/tools/functional/cli/test_cli_v4.py
@@ -42,7 +42,7 @@ def test_update_group(gitlab_cli, gl, group):
 
     assert ret.success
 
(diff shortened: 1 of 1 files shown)


nejch:
> Since there are two groups we can't depend on the one we changed to
> always be the first one returned.
> 
> Instead fetch the group we want and then test our assertion against
> that group.

Thanks, this will fix a lot of the flakiness.

There's probably some more weirdness and unnecessary dependency on execution order in functional tests, because they were rewritten from a single module with plain asserts. I'll try to get to that at some point.

PR #1316, opened 2021-02-21
fix: extend wait timeout for test_delete_user()

JohnVillalovos:


nejch:
Thanks for looking into this one. TBH I'm not sure anymore if the sidekiq metrics are a reliable source for the slow async operations - the point was to avoid hardcoded sleeps due to the underpowered CI VMs. I haven't taken the time to investigate it, so if you can solve it that's great, but don't lose too much time over this :)

JohnVillalovos:
> 
> 
> Thanks for looking into this one. TBH I'm not sure anymore if the sidekiq metrics are a reliable source for the slow async operations - the point was to avoid hardcoded sleeps due to the underpowered CI VMs. I haven't taken the time to investigate it, so if you can solve it that's great, but don't lose too much time over this :)

Thanks. Doing some debugging at the moment. I'll see what I figure out.

JohnVillalovos:
Not sure if this fixes the issue. Especially if the sidekick operations can not be counted on. 

But it should not hurt.

nejch:
> Not sure if this fixes the issue. Especially if the sidekick operations can not be counted on.
> 
> But it should not hurt.

Cool, I guess the only thing that might happen is that some rogue sidekiq processes will drag this longer, but this was probably already the case and I forget what I was trying to achieve with relying only on the first process :grin:

PR #1318, opened 2021-02-22
chore: remove usage of 'from ... import *' in client.py

JohnVillalovos:
In gitlab/client.py remove usage of:
      * from gitlab.const import *
      * from gitlab.exceptions import *

    Change them to:
      * import gitlab.const
      * import gitlab.exceptions

    Update code to explicitly reference things in gitlab.const and
    gitlab.exceptions

    A flake8 run no longer lists any undefined variables. Before it listed
    possible undefined variables.

### DIFF
--- a/gitlab/client.py
+++ b/gitlab/client.py
@@ -22,9 +22,9 @@ import requests
 import requests.utils
 
 import gitlab.config
-from gitlab.const import *  # noqa
-from gitlab.exceptions import *  # noqa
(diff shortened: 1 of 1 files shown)


nejch (approved):
Thanks! Would make sense to get rid of these blanket imports even in init but there are many things advertised publicly from that namespace even in docs (https://python-gitlab.readthedocs.io/en/stable/search.html?q=constants).

PR #1319, opened 2021-02-22
chore: remove usage of 'from ... import *'

JohnVillalovos:
In gitlab/v4/objects/*.py remove usage of:
  * from gitlab.base import *
  * from gitlab.mixins import *

    Change them to:
      * from gitlab.base import CLASS_NAME
      * from gitlab.mixins import CLASS_NAME

    Programmatically update code to explicitly import needed classes only.

After the change the output of:
  $ flake8 gitlab/v4/objects/*py | grep 'REST\|Mixin'

Is empty. Before many messages about unable to determine if it was a
valid name.

### DIFF
--- a/gitlab/v4/objects/access_requests.py
+++ b/gitlab/v4/objects/access_requests.py
@@ -1,5 +1,11 @@
-from gitlab.base import *  # noqa
-from gitlab.mixins import *  # noqa
+from gitlab.base import RESTManager, RESTObject
+from gitlab.mixins import (
+    AccessRequestMixin,
+    CreateMixin,
+    DeleteMixin,
+    ListMixin,
+    ObjectDeleteMixin,
+)
 
 
 __all__ = [
--- a/gitlab/v4/objects/appearance.py
+++ b/gitlab/v4/objects/appearance.py
@@ -1,6 +1,6 @@
 from gitlab import exceptions as exc
-from gitlab.base import *  # noqa
-from gitlab.mixins import *  # noqa
+from gitlab.base import RESTManager, RESTObject
+from gitlab.mixins import GetWithoutIdMixin, SaveMixin, UpdateMixin
 
 
 __all__ = [
--- a/gitlab/v4/objects/applications.py
+++ b/gitlab/v4/objects/applications.py
@@ -1,5 +1,5 @@
-from gitlab.base import *  # noqa
-from gitlab.mixins import *  # noqa
+from gitlab.base import RESTManager, RESTObject
+from gitlab.mixins import CreateMixin, DeleteMixin, ListMixin, ObjectDeleteMixin
 
 __all__ = [
     "Application",
--- a/gitlab/v4/objects/audit_events.py
+++ b/gitlab/v4/objects/audit_events.py
@@ -3,8 +3,8 @@ GitLab API:
 https://docs.gitlab.com/ee/api/audit_events.html#project-audit-events
 """
 
-from gitlab.base import *  # noqa
-from gitlab.mixins import *  # noqa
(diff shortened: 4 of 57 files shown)


JohnVillalovos:
I updated the files using this script:

```
#!/usr/bin/python3 -ttu
import argparse
import re
import sys


def main() -> int:
    args = parse_args()

    fixup_file(filename=args.filename)
    return 0


def fixup_file(*, filename: str):
    content = []
    with open(filename, "r") as in_file:
        for line in in_file.readlines():
            line = line.rstrip()
            if line.startswith("import gitlab.base"):
                print("This file seems to already have been processed. Exiting")
                return
            if line.startswith("import gitlab.mixins"):
                print("This file seems to already have been processed. Exiting")
                return
            line = fixed_line(line=line)
            content.append(line)

    with open(filename, "w") as out_file:
        for line in content:
            print(line, file=out_file)


def fixed_line(*, line: str) -> str:
    if line.startswith("from gitlab.base import"):
        line = "import gitlab.base"

    if line.startswith("from gitlab.mixins import"):
        line = "import gitlab.mixins"

    for rest_obj in (
        "RESTObject",
        "RESTObjectList",
        "RESTManager",
    ):
        line = line.replace(rest_obj, "gitlab.base.{}".format(rest_obj))

    for mixin_obj in (
        "AccessRequestMixin",
        "BadgeRenderMixin",
        "CreateMixin",
        "CRUDMixin",
        "DeleteMixin",
        "DownloadMixin",
        "GetMixin",
        "GetWithoutIdMixin",
        "ListMixin",
        "NoUpdateMixin",
        "ObjectDeleteMixin",
        "ParticipantsMixin",
        "RefreshMixin",
        "RetrieveMixin",
        "SaveMixin",
        "SetMixin",
        "SubscribableMixin",
        "TimeTrackingMixin",
        "TodoMixin",
        "UpdateMixin",
        "UserAgentDetailMixin",
    ):
        line = re.sub(
            r"\b" + mixin_obj + r"\b", "gitlab.mixins.{}".format(mixin_obj), line
        )

    return line


def parse_args() -> argparse.Namespace:
    parser = argparse.ArgumentParser()

    parser.add_argument("-f", "--filename", required=True)

    args = parser.parse_args()
    return args


if "__main__" == __name__:
    sys.exit(main())
```

nejch (changes requested):
Thanks for this @JohnVillalovos! I'm happy to see more star imports gone :grin: 

I'm just thinking that since these will be used in every new module as resources are added, perhaps we could import them all as needed into the module's namespace to keep a smaller footprint in the class definitions (and this diff would also be smaller). Especially since some of them will also grow with type annotations. Like this:

https://github.com/encode/httpx/blob/0f280af8b170ed5cc48c12a894f71a8b5762f748/httpx/_client.py#L34-L47 and
https://github.com/encode/httpx/blob/0f280af8b170ed5cc48c12a894f71a8b5762f748/httpx/__init__.py#L6-L35

And similar in requests, although I prefer the above formatting in httpx (black probably won't let us do otherwise anyway :smile: ):

https://github.com/psf/requests/blob/bdc00eb0978dd3cdd43f7cd1f95ced7aa75fab39/requests/__init__.py#L124-L128

I think this would be a bit more readable for future contributors while still explicit. WDYT? Since you were already able to script this it might be a quick change for you :) The only downside would be adding mixins to imports as they get added to modules each time.

JohnVillalovos:
Sure. I think I could modify the script to do that...  Let me look into it when I finish my work day.

Thanks for the review!

nejch (approved):

JohnVillalovos:
Sweet. Less 'import *' in the code 😊  Thanks for the merge!

PR #1321, opened 2021-02-23
chore: remove unused function _construct_url()

JohnVillalovos:
The function _construct_url() was used by the v3 API. All usage of the
function was removed in commit
fe89b949922c028830dd49095432ba627d330186


nejch (approved):
Agreed, was wondering some time ago if this was ever still going to be useful :)

PR #1322, opened 2021-02-23
fix: undefined name errors

JohnVillalovos:
Discovered that there were some undefined names.


nejch (approved):
Wow thanks, really need more test coverage for this.

Some of this must have just worked before some of the refactoring because of those blanket imports. Or no one ever noticed :thinking:

JohnVillalovos:
> 
> 
> Wow thanks, really need more test coverage for this.
> 
> Some of this must have just worked before some of the refactoring because of those blanket imports. Or no one ever noticed 🤔

I ran flake8 locally after doing the patch to remove the 'import *' lines. And found these missing names.

PR #1325, opened 2021-02-24
fix: tox pep8 target, so that it can run

JohnVillalovos:
Previously running the pep8 target would fail as flake8 was not
installed.

Now install flake8 for the pep8 target.

NOTE: Running the pep8 target fails as there are many warnings/errors.
But it does allow us to run it and possibly work on reducing these
warnings/errors in the future.

In addition, add two checks to the ignore list as black takes care of
formatting. The two checks added to the ignore list are:
  * E501: line too long
  * W503: line break before binary operator

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -16,6 +16,10 @@ commands =
   pytest gitlab/tests {posargs}
 
 [testenv:pep8]
+basepython = python3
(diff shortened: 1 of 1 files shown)


max-wittig:
That's a nice change. We should also add that to CI in the future.

PR #1327, opened 2021-02-24
feat(projects): add project access token api

max-wittig:
Fun Trivia: The **create** endpoint is totally useless btw. as it doesn't return the token and there is no way to get it

https://gitlab.com/gitlab-org/gitlab/-/issues/238991#note_516433443

### DIFF
--- a/gitlab/tests/objects/test_project_access_tokens.py
+++ b/gitlab/tests/objects/test_project_access_tokens.py
@@ -57,32 +57,6 @@ def resp_create_project_access_token():
 
 
 @pytest.fixture
-def resp_list_project_access_token():
-    content = [
-        {
-            "user_id": 141,
-            "scopes": ["api"],
-            "name": "token",
-            "expires_at": "2021-01-31",
-            "id": 42,
-            "active": True,
-            "created_at": "2021-01-20T22:11:48.151Z",
-            "revoked": False,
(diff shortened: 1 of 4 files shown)


gitlab/tests/objects/test_project_access_tokens.py
```diff
@@ -0,0 +1,139 @@
...
+        rsps.add(
+            method=responses.GET,
+            url="http://localhost/api/v4/projects/1/access_tokens",
+            json=content,
+            content_type="application/json",
+            status=200,
+        )
+        yield rsps
+
+
```
nejch, line 117: Looks like this is already on lines 9-32 :smile: 
```suggestion
```

nejch (changes requested):
Nice! Just found one little duplicate.

Also lol :facepalm: I had high hopes for this endpoint in the release notes but looks pointless to me :(

gitlab/tests/objects/test_project_access_tokens.py
```diff
@@ -0,0 +1,139 @@
...
```
nejch, line 84:
```suggestion
```

max-wittig: What did I do 😄

PR #1332, opened 2021-02-25
fix: wrong variable name

JohnVillalovos:
Discovered this when I ran flake8 on the file. Unfortunately I was the
one who introduced this wrong variable name :(


nejch (approved):
Haha well if someone tried to use the v3 API at this point they wouldn't get very far anyway :)

PR #1333, opened 2021-02-25
feat(users): add follow/unfollow API

max-wittig:


docs/gl_objects/users.rst
```diff
@@ -84,6 +89,15 @@ Delete an external identity by provider name::
 
     user.identityproviders.delete('oauth2_generic')
 
+Get the followers of a user
+
+    user.followers_users.list()
+
+Get the followings of a user
+
+    user.following_users.list()
```
max-wittig, line 98: @nejch I couldn't use `following` and `followers` directly, as it has the same name as it's user attribute.

max-wittig: What do you think about this naming?

gitlab/tests/objects/test_users.py
```diff
@@ -135,3 +201,17 @@ def test_user_activate_deactivate(user, resp_activate):
 
 def test_delete_user_identity(user, resp_delete_user_identity):
     user.identityproviders.delete("test_provider")
+
+
+def test_user_follow_unfollow(user, resp_follow_unfollow):
+    user.follow()
```
max-wittig, line 207: sadly in the test case only the `dict` is returned, so I cannot really assert the attributes of the `User` object.

Any ideas how to convert this to a `User` in the tests? It does work normally

nejch (changes requested):
Cool! It all looks good to me, just that little docstring return (unless you want to return a User, see other thread).

gitlab/tests/objects/test_users.py
max-wittig: Okay that makes sense. Let's keep the conventions then.

docs/gl_objects/users.rst
max-wittig: Then let's keep the clearer name I guess 😄

nejch (approved):

PR #1335, opened 2021-02-26
fix: remove duplicate class definitions in v4/objects/users.py

JohnVillalovos:
The classes UserStatus and UserStatusManager were each declared twice.
Remove the duplicate declarations.


nejch (approved):
Thanks!

Oops. Probably a good sign I should avoid massive refactors in one PR :rofl:

PR #1339, opened 2021-02-26
chore: add type-hints to gitlab/client.py

JohnVillalovos:
Adding some initial type-hints to gitlab/client.py

Unsure about my usage of assert to identify the type of the object. Feedback appreciated.

### DIFF
--- a/gitlab/client.py
+++ b/gitlab/client.py
@@ -17,6 +17,7 @@
 """Wrapper for the GitLab API."""
 
 import time
+from typing import cast, Any, Dict, List, Optional, Tuple, Union
 
 import requests
 import requests.utils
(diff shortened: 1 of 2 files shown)


gitlab/client.py
```diff
@@ -793,51 +849,57 @@ def _query(self, url, query_data=None, **kwargs):
...
+        assert self._total is not None
         return int(self._total)
 
-    def __iter__(self):
+    def __iter__(self) -> "GitlabList":
         return self
 
-    def __len__(self):
+    def __len__(self) -> int:
+        if self._total is None:
```
max-wittig, line 895: isn't that gonna break somebodies workflow? 😄

max-wittig:
@JohnVillalovos I love these type merge requests. Just a quick question. What about the depreciation of the typing module in Python 3.9? We may need some feature imports then right?

gitlab/client.py
JohnVillalovos: I don't think so. So what I was noticing is that sometimes __len__ gets called when it is None and thus a TypeError is raised when doing `int(None)`. But the caller (maybe `__iter__`) ignores it. But if I had an `assert self._total is not None` that blew up.

So it seems to be working but please let me know if you think something will break and can try to figure out a better way.

JohnVillalovos:
> 
> 
> @JohnVillalovos I love these type merge requests. Just a quick question. What about the depreciation of the typing module in Python 3.9? We may need some feature imports then right?

I was not aware of that. But I just read this: https://docs.python.org/3/library/typing.html

> The deprecated types will be removed from the typing module in the first Python version released 5 years after the release of Python 3.9.0.

Which sounds like we will have 5 years to take care of this. Maybe I misunderstand that.

gitlab/client.py
JohnVillalovos: So this is a sample program which shows how Python it ignores the type error, when using list()

```
$ cat test.py
import sys

class Testing:
    def __len__(self):
        print("Entered __len__")
        raise TypeError("we ignore this")
        return int(None)

    def __iter__(self):
        print("Entered __iter__")
        return self

    def __next__(self):
        raise StopIteration


def main() -> int:
    test = Testing()

    print(list(test))
    return 0


if "__main__" == __name__:
    sys.exit(main())
```

Run it:
```
$ python3 test.py
Entered __len__
Entered __iter__
Entered __len__
[]
```

max-wittig:
Okay then it should be fine.

max-wittig:
Thanks for the MR!

JohnVillalovos:
> 
> 
> Thanks for the MR!

Slowly coming along. One file at a time!

PR #1343, opened 2021-02-27
chore: add and fix some type-hints in gitlab/client.py 

JohnVillalovos:
Was able to figure out better type-hints for gitlab/client.py


nejch (approved):
Cool, looks like all method returns are now covered in client.py.

Issue #1306, opened 2021-02-18
Support Generate Changelog API


nejch:
Coming soon in 13.9 🚀 Automated releases are the best releases

https://docs.gitlab.com/ee/api/repositories.html#generate-changelog-data

KraMorK:
This feature is really needed, please implement it

PR #1347, opened 2021-02-28
chore(api): move repository endpoints into separate module

nejch:
Prep work for #1306 because `projects.py` is getting bloated again already (I'm splitting this into two PRs just for clarity).

I noticed all the `/repository/` endpoints were just kind of slapped into the project manager as custom methods. For now this just splits the methods into a separate Mixin, but perhaps later namespacing them with a separate manager would make it cleaner:

```python
project.repository.tree()
project.repository.blob()
project.repository.compare()
project.repository.contributors()
# ...
```
The old methods could then just call the manager, and maybe raise a deprecation warning.
WDYT @max-wittig would that make sense? I'd open a follow-up later for the repository manager.

### DIFF
--- a/gitlab/v4/objects/projects.py
+++ b/gitlab/v4/objects/projects.py
@@ -47,6 +47,7 @@ from .pipelines import (
 )
 from .push_rules import ProjectPushRulesManager
 from .releases import ProjectReleaseManager
+from .repositories import RepositoryMixin
 from .runners import ProjectRunnerManager
 from .services import ProjectServiceManager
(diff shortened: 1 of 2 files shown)


max-wittig:
I guess we don't need the depreciation warning. The next release is gonna be called 3.0.0 anyway. We already did a couple of breaking changes and that would make more sense, instead of maintaining the old behaviour. We're just a small library. People who don't want to switch can stay on the 2.X version.

max-wittig:
Looks very good!

PR #1349, opened 2021-02-28
WIP: Testing refactoring base classes of Mixins

JohnVillalovos:
NOTE: May not be necessary. I think if I change the ordering inside the class definitions I can fix things. This is due to MRO (Method Resolution Order) things. Basically RESTObject appears like it should be the last class in a multiple inheritance declaration.

Use RESTObject and RESTManager as the base classes for all the Mixin
classes.

Benefits are:
  * Simplifies usage of Mixins as don't need to add either RESTObject
    or RESTManager to every gitlab/v4/objects/ class
  * Type checking will work easier. Have attempted two different ways
    of type checking of Mixins and their derived classes and both
    failed :(

### DIFF
--- a/gitlab/mixins.py
+++ b/gitlab/mixins.py
@@ -16,7 +16,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import gitlab
-from gitlab import base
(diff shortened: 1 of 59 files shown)


JohnVillalovos:
I don't think we need this. So going to close.

PR #1350, opened 2021-03-01
chore: Put assert statements inside 'if TYPE_CHECKING:'

JohnVillalovos:
To be safe that we don't assert while running, put the assert
statements, which are used by mypy to check that types are correct,
inside an 'if TYPE_CHECKING:' block.

Also, instead of asserting that the item is a dict, instead assert
that it is not a requests.Response object. Theoretically the JSON
could return as a list or dict, though at this time we are assuming a
dict.

### DIFF
--- a/gitlab/client.py
+++ b/gitlab/client.py
@@ -17,7 +17,7 @@
 """Wrapper for the GitLab API."""
 
 import time
-from typing import cast, Any, Dict, List, Optional, Tuple, Union
+from typing import cast, Any, Dict, List, Optional, Tuple, TYPE_CHECKING, Union
 
 import requests
 import requests.utils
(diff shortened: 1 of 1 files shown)


gitlab/client.py
```diff
@@ -266,7 +266,8 @@ def lint(self, content: str, **kwargs: Any) -> Tuple[bool, List[str]]:
         """
         post_data = {"content": content}
         data = self.http_post("/ci/lint", post_data=post_data, **kwargs)
-        assert isinstance(data, dict)
+        if TYPE_CHECKING:
+            assert not isinstance(data, requests.Response)
```
nejch, line 270: out of curiosity: we didn't actually see this anywhere right, all the http_* methods process the result in some way?

nejch (approved):
Thanks @JohnVillalovos that definitely feels a bit safer not having asserts in production for older code, who knows :)

gitlab/client.py
JohnVillalovos: > 
> 
> out of curiosity: we didn't actually see this anywhere right, all the http_* methods process the result in some way?

There are some calls where they do want the `requests.Response` object back from the http_* methods.  As far as I can tell things want either one or the other. I don't see any (yet) that might expect either back.

PR #1352, opened 2021-03-01
fix: add a check to ensure the MRO is correct

JohnVillalovos:
~~This should be merged after:
https://github.com/python-gitlab/python-gitlab/pull/1344~~  DONE

Add a check to ensure the MRO (Method Resolution Order) is correct for classes in
gitlab.v4.objects.

```
An example of an incorrect definition:
    class ProjectPipeline(RESTObject, RefreshMixin, ObjectDeleteMixin):
                          ^^^^^^^^^^ This should be at the end.

Correct way would be:
    class ProjectPipeline(RefreshMixin, ObjectDeleteMixin, RESTObject):
                                      Correctly at the end ^^^^^^^^^^
```

Also fix classes which have the issue.

### DIFF
--- a/gitlab/tests/objects/test_mro.py
+++ b/gitlab/tests/objects/test_mro.py
@@ -106,7 +106,6 @@ def test_mros():
                 if obj.__module__ == "gitlab.base":
                     has_base = True
                     base_classname = obj.__name__
-                # print(f"\t{obj.__name__}: {obj.__module__}")
             if has_base:
                 filename = inspect.getfile(class_value)
                 # NOTE(jlvillal): The very last item 'mro[-1]' is always going
--- a/gitlab/tests/objects/test_mro.py
+++ b/gitlab/tests/objects/test_mro.py
@@ -1,0 +1,123 @@
+"""
+Ensure objects defined in gitlab.v4.objects have REST* as last item in class
+definition
+
+Original notes by John L. Villalovos
+
+An example of an incorrect definition:
+    class ProjectPipeline(RESTObject, RefreshMixin, ObjectDeleteMixin):
+                          ^^^^^^^^^^ This should be at the end.
+
+Correct way would be:
+    class ProjectPipeline(RefreshMixin, ObjectDeleteMixin, RESTObject):
+                                      Correctly at the end ^^^^^^^^^^
+
+
+Why this is an issue:
+
+  When we do type-checking for gitlab/mixins.py we make RESTObject or
+  RESTManager the base class for the mixins
+
+  Here is how our classes look when type-checking:
+
+      class RESTObject(object):
+          def __init__(self, manager: "RESTManager", attrs: Dict[str, Any]) -> None:
+              ...
+
+      class Mixin(RESTObject):
+          ...
+
+      # Wrong ordering here
+      class Wrongv4Object(RESTObject, RefreshMixin):
+          ...
+
+  If we actually ran this in Python we would get the following error:
+         class Wrongv4Object(RESTObject, Mixin):
+    TypeError: Cannot create a consistent method resolution
+    order (MRO) for bases RESTObject, Mixin
+
+  When we are type-checking it fails to understand the class Wrongv4Object
+  and thus we can't type check it correctly.
+
+Almost all classes in gitlab/v4/objects/*py were already correct before this
+check was added.
+"""
+import inspect
+
+import pytest
+
+import gitlab.v4.objects
+
+
+def test_show_issue():
+    """Test case to demonstrate the TypeError that occurs"""
+
+    class RESTObject(object):
+        def __init__(self, manager: str, attrs: int) -> None:
+            ...
+
+    class Mixin(RESTObject):
+        ...
+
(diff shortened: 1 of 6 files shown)


max-wittig:
Thanks for the MR. Could you elaborate why this matters? I know what you can override methods based on the order, but other than that, I'm unsure. Would be interesting to know @JohnVillalovos

JohnVillalovos:
> 
> 
> Thanks for the MR. Could you elaborate why this matters? I know what you can override methods based on the order, but other than that, I'm unsure. Would be interesting to know @JohnVillalovos

Sure. Let me write up a better explanation and post it. It will have to be after my work-day is over here in Oregon.

JohnVillalovos:
> > Thanks for the MR. Could you elaborate why this matters? I know what you can override methods based on the order, but other than that, I'm unsure. Would be interesting to know @JohnVillalovos
> 
> Sure. Let me write up a better explanation and post it. It will have to be after my work-day is over here in Oregon.

I have put comments in the test file that explain it better.

This PR is assuming that https://github.com/python-gitlab/python-gitlab/pull/1344 will be merged. As that is what triggered this issue.

JohnVillalovos:
@max-wittig  Hopefully I have answered your question.

JohnVillalovos:
Any more questions or comments about this?

gitlab/tests/objects/test_mro.py
```diff
@@ -0,0 +1,132 @@
...
+    """
+
+    debug = False
+    failed_messages = []
+    for module_name, module_value in inspect.getmembers(gitlab.v4.objects):
+        if not inspect.ismodule(module_value):
+            # We only care about the modules
+            continue
+        if debug:
+            print("Module:", module_name, type(module_value), module_value)
```
nejch, line 94: Not sure how you ran this but with pytest it will need capsys disabled to really print anything. Could you add a note in the module docstring on this?

E.g. `tox -e py39 -- -s` or

gitlab/tests/objects/test_mro.py
```diff
@@ -0,0 +1,132 @@
...
+
+def test_mros():
+    """Ensure objects defined in gitlab.v4.objects have REST* as last item in
+    class definition.
+
+    We do this as we need to ensure the MRO (Method Resolution Order) is
+    correct.
+    """
+
+    debug = False
```
nejch, line 87: nitpick: If we're adding all this, I'd prefer it as a pytest option e.g. in conftest.py instead of a hardcoded `debug` value (so people don't accidentally commit it when changing locally):

```python
def pytest_addoption(parser):
    parser.addoption(
        "--debug-print",
        action="store_true",
        default=False
        help="Print debug info in tests",
    )


@pytest.fixture
def debug_print(request):
    return request.config.getoption("--debug-print")
```

Then pass it to the test with `def test_mros(debug_print):` and use it there.

On the other hand I don't know if this is needed since pytest suppresses all this by default (see below) and you can just control it with capsys flags - in that case I'd just remove the check for debug. Up to you ;)

nejch (changes requested):
Thanks @JohnVillalovos! Just a few quick comments.

JohnVillalovos:
@nejch  Thanks for the feedback. I'll try to get some time to work on this in the next few days.

gitlab/tests/objects/test_mro.py
```diff
@@ -0,0 +1,132 @@
...
+#   If we actually ran this in Python we would get the following error:
+#          class Somev4Object(RESTObject, Mixin):
+#     TypeError: Cannot create a consistent method resolution
+#     order (MRO) for bases RESTObject, Mixin
+#
+#   When we are type-checking it fails to understand the class Somev4Object
+#   and thus we can't type check it correctly.
+#
+# Almost all classes in gitlab/v4/objects/*py were already correct before this
+# check was added.
```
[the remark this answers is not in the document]
JohnVillalovos: Done.

gitlab/tests/objects/test_mro.py
JohnVillalovos: I'll just go ahead and remove this debug code.

gitlab/tests/objects/test_mro.py
JohnVillalovos: I'll just go ahead and remove this debug code.

JohnVillalovos:
@nejch  Updated. I removed the debug code and moved the comments into the docstring.

gitlab/tests/objects/test_mro.py
```diff
@@ -0,0 +1,123 @@
...
+
+            mro = class_value.mro()
+
+            # We only check classes which have a 'gitlab.base' class in their MRO
+            has_base = False
+            for count, obj in enumerate(mro, start=1):
+                if obj.__module__ == "gitlab.base":
+                    has_base = True
+                    base_classname = obj.__name__
+                # print(f"\t{obj.__name__}: {obj.__module__}")
```
nejch, line 109:
```suggestion

```

nejch (approved):

PR #1359, opened 2021-03-06
feat(config): option to configure a helper to lookup the token

klorenz:
I have stored my personal access token with all other secrets in UNIX password store.  
For not having the token stored on the disk, I have implemented the option to let tokens be
looked up from a helper program.

Example:
```ini
[my_config]
url = https://gitlab.server
private_token = helper: ~/bin/pass-helper.sh
api_version = 4
```
whith a helper script ``~/bin/pass-helper.sh``:
```bash
#!/bin/bash
pass show path/to/token | head -n 1
```

### DIFF
--- a/gitlab/config.py
+++ b/gitlab/config.py
@@ -36,9 +36,7 @@ _DEFAULT_FILES: List[str] = _env_config() + [
 
 HELPER_PREFIX = "helper:"
 
-HELPER_ATTRIBUTES = [
-    "job_token", "http_password", "private_token", "oauth_token"
-]
+HELPER_ATTRIBUTES = ["job_token", "http_password", "private_token", "oauth_token"]
 
 class ConfigError(Exception):
     pass
@@ -202,11 +200,11 @@ class GitlabConfigParser(object):
             pass
 
     def _get_values_from_helper(self):
-        """Update attributes, which may get values from an external helper program
-        """
+        """Update attributes, which may get values from an external helper program"""
         for attr in HELPER_ATTRIBUTES:
             value = getattr(self, attr)
-            if isinstance(value, str) and value.lower().strip().startswith(HELPER_PREFIX):
+            _value_lower = value.lower().strip()
+            if isinstance(value, str) and _value_lower.startswith(HELPER_PREFIX):
                 helper = value[len(HELPER_PREFIX) :].strip()
                 value = subprocess.check_output([helper]).decode("utf-8").strip()
                 setattr(self, attr, value)
--- a/gitlab/config.py
+++ b/gitlab/config.py
@@ -204,8 +204,10 @@ class GitlabConfigParser(object):
         """Update attributes, which may get values from an external helper program"""
         for attr in HELPER_ATTRIBUTES:
             value = getattr(self, attr)
-            _value_lower = value.lower().strip()
-            if isinstance(value, str) and _value_lower.startswith(HELPER_PREFIX):
+            if not isinstance(value, str):
+                continue
+
+            if value.lower().strip().startswith(HELPER_PREFIX):
                 helper = expanduser(value[len(HELPER_PREFIX) :].strip())
                 value = subprocess.check_output([helper]).decode("utf-8").strip()
                 setattr(self, attr, value)
--- a/gitlab/config.py
+++ b/gitlab/config.py
@@ -17,9 +17,10 @@
 
 import os
 import configparser
+import shlex
 import subprocess
 from typing import List, Optional, Union
-from os.path import expanduser
+from os.path import expanduser, expandvars
 
 from gitlab.const import USER_AGENT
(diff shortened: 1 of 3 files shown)


klorenz:
Hi @nejch ,

to fix commitlinting issues, I have squashed my commits and created better commit messages.

gitlab/tests/test_config.py
```diff
@@ -192,6 +196,17 @@ def test_valid_data(m_open, path_exists):
...
+    fd.close = mock.Mock(return_value=None)
+    m_open.return_value = fd
+    cp = config.GitlabConfigParser(gitlab_id="five")
+    assert "five" == cp.gitlab_id
+    assert "https://five.url" == cp.url
+    assert None == cp.private_token
+    assert "foobar" == cp.oauth_token
+    assert 2 == cp.timeout
+    assert True == cp.ssl_verify
+
```
nejch, line 209: Thanks for adding tests for this :)

This test is really growing now, it might make sense to split it into a new `test_data_from_helper`. Then you could check: what error the user will see if the subprocess exits with non-zero, if it returns nothing, in case there's nothing given in the config after the prefix, etc.

Most of the other stuff in this class is wrapped in blanket try/except blocks so I think if we don't do it for helpers, it should be tested well.

gitlab/config.py
```diff
@@ -150,6 +151,16 @@ def __init__(
         except Exception:
             pass
 
+        for attr in ("job_token", "http_password", "private_token", "oauth_token"):
+            value = getattr(self, attr)
+            prefix = "lookup:"
+            if isinstance(value, str) and value.lower().strip().startswith(prefix):
+                helper = value[len(prefix) :].strip()
+                value = (
+                    subprocess.check_output(helper, shell=True).decode("utf-8").strip()
```
nejch, line 160: I think we should at least warn the user in the docs about `shell=True` or link to https://docs.python.org/3/library/subprocess.html#security-considerations. I guess you added it here so people can use pipes etc.

Also I guess this opens the user to arbitrary commands executed on the machine if someone manages to put a config file with precedence or in `PYTHON_GITLAB_CFG` env var. @max-wittig WDYT about using `shell=True` and this 2nd point here, does it matter? :thinking:

nejch (changes requested):
Thanks a lot for this feature, this is really really neat :) People have requested native keyring support but I think this might be even more flexible as it can be used with any backend/CLI client like vault etc.

I've just added a few comments, sorry it's just a very user-facing feature so I thought it'd be good to agree on the approach from the start.

Also I think this deserves a mention in the docs, would you mind adding a section in https://github.com/python-gitlab/python-gitlab/blob/master/docs/cli-usage.rst?

gitlab/config.py
klorenz: We could also rather run a simple command and leave it up to the user to create a shell script.  If I remember right, this is the behavior from docker.

gitlab/config.py
```diff
@@ -150,6 +151,16 @@ def __init__(
         except Exception:
             pass
 
+        for attr in ("job_token", "http_password", "private_token", "oauth_token"):
```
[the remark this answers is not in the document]
klorenz: Done this.

gitlab/config.py
```diff
@@ -150,6 +151,16 @@ def __init__(
...
+            value = getattr(self, attr)
+            prefix = "lookup:"
```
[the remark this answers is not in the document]
klorenz: Done this.  (is now named "helper:")

gitlab/tests/test_config.py
klorenz: Improved the test and adapted to changed helper.

gitlab/config.py
klorenz: Removed the ``shell=True`` and instead allow only a path to the helper program.  I also updated the documentation.

nejch: Thank you for the changes @klorenz and sorry for the delay! I'll be merging in a minute.

I really liked the convenience of your initial implementation, and I don't want to block this PR any longer so I just pushed a small change (I credited you as the author still) - I went for a compromise where `shell=False` is still used and recommend wrappers for any scripts that require piping etc. I hope this is ok with you. 
I also expanded the docs and wrapped the call in a try/except block for people to have a meaningful erorr message on the CLI. See https://github.com/python-gitlab/python-gitlab/pull/1359/commits/91ffb8e97e213d2f14340b952630875995ecedb2, let me know if you want to revisit this in a follow-up.

nejch (approved):

klorenz:
Hi @nejch, yes I am fine with your changes. This is a good compromise
between convenience and security. Thanks :)

Nejc Habjan ***@***.***> schrieb am So., 18. Apr. 2021, 12:11:

PR #1360, opened 2021-03-06
fix: make scopes work with multiple scope-names

klorenz:
I have found out, that it was not possible to use multiple comma-separated scopes from command line as documented.  These changes fixes that issue.


JohnVillalovos:
When I look at "Files changed" only one file is changed. Was that the intent? The second commit is a Merge commit which doesn't seem correct.

gitlab/types.py
```diff
@@ -45,6 +45,21 @@ def get_for_api(self):
...
+            self._value = [item.strip() for item in cli_value.split(",")]
+
+    def get_for_api(self):
+        # Do not comma-split single value passed as string
+        if isinstance(self._value, str):
+            return [self._value]
+
+        return self._value
+
+
```
nejch, line 62: Since you are inheriting from ListAttribute you don't need to duplicate methods. Just override the one you need to and maybe instead of copying the previous comment, add one (or a docstring) about why it needs to be different from the inherited class. Unless I missed a small difference in `set_from_cli` then please correct me :grin:

nejch (changes requested):
Thanks a lot for this @klorenz I think some people have asked for this already and will be happy to see it fixed! I'll try to find related issues so we can close them if it fixes them.

> When I look at "Files changed" only one file is changed. Was that the intent? The second commit is a Merge commit which doesn't seem correct.
> 
> It seems like this patch was started before gitlab/v4/objects/**init**.py was split up into multiple files.

I agree with @JohnVillalovos, looks like with your merge you lost the custom `_types` attributes. Could you re-add them in the split objects/ files? Also looks like you had an unrelated change in the initial commit, please submit that as a separate PR if you add it back ;)

Apart from John's comment see also my note on only overriding the right method.

And finally since I see from the other PR that you're comfortable with unit tests here could you add one to prove it fixes the issue? ;)

klorenz:
Hi, i struggled a bit with the linting checks you do and had to redo the
commits. I check today if everything is in there. Sorry for the trouble.

Regards, Kiwi

Nejc Habjan <notifications@github.com> schrieb am Sa., 6. März 2021, 18:43:

gitlab/types.py
klorenz: Hi @nejch,

indeed I lost the changes.  I implemented the original code in the __init__.py (some versions ago) and with the split I oversaw to add the changes in all the new files.

If fixed also the other requested changes.

Regards, Kiwi

tools/functional/cli/test_cli_v4.py
```diff
@@ -561,11 +561,10 @@ def test_create_project_with_values_from_file(gitlab_cli, tmpdir):
     assert description in ret.stdout
 
 
-def test_create_project_deploy_token(gitlab_cli, project):
+def do_test_create_project_deploy_token(gitlab_cli, project, scopes, expected_scopes):
```
nejch, line 564: hint: you could keep a single test function here and parametrize the input/expected output with pytest:

```python
@pytest.mark.parametrize(
    "scopes,expected_scopes",
    [
        ("read_registry", "['read_registry']"),
        ("read_registry,read_repository", "['read_repository', 'read_registry']"),
    ],
)
def test_create_project_deploy_token(gitlab_cli, project, scopes, expected_scopes):
    ...
```

Then you don't need a custom helper function at all and it will be easy to add more edge cases later. See very similar example here:
https://github.com/python-gitlab/python-gitlab/blob/af781c10db3829163f977e494e4008acf2096d64/gitlab/tests/test_config.py#L236-L249

nejch (changes requested):
Thanks for the added tests, looks good! I just left one small note.

nejch:
Hi @klorenz, coming back to this, I realized we've solved the same/similar problem in a more generic way in https://github.com/python-gitlab/python-gitlab/pull/1420 for all objects that need special list filters. 

I just ran your functional tests on the current master branch and they're passing, so I just wanted to check what your original issue was so that we can see if this is still needed for any objects. Could you let me know what CLI commands you were trying before that did not work properly?

It seems like creating deploy tokens with scopes as a comma-delimited string (`--scopes "read_registry,read_repository"` works well with the current master branch.

JohnVillalovos:
Related PR which also fixes other issues we have with array variables passed to the GitLab API

https://github.com/python-gitlab/python-gitlab/pull/1699

Issue #1107, opened 2020-05-26
Add support for group audit events


JacobHenner:
## Description of the problem, including code/CLI snippet

It does not appear as if [group audit events](https://docs.gitlab.com/ee/api/audit_events.html#group-audit-events-starter) are supported.

Presumably, these should be accessible through Group objects.

(See also: https://github.com/python-gitlab/python-gitlab/issues/948)

## Specifications

  - python-gitlab version: 2.2.0
  - API version you are using (v3/v4): v4
  - Gitlab server version (or gitlab.com): gitlab.com

Issue #948, opened 2019-11-26
Add support for audit events


idanbensha:
There's no support for the Audit Events API via the python gitlab client.

See here:
https://docs.gitlab.com/ee/api/audit_events.html

## Specifications

  - python-gitlab version: 1.9
  - API version you are using (v3/v4): v4

max-wittig:
Hi I don't have access to the enterprise version of GitLab. But if you like to have this feature, I would be happy to review any merge requests regarding this.

idanbensha:
Hello, that would be great.
I will apply a PR adding this feature ASAP!

PR #1363, opened 2021-03-07
Feat: cover all audit events

nejch:
This adds group audit events, moves all audit events into one module and makes the class naming consistent.

Closes https://github.com/python-gitlab/python-gitlab/issues/1107, closes https://github.com/python-gitlab/python-gitlab/issues/948.


max-wittig:
@nejch Sorry for ignoring that for soo long. Looks good!

PR #1365, opened 2021-03-07
chore: make _types always present in RESTManager

JohnVillalovos:
We now create _types = {} in RESTManager class.

By making _types always present in RESTManager it makes the code
simpler. We no longer have to do:
   types = getattr(self, "_types", {})

And the type checker now understands the type.


nejch (approved):
Cool, gets rid of some of the magic as well.

PR #1366, opened 2021-03-07
chore: have _create_attrs & _update_attrs be a namedtuple

JohnVillalovos:
Convert _create_attrs and _update_attrs to use a NamedTuple
(RequiredOptional) to help with code readability. Update all code to
use the NamedTuple.

### DIFF
--- a/gitlab/base.py
+++ b/gitlab/base.py
@@ -17,12 +17,13 @@
 
 import importlib
 from types import ModuleType
-from typing import Any, Dict, Optional, Tuple, Type
+from typing import Any, Dict, NamedTuple, Optional, Tuple, Type
 
 from .client import Gitlab, GitlabList
 from gitlab import types as g_types
 
 __all__ = [
+    "RequiredOptional",
     "RESTObject",
     "RESTObjectList",
     "RESTManager",
@@ -249,6 +250,11 @@ class RESTObjectList(object):
         return self._list.total
 
 
+class RequiredOptional(NamedTuple):
+    required: Tuple[str, ...] = tuple()
+    optional: Tuple[str, ...] = tuple()
+
+
 class RESTManager(object):
     """Base class for CRUD operations on objects.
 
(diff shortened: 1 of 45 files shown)


JohnVillalovos:
Breaking this up into smaller pieces.

gitlab/base.py
```diff
@@ -249,6 +250,11 @@ def total(self) -> int:
         return self._list.total
 
 
+class RequiredOptional(NamedTuple):
+    required: Tuple[str, ...]
+    optional: Tuple[str, ...]
```
nejch, line 255: Would it make sense to add default values (empty tuples) like the example in https://docs.python.org/3/library/typing.html#typing.NamedTuple?

Then people would only explicitly add required / optional args if needed. Might make the classes a bit more concise. I haven't tried though :)

JohnVillalovos: We could do that.

I set up a default value at:
https://github.com/python-gitlab/python-gitlab/pull/1366/files#diff-4d0a1fd6f7611cc6ced65bac48f9cbe658039289f350e2c5d7881674ff61dd9eR267-R272

But that doesn't help when people are setting the value.

Let me investigate it and see how it works.

JohnVillalovos: @nejch  I did what you suggested and updated all the `v4/objects/*.py` based on having default values.

JohnVillalovos:
@nejch  Updated this with your feedback.

nejch:
> @nejch Updated this with your feedback.

Thanks a lot for the hard work here, looks a lot clearer than the old code! Sorry for the delay, again :)
I didn't want to block this PR any longer so I just force-pushed a tiny change to delete a leftover tt.sh script, I hope that's ok. I'll merge in a minute.

JohnVillalovos:
> 
> 
> > @nejch Updated this with your feedback.
> 
> Thanks a lot for the hard work here, looks a lot clearer than the old code! Sorry for the delay, again :)
> I didn't want to block this PR any longer so I just force-pushed a tiny change to delete a leftover tt.sh script, I hope that's ok. I'll merge in a minute.

Oh! Whoops on my part leaving the `tt.sh` in there 😟

Thanks for the merge!

PR #1371, opened 2021-03-10
chore: add _create_attrs & _update_attrs to RESTManager

JohnVillalovos:
Add the attributes: _create_attrs and _update_attrs to the RESTManager
class. This is so that we stop using getattr() if we don't need to.

This also helps with type-hints being available for these attributes.

### DIFF
--- a/gitlab/base.py
+++ b/gitlab/base.py
@@ -17,7 +17,7 @@
 
 import importlib
 from types import ModuleType
-from typing import Any, Dict, Optional, Type
+from typing import Any, Dict, Optional, Tuple, Type
 
 from .client import Gitlab, GitlabList
 from gitlab import types as g_types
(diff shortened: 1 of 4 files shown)


nejch (approved):
Hey @JohnVillalovos sorry for the delay, thanks a lot!

I really liked the namedtuple approach in your earlier PR https://github.com/python-gitlab/python-gitlab/pull/1366, is the idea to introduce it back in a follow-up so this can be a smaller change for now?

It would take out some of the mystery in all those attributes for new contributors I think.

JohnVillalovos:
> 
> Hey @JohnVillalovos sorry for the delay, thanks a lot!

No worries! Thanks for reviewing 😀

> I really liked the namedtuple approach in your earlier PR #1366, is the idea to introduce it back in a follow-up so this can be a smaller change for now?

You read my mind! I thought let's get the small change in first and wait on the giant change.

> 
> It would take out some of the mystery in all those attributes for new contributors I think.

Agreed. It took me awhile to figure out what they were for. I think with the NamedTuple it should be clearer.

Thanks again.

JohnVillalovos:
@nejch   Once this is merged I will work on the follow-up patch to use namedtuple.

PR #1373, opened 2021-03-11
feat: add ProjectPackageFile

### LINKED ISSUE #1372: Cannot list package files

[Listing package files](https://docs.gitlab.com/ee/api/packages.html#list-package-files) appears to be unsupported. The API endpoint was introduced in GitLab 11.8.

## Expected Behavior

Listing package files should be possible.

## Actual Behavior

Listing package files is not possible.

## Specifications

  - python-gitlab version: 2.6.0

PR incoming.

JacobHenner:
Add ProjectPackageFile and the ability to list project package package_files.

Fixes #1372

### DIFF
--- a/gitlab/tests/objects/test_packages.py
+++ b/gitlab/tests/objects/test_packages.py
@@ -179,7 +179,7 @@ def test_delete_project_package(project, resp_delete_package):
     package.delete()
 
 
-def test_list_project_packages(project, resp_list_package_files):
+def test_list_project_package_files(project, resp_list_package_files):
     package = project.packages.get(1, lazy=True)
     package_files = package.package_files.list()
     assert isinstance(package_files, list)
--- a/gitlab/tests/objects/test_packages.py
+++ b/gitlab/tests/objects/test_packages.py
@@ -6,7 +6,7 @@ import re
 import pytest
 import responses
 
-from gitlab.v4.objects import GroupPackage, ProjectPackage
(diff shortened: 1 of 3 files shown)


gitlab/tests/objects/test_packages.py
```diff
@@ -117,3 +177,11 @@ def test_get_project_package(project, resp_get_package):
 def test_delete_project_package(project, resp_delete_package):
     package = project.packages.get(1, lazy=True)
     package.delete()
+
+
+def test_list_project_packages(project, resp_list_package_files):
```
nejch, line 182: Just a small one here so it doesn't clash with the test name above:
 
```suggestion
def test_list_project_package_files(project, resp_list_package_files):
```

nejch (changes requested):
Thanks for this really clean PR @JacobHenner!

Just a small note on the test name.

Not sure how this one slipped when the Packages API was initially added :see_no_evil:

gitlab/tests/objects/test_packages.py
JacobHenner: Ah, that was an oversight when I had copied my test back into `test_packages.py`. I see you added the change to the PR - is anything else required?

nejch: Sorry, forgot to merge when I was waiting for the tests ;) will merge now, thanks for the ping!

JacobHenner: Thanks!

PR #1375, opened 2021-03-14
chore: remove usage of getattr()

JohnVillalovos:
```
Remove usage of getattr(self, "_update_uses_post", False)

Instead add it to class and set default value to False.

Add a test that shows it is set to True for the
ProjectMergeRequestApprovalManager class.
```

### DIFF
--- a/gitlab/tests/objects/test_project_merge_request_approvals.py
+++ b/gitlab/tests/objects/test_project_merge_request_approvals.py
@@ -2,9 +2,12 @@
 Gitlab API: https://docs.gitlab.com/ee/api/merge_request_approvals.html
 """
 
+import copy
+
 import pytest
 import responses
-import copy
+
+import gitlab
 
 
 approval_rule_id = 1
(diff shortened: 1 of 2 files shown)


JohnVillalovos:
@nejch  If you have time to review, it would be appreciated.

gitlab/tests/objects/test_project_merge_request_approvals.py
```diff
@@ -230,6 +233,17 @@ def resp_snippet():
         yield rsps
 
 
+def test_project_approval_manager_update_uses_post(project, resp_snippet):
+    """Ensure the
+    gitlab.v4.objects.merge_request_approvals.ProjectApprovalManager object has
+    _update_uses_post set to True"""
```
nejch, line 239: I was originally thinking this would be more appropriate in a separate test/fake manager, to test just this specific aspect of it, e.g. in `test_object_mixin_attributes.py`. But since it's only used here I think it's ok, maybe for a follow-up PR :)

nejch (approved):
Thanks again! :) Left a side note for a follow-up if we get to it.

gitlab/tests/objects/test_project_merge_request_approvals.py
JohnVillalovos: I can do that, though I'm not quite sure what you want me to do 😕

Do you want a test that shows that `update_uses_post` exists and is False in `test_object_mixin_attributes.py` ?

PR #1383, opened 2021-03-23
fix(types): prevent __dir__ in RestObject from producing duplicates

spyoungtech:
Fixes a small issue with the feature added in #1072 for `__dir__` to display rest object attributes.

In at least one case, classes have overlapping attributes with the attributes listed in `.attributes`. For example, the `Group` object has its own `projects` property for the `GroupProjectManager`. This shadows the name that would normally access `.attributes['projects']`. This causes `dir(some_group)`, as currently implemented, to list `projects` twice.

While this is unlikely to be problematic, it's ideal if duplicate attributes do not appear in the response for `__dir__`. This PR ensures that duplicates are not produced by the `__dir__` method.

Instead of _adding_ the list of attributes, we use the `|` operator to produce a set with the union of all the elements, ensuring there are no duplicates.

### DIFF
--- a/gitlab/base.py
+++ b/gitlab/base.py
@@ -132,7 +132,7 @@ class RESTObject(object):
         return super(RESTObject, self) != other
 
     def __dir__(self):
-        return super(RESTObject, self).__dir__() + list(self.attributes)
+        return super(RESTObject, self).__dir__() | self.attributes.keys()
 
     def __hash__(self) -> int:
         if not self.get_id():
--- a/gitlab/tests/test_base.py
+++ b/gitlab/tests/test_base.py
@@ -135,6 +135,10 @@ class TestRESTObject:
         assert {"foo": "foo"} == obj._attrs
         assert {} == obj._updated_attrs
 
+    def test_dir_unique(self, fake_manager):
+        obj = FakeObject(fake_manager, {"manager": "foo"})
+        assert len(dir(obj)) == len(set(dir(obj)))
+
     def test_create_managers(self, fake_gitlab, fake_manager):
         class ObjectWithManager(FakeObject):
             _managers = (("fakes", "FakeManager"),)


JohnVillalovos (changes requested):
I think this needs a unit test. 

In my quick testing of this idea it doesn't seem to work.

```
>>> a = [1,2] | [3,4]
TypeError: unsupported operand type(s) for |: 'list' and 'list'
```

nejch:
@JohnVillalovos I think this would work, because dict `keys()` returns a set-like object https://docs.python.org/3/library/stdtypes.html#dictionary-view-objects.
Although true, if you like @spyoungtech, maybe a quick test against the original issue would be great :)

JohnVillalovos:
> @JohnVillalovos I think this would work, because dict `keys()` returns a set-like object https://docs.python.org/3/library/stdtypes.html#dictionary-view-objects.
> Although true, if you like @spyoungtech, maybe a quick test against the original issue would be great :)

Ah correct. I thought both were lists.

gitlab/base.py
```diff
@@ -132,7 +132,7 @@ def __ne__(self, other: object) -> bool:
         return super(RESTObject, self) != other
 
     def __dir__(self):
-        return super(RESTObject, self).__dir__() + list(self.attributes)
+        return super(RESTObject, self).__dir__() | self.attributes.keys()
```
JohnVillalovos, line 135: Can the result be converted to a list? As that is what it returned before.

```
    return list(super(RESTObject, self).__dir__() | self.attributes.keys())
```

spyoungtech: The `__dir__` method can return any sequence and will be converted into a list and sorted automatically. For example, if you return a set or any other sequence, `dir(obj)` will always be a list.

```python
class MyObj:
    def __dir__(self):
        return {'abc', 'xyz', 'def',}
obj = MyObj()
dir(obj)
['abc', 'def', 'xyz']
```

If you like we can convert it in the code, if for nothing else, to be more explicit.

JohnVillalovos: I (for selfish reasons) would like it to return a list. As I am going to be working on adding type-hints to that code in the future 😊 But not necessary.

JohnVillalovos: And not necessary, because I think I will be adding a type-hint for the return value of `Iterable[something]`, most likely.

JohnVillalovos (approved):

gitlab/base.py
spyoungtech: Sounds good to me. Feel free to apply the above suggestion (or not)

Looking forward to type hinting!

Worth mentioning that it's _usually_ not necessary to provide type annotations or type hints for certain magic methods, since their return type is implied because there's no way to implement them to return a different type.

For example, the return type for methods like `__str__`, `__repr__`, `__len__`, `__dir__`, are known to IDEs and type checkers like MyPy automatically because the returned type must be valid, otherwise a `TypeError` is raised.

spyoungtech:
Went ahead and added test case @nejch 

Confirmed it fails on the main branch, but passes with the changes this PR adds.

JohnVillalovos (approved):

nejch:
> 
> 
> Went ahead and added test case @nejch
> 
> Confirmed it fails on the main branch, but passes with the changes this PR adds.

Great, thanks @spyoungtech!

PR #1392, opened 2021-04-07
Improvements to HTTP requests

### LINKED ISSUE #1386: Request URL is getting too long because of query_parameters

I was using the following solution to iterate over the jobs:

> @duncz there is the solution I used in my project:
> 
> ```py
> scopes = ['pending', 'running', 'failed', 'success']
> jobs = git_project.jobs.list(query_parameters={'scope[]': scopes})
> ```
> 
> Should work.

_Originally posted by @romansvozil in https://github.com/python-gitlab/python-gitlab/issues/806#issuecomment-607449709_

When using it with `as_list=False` the scope is added to each `next_url` again. This results at some point to a 502 Bad Gateway error, because the URL is getting too long.

Here is a small example how I was able to reproduce it:

```py
try:
    for idx, job in enumerate(project.jobs.list(query_parameters={'scope[]': ['failed', 'success']}, as_list=False), start=1):
        if idx == 500:
            break
except Exception as exc:
    print(exc)
```

The URL is getting too long at idx == 400.

I think it started to appear after #1124 was merged. Anyhow it seems related to the brackets in the scope parameter, maybe because they get encoded in the URL.

## Expected Behavior

Iterating through all jobs.

## Actual Behavior

Getting a 502 Bad Gateway after some iterations because the URL is getting too long.

## Specifications

  - python-gitlab version: 2.6.0
  - Gitlab server version (or gitlab.com): 13.7.4

bbatliner:
Two related commits that fixed #1386 for me. Only the first commit is strictly necessary. The second is a "technically correct" fix so that `http_request` does not send extra query parameters.

### DIFF
--- a/gitlab/client.py
+++ b/gitlab/client.py
@@ -821,6 +821,9 @@ class GitlabList(object):
         self._query(url, query_data, **self._kwargs)
         self._get_next = get_next
 
+        # Remove query_parameters from kwargs, which are saved via the `next` URL
+        self._kwargs.pop("query_parameters", None)
+
     def _query(
         self, url: str, query_data: Optional[Dict[str, Any]] = None, **kwargs: Any
     ) -> None:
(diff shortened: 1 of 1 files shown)


JohnVillalovos (approved):

max-wittig:
That's much nicer. Thanks 👍

Issue #1267, opened 2021-01-15
Create MR with reviewers


mary3000:
I didn't find an option to assign reviewers while creating MR. 
As I see, gitlab REST API also [doesn't have](https://docs.gitlab.com/ee/api/merge_requests.html#create-mr) this option yet (should I also file an issue about it [here](https://gitlab.com/gitlab-org/gitlab/-/issues)?)

JoelFeiner:
@mary3000 I see `reviewer_ids` in the API. Perhaps the documentation just hadn't been updated yet?

m00lecule:
Hello! I am also struggling to find this feature in `python-gitlab 2.6.0`

mattiasb:
This should be in now since a couple of days. See [here](https://docs.gitlab.com/ee/api/merge_requests.html#update-mr)

spyoungtech:
It would also be nice to be able to change the reviewers on an existing merge request.

spyoungtech:
As a workaround for this, you can do the following:

```python
# update the reviewers on the MR
updated_data = mr.manager.update(mr.get_id(), {'reviewer_ids': [1,2,3]})

#  for completeness, update the object from the server response
if updated_data:
    mr._update_attrs(updated_data)
```

mykolasolodukha:
This might've been closed now since the #1396 is merged.

nejch:
You're right @mykolasolodukha, this does already work. Thanks for the catch!

```python
mr = project.mergerequests.create(
    {
        "source_branch": "your-branch",
        "target_branch": "main",
        "title": "your MR",
        "reviewer_ids": [reviewer_id]
    }
)

mr.reviewer_ids = [reviewer_id, another_reviewer_id]
mr.save()
```

PR #1396, opened 2021-04-14
feat(api):add support for creating/editing reviewers in project MRs

spyoungtech:
This is a first attempt at allowing users to create MRs with reviewers or edit an existing MR with the `reviewer_ids` parameter.

This is kind of unique in the sense that `reviewer_ids` doesn't appear in the rest response, but it is an accepted parameter.

A manager might be a better way to implement this, but I'm not sure it's appropriate since reviewers doesn't have its own API path, so reviewers are not really a rest object in their own right. The only example of a manager being used as a helper is the sidekiq manager.

API:

```python
mr.reviewer_ids = [1,2,3]
mr.save()
```

The property can be omitted and the same code as above works, however `mr.reviewer_ids` would be an attribute error, which may be an odd interface.

What I have here _works_ but I'm not sure it's an ideal interface. Mainly raising this to get feedback from maintainers on how this should be implemented. Feedback appreciated.

Relates to #1267

### DIFF
--- a/gitlab/v4/objects/merge_requests.py
+++ b/gitlab/v4/objects/merge_requests.py
@@ -138,6 +138,22 @@ class ProjectMergeRequest(
 ):
     _id_attr = "iid"
 
+    @property
+    def reviewer_ids(self):
+        return [reviewer["id"] for reviewer in self.reviewers]
+
+    @reviewer_ids.setter
+    def reviewer_ids(self, new_reviewer_ids):
+        new_reviewers = [{"id": id} for id in set(new_reviewer_ids)]
+        new_reviewers.extend(
+            [
+                reviewer
+                for reviewer in self.reviewers
+                if reviewer["id"] in new_reviewer_ids
+            ]
+        )
+        self.reviewers = new_reviewers
+
     _managers = (
         ("approvals", "ProjectMergeRequestApprovalManager"),
         ("approval_rules", "ProjectMergeRequestApprovalRuleManager"),
@@ -373,6 +389,7 @@ class ProjectMergeRequestManager(CRUDMixin, RESTManager):
             "remove_source_branch",
             "allow_maintainer_to_push",
             "squash",
+            "reviewer_ids",
         ),
     )
     _update_attrs = RequiredOptional(
@@ -388,6 +405,7 @@ class ProjectMergeRequestManager(CRUDMixin, RESTManager):
             "discussion_locked",
             "allow_maintainer_to_push",
             "squash",
+            "reviewer_ids",
         ),
     )
     _list_filters = (
--- a/gitlab/v4/objects/merge_requests.py
+++ b/gitlab/v4/objects/merge_requests.py
@@ -138,22 +138,6 @@ class ProjectMergeRequest(
 ):
     _id_attr = "iid"
 
-    @property
-    def reviewer_ids(self):
-        return [reviewer["id"] for reviewer in self.reviewers]
-
-    @reviewer_ids.setter
-    def reviewer_ids(self, new_reviewer_ids):
-        new_reviewers = [{"id": id} for id in set(new_reviewer_ids)]
-        new_reviewers.extend(
-            [
-                reviewer
-                for reviewer in self.reviewers
-                if reviewer["id"] in new_reviewer_ids
-            ]
-        )
-        self.reviewers = new_reviewers
-
     _managers = (
         ("approvals", "ProjectMergeRequestApprovalManager"),
         ("approval_rules", "ProjectMergeRequestApprovalRuleManager"),


gitlab/v4/objects/merge_requests.py
```diff
@@ -112,6 +112,22 @@ class ProjectMergeRequest(
...
+        new_reviewers = [{"id": id} for id in set(new_reviewer_ids)]
+        new_reviewers.extend(
+            [
+                reviewer
+                for reviewer in self.reviewers
+                if reviewer["id"] in new_reviewer_ids
+            ]
+        )
+        self.reviewers = new_reviewers
+
```
spyoungtech, line 130: All this _could_ be omitted and the functionality could still work, but that causes attribute lookup for `mr.reviewer_ids` to raise an `AttributeError`, unless it was specifically set, which might be kind of odd.

JohnVillalovos: Yeah, I don't see any other usage of `@property` in the `gitlab/v4/objects/*.py` files. So I'm not sure about that.

@nejch What do you think?

gitlab/v4/objects/merge_requests.py
```diff
@@ -373,6 +389,7 @@ class ProjectMergeRequestManager(CRUDMixin, RESTManager):
             "remove_source_branch",
             "allow_maintainer_to_push",
             "squash",
+            "reviewer_ids",
```
JohnVillalovos, line 392: As long as we are updating this I think it would be good to bring it up to date with the current API

https://docs.gitlab.com/ee/api/merge_requests.html#create-mr

These seem missing to me:

- assignee_ids | integer array | no | The ID of the user(s) to assign the MR to. Set to 0 or provide an empty value to unassign all assignees.
- reviewer_ids | integer array | no | The ID of the user(s) added as a reviewer to the MR. If set to 0 or left empty, no reviewers are added.
- remove_source_branch | boolean | no | Flag indicating if a merge request should remove the source branch when merging.

JohnVillalovos (changes requested):
Thanks a lot for this! Sorry for the delay in getting a review done.

I have left some comments.

I also rebased the patch to bring it in sync with the upstream branch.

gitlab/v4/objects/merge_requests.py
```diff
@@ -388,6 +405,7 @@ class ProjectMergeRequestManager(CRUDMixin, RESTManager):
             "discussion_locked",
             "allow_maintainer_to_push",
             "squash",
+            "reviewer_ids",
```
JohnVillalovos, line 408: This list would also be nice to be updated.

https://docs.gitlab.com/ee/api/merge_requests.html#update-mr

The list of `update` attributes differs slightly from the list of `create` attributes.

gitlab/v4/objects/merge_requests.py
spyoungtech: Thanks for taking a look at this.

Yeah, what's interesting about this particular API is that, unlike most other endpoints, the POST body includes parameters that are _not_ included in the response.

One consequence I noticed of adding a name to the list is that the framework will try to make that name an accessible attribute on the RestObject. However, because `reviewer_ids` is not present in the response, you get a (arguably unexpected) error when trying to access it. 

This is not a _huge_ problem, but was a motivation behind the `@property` getter/setter for `reviewer_ids` in an attempt to keep the Python API more consistent.

If the idea is to be able to do something like this:
```python
mr.reviewer_ids = [1,2,3]
mr.save()
```

Then `mr.reviewer_ids` should always be an accessible attribute even on a freshly retrieved object, in my opinion (without the property, this will be an attribute error because it's not in the response as such)

If this is seen as a good approach, I'd be happy to add the same for assignees (it has the same oddities, IIRC).

I'll also be happy to take care of the chore of adding `remove_source_branch` and double-checking for any other missing parameters for MRs.

gitlab/v4/objects/merge_requests.py
nejch: Sorry for the delay here people. Thanks for this, although I had some concerns originally so it took me a while to get back to this.
1. This will be a problem for assignees as well then, right? POST/PUT take `assignee_ids`, GET returns `assignees` (edit: ah I see it in the other comment)
2. If so it's also a problem for issues, not just MRs. Again POST/PUT take `assignee_ids`, GET returns `assignees`. There may be other resources we'd need to find

So I think we're getting into the territory of GitLab API quirks kind of, where we already closed the `public_builds` (POST/PUT) vs `public_jobs` (GET) PR, https://github.com/python-gitlab/python-gitlab/pull/1439. In that case we opened an upstream issue instead. We've also recently decided not to do too much custom work around just wrapping the API, see e.g. https://github.com/python-gitlab/python-gitlab/pull/1443.

I'd like to open an upstream issue first just to see if they'd be willing to make the naming consistent though I doubt it as it's quite explicit this way for them at least, and I think they also use `*_usernames` in list filters. If we decide to do it here though I'd like it to be as generic as possible (maybe via mixins or custom attributes, if it's feasible) for all resources where this might happen. Thanks again, I'll update with the upstream issue :) Any thoughts so far on this?

spyoungtech: Thanks for taking the time to review @nejch  

My thoughts about the upstream change: relying on upstream changes is probably sufficient for users of GitLab.com -- they'll eventually be fixed and the need for any other solution obviated. However, for self-hosted users who may stick to earlier versions of GitLab may still yet need an interface for the API v4 capabilities of GitLab for 12.x, 13.x, etc.

I think it would be nice to see the wrapper handle this, but I would understand and support the decision to not need to maintain that code into the future.

In the meantime, perhaps we can split up the two concerns:

1. The capability to use the _assignee_ids_ for put/post
2. Whether or not we would smooth out the API quirks (e.g. with a property, or generic wrapper of some kind)

So, maybe this PR can remove the `@property` bits for now so we can proceed with adding `assignee_ids` to the list of parameters to get the basic functionality working. Then, perhaps we can discuss smoothing the API issues in a separate MR, if we still feel it's worth pursuing.

```suggestion
```

gitlab/v4/objects/merge_requests.py
spyoungtech: It would be really nice if GitLab published an OpenAPI spec, then this could probably be handled dynamically, or at least more easily 🤔

nejch: I see you've found the upstream issue :)

gitlab/v4/objects/merge_requests.py
nejch: Yes that would make sense so we don't block this for now, thanks :) and we can discuss the rest in the issue/follow-up.

Maintaining compatibility based on gitlab versions is a whole other topic IMO as gitlab keeps breaking the API behavior in subtle ways even on minor/patch releases as we saw with 13.12, so we'd potentially have to litter our code with conditionals to cover all use cases, but I understand your point, another thing to discuss :)

nejch (approved):
Sorry, somehow I missed this!

Since we reduced the scope I think we can go ahead with this PR and I'll open a follow-up/upstream issue to see if we can align the parameters. The OpenAPI topic is already discussed here https://github.com/python-gitlab/python-gitlab/issues/1446 and we can hopefully come up with a general solution that covers everything there.

P-N-L:
Hello! I just read through the changes and discussions in this MR, and would be interested in following the upstream issue(s) with GitLab regarding aligning the parameters. @nejch apologies if I missed a link somewhere in the discussions, but could you please share the GitLab upstream issue? Appreciate it!

PR #1398, opened 2021-04-18
fix: correct some type-hints in gitlab/mixins.py

JohnVillalovos:
Commit baea7215bbbe07c06b2ca0f97a1d3d482668d887 introduced type-hints
for gitlab/mixins.py.

After starting to add type-hints to gitlab/v4/objects/users.py
discovered a few errors.

Main error was using '=' instead of ':'. For example:
  _parent = Optional[...] should be _parent: Optional[...]

Resolved those issues.


max-wittig:
Thanks for the fixes. That makes much more sense now 😄

PR #1399, opened 2021-04-18
fix: argument type was not a tuple as expected

JohnVillalovos:
While adding type-hints mypy flagged this as an issue. The third
argument to register_custom_action is supposed to be a tuple. It was
being passed as a string rather than a tuple of strings.


max-wittig:
That mistake happens too often in Python 😄 

LGTM 👍

JohnVillalovos:
> That mistake happens too often in Python 😄

One of the things I like with adding the static type-hints and then using mypy. That is how I caught this one. May find more issues as more and more type-hints are added.

PR #1400, opened 2021-04-18
chore: remove unused function sanitize_parameters()

JohnVillalovos:
The function sanitize_parameters() was used when the v3 API was in
use. Since v3 API support has been removed there are no more users of
this function.


max-wittig:
Removing code is always nice! Thanks 👍

PR #1404, opened 2021-04-22
fix: update user's bool data and avatar

DylannCordel:
Hi !

If we want to update email, avatar and do not send email confirmation change (`skip_reconfirmation` = True), `MultipartEncoder` will try to encode everything except None and bytes. So it tries to encode bools. 

Casting bool's values to their stringified int representation fix it.

Exemple to reproduce it:

```
user.skip_reconfirmation = True
user.email = user_data['email']
user.avatar = b'\xff\xd8\xff\xdb\x00C\x00\x06\x04\x0...'
user.save()
```

### DIFF
--- a/gitlab/client.py
+++ b/gitlab/client.py
@@ -505,6 +505,12 @@ class Gitlab(object):
             json = None
             if post_data is None:
                 post_data = {}
+            else:
+                # booleans does not exists for data (neither for MultipartEncoder):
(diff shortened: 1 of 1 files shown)


JohnVillalovos (approved):
Makes sense to me.

Thanks

max-wittig:
Thanks LGTM 👍

Issue #1403, opened 2021-04-22
Wrong return type in ProjectFile.decode()


Nov1kov:
## Description of the problem, including code/CLI snippet
https://python-gitlab.readthedocs.io/en/stable/api/gitlab.v4.html#gitlab.v4.objects.ProjectFile.decode

The method should return the content of a file in "str" type.

## Expected Behavior
```
>> project.files.get(ref='master', file_path='some/path').decode()
>> 'content in file'
```

## Actual Behavior
```
>> project.files.get(ref='master', file_path='some/path').decode()
>> b"content in file"
```

## Specifications

  - python-gitlab version: 2.6.0
  - API version you are using (v3/v4): v4
  - Gitlab server version (or gitlab.com): 13.6.1

JohnVillalovos:
I don't understand the bug description 😕

What is incorrect?  The documentation? The implementation? Both?

Why is it incorrect?

Nov1kov:
@JohnVillalovos Sorry. Updated the theme. I think the implementation is wrong.
The .decode () method should return content in str format instead bytes.

JohnVillalovos:
Seems like a documentation error in python-gitlab to me.

The code uses the function `base64.b64decode` to decode the content:

According to: https://docs.python.org/3/library/base64.html#base64.b64decode

The function `base64.b64decode` returns a `bytes` object.

JohnVillalovos:
Привет @Nov1kov !

I've submitted a patch to update the documentation for the function.

Now if you want to convert that `bytes` content to `str` content, I tend to use `os.fsdecode()` for that. But I'm not sure if it works for all cases.

Пока Пока,
John

Nov1kov:
Thanks!

PR #1405, opened 2021-04-23
fix: correct ProjectFile.decode() documentation

JohnVillalovos:
ProjectFile.decode() returns 'bytes' and not 'str'.

Update the method's doc-string and add a type-hint.

ProjectFile.decode() returns the result of a call to
base64.b64decode()

The docs for that function state it returns 'bytes':
https://docs.python.org/3/library/base64.html#base64.b64decode

Fixes: #1403


max-wittig:
Thanks 👍

PR #1409, opened 2021-04-24
chore: mypy: Disallow untyped definitions

JohnVillalovos:
Be more strict and don't allow untyped definitions on the files we
check.

Also this adds type-hints for two of the decorators so that now
functions/methods decorated by them will have their types be revealed
correctly.

### DIFF
--- a/gitlab/cli.py
+++ b/gitlab/cli.py
@@ -21,7 +21,7 @@ import argparse
 import functools
 import re
 import sys
-from typing import Any, Callable, Dict, Optional, Tuple, Union
+from typing import Any, Callable, cast, Dict, Optional, Tuple, TypeVar, Union
 
 import gitlab.config  # noqa: F401
 
@@ -35,14 +35,21 @@ camel_re = re.compile("(.)([A-Z])")
 custom_actions: Dict[str, Dict[str, Tuple[Tuple[str, ...], Tuple[str, ...], bool]]] = {}
 
 
+# For an explanation of how these type-hints work see:
+# https://mypy.readthedocs.io/en/stable/generics.html#declaring-decorators
+#
+# The goal here is that functions which get decorated will retain their types.
+__F = TypeVar("__F", bound=Callable[..., Any])
+
+
 def register_custom_action(
     cls_names: Union[str, Tuple[str, ...]],
     mandatory: Tuple[str, ...] = tuple(),
     optional: Tuple[str, ...] = tuple(),
-) -> Callable:
-    def wrap(f: Callable) -> Callable:
+) -> Callable[[__F], __F]:
+    def wrap(f: __F) -> __F:
         @functools.wraps(f)
-        def wrapped_f(*args, **kwargs):
+        def wrapped_f(*args: Any, **kwargs: Any) -> Any:
             return f(*args, **kwargs)
 
         # in_obj defines whether the method belongs to the obj or the manager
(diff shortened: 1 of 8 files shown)


gitlab/cli.py
```diff
@@ -241,6 +244,6 @@ def main():
     if debug:
         gl.enable_debug()
 
-    gitlab.v4.cli.run(gl, what, action, args, verbose, output, fields)
+    gitlab.v4.cli.run(gl, what, action, args_dict, verbose, output, fields)
 
-    sys.exit(0)
+    return 0
```
max-wittig, line 249: is there even a need to return anything here? This is Python, not C. Maybe we can delete the line?

max-wittig:
Just a small question

gitlab/cli.py
JohnVillalovos: Actually we should.

```
$ cat ~/.local/bin/gitlab
#!/usr/bin/python3
# -*- coding: utf-8 -*-
import re
import sys
from gitlab.cli import main
if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw|\.exe)?$', '', sys.argv[0])
    sys.exit(main())
```

So it returns the result code via the sys.exit(main())

JohnVillalovos: > is there even a need to return anything here? This is Python, not C. Maybe we can delete the line?

Probably stuff already known: As a note. All programs return an exit code. Exit code value is from 0-255 (I believe). Unix considers an exit code of 0 to be success and any other result a failure.

In Python if you return None, which is the default if a function doesn't return anything, then sys.exit(None) will return an exit code of 0.

JohnVillalovos:
> Just a small question

@max-wittig  I think I answered your question, but let me know if you need more information.

Thanks

gitlab/cli.py
nejch: Agreed we probably don't even need this line. From what I've seen elsewhere the `main` of the CLI module usually explicitly exits rather than returns (something like `sys.exit(return_code)`) but in our case the `die()` function already does a `sys.exit(1)` if there are any exceptions from the CLI so if it gets to this point it's definitely zero.

JohnVillalovos: I have removed the `return 0`

gitlab/cli.py
```diff
@@ -135,12 +142,16 @@ def _get_base_parser(add_help: bool = True) -> argparse.ArgumentParser:
...
 
-def _get_parser(cli_module):
+def _get_parser() -> argparse.ArgumentParser:
+    # NOTE: We must delay import of gitlab.v4.cli until now or
+    # otherwise it will cause circular import errors
+    import gitlab.v4.cli
+
     parser = _get_base_parser()
-    return cli_module.extend_parser(parser)
+    return gitlab.v4.cli.extend_parser(parser)
```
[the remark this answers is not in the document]
JohnVillalovos: @nejch 

I can change it back if desired. I prefer to simplify the code, and it makes it easier to understand (in my opinion). If in the future the need changes then we can change the code.

Let me know what you would like done.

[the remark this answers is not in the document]
nejch: Ok I'm fine with this, we can reintroduce it if we have a 2nd module (e.g. graphql) when we start on that :)

nejch (approved):

PR #1410, opened 2021-04-24
chore: make RESTObject._short_print_attrs always present

JohnVillalovos:
Always create RESTObject._short_print_attrs with a default value of
None.

This way we don't need to use hasattr() and we will know the type of
the attribute.


max-wittig:
Makes sense

Issue #1407, opened 2021-04-23
Trying to Fetch multiple ProjectIssues in a single batch call returns only the last ProjectIssue specified in the list


aaronalphonso:
## Description of the problem, including code/CLI snippet
According to [Gitlab](https://docs.gitlab.com/ee/api/issues.html#list-project-issues), we can make a batch GET call to fetch multiple issues by passing in the list of issue_ids in parameter `iids`. 

However, while using the library this method call always returns a list with a single element (which happens to be the last element in the `iids` list.

```python
project = gl.projects.get(project_id)
project.issues.list(all=True, iids=[1, 2, 3])   # Returns a list with a single element ProjectIssue[iid=3]
project.issues.list(all=True, iids=[3, 2, 1])   # Returns a list with a single element ProjectIssue[iid=1]

```

Are batch requests not supported yet or is this a bug?  

## Expected Behavior
The batch call should return a list of all ProjectIssues as mentioned in the `iids` field.
```python
project = gl.projects.get(project_id)
project.issues.list(all=True, iids=[1, 2, 3])   # Should return [ProjectIssue(iid=1), ProjectIssue(iid=2), ProjectIssue(iid=3)]
```

## Actual Behavior
Only the ProjectIssue corresponding to the last id in the `iids` list is returned 
```python
project = gl.projects.get(project_id)
project.issues.list(all=True, iids=[1, 2, 3])    # Returns a list with a single element ProjectIssue[iid=3]
```

## Specifications

  - python-gitlab version: 2.6.0
  - API version you are using (v3/v4): v4
  - Gitlab server version (or gitlab.com): GitLab Community Edition 13.10.2

nejch:
This seems related to some other issues where lists aren't passed to query params correctly, I'll check. I think we have a similar issue with scopes.

Out of curiosity, have you tried passing the `iids` as a comma-delimited string? Would need to check what the current state of the gitlab API is with this (see e.g. https://gitlab.com/gitlab-org/gitlab-foss/-/issues/48007)

aaronalphonso:
I just tried passing in `iids` as a comma-delimted string and it gives me the expected results. Thank you very much for the suggestion! :)

JohnVillalovos:
I have proposed a patch for this issue in https://github.com/python-gitlab/python-gitlab/pull/1413

I'm hopeful that it will work though I have not tested it.

JohnVillalovos:
[...]

PR #1415, opened 2021-04-25
feat: add support for lists of integers to ListAttribute

JohnVillalovos:
Previously ListAttribute only support lists of integers. Now be more
flexible and support lists of items which can be coerced into strings,
for example integers.

This will help us fix issue #1407 by using ListAttribute for the
'iids' field.


nejch (approved):
I think it's a good idea to merge this first :) later when it comes to individual endpoints it might be good to add int. tests for those to see how gitlab behaves.

Issue #1416, opened 2021-04-26
v2.7.0 breaks when trying to get pathed files


xgui3783:
## Description of the problem, including code/CLI snippet

when trying to 

```python
from gitlab import Gitlab
project=Gitlab(GITLAB_SRC).projects.get(GITLAB_PROJECT_ID)
project.files.get(file_path=PATH_TO_FILE, ref=GITLAB_REF)
```

where `PATH_TO_FILE` is in the form of `{directory}/{filename}`, client returns not found error. (see github action: https://github.com/xgui3783/python-gitlab-exmaple/actions/runs/786113706)

I suspect this is due to double escaping the `/` character from:
-  `GetMixin.get` (https://github.com/python-gitlab/python-gitlab/blob/8933113/gitlab/mixins.py#L104) and 
- `ProjectFileManager.get` (https://github.com/python-gitlab/python-gitlab/blob/89331131b3337308bacb0c4013e80a4809f3952c/gitlab/v4/objects/files.py#L97)

## Expected Behavior

return file descriptor

## Actual Behavior

not found error

## Specifications

  - python-gitlab version: v2.7.0
  - API version you are using (v3/v4): v4
  - Gitlab server version (or gitlab.com): gitlab.com

JohnVillalovos:
This might be related to commit b4dac5ce33843cf52badeb9faf0f7f52f20a9a6a done by @em-

JohnVillalovos:
I do notice that this will happen:

```
path/to/file
path%2fto%2ffile
path%252fto%252ffile
```

With this code:
```
    a = "path/to/file"
    print(a)
    a = a.replace("/", "%2f")
    print(a)
    a = urllib.parse.quote(a, safe="")
    print(a)
```

nejch:
Yes, I think the `get()` override for ProjectFileManager (or at least the extra replace) is unnecessary now that the ID is already url-encoded. @JohnVillalovos are you already working on this? Otherwise I'll be pushing a fix tonight.

nejch:
Related: https://github.com/python-gitlab/python-gitlab/issues/1356

JohnVillalovos:
> @JohnVillalovos are you already working on this? Otherwise I'll be pushing a fix tonight.

@nejch  No. Sorry I am doing my day-job at the moment. So I can't work on it 😢

xgui3783:
Thanks for the quick fix! 👏👏👏

Issue #1356, opened 2021-03-02
Get rid of all custom URL encoding


nejch:
See also https://github.com/python-gitlab/python-gitlab/pull/1336, https://github.com/python-gitlab/python-gitlab/pull/1080, https://github.com/python-gitlab/python-gitlab/issues/1079.

On the other hand,  need to check https://github.com/python-gitlab/python-gitlab/issues/1006 and the current state of https://github.com/python-gitlab/python-gitlab/blob/96d2805b5bf372cb79c2b7db5c1e499c41e477c1/gitlab/client.py#L513-L524
If proxies handle it differently maybe an additional arg for e.g. `encode_xyz=True` might help.

---

```
gitlab/utils.py:    return id.replace("/", "%2F").replace("#", "%23")
gitlab/utils.py:        return value.replace("/", "%2F")
gitlab/utils.py:    new_path = parsed.path.replace(".", "%2E")
gitlab/v4/objects/branches.py:        id = self.get_id().replace("/", "%2F")
gitlab/v4/objects/branches.py:        id = self.get_id().replace("/", "%2F")
gitlab/v4/objects/features.py:        path = "%s/%s" % (self.path, name.replace("/", "%2F"))
gitlab/v4/objects/files.py:        self.file_path = self.file_path.replace("/", "%2F")
gitlab/v4/objects/files.py:        file_path = self.get_id().replace("/", "%2F")
gitlab/v4/objects/files.py:        file_path = file_path.replace("/", "%2F")
gitlab/v4/objects/files.py:        file_path = new_data.pop("file_path").replace("/", "%2F")
gitlab/v4/objects/files.py:        file_path = file_path.replace("/", "%2F")
gitlab/v4/objects/files.py:        path = "%s/%s" % (self.path, file_path.replace("/", "%2F"))
gitlab/v4/objects/files.py:        file_path = file_path.replace("/", "%2F").replace(".", "%2E")
gitlab/v4/objects/files.py:        file_path = file_path.replace("/", "%2F").replace(".", "%2E")
gitlab/v4/objects/repositories.py:        submodule = submodule.replace("/", "%2F")  # .replace('.', '%2E')
gitlab/v4/objects/tags.py:        id = self.get_id().replace("/", "%2F")
```

oux:
Hi, I made a pullrequest to fix the lazy object deletion (fix #1494): https://github.com/python-gitlab/python-gitlab/pull/1541
Let me know if you need any action from me.

PR #1418, opened 2021-04-26
fix(files): do not url-encode filepaths twice

nejch:
Closes https://github.com/python-gitlab/python-gitlab/issues/1416.

This is a quick fix for 2.7.0 but we should clean up with https://github.com/python-gitlab/python-gitlab/issues/1356 eventually :)


max-wittig:
@nejch Thanks for the fix! I will push out a release right after the pipeline passes 👍

PR #1421, opened 2021-04-28
feat: indicate that we are a typed package

JohnVillalovos:
By adding the file: py.typed
it indicates that python-gitlab is a typed package and contains
type-hints.

https://www.python.org/dev/peps/pep-0561/

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,6 +27,7 @@ setup(
     url="https://github.com/python-gitlab/python-gitlab",
     packages=find_packages(),
     install_requires=["requests>=2.22.0", "requests-toolbelt>=0.9.1"],
+    package_data = {'gitlab': ['py.typed'], },
     python_requires=">=3.6.0",
     entry_points={"console_scripts": ["gitlab = gitlab.cli:main"]},
     classifiers=[


nejch (commented):
question: does this mean we should probably do in `.mypy.ini`:

```ini
[mypy]
files = gitlab/**/*.py
```

So that we check this recursively with mypy as `py.typed` applies recursively, just in case.

Currently, this fails because of unit tests, but I've long thought we should unit and functional tests together, outside the source dir. (e.g. `tools/functional` -> `tests/functional`, `gitlab/tests` -> `tests/unit`)

JohnVillalovos:
> question: does this mean we should probably do in `.mypy.ini`:
> 
> ```ini
> [mypy]
> files = gitlab/**/*.py
> ```
> 
> So that we check this recursively with mypy as `py.typed` applies recursively, just in case.
> 
> Currently, this fails because of unit tests, but I've long thought we should unit and functional tests together, outside the source dir. (e.g. `tools/functional` -> `tests/functional`, `gitlab/tests` -> `tests/unit`)

I would like to wait on saying we mypy check everything. As I don't think we are ready.

This PR in my mind is just saying we do have type-hints so someone can use them. Also `mypy` will see that file and then not complain that there aren't type-hints.

Do we have type-hints for everything? No, but progress is being made and the type-hints we have are useful already and can be used by users of this library to help type-hint their code.

nejch:
Ok makes sense :) Thanks!

Issue #1425, opened 2021-04-28
AttributeError: 'NoneType' object has no attribute 'replace' when doing project.mergerequests.get


nicoleepp:
## Description of the problem, including code/CLI snippet

```
Traceback (most recent call last):
--
File "/var/task/tagbotgitlab/tagbot.py", line 31, in handler
status, msg = 200, handle_event(json.loads(evt.get("body", "{}")))
File "/var/task/tagbotgitlab/tagbot.py", line 53, in handle_event
return handle_open(payload)
File "/var/task/tagbotgitlab/tagbot.py", line 68, in handle_open
mr = p.mergerequests.get(mr_id, lazy=True)
File "/var/task/gitlab/exceptions.py", line 279, in wrapped_f
return f(*args, **kwargs)
File "/var/task/gitlab/mixins.py", line 46, in get
id = utils.clean_str_id(id)
File "/var/task/gitlab/utils.py", line 51, in clean_str_id
return id.replace("/", "%2F").replace("#", "%23")
AttributeError: 'NoneType' object has no attribute 'replace'
STATUS : 500
ERROR : Runtime error
```

## Expected Behavior
p.mergerequests.get(mr_id, lazy=True) should return an MR instead of throwing an error

## Actual Behavior
AttributeError: 'NoneType' object has no attribute 'replace'

## Specifications

  - python-gitlab version: 2.5.0
  - API version you are using (v3/v4): v4 
  - Gitlab server version (or gitlab.com): 13.11.1-ee

JohnVillalovos:
As a note the current version is 2.7.1.  Does the issue occur with that version too?

JohnVillalovos:
I created a Pull Request adding a functional test that shows in the current version that it is working.

https://github.com/python-gitlab/python-gitlab/pull/1426

nicoleepp:
Thanks, after updating to the newest version this works

PR #1426, opened 2021-04-28
chore: add functional test of mergerequest.get() 

JohnVillalovos:
Add a functional test of test mergerequest.get() and
mergerequest.get(..., lazy=True)

Closes: #1425

### DIFF
--- a/tools/functional/api/test_merge_requests.py
+++ b/tools/functional/api/test_merge_requests.py
@@ -29,6 +29,18 @@ def test_merge_requests(project):
     )
 
 
+def test_merge_requests_get(project):
+    """Test variations of mergerequests.get()"""
+    mr_iid = 1
+    mr = project.mergerequests.get(mr_iid)
+    assert mr.iid == mr_iid
+    mr = project.mergerequests.get("1")
+    assert mr.iid == mr_iid
+    # Make sure a 'lazy' get works
+    mr = project.mergerequests.get(mr_iid, lazy=True)
+    assert mr.iid == mr_iid
+
+
 def test_merge_request_discussion(project):
     mr = project.mergerequests.list()[0]
     size = len(mr.discussions.list())


tools/functional/api/test_merge_requests.py
```diff
@@ -29,6 +29,16 @@ def test_merge_requests(project):
     )
 
 
+def test_merge_request_get(project):
```
max-wittig, line 32: What special case is tested here? Maybe we can get rid of that comment by having a more elaborate test name?

```suggestion
def test_merge_request_get_lazy_returns_mr_iid(project):
```

tools/functional/api/test_merge_requests.py
```diff
@@ -29,6 +29,16 @@ def test_merge_requests(project):
...
+    mr = project.mergerequests.list()[0]
+    mr_iid = mr.iid
+    mr = project.mergerequests.get(mr_iid)
+    assert mr.iid == mr_iid
```
max-wittig, line 36: that's not really part of this test. That should be tested somewhere else already. I think we can safely get rid of that first assert

JohnVillalovos: In the functional tests I don't see any tests for mergerequests.get(), except in the EE part there happens to be a get.

tools/functional/api/test_merge_requests.py
JohnVillalovos: > What special case is tested here?

No special case, just testing that mergerequest.get() works for both a standard get() and get(..., lazy=True).

I would like to put them both in the same function as it takes less time as no need to setup/tear-down with a second function.

Does that answer your question?

Thanks.

tools/functional/api/test_merge_requests.py
max-wittig: Mhmm okay, maybe then we keep it in a separate test.

tools/functional/api/test_merge_requests.py
max-wittig: The time doesn't really matter. Code quality matters a bit more imo.

tools/functional/api/test_merge_requests.py
JohnVillalovos: > Mhmm okay, maybe then we keep it in a separate test.

Sorry, I don't quite understand what you would like me to do. Happy to change things but I'm a bit confused.

tools/functional/api/test_merge_requests.py
nejch: to move this along, @JohnVillalovos since this PR was originally opened you have added some more specific tests here https://github.com/python-gitlab/python-gitlab/blob/master/tests/functional/api/test_merge_requests.py#L103, maybe the name could be similar in this one to reflect it's a regression test for https://github.com/python-gitlab/python-gitlab/issues/1425.

tools/functional/api/test_merge_requests.py
nejch: Could this make use of the new merge request fixture that you added @JohnVillalovos, that might clean it up a bit? https://github.com/python-gitlab/python-gitlab/blob/master/tests/functional/conftest.py#L232

PR #1427, opened 2021-04-28
chore(ci): automate releases

nejch:
Adds an automated release ~~every week~~ once a month + manual workflow if quick fixes are needed. I thought this would be a good compromise. I didn't want this on every push as otherwise all release notes just have only 1 PR entry and it's just really noisy. I see other people are also reconsidering that: https://github.com/renovatebot/renovate/issues/9709.

Also I just chose Monday so there's no Friday releases :rofl: but open to suggestions.

I tested this here (including all my derpy trial&error pushes):
https://github.com/nejch/python-gitlab-semantic-release

Resulting releases: https://github.com/nejch/python-gitlab-semantic-release/releases
testpypi package: https://test.pypi.org/project/python-gitlab/
changelog: https://github.com/nejch/python-gitlab-semantic-release/blob/master/CHANGELOG.md

I've also already set up the `PYPI_TOKEN` secret here.

Initially I wasn't a fan of the auto-committed `CHANGELOG.md` but now I'm thinking it gives us a chance to consolidate the old changelog and release notes that are just sitting there since the GH Releases move. I can push another PR for that to add the old changes it to the new changelog.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -213,3 +213,27 @@ To cleanup the environment delete the container:
    docker rm -f gitlab-test
    docker rm -f gitlab-runner-test
 
+Releases
+--------
+
+A release is automatically published once a month on the 28th if any commits merged
+to the main branch contain commit message types that signal a semantic version bump
+(``fix``, ``feat``, ``BREAKING CHANGE:``).
+
+Additionally, the release workflow can be run manually by maintainers to publish urgent
+fixes, either on GitHub or using the ``gh`` CLI with ``gh workflow run release.yml``.
+
+**Note:** As a maintainer, this means you should carefully review commit messages
+used by contributors in their pull requests. If scopes such as ``fix`` and ``feat``
+are applied to trivial commits not relevant to end users, it's best to squash their
+pull requests and summarize the addition in a single conventional commit.
+This avoids triggering incorrect version bumps and releases without functional changes.
+
+The release workflow uses `python-semantic-release
+<https://python-semantic-release.readthedocs.io>`_ and does the following:
+
+* Bumps the version in ``__version__.py`` and adds an entry in ``CHANGELOG.md``,
+* Commits and tags the changes, then pushes to the main branch as the ``github-actions`` user,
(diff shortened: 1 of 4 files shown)


README.rst
```diff
@@ -213,3 +213,27 @@ To cleanup the environment delete the container:
    docker rm -f gitlab-test
    docker rm -f gitlab-runner-test
 
+Releases
+--------
+
+A release workflow is triggered every Monday and automatically creates a new
+release if any commits merged to the main branch contain commit message scopes
+that would result in a semantic version bump.
```
JohnVillalovos, line 221: What are the "commit message scopes" which trigger a new release?  Is it only ``fix`` and ``feat``?

nejch: Oops thanks, that doesn't make any sense, it's not really scopes but types. Maybe I can link to https://github.com/semantic-release/semantic-release#how-does-it-work here, it should be:
fix -> patch bump
feat -> minor bump

Then any kind of type that includes `BREAKING CHANGE:` in the commit message -> major.

I'll reword this.

max-wittig:
Often it's not really required to release every week. Gitlab releases every month. Maybe that makes more sense? To release shortly after Gitlab's monthly release?

nejch:
> Often it's not really required to release every week. Gitlab releases every month. Maybe that makes more sense? To release shortly after Gitlab's monthly release?

Ok, makes sense :+1: Originally I was thinking so that people don't have to wait after things get merged as they are sometimes asking in PRs, but it might get too spammy for others like that. I'll change the cron for the end of the month.

README.rst
nejch: Done, added types so I hope this is clearer now.

nejch:
I've reworded this and made the cron schedule to be on the 28th of the month, that should give us a few days after the 22nd to add some popular features after a GitLab release.

JohnVillalovos (approved):
Thanks

max-wittig (approved):

PR #1429, opened 2021-04-29
chore: have flake8 check the entire project

JohnVillalovos:
Have flake8 run at the top-level of the projects instead of just the
gitlab directory.

### DIFF
--- a/tools/functional/api/test_issues.py
+++ b/tools/functional/api/test_issues.py
@@ -39,7 +39,7 @@ def test_issue_notes(issue):
 
 
 def test_issue_labels(project, issue):
-    label = project.labels.create({"name": "label2", "color": "#aabbcc"})
+    project.labels.create({"name": "label2", "color": "#aabbcc"})
     issue.labels = ["label2"]
     issue.save()
 
--- a/tools/functional/api/test_merge_requests.py
+++ b/tools/functional/api/test_merge_requests.py
@@ -14,7 +14,7 @@ def test_merge_requests(project):
     )
 
     source_branch = "branch1"
-    branch = project.branches.create({"branch": source_branch, "ref": "master"})
+    project.branches.create({"branch": source_branch, "ref": "master"})
 
     project.files.create(
         {
@@ -24,7 +24,7 @@ def test_merge_requests(project):
             "commit_message": "New commit in new branch",
         }
     )
-    mr = project.mergerequests.create(
+    project.mergerequests.create(
         {"source_branch": "branch1", "target_branch": "master", "title": "MR readme2"}
     )
 
--- a/tools/functional/api/test_releases.py
+++ b/tools/functional/api/test_releases.py
@@ -29,7 +29,7 @@ def test_delete_project_release(project, release):
 
 
 def test_create_project_release_links(project, release):
-    link = release.links.create(link_data)
+    release.links.create(link_data)
 
     release = project.releases.get(release.tag_name)
     assert release.assets["links"][0]["url"] == link_data["url"]
--- a/tools/functional/api/test_repository.py
+++ b/tools/functional/api/test_repository.py
@@ -74,7 +74,7 @@ def test_create_commit(project):
 def test_create_commit_status(project):
     commit = project.commits.list()[0]
     size = len(commit.statuses.list())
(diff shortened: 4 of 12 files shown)


tools/functional/api/test_issues.py
```diff
@@ -39,7 +39,7 @@ def test_issue_notes(issue):
 
 
 def test_issue_labels(project, issue):
-    label = project.labels.create({"name": "label2", "color": "#aabbcc"})
+    _ = project.labels.create({"name": "label2", "color": "#aabbcc"})
```
max-wittig, line 42:
```suggestion
project.labels.create({"name": "label2", "color": "#aabbcc"})
```

tools/functional/api/test_merge_requests.py
```diff
@@ -14,7 +14,7 @@ def test_merge_requests(project):
     )
 
     source_branch = "branch1"
-    branch = project.branches.create({"branch": source_branch, "ref": "master"})
+    _ = project.branches.create({"branch": source_branch, "ref": "master"})
```
max-wittig, line 17:
```suggestion
    project.branches.create({"branch": source_branch, "ref": "master"})
```

tools/functional/api/test_releases.py
```diff
@@ -29,7 +29,7 @@ def test_delete_project_release(project, release):
 
 
 def test_create_project_release_links(project, release):
-    link = release.links.create(link_data)
+    _ = release.links.create(link_data)
```
max-wittig, line 32:
```suggestion
    release.links.create(link_data)
```

tools/functional/api/test_repository.py
```diff
@@ -74,15 +74,15 @@ def test_create_commit(project):
 def test_create_commit_status(project):
     commit = project.commits.list()[0]
     size = len(commit.statuses.list())
-    status = commit.statuses.create({"state": "success", "sha": commit.id})
+    _ = commit.statuses.create({"state": "success", "sha": commit.id})
```
max-wittig, line 77:
```suggestion
    commit.statuses.create({"state": "success", "sha": commit.id})
```

tools/functional/api/test_repository.py
```diff
@@ -74,15 +74,15 @@ def test_create_commit(project):
...
     assert len(commit.statuses.list()) == size + 1
 
 
 def test_commit_signature(project):
     commit = project.commits.list()[0]
 
     with pytest.raises(gitlab.GitlabGetError) as e:
-        signature = commit.signature()
+        _ = commit.signature()
```
max-wittig, line 85:
```suggestion
        commit.signature()
```

max-wittig (changes requested):
Nice catch!

Do we need to assign variables that we're not using to `_`?

JohnVillalovos:
> 
> 
> Nice catch!
> 
> Do we need to assign variables that we're not using to `_`?

No. The only reason I had for doing that was to make it clear that the call returns something.

I was undecided if I should just delete the assignment or use `_`. So I went with `_` 😀

But I will re-do it and delete the assignments.

max-wittig:
If we want to check, if it's returning something we should assert instead. But I think this is already checked somewhere else.

You can also apply the suggestions and I squash after, if you like 😃

tools/functional/api/test_issues.py
JohnVillalovos: Done

tools/functional/api/test_releases.py
JohnVillalovos: Done

tools/functional/api/test_repository.py
JohnVillalovos: Done

tools/functional/api/test_repository.py
JohnVillalovos: Done

tools/functional/api/test_merge_requests.py
JohnVillalovos: Done

JohnVillalovos:
@max-wittig  Updated. I think I have changed everything as requested.

max-wittig (approved):

Issue #1431, opened 2021-04-30
Can't manage GitLab Group attributes : membership_lock or shared_runners_setting


Stan0304:
## Description of the problem, including code/CLI snippet

I tried to update a GitLab group and I can't update or even read the following attributes even if those settings exists on the API side (https://docs.gitlab.com/ee/api/groups.html#update-group):

- membership_lock
- shared_runners_setting

Bellow the code to reproduce the error.

```python
#!/usr/bin/env python3
import gitlab
import os

gl = gitlab.Gitlab('https://gitlab.com', private_token=os.environ['GITLAB_TOKEN'])
gl.auth()

# Load a gitlab group
group_id = 123456 # TODO use an existing group_id here
gitlab_group = gl.groups.get(group_id)

# show the request_access_enabled -> that call is OK
print('request_access_enabled', gitlab_group.request_access_enabled)

# show the membership_lock -> that call is NOT OK
print('membership_lock', gitlab_group.membership_lock)

# show the shared_runners_setting -> that call is NOT OK
print('shared_runners_setting', gitlab_group.shared_runners_setting)

```

The `membership_lock` attribute seems to be implemented, but the `shared_runners_setting` does not appear at all.

- https://github.com/python-gitlab/python-gitlab/blob/master/gitlab/v4/objects/groups.py#L231

## Expected Behavior

Having the ability to read and/or update the `membership_lock` and  `shared_runners_setting` attributes of GitLab groups.

## Actual Behavior

Actually when I try to read the  `membership_lock` or  `shared_runners_setting`, I get a `AttributeError`

```
request_access_enabled False # <- the first call that works
Traceback (most recent call last):
  File "/home/speyssard/.local/lib/python3.6/site-packages/gitlab/base.py", line 77, in __getattr__
    return self.__dict__["_updated_attrs"][name]
KeyError: 'membership_lock'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/speyssard/.local/lib/python3.6/site-packages/gitlab/base.py", line 80, in __getattr__
    value = self.__dict__["_attrs"][name]
KeyError: 'membership_lock'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/speyssard/.local/lib/python3.6/site-packages/gitlab/base.py", line 100, in __getattr__
    return self.__dict__["_parent_attrs"][name]
KeyError: 'membership_lock'

During handling of the above exception, another exception occurred:

[...]

PR #1438, opened 2021-05-01
fix(objects): add missing group attributes

nejch:
Closes https://github.com/python-gitlab/python-gitlab/issues/1431.


JohnVillalovos (approved):
Thanks. LGTM.

PR #1440, opened 2021-05-01
test(functional): start tracking functional test coverage

nejch:
Trying out codecov flags to see how much the whole suite covers.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -55,6 +55,12 @@ jobs:
         env:
           TOXENV: ${{ matrix.toxenv }}
         run: tox
+      - name: Upload codecov coverage
+        uses: codecov/codecov-action@v1
+        with:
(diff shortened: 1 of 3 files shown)


nejch:
WDYT? I wanted to get a whole overview of what code covered by tests, including functional. Using flags you can really nicely split the suites in codecov, see https://codecov.io/gh/python-gitlab/python-gitlab/src/f875786ce338b329421f772b181e7183f0fcb333/gitlab/v4/cli.py

I think in docker tests the real bottleneck is the docker container, so I thought it wouldn't hurt adding XML coverage to the runs by default. Most people run unit tests locally I assume.

I also tweaked the codecov commenter a bit as it's not very helpful getting spammed with the comment if there's no change in coverage.

JohnVillalovos (approved):
Looks good to me. I like it.

Thanks.

Issue #1370, opened 2021-03-08
2.6.0: man pages generation warnings


kloczek:
```
+ sphinx-build -b man -d gitlab docs .
Running Sphinx v3.4.3
building [mo]: targets for 0 po files that are out of date
building [man]: all manpages
updating environment: [new config] 59 added, 0 changed, 0 removed
reading sources... [100%] switching-to-v4
/home/tkloczko/rpmbuild/BUILD/python-gitlab-2.6.0/gitlab/v4/objects/__init__.py:docstring of gitlab.v4.objects.ProjectManager.import_github:30: WARNING: Unexpected indentation.
/home/tkloczko/rpmbuild/BUILD/python-gitlab-2.6.0/gitlab/v4/objects/__init__.py:docstring of gitlab.v4.objects.ProjectManager.import_github:28: WARNING: Inline literal start-string without end-string.
/home/tkloczko/rpmbuild/BUILD/python-gitlab-2.6.0/gitlab/v4/objects/__init__.py:docstring of gitlab.v4.objects.ProjectManager.import_github:28: WARNING: Inline interpreted text or phrase reference start-string without end-string.
/home/tkloczko/rpmbuild/BUILD/python-gitlab-2.6.0/gitlab/v4/objects/__init__.py:docstring of gitlab.v4.objects.ProjectManager.import_github:33: WARNING: Unexpected indentation.
/home/tkloczko/rpmbuild/BUILD/python-gitlab-2.6.0/gitlab/v4/objects/__init__.py:docstring of gitlab.v4.objects.ProjectManager.import_github:35: WARNING: Block quote ends without a blank line; unexpected unindent.
/home/tkloczko/rpmbuild/BUILD/python-gitlab-2.6.0/gitlab/v4/objects/__init__.py:docstring of gitlab.v4.objects.ProjectManager.import_github:38: WARNING: Unexpected indentation.
/home/tkloczko/rpmbuild/BUILD/python-gitlab-2.6.0/gitlab/v4/objects/__init__.py:docstring of gitlab.v4.objects.ProjectManager.import_github:40: WARNING: Block quote ends without a blank line; unexpected unindent.
/home/tkloczko/rpmbuild/BUILD/python-gitlab-2.6.0/gitlab/v4/objects/__init__.py:docstring of gitlab.v4.objects.ProjectManager.import_github:41: WARNING: Block quote ends without a blank line; unexpected unindent.
[...]

PR #1441, opened 2021-05-01
docs: fail on warnings during sphinx build

nejch:
I was hesitant to make it this strict first, but realized this is useful when docs aren't included in the toctree and the articles don't show up on RTD, unless users specifically search for those keywords. So we actually had some dead documentation this whole time.

Closes https://github.com/python-gitlab/python-gitlab/issues/1370.


JohnVillalovos (approved):
I like being strict.

LGTM and Thanks!

Issue #1395, opened 2021-04-12
Objects lose their custom_attributes during a save()


deveaud-m:
## Description of the problem, including code/CLI snippet

When saving an object (e.g. User) with `custom_attributes`, those attributes get lost an the object needs to be synced again to recover all needed attributes:

```python
import gitlab

gl = gitlab.Gitlab(...)
bob = gl.users.get(42, with_custom_attributes=True)
print(bob.custom_attributes)
bob.save()
print(bob.custom_attributes)
```

This raises `AttributeError`:

```
Traceback (most recent call last):
  File ".../lib/python3.9/site-packages/gitlab/base.py", line 59, in __getattr__
    return self.__dict__["_updated_attrs"][name]
KeyError: 'custom_attributes'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File ".../lib/python3.9/site-packages/gitlab/base.py", line 62, in __getattr__
    value = self.__dict__["_attrs"][name]
KeyError: 'custom_attributes'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File ".../lib/python3.9/site-packages/gitlab/base.py", line 82, in __getattr__
    return self.__dict__["_parent_attrs"][name]
KeyError: 'custom_attributes'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File ".../attr_test.py", line 8, in <module>
    print(f"Bob after {json.dumps(bob.custom_attributes, indent=4)}")
  File ".../lib/python3.9/site-packages/gitlab/base.py", line 84, in __getattr__
    raise AttributeError(name)
AttributeError: custom_attributes
```

## Expected Behavior

Restore objects' `custom_attributes` after calling `save()`.
This can be reproduced with Projects and Groups.

## Actual Behavior

The objects must be synced again with the option `with_custom_attributes=True` to access `custom_attributes`.

## Specifications

  - python-gitlab version: 2.6.0
  - API version you are using (v3/v4): v4
  - Gitlab server version (or gitlab.com): 13.10.1

nejch:
Thanks @deveaud-m! Note to self:
- for `save()/update()` methods the new server data alone shouldn't be used as it will mostly be just a subset of what's returned via GET.
- for `refresh()`, the arguments used for the original `get()` should maybe be stored in the object and reused for subsequent GET requests or there will be missing attributes again.

[...]

PR #1443, opened 2021-05-02
fix(base): allow persisting local attributes when updating object

nejch:
Related to https://github.com/python-gitlab/python-gitlab/issues/1395.


max-wittig:
I'm not sure, if this makes sense. `python-gitlab` should be a light wrapper around the GitLab API. If we start adding more custom things (other than the required ones e.g. `lazy`) that might become less maintainable and maybe also confusing for the users? WDYT?

nejch:
Closing this in favor of https://github.com/python-gitlab/python-gitlab/pull/1448 as discussed.

Issue #825, opened 2019-07-12
Job retry method should return created job


PavelGloba:
## Description of the problem, including code/CLI snippet

There should be a way to get an id of a freshly created job, when the retry method is used, like it's done for creating a new pipeline.

```
>>> job = project.jobs.get('7659')
>>> q = job.retry()
>>> print(q)
None
```

## Expected Behavior
Job retrying returns RESTObject

## Actual Behavior
Job retrying returns nothing

## Specifications

  - python-gitlab version: 1.9.0
  - API version you are using (v3/v4): v4
  - Gitlab server version (or gitlab.com): 11.8.2

ahmet2mir:
Same here

In the code you have

```python
@cli.register_custom_action("ProjectJob")
@exc.on_http_error(exc.GitlabJobRetryError)
def retry(self, **kwargs):
    """Retry the job.
    Args:
        **kwargs: Extra options to send to the server (e.g. sudo)
    Raises:
        GitlabAuthenticationError: If authentication is not correct
        GitlabJobRetryError: If the job could not be retried
    """
    path = "%s/%s/retry" % (self.manager.path, self.get_id())
    self.manager.gitlab.http_post(path)
```

A simple `return` before `self.manager` and it's fine.

And it's the same for all thoses resources.
No idea if it's an issue or a feature.

But to fix this I use directly the manager

```python
>>> project = client.projects.get(project_id)
>>> job = project.jobs.get(job_id)

>>> output = job.retry()
>>> output is None
True

>>> output = job.manager.gitlab.http_post("%s/%s/retry" % (job.manager.path, job.get_id()))
>>> output is None
False
>>> output['id']
1234
```

PR #1444, opened 2021-05-02
fix(objects): return server data in cancel/retry methods

nejch:
Closes https://github.com/python-gitlab/python-gitlab/issues/825. It seems like it's something people want and saves a few extra API calls in some cases.


JohnVillalovos (approved):
LGTM. Thanks.

Issue #682, opened 2019-01-14
[feature] Implement a `count()` method for listable objects


gpocentek:
This is a followup on #670 

GitLab returns the number of available objects in the response headers. Today if a user just wants to count the objects they still have to use the `list()` method and retrieve some data. We could provide a `count()` method that would return the number of available objects using a HEAD method.

To be discussed: should we also return the pagination information (as we do when `as_list` is false)? Does this need another method?

See https://github.com/python-gitlab/python-gitlab/tree/feat/count for the simple count implementation

max-wittig:
It does make sense. I've also used a workaround before to see the available X-Page-Items or how it's called in GitLab.

matthew-a-dunlap:
@gpocentek Thanks for posting this, I found it useful for my fork adding the option to get the repository file metadata without the actual file contents https://github.com/python-gitlab/python-gitlab/compare/master...matthew-a-dunlap:blame-head

Issue #1447, opened 2021-05-06
No possibility to check if file exists


valentingregoire:
## Description of the problem, including code/CLI snippet
When wanting to check if a file exists, you either have to get the entire file, the raw file content or the blame history. None of these seem to fit for the purpose to see if a file exists or not.

## Expected Behavior
I am able to check if a file exists by running the `HEAD` method described in the [official GitLab Rest API documentation](https://docs.gitlab.com/ee/api/repository_files.html#get-file-from-repository).

## Actual Behavior
In order not to get the full content of the file, I kind of cheat by fetching the blame history.

## Specifications

  - python-gitlab version: 2.5.0
  - API version you are using (v3/v4): v4
  - Gitlab server version (or gitlab.com): newset

nejch:
Related to https://github.com/python-gitlab/python-gitlab/issues/682, which already has a HEAD implementation in https://github.com/python-gitlab/python-gitlab/commit/f404d394960683d68ccaf3c129f75a2582a2fb77 (unfinished)

JohnVillalovos:
To fix this we will need to have a method which does a `HEAD` request. Currently we don't have any methods which do a `HEAD` request.

valentingregoire:
> To fix this we will need to have a method which does a `HEAD` request. Currently we don't have any methods which do a `HEAD` request.

If we look at @nejch's answer, there is an issue already being solved with a `HEAD` implementation.

PR #1457, opened 2021-05-15
feat(api): implement a count method

Shkurupii:
Hi, I found an issue https://github.com/python-gitlab/python-gitlab/issues/682 which I think I can finish. At the moment work is in progress. I have added tests for http_head(), count(), gl.projects.count() and others.
@nejch, does my change make sense or what else needs to be done except docs? Documentation part is still on me.
Br, Alex

### DIFF
--- a/gitlab/client.py
+++ b/gitlab/client.py
@@ -21,6 +21,7 @@ from typing import cast, Any, Dict, List, Optional, Tuple, TYPE_CHECKING, Union
 
 import requests
 import requests.utils
+from requests.structures import CaseInsensitiveDict
 
 import gitlab.config
 import gitlab.const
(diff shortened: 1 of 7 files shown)


nejch (changes requested):
Thanks for this @Shkurupii, looks great already!

The `count()` method comes with one big caveat. It relies on the `x-total` header which is not available with keyset pagination, and in some cases might not be returned at all on gitlab.com I believe, as they have a custom proxy setup. Have you tried this on any endpoints at gitlab.com?

So this should be clearly documented when you add the docs. And I think we should handle the KeyError people will get if it's not present with a more informative error message (maybe raise GitLabCountError as well). See https://gitlab.com/gitlab-org/gitlab/-/merge_requests/34273 for more discussion on this.

max-wittig:
There is no need for this. We already have this functionality, when using a generator. 

See: https://python-gitlab.readthedocs.io/en/stable/api-usage.html#pagination

Please also keep in mind that GitLab will no longer send these `X-Something` headers in the future!

Shkurupii:
Thanks. Now I see. Maybe that explains why https://github.com/python-gitlab/python-gitlab/issues/682 is stuck. The reason why I chose it is another issue https://github.com/python-gitlab/python-gitlab/issues/1447 where @nejch mentioned https://github.com/python-gitlab/python-gitlab/issues/682. I was thinking it's a good idea to bring http_head method to master from the first citation.
PR then can be closed, isn't? I will open new one just for https://github.com/python-gitlab/python-gitlab/issues/1447. What would you suggest?

nejch:
Damn totally missed that we have this already. :facepalm:  https://github.com/python-gitlab/python-gitlab/blob/ce0e6427d448dfc18085a5403c793d9208ac3cf2/gitlab/client.py#L861

But yes my reference was to the HEAD request for people who don't want to fetch (presumably large) files just to see if they exist. So yes maybe open a PR just for that @Shkurupii based on the work here, thanks!

Issue #751, opened 2019-04-03
CLI: how to list commit in all branches in a project?


hapablanha:
## Description of the problem, including code/CLI snippet
I ran gitlab project-commit list and got something like below, which is lack of details.
id: 5ae8bfc5806c13d47e932c39ae164b58b864cd0b
title: 修复getallfilename不兼容win10的问题

id: 160436738016963da9b043cfdc24622abe94642c
title: 优化代码

id: 021fb3a743c38c13094707d5f9560da262601d64
title: 添加项目文件。

id: 40c84dff754a9bbb3b8c56db9fc89cbc2d513d9a
title: 添加 .gitignore 和 .gitattributes。

## Specifications

  - python-gitlab version:
  - API version you are using (v3/v4):
  - Gitlab server version (or gitlab.com):

hapablanha:
I used CLI command below.
GitLab -o json project-commit -project 1 --all

I only got the commits in master branch.

zoidbergwill:
`all` sounds like it should do the right thing: https://docs.gitlab.com/ee/api/commits.html#list-repository-commits

zoidbergwill:
Oh, I think it's because the `all` is being swallowed by the library as "fetch all pages for this resource" and not being sent as a query param to the API.

Which you can see with the following:

```
 GitLab -d -o json project-commit list --project-id 1 --all  
```

monsdar:
I stumbled across the same issue, but using the Python API. The docs mention the conflict, but did not help very much. I solved it by doing the following: `project.commits.list(all=True, query_parameters={'all': True, 'since': '2016-01-01T00:00:00Z'})`

nejch:
@monsdar would the proposed solution in https://github.com/python-gitlab/python-gitlab/pull/1458/files work better for you?

monsdar:
Sounds like this will cause a lot of confusion where users expect `all=True` to return all pages while in only some cases it returns something else.

I did not check the code in depth: Is it possible to always use `all_pages=True` instead of `all=True`? I'd then change the docs etc to have `all_pages=True` as the new default and only for backwards-compatibility keep support for `all=True` to return all pages when the API does not conflict. This way there is no breaking change, but users won't get confused as easily.

Issue #593, opened 2018-09-21
[CLI] runner-all command is bugged


weakcamel:
## Description of the problem, including code/CLI snippet

There are 2 problems with `runner all` command in CLI; I'm not sure if they're related to each other or completely independent.

The `--id` argument is required - while it shouldn't be - and its value seems to be ignored.

With a workaround (added a dummy `--id foobar` I get the output in the "legacy" format - which is slightly better - but when trying to retrieve it in JSON, it fails again with an exception. Attaching examples below.

## Expected Behavior

```
$ gitlab  runner all --scope active
```

Expected: successful list of all runners in "legacy" format (whatever it might be)

```
$ gitlab -o json runner all --scope active
```

Expected: successful list of all runners in JSON.

## Actual Behavior

a) CLI usage failure: 

```bash
$ gitlab  runner all --scope active
usage: gitlab runner all [-h] --id ID [--scope SCOPE]
gitlab runner all: error: the following arguments are required: --id
```

b) with a workaround but in JSON or YAML output format:

```
$ gitlab  -o yaml runner all --scope active --id foobar
Traceback (most recent call last):
  File "/Users/weakcamel/git/gitlab/scripts/venv/bin/gitlab", line 12, in <module>
    sys.exit(main())
  File "/Users/weakcamel/git/gitlab/scripts/venv/lib/python3.7/site-packages/gitlab/cli.py", line 179, in main
    cli_module.run(gl, what, action, args, verbose, output, fields)
  File "/Users/weakcamel/git/gitlab/scripts/venv/lib/python3.7/site-packages/gitlab/v4/cli.py", line 386, in run
    printer.display_list(data, fields, verbose=verbose)
  File "/Users/weakcamel/git/gitlab/scripts/venv/lib/python3.7/site-packages/gitlab/v4/cli.py", line 311, in display_list
    [get_dict(obj, fields) for obj in data],
  File "/Users/weakcamel/git/gitlab/scripts/venv/lib/python3.7/site-packages/gitlab/v4/cli.py", line 311, in <listcomp>
    [get_dict(obj, fields) for obj in data],
  File "/Users/weakcamel/git/gitlab/scripts/venv/lib/python3.7/site-packages/gitlab/v4/cli.py", line 290, in get_dict
    return obj.attributes
AttributeError: 'dict' object has no attribute 'attributes'

$ gitlab  -o json runner all --scope active --id foobar
Traceback (most recent call last):
[...]

PR #1458, opened 2021-05-15
fix: add `get_all` param (and `--get-all`) to allow passing all to API

nejch:
When the `all=True` option was originally added, it probably wasn't anticipated that the GitLab API would have the same for parameters, making it unusable in some cases as the library consumes the argument (especially with the CLI). Closes https://github.com/python-gitlab/python-gitlab/issues/751.

This is a follow-up in trying to lower the confusion around `all` params and endpoints (getting rid of the `all()` method was the first step IMO). 

Although currently there is an ugly workaround for this (https://github.com/python-gitlab/python-gitlab/issues/751#issuecomment-698855273), I think it'd be best if we have more explicit arguments that don't clash with GitLab API params. At the same time I didn't want to change this completely as `all=True` is probably the most widely used argument for most list endpoints.

But I'd like to hear opinion on this approach  @max-wittig @JohnVillalovos :) I _guess_ gitlab could also introduce an `get_all` API param and then we're screwed again but I think that's unlikely.

Closes https://github.com/python-gitlab/python-gitlab/issues/751.
Related to https://github.com/python-gitlab/python-gitlab/issues/593.

### DIFF
--- a/docs/api-usage.rst
+++ b/docs/api-usage.rst
@@ -110,7 +110,7 @@ Examples:
 
 .. warning::
    Calling ``list()`` without any arguments will by default not return the complete list
-   of items. Use either the ``all=True`` or ``iterator=True`` parameters to get all the
+   of items. Use either the ``get_all=True`` or ``iterator=True`` parameters to get all the
    items when using listing methods. See the :ref:`pagination` section for more
    information.
 
@@ -144,7 +144,7 @@ Some objects also provide managers to access related GitLab resources:
 
    # list the issues for a project
    project = gl.projects.get(1)
-   issues = project.issues.list(all=True)
+   issues = project.issues.list(get_all=True)
 
 python-gitlab allows to send any data to the GitLab server when making queries.
 In case of invalid or missing arguments python-gitlab will raise an exception
@@ -282,13 +282,14 @@ listing methods support the ``page`` and ``per_page`` parameters:
 
    The first page is page 1, not page 0.
 
-By default GitLab does not return the complete list of items. Use the ``all``
+By default GitLab does not return the complete list of items. Use the ``get_all``
 parameter to get all the items when using listing methods:
 
 .. code-block:: python
 
-   all_groups = gl.groups.list(all=True)
-   all_owned_projects = gl.projects.list(owned=True, all=True)
+   all_groups = gl.groups.list(get_all=True)
+
+   all_owned_projects = gl.projects.list(owned=True, get_all=True)
 
 You can define the ``per_page`` value globally to avoid passing it to every
 ``list()`` method call:
--- a/tests/functional/api/test_gitlab.py
+++ b/tests/functional/api/test_gitlab.py
@@ -18,7 +18,7 @@ def test_broadcast_messages(gl):
     msg.save()
     msg_id = msg.id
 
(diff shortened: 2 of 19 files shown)


JohnVillalovos (approved):
Thanks. This looks good to me.

JohnVillalovos:
@nejch Any updates on this? I approved this back in May. Do you want to rebase and we can work on getting it merged.

nejch:
> @nejch Any updates on this? I approved this back in May. Do you want to rebase and we can work on getting it merged.

I also received a comment on the initial issue that it might be confusing to have both, which was a good point. So I might replace this in all the docs as well, and just keep `all` for compatibility only.

JohnVillalovos:
I sometimes wonder if the real long term fix for this type of issue is to only use `query_params` for extra parameters to send to GitLab. I realize it is convenient to just pass `**kwargs` around, but I'm not sure that is the best thing to do.

nejch:
> I sometimes wonder if the real long term fix for this type of issue is to only use `query_params` for extra parameters to send to GitLab. I realize it is convenient to just pass `**kwargs` around, but I'm not sure that is the best thing to do.

I was thinking about this as well but it would really diminish the convenience of using this library vs. just using `requests` for example, which some people do. It would be a lot more verbose IMO.

docs/api-usage.rst
```diff
@@ -205,11 +207,16 @@ listing methods support the ``page`` and ``per_page`` parameters:
...
-parameter to get all the items when using listing methods:
+parameter to get all the items when using listing methods. Alternatively, if
+the endpoint also accepts an ``all`` parameter itself, you can use the
+``all_pages`` parameter for listing methods to avoid this conflict:
 
 .. code-block:: python
 
    all_groups = gl.groups.list(all=True)
+   # or
+   all_groups = gl.groups.list(all_pages=True)
```
JohnVillalovos, line 218: Curious. Can they also use:
```
gl.groups.list(all=True, query_parameters={"all": True})
```

Curious, but also wondering if we should document that, if it does work.

tests/functional/api/test_gitlab.py
```diff
@@ -217,8 +217,8 @@ def test_list_all_false_nowarning(gl, recwarn):
 
 
 def test_list_all_true_nowarning(gl, recwarn):
-    """Using `all=True` will disable the warning"""
-    items = gl.gitlabciymls.list(all=True)
+    """Using `get_all=True` will disable the warning"""
+    items = gl.gitlabciymls.list(get_all=True)
     assert not recwarn
```
JohnVillalovos, line 222: Can we also make sure to have a test with just `all=True` to make sure we aren't breaking anything?

JohnVillalovos (changes requested):

tests/functional/api/test_gitlab.py
nejch: Might seem a little lazy but I parametrized the existing tests to test both in https://github.com/python-gitlab/python-gitlab/pull/1458/commits/6d69d3317894bd9a4b4c289b1ed9d8a87871ca1f in all our tests where we had `all=True` before.

Still not entirely sure how I feel about this PR even after over a year. But - it fixes things! :sweat_smile:

JohnVillalovos (approved):
Thanks @nejch

Issue #1389, opened 2021-04-01
RFE: Add support to upload generic package


JohnVillalovos:
Request to add support for uploading via the packages API a "generic" package.

Documentation about this is here:

https://docs.gitlab.com/ee/user/packages/generic_packages/index.html

Sup3rGeo:
Great to see this was implemented! Any ideas when a release is coming so we can use this without installing from source?

max-wittig:
@nejch What's the status of the automatic release system?

nejch:
@max-wittig it was ready, we just forgot to merge https://github.com/python-gitlab/python-gitlab/pull/1489 :grinning: ran the workflow now and 2.8.0 is released. I hope my inbox doesn't explode in the morning :P

PR #1460, opened 2021-05-16
feat(objects): add support for generic packages API

nejch:
Closes https://github.com/python-gitlab/python-gitlab/issues/1389.

I had to add a `raw` param to put/post like we have for get, instead of using `files` with the multipart encoder, to really ensure only the raw data ended up as the uploaded package :thinking:

Maybe it's just how `files` are handled currently and can still be refactored to take out redundant stuff if we want a raw upload. In the meantime I also cleaned out some code from `http_request` into a helper method and centralized the content_types there (`http_request` would be getting quite long and messy otherwise).

### DIFF
--- a/gitlab/client.py
+++ b/gitlab/client.py
@@ -393,15 +393,9 @@ class Gitlab(object):
         requests_log.setLevel(logging.DEBUG)
         requests_log.propagate = True
 
-    def _create_headers(self, content_type: Optional[str] = None) -> Dict[str, Any]:
-        request_headers = self.headers.copy()
-        if content_type is not None:
-            request_headers["Content-type"] = content_type
-        return request_headers
-
-    def _get_session_opts(self, content_type: str) -> Dict[str, Any]:
+    def _get_session_opts(self) -> Dict[str, Any]:
         return {
-            "headers": self._create_headers(content_type),
+            "headers": self.headers.copy(),
             "auth": self._http_auth,
             "timeout": self.timeout,
             "verify": self.ssl_verify,
@@ -441,12 +435,39 @@ class Gitlab(object):
                 if location and location.startswith("https://"):
                     raise gitlab.exceptions.RedirectError(REDIRECT_MSG)
 
+    def _prepare_send_data(
+        self,
+        files: Dict[str, Any] = None,
+        post_data: Dict[str, Any] = None,
+        raw: Optional[bool] = False,
+    ) -> Tuple:
+        if files:
+            if post_data is None:
+                post_data = {}
+            else:
+                # booleans does not exists for data (neither for MultipartEncoder):
+                # cast to string int to avoid: 'bool' object has no attribute 'encode'
+                for k, v in post_data.items():
+                    if isinstance(v, bool):
+                        post_data[k] = str(int(v))
+            post_data["file"] = files.get("file")
+            post_data["avatar"] = files.get("avatar")
+
+            data = MultipartEncoder(post_data)
+            return (None, data, data.content_type)
+
+        if raw and post_data:
+            return (None, post_data, "application/octet-stream")
+
+        return (post_data, None, "application/json")
+
     def http_request(
         self,
         verb: str,
         path: str,
         query_data: Optional[Dict[str, Any]] = None,
         post_data: Optional[Dict[str, Any]] = None,
+        raw: Optional[bool] = False,
         streamed: bool = False,
         files: Optional[Dict[str, Any]] = None,
         timeout: Optional[float] = None,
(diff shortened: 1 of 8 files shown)


gitlab/client.py
```diff
@@ -441,12 +435,39 @@ def _check_redirects(self, result: requests.Response) -> None:
                 if location and location.startswith("https://"):
                     raise gitlab.exceptions.RedirectError(REDIRECT_MSG)
 
+    def _prepare_send_data(
+        self,
+        files: Dict[str, Any] = None,
+        post_data: Dict[str, Any] = None,
```
JohnVillalovos, line 441:
```suggestion
        files: Optional[Dict[str, Any]] = None,
        post_data: Optional[Dict[str, Any]] = None,
```

gitlab/client.py
```diff
@@ -441,12 +435,39 @@ def _check_redirects(self, result: requests.Response) -> None:
...
+        raw: Optional[bool] = False,
+    ) -> Tuple:
```
JohnVillalovos, line 443:
```suggestion
    ) -> Tuple:[Optional[Dict[str, Any], Optional[Union[Dict[str, Any], requests_toolbelt.multipart.encoder.MultipartEncoder]], str]
```

gitlab/client.py
```diff
@@ -441,12 +435,39 @@ def _check_redirects(self, result: requests.Response) -> None:
...
+
+        return (post_data, None, "application/json")
+
     def http_request(
         self,
         verb: str,
         path: str,
         query_data: Optional[Dict[str, Any]] = None,
         post_data: Optional[Dict[str, Any]] = None,
+        raw: Optional[bool] = False,
```
JohnVillalovos, line 470:
```suggestion
        raw: bool = False,
```

gitlab/client.py
```diff
@@ -441,12 +435,39 @@ def _check_redirects(self, result: requests.Response) -> None:
                 if location and location.startswith("https://"):
                     raise gitlab.exceptions.RedirectError(REDIRECT_MSG)
 
+    def _prepare_send_data(
+        self,
+        files: Dict[str, Any] = None,
+        post_data: Dict[str, Any] = None,
+        raw: Optional[bool] = False,
```
JohnVillalovos, line 442: Since we have a default value of `False` don't need to support `None` value most likely. Not 100% sure on that.

```suggestion
        raw: bool = False,
```

gitlab/client.py
```diff
@@ -683,6 +690,7 @@ def http_post(
         path: str,
         query_data: Optional[Dict[str, Any]] = None,
         post_data: Optional[Dict[str, Any]] = None,
+        raw: Optional[bool] = False,
```
JohnVillalovos, line 693:
```suggestion
        raw: bool = False,
```

gitlab/client.py
```diff
@@ -730,6 +739,7 @@ def http_put(
...
```
JohnVillalovos, line 742:
```suggestion
        raw: bool = False,
```

gitlab/v4/objects/packages.py
```diff
@@ -11,6 +21,110 @@
...
+                file_data = f.read()
+        except OSError:
+            raise exc.GitlabUploadError(f"Failed to read package file {path}")
+
+        url = f"{self._computed_path}/{package_name}/{package_version}/{file_name}"
+        server_data = self.gitlab.http_put(url, post_data=file_data, raw=True, **kwargs)
+
+        return self._obj_cls(
+            self,
+            {
```
JohnVillalovos, line 76: Nit / idea

```suggestion
            attrs = {
```

gitlab/v4/objects/packages.py
```diff
@@ -11,6 +21,110 @@
...
+        )
+
+    @cli.register_custom_action(
+        "GenericPackageManager",
+        ("package_name", "package_version", "file_name"),
+    )
+    @exc.on_http_error(exc.GitlabGetError)
+    def download(
+        self,
+        package_name: str,
```
JohnVillalovos, line 92: What would think of:

```suggestion
        *,
        package_name: str,
```

To enforce using keyword arguments.

JohnVillalovos (changes requested):

gitlab/v4/objects/packages.py
nejch: Maybe we should discuss this in a separate issue for all methods as it's not currently enforced in any custom user-facing methods.

Although I feel like this ends up in a lot of ad-hoc/CI scripts and having the convenience of quick one-liners there still has value IMO, e.g. `project.generic_packages.download(package, version, file)` is not too bad?

nejch (commented):
Thanks! :) I've applied all the typing suggestions, for enforcing keyword arguments we can maybe discuss separately

gitlab/v4/objects/packages.py
JohnVillalovos: Sounds good. Can discuss later!

JohnVillalovos (approved):
LGTM. Thanks!

PR #1463, opened 2021-05-22
chore: add isort as a checker

JohnVillalovos:
isort is a tool to make sure that imports are ordered correctly.

Add a tox environment for it and also add it to the Github workflow to
ensure it is checked for Pull Requests.

Run isort on all the code and commit changes it made.

https://pycqa.github.io/isort/

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -71,7 +71,6 @@ per-file-ignores =
 [isort]
 profile = black
 multi_line_output = 3
-force_sort_within_sections = True
 
 [testenv:docs]
 deps = -r{toxinidir}/rtd-requirements.txt
--- a/tox.ini
+++ b/tox.ini
@@ -1,7 +1,7 @@
 [tox]
 minversion = 1.6
 skipsdist = True
-envlist = py39,py38,py37,py36,pep8,black,twine-check,mypy
+envlist = py39,py38,py37,py36,pep8,black,twine-check,mypy,isort
 
 [testenv]
 passenv = GITLAB_IMAGE GITLAB_TAG
@@ -31,6 +31,14 @@ deps = -r{toxinidir}/requirements.txt
 commands =
   black {posargs} .
 
+[testenv:isort]
+basepython = python3
+deps = -r{toxinidir}/requirements.txt
+       -r{toxinidir}/test-requirements.txt
+       isort
+commands =
+  isort --dont-order-by-type {posargs} {toxinidir}
+
 [testenv:twine-check]
 basepython = python3
 deps = -r{toxinidir}/requirements.txt
@@ -59,6 +67,11 @@ max-line-length = 88
 ignore = E203,E501,W503
 per-file-ignores =
     gitlab/v4/objects/__init__.py:F401,F403
+
+[isort]
+profile = black
+multi_line_output = 3
+force_sort_within_sections = True
 
 [testenv:docs]
 deps = -r{toxinidir}/rtd-requirements.txt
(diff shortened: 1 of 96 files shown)


tox.ini
```diff
@@ -31,6 +31,14 @@ deps = -r{toxinidir}/requirements.txt
 commands =
   black {posargs} .
 
+[testenv:isort]
+basepython = python3
+deps = -r{toxinidir}/requirements.txt
+       -r{toxinidir}/test-requirements.txt
+       isort
+commands =
+  isort --dont-order-by-type {posargs} {toxinidir}
```
nejch, line 40: The isort docs are a bit confusing on this, they recommend both options for projects following PEP8:

https://pycqa.github.io/isort/docs/configuration/options/#order-by-type
> If your project follows PEP8 or a related coding standard and has many imports this is a good default, otherwise you likely will want to turn it off. From the CLI the --dont-order-by-type option will turn this off.

https://pycqa.github.io/isort/docs/configuration/options/#dont-order-by-type
> If your project follows PEP8 or a related coding standard and has many imports this is a good default. You can turn this on from the CLI using --order-by-type.

The commit that introduced it (https://github.com/PyCQA/isort/commit/de127f5c5ea3acd6892933b0dbb4fac48f48aa73) doesn't make it clearer but I think it was a copy/paste error and it seems to me like `order-by-type` is the one really recommended for PEP8? WDYT?

tox.ini
```diff
@@ -60,6 +68,11 @@ ignore = E203,E501,W503
 per-file-ignores =
     gitlab/v4/objects/__init__.py:F401,F403
 
+[isort]
+profile = black
+multi_line_output = 3
+force_sort_within_sections = True
```
nejch, line 74: I think that historically this package has sorted straight imports before from-style (also judging by the diff this PR created). It also seems implied in PEP8 (but not mentioned) and other styleguides. Am I missing something here and this other approach is more common now? Otherwise I think I'd personally go with the default :)

nejch (commented):
Nice, finally don't need to worry about whether the first 10 lines in new files make sense 😁 Just have a few thoughts.

tox.ini
JohnVillalovos: I did `--dont-order-by-type` as I preferred it always being alphabetical regardless of case.

Otherwise it will do things differently depending on the capitalization.  Which I don't like as much.

That was my personal preference.

tox.ini
JohnVillalovos: Again, my personal preference.

As then the imports are ordered by the module name. Not changed depending of if prefaced by `import` or `from`.

When I worked on OpenStack we tended to do it in this style. So it is what I am use to.

tox.ini
nejch: Ok, makes sense :)

tox.ini
nejch: Ok, thanks :) I think in this case I'd agree with the answers in https://softwareengineering.stackexchange.com/q/418329 and stay with defaults for this one (but use the other type-sort setting above).

It will also make the diff a bit smaller, we already need a `.git-blame-ignore-revs` for all the refactoring we've done in the past year IMO :grin:

JohnVillalovos: As a counterpoint: https://docs.openstack.org/hacking/latest/user/hacking.html

OpenStack is probably the largest Python project in the world

JohnVillalovos: I'll update this patch later today and remove the `force_sort_within_sections` option.

JohnVillalovos: Done! 😊

nejch (approved):
Awesome! Thanks again :)

PR #1465, opened 2021-05-23
Switch mr.merge() to use post_data (was using query_data)

JohnVillalovos:
MR https://github.com/python-gitlab/python-gitlab/pull/1121 changed
mr.merge() to use 'query_data'. This appears to have been wrong.

From the Gitlab docs they state it should be sent in a payload body
https://docs.gitlab.com/ee/api/README.html#request-payload since
mr.merge() is a PUT request.

```
Request Payload

API Requests can use parameters sent as query strings or as a
payload body. GET requests usually send a query string, while PUT
or POST requests usually send the payload body
```

### DIFF
--- a/tools/functional/api/test_merge_requests.py
+++ b/tools/functional/api/test_merge_requests.py
@@ -154,3 +154,62 @@ def test_merge_request_should_remove_source_branch(
 
     mr = project.mergerequests.get(mr.iid)
     assert mr.merged_at is not None  # Now is merged
+
+
+def test_merge_request_large_commit_message(
+    project: gitlab.v4.objects.Project, wait_for_sidekiq
+):
+    """Test to ensure https://github.com/python-gitlab/python-gitlab/issues/1452
+    is fixed"""
+    source_branch = "large_commit_message"
+    project.branches.create({"branch": source_branch, "ref": "master"})
+
+    # NOTE(jlvillal): Must create a commit in the new branch before we can
+    # create an MR that will work.
+    project.files.create(
+        {
+            "file_path": f"README.{source_branch}",
+            "branch": source_branch,
+            "content": "Initial content",
+            "commit_message": "New commit in new branch",
+        }
+    )
+
+    mr = project.mergerequests.create(
+        {
+            "source_branch": source_branch,
+            "target_branch": "master",
+            "title": "Large Commit Message",
+            "remove_source_branch": True,
+        }
+    )
+
+    result = wait_for_sidekiq(timeout=60)
+    assert result is True, "sidekiq process should have terminated but did not"
+
+    mr_iid = mr.iid
+    for _ in range(60):
+        mr = project.mergerequests.get(mr_iid)
+        if mr.merge_status != "checking":
+            break
+        time.sleep(0.5)
+    assert mr.merge_status != "checking"
+
+    # Ensure we can get the MR branch
+    project.branches.get(source_branch)
+
+    commit_message = "large_message\r\n" * 1_000
+    assert len(commit_message) > 10_000
+    assert mr.merged_at is None  # Not yet merged
+
+    mr.merge(merge_commit_message=commit_message, should_remove_source_branch=True)
+
+    result = wait_for_sidekiq(timeout=60)
+    assert result is True, "sidekiq process should have terminated but did not"
+
+    # Ensure we can NOT get the MR branch
+    with pytest.raises(gitlab.exceptions.GitlabGetError):
+        project.branches.get(source_branch)
+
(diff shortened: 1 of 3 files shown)


tools/functional/api/test_merge_requests.py
```diff
@@ -95,3 +98,116 @@ def test_merge_request_merge(project):
     with pytest.raises(gitlab.GitlabMRClosedError):
         # Two merge attempts should raise GitlabMRClosedError
         mr.merge()
+
+
+def merge_request_create_helper(
```
nejch, line 103: This could be turned into a fixture factory in `conftest.py` (scoped to functions) that we can then use in any other test as well. The convention seems to be to create a `make_merge_request` fixture with an inner `_make_merge_request` function.
https://docs.pytest.org/en/6.2.x/fixture.html#factories-as-fixtures

Although not sure all of the logic should then stay in that fixture, maybe some asserts would belong in the test 🤔 

We have something similar here
https://github.com/python-gitlab/python-gitlab/blob/1508eb78a03b8d9429e474b7a6814ffe74517abb/tools/functional/conftest.py#L68-L78

nejch (commented):
Thanks @JohnVillalovos!

The folks at `pyup` introduced that MR (see https://github.com/pyupio/pyup/issues/384) so we should maybe check with them why that was done (/cc @ferhat-aram @rafaelpivato). I agree it should be in post data for POST/PUT though, I had the same doubts in the original issue as I couldn't reproduce the issue.

The fact they had issues with 405 Method Not Allowed and that you have a lot of `wait_for_sidekiq` makes me think they were more likely hitting a race condition when applying the merge, but not sure why query data would work in that case.

nejch:
> The folks at `pyup` introduced that MR (see [pyupio/pyup#384](https://github.com/pyupio/pyup/issues/384)) so we should maybe check with them why that was done (/cc @ferhat-aram @rafaelpivato). I agree it should be in post data for POST/PUT though, I had the same doubts in the original issue as I couldn't reproduce the issue.

Actually I think we can go ahead with this, since the previous PR broke behavior for other people and POST is the documented approach by GitLab. If it doesn't work with POST then it should be addressed upstream in GitLab.

Maybe if we can just have that helper as a fixture it'd be nice :)

JohnVillalovos:
> Maybe if we can just have that helper as a fixture it'd be nice :)

Okay. Can we do that later?

Though I'm not sure I totally understand the benefit of making it a fixture as two functions use and both of those are in the same file.

I can do it, but I won't have time to do it until tomorrow at the earliest. Sorry work is a bit busy today...

nejch:
> > Maybe if we can just have that helper as a fixture it'd be nice :)
> 
> Okay. Can we do that later?
> 
> Though I'm not sure I totally understand the benefit of making it a fixture as only two functions use it and both of those are in the same file.
> 
> I can do it, but I won't have time to do it until tomorrow at the earliest. Sorry work is a bit busy today...

For example in the CLI functional tests, we're relying on test order because we don't have fixtures to create a merge request from a new branch:
https://github.com/python-gitlab/python-gitlab/blob/1508eb78a03b8d9429e474b7a6814ffe74517abb/tools/functional/cli/test_cli_v4.py#L189-L224

There are other potential tests we currently don't do because it's not convenient (merge request deployments, merge request <resource a/b/c>) and things like this would really help. But it's a wider issue for all functional tests so definitely can wait.

Also no rush of course, my comment was only that we don't wait for feedback that I initially suggested :)

JohnVillalovos:
@nejch  That makes a lot of sense. Thanks for the explanation!

tools/functional/api/test_merge_requests.py
JohnVillalovos: @nejch 

I have created a pytest.fixture for merge_request.

Let me know what you think! 😊

nejch (commented):
Thanks! Should be much more reusable later with the fixture. Just a tiny comment.

tools/functional/conftest.py
```diff
@@ -201,6 +201,73 @@ def project(gl):
...
+
+    def _merge_request(*, source_branch: str):
+        # Wait for processes to be done before we start...
+        # NOTE(jlvillal): Sometimes the CI would give a "500 Internal Server
+        # Error". Hoping that waiting until all other processes are done will
+        # help with that.
+        result = wait_for_sidekiq(timeout=60)
+        assert result is True, "sidekiq process should have terminated but did not"
+
+        project.branches.create({"branch": source_branch, "ref": "master"})
```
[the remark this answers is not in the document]
JohnVillalovos: Good idea!  Done.

nejch (approved):

PR #1469, opened 2021-05-26
chore: rename 'tools/functional/' to 'tests/functional/'

JohnVillalovos:
Rename the 'tools/functional/' directory to 'tests/functional/'

This makes more sense as these are functional tests and not tools.


nejch (changes requested):
Thanks John! Could you also change the paths in the renovate config:

https://github.com/python-gitlab/python-gitlab/blob/eabe091945d3fe50472059431e599117165a815a/.renovaterc.json#L7

And the Manifest (especially if we move unit tests there also):

https://github.com/python-gitlab/python-gitlab/blob/eabe091945d3fe50472059431e599117165a815a/MANIFEST.in#L3

JohnVillalovos:
> Thanks John! Could you also change the paths in the renovate config:
> 
> https://github.com/python-gitlab/python-gitlab/blob/eabe091945d3fe50472059431e599117165a815a/.renovaterc.json#L7
> 
> And the Manifest (especially if we move unit tests there also):
> 
> https://github.com/python-gitlab/python-gitlab/blob/eabe091945d3fe50472059431e599117165a815a/MANIFEST.in#L3

Done! Thanks.

nejch (approved):

PR #1470, opened 2021-05-26
chore: clean up tox, pre-commit and requirements

nejch:
As we add more things like mypy and isort to the development process it also raises the barrier for new contributors who might get confused by CI failures, so just wanted to add this to pre-commit and hopefully people use it.

Did some cleanup on the dependencies so we can maybe pin them later.


.pre-commit-config.yaml
```diff
@@ -6,17 +6,13 @@ repos:
...
-
-  - repo: https://github.com/alessandrojcm/commitlint-pre-commit-hook
-    rev: v5.0.0
+  - repo: https://github.com/pycqa/isort
+    rev: 5.8.0
     hooks:
-      - id: commitlint-travis
-        additional_dependencies: ['@commitlint/config-conventional']
-        stages: [manual]
+      - id: isort
```
JohnVillalovos, line 18: How does this get the required arguments?

tox.ini
```diff
@@ -11,46 +11,39 @@ usedevelop = True
...
+deps = -r{toxinidir}/requirements-lint.txt
 commands =
   flake8 {posargs} .
 
 [testenv:black]
 basepython = python3
-deps = -r{toxinidir}/requirements.txt
-       -r{toxinidir}/test-requirements.txt
-       black
+deps = -r{toxinidir}/requirements-lint.txt
```
JohnVillalovos, line 26: If going to do this then we might want to use a common tox env dir. Then they can share the same space.

```
envdir={toxworkdir}/pep8
```

tox.ini
```diff
@@ -11,46 +11,39 @@ usedevelop = True
...
 commands =
   black {posargs} .
 
 [testenv:isort]
 basepython = python3
-deps = -r{toxinidir}/requirements.txt
-       -r{toxinidir}/test-requirements.txt
-       isort
+deps = -r{toxinidir}/requirements-lint.txt
```
JohnVillalovos, line 32: If going to do this then we might want to use a common tox env dir. Then they can share the same space.

```
envdir={toxworkdir}/pep8
```

tox.ini
```diff
@@ -11,46 +11,39 @@ usedevelop = True
...
+       -r{toxinidir}/requirements-test.txt
        twine
 commands =
   twine check dist/*
 
 [testenv:mypy]
 basepython = python3
-deps = -r{toxinidir}/requirements.txt
-       -r{toxinidir}/test-requirements.txt
+deps = -r{toxinidir}/requirements-lint.txt
```
JohnVillalovos, line 46: If going to do this then we might want to use a common tox env dir. Then they can share the same space.

```
envdir={toxworkdir}/pep8
```

.pre-commit-config.yaml
nejch: Oops good point, moving all the options to `pyproject.toml` so it's picked up consistently.

tox.ini
nejch: Was thinking that was a bit inefficient, that makes sense thanks!

JohnVillalovos (approved):
Thanks! Looks nice.

PR #1474, opened 2021-05-27
chore: move 'gitlab/tests/' dir to 'tests/unit/'

JohnVillalovos:
Move the 'gitlab/tests/' directory to 'tests/unit/' so we have all the
tests located under the 'tests/' directory.


nejch (approved):
Nice, thanks. Things are starting to look cleaner :)

Heads-up for distro package maintainers this will need a change if you're testing the package outside of tox. I recommend only running `tests/unit` if you want to avoid the docker dependency.
@kloczek @derekschrock @svenstaro @williamh @FedericoCeratto

PR #1478, opened 2021-05-27
Optionally keep containers after running integration tests

benjamb:
While developing tests, waiting for a new GitLab container to spin up every run became a little frustrating, especially given I only wanted to run one file within the suite.

Added an optional flag which may be passed to pytest to keep the containers around after a test run.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -186,6 +186,21 @@ To run these tests:
    # run the python API tests:
    tox -e py_func_v4
 
+When developing tests it can be a little frustrating to wait for GitLab to spin
+up every run. To prevent the containers from being cleaned up afterwards, pass
+`--keep-containers` to pytest, i.e.:
+
+.. code-block:: bash
+
+   tox -e py_func_v4 -- --keep-containers
+
+If you then wish to test against a clean slate, you may perform a manual clean
+up of the containers by running:
+
+.. code-block:: bash
+
+   docker-compose -f tests/functional/fixtures/docker-compose.yml -p pytest-python-gitlab down -v
+
 By default, the tests run against the latest version of the ``gitlab/gitlab-ce``
 image. You can override both the image and tag by providing either the
 ``GITLAB_IMAGE`` or ``GITLAB_TAG`` environment variables.
--- a/gitlab/mixins.py
+++ b/gitlab/mixins.py
@@ -576,7 +576,7 @@ class ObjectDeleteMixin(_RestObjectBase):
         """
         if TYPE_CHECKING:
             assert isinstance(self.manager, DeleteMixin)
-        self.manager.delete(self.get_id())
+        self.manager.delete(self.get_id(), **kwargs)
 
 
 class UserAgentDetailMixin(_RestObjectBase):
--- a/tests/functional/cli/test_cli_artifacts.py
+++ b/tests/functional/cli/test_cli_artifacts.py
@@ -1,11 +1,8 @@
 import subprocess
-import sys
 import textwrap
 import time
 from io import BytesIO
 from zipfile import is_zipfile
-
-import pytest
 
 content = textwrap.dedent(
     """\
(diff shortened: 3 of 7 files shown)


tests/functional/cli/test_cli_artifacts.py
```diff
@@ -27,7 +27,9 @@
 def test_cli_artifacts(capsysbinary, gitlab_config, gitlab_runner, project):
     project.files.create(data)
 
-    while not (jobs := project.jobs.list(scope="success")):
+    jobs = None
+    while not jobs:
+        jobs = project.jobs.list(scope="success")
```
nejch, line 32: Was this change meant to be part of this PR or for a follow-up?

Also maybe we can just get rid of the skip marker then if your idea was to support < 3.8.

tests/functional/fixtures/set_token.rb
```diff
@@ -2,7 +2,7 @@
 
 user = User.find_by_username('root')
 
-token = user.personal_access_tokens.create(scopes: [:api, :sudo], name: 'default');
+token = user.personal_access_tokens.first_or_create(scopes: [:api, :sudo], name: 'default');
```
nejch, line 5: Nice!

tests/functional/conftest.py
```diff
@@ -65,6 +73,21 @@ def docker_compose_file(test_dir):
     return test_dir / "fixtures" / "docker-compose.yml"
 
 
+@pytest.fixture(scope="session")
+def docker_compose_project_name():
+    """Set a consistent project name to enable optional reuse of containers."""
+    return "pytest"
```
nejch, line 79: What do you think about making it a bit more specific to avoid clashes in case people have the same idea with other local projects using `pytest-docker`? Unlikely, I know.
```suggestion
    return "pytest-python-gitlab"
```

README.rst
```diff
@@ -185,6 +185,14 @@ To run these tests:
    # run the python API tests:
    tox -e py_func_v4
 
+When developing tests it can be a little frustrating to wait for GitLab to spin
+up every run. To prevent the containers from being cleaned up afterwards, pass
```
nejch, line 189: Could you just add a quick note that a manual docker-compose down/cleanup might be needed if the user wants a clean start after running this?

nejch (changes requested):
Awesome @benjamb, we lost this when switching from the bash scripts as the cleanup fixture wasnt available in pytest-docker at the time. Was going to get to this but you beat me to it ;)

Just a few tiny comments. BTW, have you tried running the full suites (both API and CLI) a few times in a row to see if our tests actually clean up all resources for consistent results? The functional tests are still in kind of poor shape so I'm not sure. I'll try to play a bit later today but just checking. :) Thanks

tests/functional/cli/test_cli_artifacts.py
benjamb: It was the only thing preventing me from running the full test suite in my debian VM, figured it was insignificant enough to include it in this PR, but I can create a separate one for this if preferred?

tests/functional/conftest.py
benjamb: Yup, makes sense.

README.rst
benjamb: Sure thing.

tests/functional/cli/test_cli_artifacts.py
nejch: No it's ok, I guess I was a little too excited when adding that and 3.8 was still new lol. But we can just remove the skip

benjamb:
> Just a few tiny comments. BTW, have you tried running the full suites (both API and CLI) a few times in a row to see if our tests actually clean up all resources for consistent results? The functional tests are still in kind of poor shape so I'm not sure. I'll try to play a bit later today but just checking. :) Thanks

I could have sworn I had, though after running the full API tests twice in a row I hit a few errors and have since included some additional fixes to address these on top of this PR.

tests/functional/cli/test_cli_artifacts.py
benjamb: Done.

gitlab/mixins.py
```diff
@@ -576,7 +576,7 @@ def delete(self, **kwargs: Any) -> None:
         """
         if TYPE_CHECKING:
             assert isinstance(self.manager, DeleteMixin)
-        self.manager.delete(self.get_id())
+        self.manager.delete(self.get_id(), **kwargs)
```
nejch, line 579: Ah, thanks, nice catch.

nejch (approved):

PR #1480, opened 2021-05-28
chore: correct a type-hint

JohnVillalovos:


nejch (approved):
Thanks! Wasn't sure what to do with the module there yeah. One day I'll read up more on typing :D

PR #1481, opened 2021-05-28
chore: use built-in function issubclass() instead of getmro()

JohnVillalovos:
Code was using inspect.getmro() to replicate the functionality of the
built-in function issubclass()

Switch to using issubclass()

### DIFF
--- a/gitlab/v4/cli.py
+++ b/gitlab/v4/cli.py
@@ -16,7 +16,6 @@
 # You should have received a copy of the GNU Lesser General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-import inspect
 import operator
 import sys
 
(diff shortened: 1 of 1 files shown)


gitlab/v4/cli.py
```diff
@@ -268,12 +267,11 @@ def extend_parser(parser):
...
-            if gitlab.base.RESTManager in inspect.getmro(cls):
-                if cls._obj_cls is not None:
-                    classes.append(cls._obj_cls)
-        except AttributeError:
-            pass
+        if not isinstance(cls, type):
+            continue
+        if issubclass(cls, gitlab.base.RESTManager):
+            if cls._obj_cls is not None:
+                classes.append(cls._obj_cls)
```
nejch, line 274: Do you think this could be simplified even more? Or did I miss something?

```suggestion
        if not issubclass(cls, gitlab.base.RESTManager):
            continue
        if cls._obj_cls is not None:
            classes.append(cls._obj_cls)
```

nejch (commented):
Thanks. Looking at this reminds me there's a lot of V3 cruft left in the code (like this try/except). Just one little question :)

gitlab/v4/cli.py
JohnVillalovos: Need the isinstance() check or it blows up.

```
>>> issubclass("hello", object)
TypeError: issubclass() arg 1 must be a class
>>>
```

nejch: Ah, makes sense then. Thanks!

nejch (approved):

Issue #1479, opened 2021-05-28
decendant groups API endpoint not usable


petzsch:
## Description of the problem, including code/CLI snippet

I would like to access decendant groups of a certain group (to change the visibility of all decendant groups of a certain group).

## Expected Behavior

There should be an API endpoint for this according to: https://docs.gitlab.com/ce/api/groups.html#list-a-groups-descendant-groups

I understand that's a fairly new feature (Oct 2020 in Gitlab 13.5).

## Actual Behavior

There is no implementation for descendant groups

## Specifications

  - python-gitlab version: stable
  - API version you are using (v3/v4): v4
  - Gitlab server version (or gitlab.com): 13.11.3

PR #1482, opened 2021-05-29
feat(objects): add support for descendant groups API

nejch:
Closes https://github.com/python-gitlab/python-gitlab/issues/1479.

I'm struggling to understand why gitlab didn't just add a `recursive` param to subgroups instead of a whole new endpoint, but maybe I'm missing something ¯\\\_(ツ)\_/¯ :grin:


JohnVillalovos (approved):
Thanks. Looks good to me!

Issue #1224, opened 2020-10-30
Question: Is Group wikis supported?


prokher:
Dear Colleagues,

Recently GitLab has [introduced Group wikis](https://about.gitlab.com/releases/2020/10/22/gitlab-13-5-released/#group-wikis). Does `Python GitLab` support it?

Regards,
Alexander.

max-wittig:
I don't have access to EE features, so I'm not implementing them. But I would be happy to review a MR regarding this topic.

PR #1484, opened 2021-05-30
feat(objects): add support for Group wikis

spamsch:
This pull request adds support for wikis on Group level

Closes https://github.com/python-gitlab/python-gitlab/issues/1224

### DIFF
--- a/docs/gl_objects/wikis.rst
+++ b/docs/gl_objects/wikis.rst
@@ -15,24 +15,33 @@ References
   + :class:`gitlab.v4.objects.GroupWikiManager`
   + :attr:`gitlab.v4.objects.Group.wikis`
 
-* GitLab API: https://docs.gitlab.com/ce/api/wikis.html
+* GitLab API for Projects: https://docs.gitlab.com/ce/api/wikis.html
+* GitLab API for Groups: https://docs.gitlab.com/ee/api/group_wikis.html
 
 Examples
 --------
 
-Get the list of wiki pages for a project::
+Get the list of wiki pages for a project. These do not contain the contents of the wiki page. You will need to call get(slug) to retrieve the content by accessing the content attribute::
 
     pages = project.wikis.list()
 
-Get the list of wiki pages for a group::
+Get the list of wiki pages for a group. These do not contain the contents of the wiki page. You will need to call get(slug) to retrieve the content by accessing the content attribute::
 
     pages = group.wikis.list()
 
-Get a single wiki page::
+Get a single wiki page for a project::
 
     page = project.wikis.get(page_slug)
 
-Create a wiki page::
+Get a single wiki page for a group::
+
+    page = group.wikis.get(page_slug)
+
+Get the contents of a wiki page::
+
+    print(page.content)
+
+Create a wiki page on a project level::
 
     page = project.wikis.create({'title': 'Wiki Page 1',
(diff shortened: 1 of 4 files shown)


docs/gl_objects/wikis.rst
```diff
@@ -11,6 +11,9 @@ References
   + :class:`gitlab.v4.objects.ProjectWiki`
   + :class:`gitlab.v4.objects.ProjectWikiManager`
   + :attr:`gitlab.v4.objects.Project.wikis`
+  + :class:`gitlab.v4.objects.GroupWiki`
+  + :class:`gitlab.v4.objects.GroupWikiManager`
+  + :attr:`gitlab.v4.objects.Group.wikis`
 
 * GitLab API: https://docs.gitlab.com/ce/api/wikis.html
```
JohnVillalovos, line 19: Please add link to the API page:

`https://docs.gitlab.com/ee/api/group_wikis.html`

docs/gl_objects/wikis.rst
```diff
@@ -21,6 +24,10 @@ Get the list of wiki pages for a project::
 
     pages = project.wikis.list()
 
+Get the list of wiki pages for a group::
+
+    pages = group.wikis.list()
+
 Get a single wiki page::
 
     page = project.wikis.get(page_slug)
```
JohnVillalovos, line 33: Can we also get an example of doing a "get" of a group wiki page?

And an example of creating a group wiki page too.

JohnVillalovos (changes requested):

JohnVillalovos:
Thanks for this PR!

I left a few comments, but overall it is looking great 👍

max-wittig:
Would it be possible for you to add some unit tests as well for this?

spamsch:
> Would it be possible for you to add some unit tests as well for this?

There are no unit tests for project wikis. Not sure it makes sense to have unit tests for groups only. Could think about adding functional tests. Makes sense?

nejch:
> 
> 
> > Would it be possible for you to add some unit tests as well for this?
> 
> There are no unit tests for project wikis. Not sure it makes sense to have unit tests for groups only. Could think about adding functional tests. Makes sense?

I just triggered the workflow. Group Wikis are a Premium/EE feature so let's see if this works in the CE container 😁 And you're right, we'd need to create a new `test_wikis.py` module with a group wiki mock response for this. There are examples in other modules if the functional tests fail (if it fails, maybe add `@pytest.mark.skip` to the functional test that you added, for now)

JohnVillalovos:
FYI: Running `tox -e black` locally should resolve the black formatting issues.

JohnVillalovos:
Thanks a lot @spamsch 😊

Looks good to me. Personally I would like it if it was rebased and squashed. As right now there are six commits and one of them is a merge commit. I think it will make the project history be better if it was rebased and squashed.

Not sure what @nejch and @max-wittig  think.

nejch:
> Thanks a lot @spamsch 😊
> 
> Looks good to me. Personally I would like it if it was rebased and squashed. As right now there are six commits and one of them is a merge commit. I think it will make the project history be better if it was rebased and squashed.
> 
> Not sure what @nejch and @max-wittig  think.

Ok, i can also squash on merge and add the unit tests along with project wiki tests in a follow-up. Just not from my phone on my commute right now.. :D

Issue #1485, opened 2021-05-30
RFE: We should catch all 301, 302 redirects


JohnVillalovos:
Currently we raise an error if there is a 301, 302 redirect from an `http` URL to an `httpS` URL for any non `GET` methods.

But we don't raise an error for any other redirects.

This causes two problems:

1. `PUT` requests that are redirected get changed to `GET` requests which don't perform the correct action but raise no error. This is because the `GET` response succeeds but since it wasn't a `PUT` it doesn't update. This can be seen by updating a user status. It raises no errors but the status does not change. See issue https://github.com/python-gitlab/python-gitlab/issues/1432
2. `POST` requests blow up with hard to debug tracebacks.  See below. Also see issue https://github.com/python-gitlab/python-gitlab/issues/1477

An example of attempting to upload an SSH key on a URL that redirects with 302 for the `POST` request:
```
Traceback (most recent call last):
  File "/home/jlvillal/.local/lib/python3.9/site-packages/gitlab/base.py", line 77, in __getattr__
    return self.__dict__["_updated_attrs"][name]
KeyError: '_managers'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/jlvillal/sources/local/py-gitlab-testing/./tt.py", line 109, in <module>
    sys.exit(main())
  File "/home/jlvillal/sources/local/py-gitlab-testing/./tt.py", line 82, in main
    k = user.keys.create({"title": "My key", "key": SSH_KEY})
  File "/home/jlvillal/.local/lib/python3.9/site-packages/gitlab/exceptions.py", line 287, in wrapped_f
    return f(*args, **kwargs)
  File "/home/jlvillal/.local/lib/python3.9/site-packages/gitlab/mixins.py", line 325, in create
    return self._obj_cls(self, server_data)
  File "/home/jlvillal/.local/lib/python3.9/site-packages/gitlab/base.py", line 62, in __init__
    self._create_managers()
  File "/home/jlvillal/.local/lib/python3.9/site-packages/gitlab/base.py", line 145, in _create_managers
    managers = getattr(self, "_managers", None)
  File "/home/jlvillal/.local/lib/python3.9/site-packages/gitlab/base.py", line 80, in __getattr__
    value = self.__dict__["_attrs"][name]
TypeError: list indices must be integers or slices, not str
```

max-wittig:
[...]

Issue #1477, opened 2021-05-27
Not possible to create note in issue


Cynerd:
## Description of the problem, including code/CLI snippet
I am unable to create notes. I am simply trying to do:
```
    project = ctx.projects.get(265)
    issue = project.issues.get(729)
    issue.notes.create({'body': 'note content'})
```
(Lifted pretty much strait from documentation)

## Expected Behavior
Note is posted

## Actual Behavior
An exception is thrown:
```
Traceback (most recent call last):
  File "./turris-os-milestone-rename.py", line 60, in <module>
    main()
  File "./turris-os-milestone-rename.py", line 20, in main
    issue.notes.create({'body': 'note content'})
  File "/usr/lib/python3.8/site-packages/gitlab/exceptions.py", line 287, in wrapped_f
    return f(*args, **kwargs)
  File "/usr/lib/python3.8/site-packages/gitlab/mixins.py", line 325, in create
    return self._obj_cls(self, server_data)
  File "/usr/lib/python3.8/site-packages/gitlab/base.py", line 62, in __init__
    self._create_managers()
  File "/usr/lib/python3.8/site-packages/gitlab/base.py", line 151, in _create_managers
    manager = cls(self.manager.gitlab, parent=self)
  File "/usr/lib/python3.8/site-packages/gitlab/base.py", line 290, in __init__
    self._computed_path = self._compute_path()
  File "/usr/lib/python3.8/site-packages/gitlab/base.py", line 305, in _compute_path
    data = {
  File "/usr/lib/python3.8/site-packages/gitlab/base.py", line 306, in <dictcomp>
    self_attr: getattr(self._parent, parent_attr, None)
  File "/usr/lib/python3.8/site-packages/gitlab/base.py", line 80, in __getattr__
    value = self.__dict__["_attrs"][name]
TypeError: list indices must be integers or slices, not str
```

## Specifications

  - python-gitlab version: 2.7.1
  - API version you are using (v3/v4): v4
  - Gitlab server version (or gitlab.com): 13.11.3

JohnVillalovos:
I tried it without issue using gitlab.com:

```
    GL = gitlab.Gitlab(url="https://gitlab.com/", private_token=TOKEN)
    print(GL)
    print()

    project = GL.projects.get(my_project)
    print(project)
    print()

    issue = project.issues.get(1)
    print(issue)
    print()

    result = issue.notes.create({'body': 'note content'})
    print(result)
```

[...]

Issue #1432, opened 2021-04-30
Changing milestone of issue does not work


Cynerd:
## Description of the problem, including code/CLI snippet
I am trying to change  milestone of issue. I understand that I should set `milestone_id` with valid ID of milestone (reference https://github.com/python-gitlab/python-gitlab/issues/612 as well as https://docs.gitlab.com/ee/api/issues.html#edit-issue). The problem is that it seems to not work. The code passes without any error or warning but issue still has the old milestone assigned.

```
    project = ctx.projects.get(265)
    issue = project.issues.get(729)
    issue.milestone_id = 1372
    issue.save()
```

I am sure that I am setting correct milestone ID as I am getting it just few lines prior using code:
```
    group = ctx.groups.get(GROUP)
    new_milestone = next(group.milestones.list(title=new_milestone, as_list=False))
    print(new_milestone.id)
```

## Expected Behavior
Milestone changes.

## Actual Behavior
Nothing happens.

## Specifications

  - python-gitlab version: 2.6.0
  - API version you are using (v3/v4): v4
  - Gitlab server version (or gitlab.com): 13.10.3

Cynerd:
The same behavior with python-gitlab 2.7.1 and Gitlab 13.11.3.

nejch:
@Cynerd could you check if this might be related to your redirect issue in https://github.com/python-gitlab/python-gitlab/issues/1477? Then we can close at least one of these.

Cynerd:
You are right. With redirect resolution suggested in #1477 it works.

PR #1486, opened 2021-05-30
fix: raise error if there is a 301/302 redirection

JohnVillalovos:
Before we raised an error if there was a 301, 302 redirect but only
from an http URL to an https URL.  But we didn't raise an error for
any other redirects.

This caused two problems:

  1. PUT requests that are redirected get changed to GET requests
     which don't perform the desired action but raise no error. This
     is because the GET response succeeds but since it wasn't a PUT it
     doesn't update. See issue:
     #1432
  2. POST requests that are redirected also got changed to GET
     requests. They also caused hard to debug tracebacks for the user.
     See issue:
     #1477

Correct this by always raising a RedirectError exception and improve
the exception message to let them know what was redirected.

Closes: #1485
Closes: #1432
Closes: #1477

### DIFF
--- a/docs/api-usage.rst
+++ b/docs/api-usage.rst
@@ -14,7 +14,9 @@ To connect to a GitLab server, create a ``gitlab.Gitlab`` object:
    import gitlab
 
    # private token or personal token authentication
-   gl = gitlab.Gitlab('http://10.0.0.1', private_token='JVNSESs8EwWRx5yDxM5q')
+   # Note that a 'url' that results in 301/302 redirects will cause an error
+   # (see below for more information).
+   gl = gitlab.Gitlab(url='https://gitlab.example.com', private_token='JVNSESs8EwWRx5yDxM5q')
 
    # oauth token authentication
    gl = gitlab.Gitlab('http://10.0.0.1', oauth_token='my_long_token_here')
@@ -46,6 +48,13 @@ configuration files.
 
    If the GitLab server you are using redirects requests from http to https,
    make sure to use the ``https://`` protocol in the URL definition.
+
+.. note::
+
+   It is highly recommended to use the final destination in the ``url`` field.
+   What this means is that you should not use a URL which redirects as it will
+   most likely cause errors. python-gitlab will raise a ``RedirectionError``
+   when it encounters a redirect which it believes will cause an error.
 
 Note on password authentication
 -------------------------------
--- a/docs/cli-usage.rst
+++ b/docs/cli-usage.rst
@@ -89,6 +89,13 @@ You must define the ``url`` in each GitLab server section.
    If the GitLab server you are using redirects requests from http to https,
    make sure to use the ``https://`` protocol in the ``url`` definition.
 
+.. note::
+
+   It is highly recommended to use the final destination in the ``url`` field.
+   What this means is that you should not use a URL which redirects as it will
+   most likely cause errors. python-gitlab will raise a ``RedirectionError``
+   when it encounters a redirect which it believes will cause an error.
+
 Only one of ``private_token``, ``oauth_token`` or ``job_token`` should be
 defined. If neither are defined an anonymous request will be sent to the Gitlab
 server, with very limited permissions.
@@ -101,7 +108,7 @@ We recommend that you use `Credential helpers`_ to securely store your tokens.
    * - Option
      - Description
    * - ``url``
-     - URL for the GitLab server
+     - URL for the GitLab server. Do **NOT** use a URL which redirects.
    * - ``private_token``
      - Your user token. Login/password is not supported. Refer to `the
        official documentation
--- a/gitlab/client.py
+++ b/gitlab/client.py
@@ -29,8 +29,9 @@ import gitlab.exceptions
 from gitlab import utils
 
 REDIRECT_MSG = (
-    "python-gitlab detected an http to https redirection. You "
-    "must update your GitLab URL to use https:// to avoid issues."
+    "python-gitlab detected a {status_code} ({reason!r}) redirection. You must update "
+    "your GitLab URL to the correct URL to avoid issues. The redirection was from: "
+    "{source!r} to {target!r}"
 )
 
 
@@ -456,24 +457,29 @@ class Gitlab(object):
             return "%s%s" % (self._url, path)
 
     def _check_redirects(self, result: requests.Response) -> None:
-        # Check the requests history to detect http to https redirections.
-        # If the initial verb is POST, the next request will use a GET request,
-        # leading to an unwanted behaviour.
-        # If the initial verb is PUT, the data will not be send with the next
-        # request.
-        # If we detect a redirection to https with a POST or a PUT request, we
+        # Check the requests history to detect 301/302 redirections.
+        # If the initial verb is POST or PUT, the redirected request will use a
+        # GET request, leading to unwanted behaviour.
+        # If we detect a redirection with a POST or a PUT request, we
         # raise an exception with a useful error message.
-        if result.history and self._base_url.startswith("http:"):
-            for item in result.history:
-                if item.status_code not in (301, 302):
-                    continue
-                # GET methods can be redirected without issue
-                if item.request.method == "GET":
-                    continue
-                # Did we end-up with an https:// URL?
-                location = item.headers.get("Location", None)
-                if location and location.startswith("https://"):
-                    raise gitlab.exceptions.RedirectError(REDIRECT_MSG)
+        if not result.history:
+            return
+
+        for item in result.history:
+            if item.status_code not in (301, 302):
+                continue
+            # GET methods can be redirected without issue
+            if item.request.method == "GET":
+                continue
+            target = item.headers.get("location")
+            raise gitlab.exceptions.RedirectError(
+                REDIRECT_MSG.format(
+                    status_code=item.status_code,
+                    reason=item.reason,
+                    source=item.url,
+                    target=target,
+                )
+            )
 
     def _prepare_send_data(
         self,
--- a/tests/unit/test_gitlab_http_methods.py
+++ b/tests/unit/test_gitlab_http_methods.py
@@ -2,7 +2,7 @@ import pytest
 import requests
 from httmock import HTTMock, response, urlmatch
 
-from gitlab import GitlabHttpError, GitlabList, GitlabParsingError
+from gitlab import GitlabHttpError, GitlabList, GitlabParsingError, RedirectError
 
 
 def test_build_url(gl):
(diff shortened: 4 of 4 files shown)


gitlab/client.py
```diff
@@ -29,8 +29,9 @@
 from gitlab import utils
 
 REDIRECT_MSG = (
-    "python-gitlab detected an http to https redirection. You "
-    "must update your GitLab URL to use https:// to avoid issues."
+    "python-gitlab detected a {status_code} ({reason!r}) redirection. You must update "
+    "your GitLab URL to the correct URL to avoid issues. The redirection was from: "
+    "{source!r} to {target!r}"
```
max-wittig, line 34: Any reason for this `!r`. Wouldn't it work without? What does it do exactly here?

tests/unit/test_gitlab_http_methods.py
```diff
@@ -38,6 +38,100 @@ def resp_cont(url, request):
             gl.http_request("get", "/not_there")
 
 
+def test_http_request_302_get(gl):
```
max-wittig, line 41:
```suggestion
def test_http_request_302_get_does_not_throw(gl):
```

tests/unit/test_gitlab_http_methods.py
```diff
@@ -38,6 +38,100 @@ def resp_cont(url, request):
...
+            )
+        ]
+
+        content = (
+            b'{"emoji":"thumbsup","message":"Hi redirect!","availability":"not_set",'
+            b'"message_html":"Hi redirect!","clear_status_at":null}'
+        )
+        resp_obj = response(
+            status_code=200,
+            content=content,
```
max-wittig, line 74: couldn't this be just formatted as a Python object (dict) and provide it as a `json` argument?

tests/unit/test_gitlab_http_methods.py
```diff
@@ -38,6 +38,100 @@ def resp_cont(url, request):
...
+            request=request,
+        )
+        resp_obj.history = history
+        return resp_obj
+
+    with HTTMock(resp_cont):
+        gl.http_request("get", "/user/status")
+
+
+def test_http_request_302_put(gl):
```
max-wittig, line 87:
```suggestion
def test_http_request_302_put_raises_redirecterror(gl):
```

tests/unit/test_gitlab_http_methods.py
```diff
@@ -38,6 +38,100 @@ def resp_cont(url, request):
...
+
+    @urlmatch(
+        scheme="http", netloc="localhost", path="/api/v4/user/status", method="put"
+    )
+    def resp_cont(url, request):
+        req = requests.Request(
+            "PUT",
+            url="http://localhost//api/v4/user/status",
+            data={"message": "Hi there!", "emoji": "thumbsup"},
+        )
```
max-wittig, line 98: req is not used anymore, so maybe we can save a line?

```suggestion
        ).prepare()
```

tests/unit/test_gitlab_http_methods.py
```diff
@@ -38,6 +38,100 @@ def resp_cont(url, request):
...
+                    b'="white">\r\n<center><h1>302 Found</h1></center>\r\n<hr><center>'
+                    b"nginx</center>\r\n</body>\r\n</html>\r\n"
+                ),
+                headers={"Location": "https://example.com/api/v4/user/status"},
+                reason="Moved Temporarily",
+                request=prepped,
+            )
+        ]
+
+        content = (
```
max-wittig, line 115: Same as above. A Python object might be nicer here.

max-wittig (changes requested):
Thanks for your MR. Just some suggestions!

gitlab/client.py
JohnVillalovos: The `!r` just quote it for us automatically. So the message looks like this:

```
python-gitlab detected a 302 ('Moved Temporarily') redirection. You must update your 
GitLab URL to the correct URL to avoid issues. The redirection was 
from: 'https://gitlab.labs.nic.cz/api/v4/user/status'
to 'https://gitlab.nic.cz/api/v4/user/status'
```

Note: I manually word-wrapped it above so it would be easier to read.

Here is raw output from a test run:
```
gitlab.exceptions.RedirectError: python-gitlab detected a 302 ('Moved Temporarily') redirection. You must update your GitLab URL to the correct URL to avoid issues. The redirection was from: 'https://gitlab.labs.nic.cz/api/v4/user/status' to 'https://gitlab.nic.cz/api/v4/user/status'
```

tests/unit/test_gitlab_http_methods.py
JohnVillalovos: Done!

tests/unit/test_gitlab_http_methods.py
JohnVillalovos: I'm not sure. I changed things around and added some comments which hopefully make things clearer.

tests/unit/test_gitlab_http_methods.py
JohnVillalovos: Done

tests/unit/test_gitlab_http_methods.py
JohnVillalovos: Personally I like the explicitness of it as I think it helps readability. But I can change if desired.

tests/unit/test_gitlab_http_methods.py
JohnVillalovos: I'm not sure. I changed things around and added some comments which hopefully make things clearer.

tests/unit/test_gitlab_http_methods.py
```diff
@@ -38,9 +38,104 @@ def resp_cont(url, request):
...
+    http_method = http_method.upper()
+
+    # Create a history which contains our original request which is redirected
+    history = [
+        response(
+            status_code=302,
+            content=(
+                b"<html>\r\n<head><title>302 Found</title></head>\r\n<body bgcolor"
+                b'="white">\r\n<center><h1>302 Found</h1></center>\r\n<hr><center>'
+                b"nginx</center>\r\n</body>\r\n</html>\r\n"
```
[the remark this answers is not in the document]
JohnVillalovos: We don't need it. I was just putting what I saw from a real run in there. But I can just make it an empty string. Tests work fine without it.

Ah I didn't realize we had multiple mock libraries for http. I just copied from the nearby code 😊

gitlab/client.py
```diff
@@ -423,24 +424,27 @@ def _build_url(self, path: str) -> str:
...
-        # If the initial verb is PUT, the data will not be send with the next
-        # request.
-        # If we detect a redirection to https with a POST or a PUT request, we
+        # Check the requests history to detect 301/302 redirections.
+        # If the initial verb is POST or PUT, the redirected request will use a
+        # GET request, leading to unwanted behaviour.
+        # If we detect a redirection with a POST or a PUT request, we
         # raise an exception with a useful error message.
-        if result.history and self._base_url.startswith("http:"):
+        if result.history:
```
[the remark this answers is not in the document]
JohnVillalovos: Sure

tests/unit/test_gitlab_http_methods.py
JohnVillalovos: Will mark as resolved. Let me know if still have questions/issues.

nejch: How about we resolve this in a follow-up issue by removing `httmock` entirely in favor of `responses` where this can be rewritten potentially in a few lines with `rsps.add` :)

docs/api-usage.rst
```diff
@@ -47,6 +48,13 @@ configuration files.
    If the GitLab server you are using redirects requests from http to https,
    make sure to use the ``https://`` protocol in the URL definition.
 
+.. note::
+
+   It is highly recommended to use the final destination in the ``url`` field.
+   What this means is that you should NOT use a URL which redirects as it will
```
nejch, line 54: Sorry for the nitpick - we already have an admonition here so I think this works already 😁 
```suggestion
   What this means is that you should not use a URL which redirects as it will
```

docs/cli-usage.rst
```diff
@@ -89,6 +89,13 @@ You must define the ``url`` in each GitLab server section.
    If the GitLab server you are using redirects requests from http to https,
    make sure to use the ``https://`` protocol in the ``url`` definition.
 
+.. note::
+
+   It is highly recommended to use the final destination in the ``url`` field.
+   What this means is that you should NOT use a URL which redirects as it will
```
nejch, line 95:
```suggestion
   What this means is that you should not use a URL which redirects as it will
```

nejch (commented):
Only some tiny notes from me. I think this should go into the 3.0.0 milestone, right?

docs/api-usage.rst
```diff
@@ -14,7 +14,8 @@ To connect to a GitLab server, create a ``gitlab.Gitlab`` object:
    import gitlab
 
    # private token or personal token authentication
-   gl = gitlab.Gitlab('http://10.0.0.1', private_token='JVNSESs8EwWRx5yDxM5q')
+   # NOTE: Ensure the 'url' specified does NOT redirect
```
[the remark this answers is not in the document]
JohnVillalovos: Done, with tiny change in wording.

docs/api-usage.rst
JohnVillalovos: Done

docs/cli-usage.rst
JohnVillalovos: Done

JohnVillalovos:
> 
> 
> Only some tiny notes from me. I think this should go into the 3.0.0 milestone, right?

3.0.0 is fine for me. Earlier is also fine for me.

tests/unit/test_gitlab_http_methods.py
JohnVillalovos: Sounds reasonable to me to resolve this in a follow-up PR. I will get a chance to learn about `responses` 😊

JohnVillalovos:

nejch:
> 
> 

Actually no @JohnVillalovos this was ready but I had this idea of getting a bunch of things into 3.0.0 🤣 Let's just do it. Could you just rebase & amend the commit message to have a breaking change trailer for semantic-release?

JohnVillalovos:
> amend the commit message to have a breaking change trailer for semantic-release?

How do I do that? I'm not sure what goes in the message to indicate a breaking change.

Thanks.

nejch:
> 
> 
> > amend the commit message to have a breaking change trailer for semantic-release?
> 
> How do I do that? I'm not sure what goes in the message to indicate a breaking change.
> 
> Thanks.

Sure! Here are some examples https://www.conventionalcommits.org/en/v1.0.0/#examples

Actually, maybe it's easier to just amend the title as `fix!: raise error if there is a 301/302 redirection`, looks like it'll have the same effect :) Then the release action should bump it.

JohnVillalovos:
> Actually, maybe it's easier to just amend the title as `fix!: raise error if there is a 301/302 redirection`, looks like it'll have the same effect :) Then the release action should bump it.

Done!

Thanks.

tests/unit/test_gitlab_http_methods.py
nejch: I opened a follow-up issue for this: https://github.com/python-gitlab/python-gitlab/issues/1591

JohnVillalovos:
> Thanks for your MR. Just some suggestions!

@max-wittig  I believe I resolved all your comments. I marked them as resolved.

nejch (approved):

nejch:
> 
> 
> > Thanks for your MR. Just some suggestions!
> 
> @max-wittig I believe I resolved all your comments. I marked them as resolved.

I've also opened an issue to resolve the rest (by using responses instead)

PR #1487, opened 2021-05-31
fix: catch invalid type used to initialize RESTObject

JohnVillalovos:
```
Sometimes we have errors where we don't get a dictionary passed to
RESTObject.__init__() method. This breaks things but in confusing
ways.

Check in the __init__() method and raise an exception if it occurs.
```

### DIFF
--- a/gitlab/base.py
+++ b/gitlab/base.py
@@ -20,6 +20,7 @@ from types import ModuleType
 from typing import Any, Dict, Iterable, NamedTuple, Optional, Tuple, Type
 
 from gitlab import types as g_types
+from gitlab.exceptions import GitlabParsingError
 
 from .client import Gitlab, GitlabList
 
@@ -51,6 +52,12 @@ class RESTObject(object):
     manager: "RESTManager"
 
     def __init__(self, manager: "RESTManager", attrs: Dict[str, Any]) -> None:
+        if not isinstance(attrs, dict):
+            raise GitlabParsingError(
+                "Attempted to initialize RESTObject with a non-dictionary value: "
+                "{!r}\nThis likely indicates an incorrect or malformed server "
+                "response.".format(attrs)
+            )
         self.__dict__.update(
             {
                 "manager": manager,
--- a/tests/unit/test_base.py
+++ b/tests/unit/test_base.py
@@ -19,6 +19,7 @@ import pickle
 
 import pytest
 
+import gitlab
 from gitlab import base
 
 
@@ -84,6 +85,10 @@ class TestRESTObject:
         assert obj._create_managers() is None
         assert fake_manager == obj.manager
         assert fake_gitlab == obj.manager.gitlab
+
+    def test_instantiate_non_dict(self, fake_gitlab, fake_manager):
+        with pytest.raises(gitlab.exceptions.GitlabParsingError):
+            FakeObject(fake_manager, ["a", "list", "fails"])
 
     def test_picklability(self, fake_manager):
         obj = FakeObject(fake_manager, {"foo": "bar"})


JohnVillalovos:
Interested to get some feedback on this. Not sure if that is the correct exception to use.

nejch:
Is the original idea behind this due to the malformed/unexpected response we got in the 301/302 redirect issue with POST? What exactly was the traceback there? :)

JohnVillalovos:
> Is the original idea behind this due to the malformed/unexpected response we got in the 301/302 redirect issue with POST? What exactly was the traceback there? :)

Like in https://github.com/python-gitlab/python-gitlab/issues/1477

```
  File "./turris-os-milestone-rename.py", line 60, in <module>
    main()
  File "./turris-os-milestone-rename.py", line 20, in main
    issue.notes.create({'body': 'note content'})
  File "/usr/lib/python3.8/site-packages/gitlab/exceptions.py", line 287, in wrapped_f
    return f(*args, **kwargs)
  File "/usr/lib/python3.8/site-packages/gitlab/mixins.py", line 325, in create
    return self._obj_cls(self, server_data)
  File "/usr/lib/python3.8/site-packages/gitlab/base.py", line 62, in __init__
    self._create_managers()
  File "/usr/lib/python3.8/site-packages/gitlab/base.py", line 151, in _create_managers
    manager = cls(self.manager.gitlab, parent=self)
  File "/usr/lib/python3.8/site-packages/gitlab/base.py", line 290, in __init__
    self._computed_path = self._compute_path()
  File "/usr/lib/python3.8/site-packages/gitlab/base.py", line 305, in _compute_path
    data = {
  File "/usr/lib/python3.8/site-packages/gitlab/base.py", line 306, in <dictcomp>
    self_attr: getattr(self._parent, parent_attr, None)
  File "/usr/lib/python3.8/site-packages/gitlab/base.py", line 80, in __getattr__
    value = self.__dict__["_attrs"][name]
TypeError: list indices must be integers or slices, not str
```

gitlab/base.py
```diff
@@ -51,6 +52,11 @@ class RESTObject(object):
     manager: "RESTManager"
 
     def __init__(self, manager: "RESTManager", attrs: Dict[str, Any]) -> None:
+        if not isinstance(attrs, dict):
+            raise GitlabParsingError(
+                "Attempted to initialize RESTObject with a non-dictionary value: "
```
nejch, line 57: I think `GitlabParsingError` is fine as we use it elsewhere but my first thought would be simply https://docs.python.org/3/library/exceptions.html#TypeError :)

Just wondering if this message helps the user - the other parsing errors all just throw "Failed to parse the server message". attrs will be `server_data` except when lazily creating objects, so would adding something like `"Unexpected type passed to object: {attrs} This may indicate incorrect/malformed server response"`, or similar, help give more of a clue?

nejch (commented):
Ok, thanks! Just a small thought from my side.

gitlab/base.py
JohnVillalovos: So compared to what is the current case I do think it helps the user more. The previous `TypeError: list indices must be integers or slices, not str` was very confusing.

Could it be better? Yes.  I'm open to a better message on what to say here.

It also helps us as we are better able to help them in troubleshooting if they ask questions.

JohnVillalovos: I could change it to TypeError, but I guess I liked it to be a Gitlab specific exception. To make it clear it was a Gitlab/python-gitlab thing.

gitlab/base.py
```diff
@@ -51,6 +52,11 @@ class RESTObject(object):
...
+                "{!r}".format(attrs)
```
nejch, line 58: Ok, maybe just an extra sentence for the user is all I can think of right now :)

```suggestion
                "Attempted to initialize RESTObject with a non-dictionary value: "
                "{!r}\n"
                "This likely indicates an incorrect or malformed server response.".format(attrs)
```

(black might disagree with this web IDE suggestion)

PR #1490, opened 2021-06-02
feat: add keys endpoint

benjamb:
While this endpoint may be a little niche, it was useful for a project I was involved in recently.

### DIFF
--- a/docs/gl_objects/keys.rst
+++ b/docs/gl_objects/keys.rst
@@ -1,0 +1,28 @@
+####
+Keys
+####
+
+Keys
+====
+
+Reference
+---------
+
+* v4 API
+
+  + :class:`gitlab.v4.objects.Key`
+  + :class:`gitlab.v4.objects.KeyManager`
+  + :attr:`gitlab.Gitlab.keys`
+
+* GitLab API: https://docs.gitlab.com/ce/api/keys.html
+
+Examples
+--------
+
+Get an ssh key by its id (requires admin access)::
+
+    key = gl.keys.get(key_id)
+
+Get an ssh key (requires admin access) or a deploy key by its fingerprint::
+
+    key = gl.keys.get(fingerprint="SHA256:ERJJ/OweAM6jA8OjJ/gXs4N5fqUaREEJnz/EyfywfXY")
--- a/gitlab/v4/objects/keys.py
+++ b/gitlab/v4/objects/keys.py
@@ -1,0 +1,26 @@
+from gitlab.base import RESTManager, RESTObject
+from gitlab.mixins import GetMixin
+
+__all__ = [
+    "Key",
+    "KeyManager",
+]
+
+
+class Key(RESTObject):
+    pass
+
+
+class KeyManager(GetMixin, RESTManager):
+    _path = "/keys"
+    _obj_cls = Key
+
+    def get(self, id=None, **kwargs):
+        if id is not None:
+            return super(KeyManager, self).get(id, **kwargs)
+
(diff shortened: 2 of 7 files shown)


JohnVillalovos:
> On a somewhat unrelated note, is there a reason flake8 isn't included in the pre-commit config?

Good catch. @nejch recently updated it to add `isort` and `black`. But likely we should add `flake8' and `mypy``to it too.

gitlab/v4/objects/keys.py
```diff
@@ -0,0 +1,26 @@
...
+    pass
+
+
+class KeyManager(GetMixin, RESTManager):
+    _path = "/keys"
+    _obj_cls = Key
+
+    def get(self, id=None, **kwargs):
+        if id:
+            return super(KeyManager, self).get(id, kwargs)
```
JohnVillalovos, line 20:
```suggestion
        if id is not None:
            return super(KeyManager, self).get(id, **kwargs)
```

JohnVillalovos (changes requested):

gitlab/v4/objects/keys.py
benjamb: Erk, good spot.

benjamb: @JohnVillalovos I've squashed this in.

JohnVillalovos (approved):
LGTM. Thanks!

I'd like @nejch and/or @max-wittig  to also look.

gitlab/v4/objects/keys.py
```diff
@@ -0,0 +1,26 @@
...
+class KeyManager(GetMixin, RESTManager):
+    _path = "/keys"
+    _obj_cls = Key
+
+    def get(self, id=None, **kwargs):
+        if id is not None:
+            return super(KeyManager, self).get(id, **kwargs)
+
+        if "fingerprint" not in kwargs:
+            raise AttributeError("Missing attribute: fingerprint")
```
nejch, line 23: This should probably have both options:

```suggestion
            raise AttributeError("Missing attribute: id or fingerprint")
```

docs/gl_objects/keys.rst
```diff
@@ -0,0 +1,28 @@
...
+Reference
+---------
+
+* v4 API
+
+  + :class:`gitlab.v4.objects.Key`
+  + :class:`gitlab.v4.objects.KeyManager`
+  + :attr:`gitlab.Gitlab.keys`
+
+* GitLab API: https://docs.gitlab.com/ce/api/keys.html
```
nejch, line 17: Can we have just a quick note that for SSH keys, this requires admin access? Like here:
https://github.com/python-gitlab/python-gitlab/blob/eabe091945d3fe50472059431e599117165a815a/docs/gl_objects/variables.rst#L12

nejch (changes requested):
Thanks a lot, looks great! Just 2 tiny comments :)

nejch:
> > On a somewhat unrelated note, is there a reason flake8 isn't included in the pre-commit config?
> 
> Good catch. @nejch recently updated it to add `isort` and `black`. But likely we should add `flake8` and `mypy` to it too.

Yes, these were just very recent additions to CI checks but we should add pre-commit as well.

gitlab/v4/objects/keys.py
benjamb: Done.

docs/gl_objects/keys.rst
benjamb: And done.

nejch (approved):

PR #1491, opened 2021-06-02
chore: add missing linters to pre-commit and pin versions

nejch:


requirements-lint.txt
```diff
@@ -1,4 +1,4 @@
-black
-flake8
-isort
-mypy
+black==20.8b1
+flake8==3.9.2
+isort==5.8.0
+mypy==0.812
```
JohnVillalovos, line 4: I'll be honest I'm not a fan of pinning versions as then nobody will ever update these and we stay on old versions for a long time.

I rather break when an update occurs and fix it then.  Otherwise we build up technical debt for the time when we do want to update.

JohnVillalovos: But if renovatebot will be updating this, then I like it.

JohnVillalovos: As a note I don't think the requirements-*.txt files will be updated.

https://docs.renovatebot.com/modules/manager/pip_requirements/

nejch: I missed that I changed to a non-standard pattern there, it just seemed nicer like this. :open_mouth:  Pushed a filematch fix, thanks!

Yes definitely with renovate, I don't even think about updating deps manually anymore. But if you don't pin dependencies you can have things break and not notice (like https://github.com/python-gitlab/python-gitlab/pull/1450) and CI starts to randomly fail on contributors and confuse them (we had this a few months ago here with black).

JohnVillalovos (approved):

Issue #1497, opened 2021-06-03
Allow updating "forward deployment" setting


chris-peterson:
## Description of the problem, including code/CLI snippet

This setting in a project's CI configuration is not currently settable via cli
![image](https://user-images.githubusercontent.com/1980791/120686867-4ba0e800-c456-11eb-8149-a5eef3af0227.png)

## Expected Behavior

`gitlab project update --id xxxx --ci_forward_deployment_enabled false`

## Actual Behavior

`gitlab: error: unrecognized arguments: --ci_forward_deployment_enabled false`

## Specifications

  - python-gitlab version: 2.7.1
  - API version you are using (v3/v4): v4
  - Gitlab server version (or gitlab.com): 13.10.3

JohnVillalovos:
As a note when things are working I believe the command line will be:

```
gitlab project update --id xxxx --ci-forward-deployment-enabled false
```

Hyphens (`-`) between the words instead of underscores (`_`).

I have proposed a patch which should add support for this.

chris-peterson:
thank you @JohnVillalovos !

good call on the `-` instead of `_`.  I think I copied it from the json structure rather than existing CLI examples.

chris-peterson:
just verified the fix in `2.8.0`!

`gitlab project update --id 'cpeterson%2fci-test' --ci-forward-deployment-enabled false` ✅ 

a lot of my deployment jobs are multi-regional to the same logical "environment".  the default setting of `true` doesn't make sense -- glad to be able to change it more easily now.

thanks all 🙇

JohnVillalovos:
Glad to know it is working @chris-peterson 😊

PR #1499, opened 2021-06-03
chore: sync create and update attributes for Projects

JohnVillalovos:
Sync the create attributes with:
https://docs.gitlab.com/ee/api/projects.html#create-project

Sync the update attributes with documentation at:
https://docs.gitlab.com/ee/api/projects.html#edit-project

As a note the ordering of the attributes was done to match the
ordering of the attributes in the documentation.

Closes: #1497


max-wittig (approved):
Makes sense

PR #1505, opened 2021-06-08
chore: add new required type packages for mypy

JohnVillalovos:
New version of mypy flagged errors for missing types. Install the
recommended type-* packages that resolve the issues.

PR https://github.com/python-gitlab/python-gitlab/pull/1504 will fail until this is merged.

### DIFF
--- a/requirements-lint.txt
+++ b/requirements-lint.txt
@@ -2,3 +2,5 @@ black==20.8b1
 flake8==3.9.2
 isort==5.8.0
 mypy==0.812
+types-PyYAML==0.1.5
+types-requests==0.1.8


max-wittig:
Interesting thing. Seems like that's kind of a mess. So every package has to kinda publish another package with `types-*`?

nejch:
> 
> 
> Interesting thing. Seems like that's kind of a mess. So every package has to kinda publish another package with `types-*`?

These `types-*` are actually all from the typeshed repo but packaged separately (only released when the specific change). `mypy` used to bundle all the non-stdlib typing but changed this for some reason. It seems like mypy can take care of this itself with an `--install-types` flag, but maybe tracking versions is better

JohnVillalovos:
Other projects are having to do the same thing. Example: https://gitlab.com/cki-project/cki-lib/-/merge_requests/278

nejch (approved):
I guess since we will never have many dependencies for the package itself this shouldn't get out of control and renovate should keep it tidy for us at least for now.

PR #1512, opened 2021-06-12
chore: improve type-hinting for managers

JohnVillalovos:
The 'managers' are dynamically created. This unfortunately means that
we don't have any type-hints for them and so editors which understand
type-hints won't know that they are valid attributes.

 * Add the type-hints for the managers we define.
 * Add a unit test that makes sure that the type-hints and the
   '_managers' attribute are kept in sync with each other.
 * Add unit test that makes sure specified managers in '_managers'
   have a name ending in 'Managers' to keep with current convention.
 * Make RESTObject._managers always present with a default value of
   None.


JohnVillalovos:
So in my quick check I could just use the annotations to create the managers. As all the annotations which are of type `*Manager` are also in the `_managers` attribute.

But I think to be safe would probably be better to do this as a first step and then later switch to using annotations only.

Any thoughts?

gitlab/v4/objects/commits.py
```diff
@@ -17,6 +17,10 @@
 
 class ProjectCommit(RESTObject):
     _short_print_attr = "title"
+
+    comments: "ProjectCommitCommentManager"
+    discussions: ProjectCommitDiscussionManager
+    statuses: "ProjectCommitStatusManager"
```
nejch, line 23: I see you mix plain classes and strings here, should this be consistent everywhere or is there a specific reason for this?

nejch (commented):
I would personally prefer if we could define things in one place and avoid adding too much duplication, and we would also not need to create custom tests to enforce it (and just test `_create_managers` instead).

Using `__annotations__` for this would be convenient although a little magic. But the syntax is so clean it might be less cryptic to new contributors who don't care how managers are created. It could be combined with a simple check that it's a real manager, similar to what we already do for the CLI https://github.com/python-gitlab/python-gitlab/blob/master/gitlab/cli.py#L92-L96

```python
# namespace = gitlab.v4.objects
managers = [man for man in namespace.__dict__ if man.endswith("Manager")]
if manager not in managers:
  continue
...
```
(something prettier than this I guess, but you get the idea :) )

When I look at this in your PR:
```python
class ProjectDeployment(SaveMixin, RESTObject):
    mergerequests: ProjectDeploymentMergeRequestManager
    _managers = (("mergerequests", "ProjectDeploymentMergeRequestManager"),)
```

The new syntax is just so nice and reminds me of dataclasses, and the old one starts to look ugly and I'd just remove it :smile_cat: On the other hand I don't want to force something that's too magic, @max-wittig what's your opinion here?

gitlab/v4/objects/commits.py
JohnVillalovos: It is because of forward references. The class hasn't been declared yet so get an error.

Could probably work around it by using the Python 3.7 or later feature:

```
from future import __annotations__
```

But we currently support Python 3.6 which is EOL at the end of this year (2021).
https://www.python.org/dev/peps/pep-0494/#lifespan

JohnVillalovos:
> I would personally prefer if we could define things in one place and avoid adding too much duplication, and we would also not need to create custom tests to enforce it (and just test `_create_managers` instead).

Agreed.

I am going to update the PR to just use annotations. We can then decide which one we like better.

I do like the annotation only one.

gitlab/v4/objects/commits.py
nejch: I see, thanks! Maybe we could conditionally add the backport, https://pypi.org/project/future-annotations/#files for folks on 3.6 to the install requires, it's a 6kB package with a single module. It would clean up _create_managers a bit as well.

JohnVillalovos: My preference would be to leave it as is. Using strings or proper class names are both valid ways to do it. It is a more of a style issue.

Now when we drop support for Python 3.6, then sure we could add the `from future import __annotations__` though what that does is basically make them all be stored as strings.

There are some disadvantages to doing that as explained in: https://www.python.org/dev/peps/pep-0563/#resolving-type-hints-at-runtime  I am not sure I understand it enough to say which way is best.

gitlab/base.py
```diff
@@ -150,14 +150,22 @@ def __hash__(self) -> int:
...
+            if isinstance(annotation, type):
+                cls_name = annotation.__name__
+            else:
+                cls_name = annotation
+            # All *Manager classes are used except for the base "RESTManager" class
+            if cls_name.endswith("Manager") and cls_name != "RESTManager":
+                cls = getattr(self._module, cls_name)
+                manager = cls(self.manager.gitlab, parent=self)
+                # Since we have our own __setattr__ method, we can't use setattr()
+                self.__dict__[attr] = manager
```
nejch, line 168: Not sure if you're already bored of my `if not .. continue .. dedent` comments but do you something like this 😁, since we already do the same above?

```suggestion
            if cls_name == "RESTManager" or not cls_name.endswith("Manager"):
                continue

            cls = getattr(self._module, cls_name)
            manager = cls(self.manager.gitlab, parent=self)
            # Since we have our own __setattr__ method, we can't use setattr()
            self.__dict__[attr] = manager
```

gitlab/v4/objects/commits.py
nejch: Ok, I was just thinking because it's a tiny package that would only get installed and add `from future import __annotations__` for 3.6, would probably help us clean this up:
```
            if isinstance(annotation, type):
                cls_name = annotation.__name__
            else:
                cls_name = annotation
```

But true we can revisit this later :)

nejch (commented):
> 
> 
> > I would personally prefer if we could define things in one place and avoid adding too much duplication, and we would also not need to create custom tests to enforce it (and just test `_create_managers` instead).
> 
> Agreed.
> 
> I am going to update the PR to just use annotations. We can then decide which one we like better.
> 
> I do like the annotation only one.

I kind of really like this as well, despite the magic so I'll approve it (just needs a rebase and I just have a style question) 😀 When you think about it, the dynamic manager creation has always been a bit magic and considering the stuff we already do in:
https://github.com/python-gitlab/python-gitlab/blob/6abf13a7e25e368da342e7d1da6cfc19915c2dfd/gitlab/v4/cli.py#L177-L178
https://github.com/python-gitlab/python-gitlab/blob/6abf13a7e25e368da342e7d1da6cfc19915c2dfd/gitlab/v4/cli.py#L315-L321
https://github.com/python-gitlab/python-gitlab/blob/6abf13a7e25e368da342e7d1da6cfc19915c2dfd/gitlab/v4/cli.py#L49

... this is not really outrageous compared to that :) But would like also @max-wittig to take a quick look.

gitlab/base.py

gitlab/base.py
```diff
@@ -150,13 +150,22 @@ def __hash__(self) -> int:
...
+        # NOTE(jlvillal): We are creating our managers by looking at the class
+        # annotations. If an attribute is annotated as being a *Manager type
+        # then we create the manager and assign it to the attribute.
+        for attr, annotation in sorted(self.__annotations__.items()):
+            if not isinstance(annotation, (type, str)):
+                continue
+            if isinstance(annotation, type):
+                cls_name = annotation.__name__
+            else:
+                cls_name = annotation
```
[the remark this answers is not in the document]
JohnVillalovos: We are checking to make sure it is one of two types. Either of type `type` or of type `str`. If not we continue.

If it is of type `type` we use the `annotation.__name__` attribute. If it is of type `str` then we use the `str` value directly.

JohnVillalovos:

nejch:
> 
> 

I think it'll also be ready @JohnVillalovos just ping me when you're finished with the current rebase

JohnVillalovos:
> I think it'll also be ready @JohnVillalovos just ping me when you're finished with the current rebase

@nejch  It is ready for review. I added a fix for flakiness in one of the functional tests.

gitlab/base.py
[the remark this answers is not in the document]
nejch: For context, see the other discussion (https://github.com/python-gitlab/python-gitlab/pull/1512/files#r650563002)

tests/functional/api/test_merge_requests.py
```diff
@@ -125,10 +125,18 @@ def test_merge_request_should_remove_source_branch(
...
     # Ensure we can NOT get the MR branch
     with pytest.raises(gitlab.exceptions.GitlabGetError):
-        project.branches.get(source_branch)
+        result = project.branches.get(source_branch)
+        # Help to debug in case the expected exception doesn't happen.
+        import pprint
+
+        print("mr:", pprint.pformat(mr))
+        print("mr.merged_at:", pprint.pformat(mr.merged_at))
+        print("result:", pprint.pformat(result))
```
nejch, line 139: Was this meant to stay here or was it temporary during debugging?

I thought pytest suppressed normal prints so wasn't sure. Maybe we can add `logging` if we want something more long term? Seems like it's natively picked up (`warn` and above): https://docs.pytest.org/en/latest/how-to/logging.html

JohnVillalovos: Well I was thinking about leaving it.

1. This code should never be executed if the test works as the call above will raise an exception.
2. If a test fails then pytest will print stdout. If a test succeeds then stdout is suppressed.
3. So if the test fails then it will print the stdout that we generate.

But more than willing to remove it if preferred.

nejch: Ah, that's right it gets dumped on failure, I forgot. Sure, we can remove it after we finally get rid of the flakiness 😁

nejch (approved):

PR #1515, opened 2021-06-13
chore: add type-hints to gitlab/v4/objects/users.py

JohnVillalovos:
Adding type-hints to gitlab/v4/objects/users.py

### DIFF
--- a/gitlab/v4/objects/users.py
+++ b/gitlab/v4/objects/users.py
@@ -1,7 +1,11 @@
+from typing import Any, cast, Dict, List, Union
+
+import requests
+
 from gitlab import cli
 from gitlab import exceptions as exc
 from gitlab import types
-from gitlab.base import RequiredOptional, RESTManager, RESTObject
+from gitlab.base import RequiredOptional, RESTManager, RESTObject, RESTObjectList
 from gitlab.mixins import (
     CreateMixin,
     CRUDMixin,
@@ -129,7 +133,7 @@ class User(SaveMixin, ObjectDeleteMixin, RESTObject):
 
     @cli.register_custom_action("User")
     @exc.on_http_error(exc.GitlabBlockError)
-    def block(self, **kwargs):
+    def block(self, **kwargs: Any) -> Union[Dict[str, Any], requests.Response]:
         """Block the user.
 
         Args:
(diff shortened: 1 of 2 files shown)


gitlab/v4/objects/users.py
```diff
@@ -319,6 +323,9 @@ class UserManager(CRUDMixin, RESTManager):
     )
     _types = {"confirm": types.LowercaseStringAttribute, "avatar": types.ImageAttribute}
 
+    def get(self, id: Union[str, int], lazy: bool = False, **kwargs: Any) -> User:
+        return cast(User, super().get(id=id, lazy=lazy, **kwargs))
```
nejch, line 327: question: I already merged this for projects but now it has me thinking, is there any way to avoid adding all this boilerplate for typing? There will be a _lot_ of these I imagine if as we go module by module, and I guess will need to be copy/pasted for a lot of mixin usage.

nejch (commented):
I just have a question about how this will go if we continue this way for all modules. I haven't looked at alternatives at all so just asking atm.

gitlab/v4/objects/users.py
JohnVillalovos: I have tried and tried again and tried again... But so far I can't figure out a way to do it. If you can figure it out it will be awesome!

There are ways to do it if you return a `self` object. But since we are returning `self._obj_cls()` it doesn't seem like things are smart enough to understand that yet. Well at least not `mypy`

This will help editors and the type-checker in knowing that `gl.users.get()` returns a `User` object. Otherwise it will think it returns a `RESTObject` which isn't as useful.

JohnVillalovos: Oh, you will probably like my idea for handling `gl.users.list()` even less 😜

At the moment my best solution for that seems to be adding a `UserList` class or something like that. I haven't figured it out yet but thinking about it.

JohnVillalovos:

nejch:
> 
> 

Sorry @JohnVillalovos I never ended up diving into whether we could avoid the extra code in the discussion above. Maybe I could try to find any typed projects that might rely heavily on managers and mixins?

JohnVillalovos:
> 
> Sorry @JohnVillalovos I never ended up diving into whether we could avoid the extra code in the discussion above. Maybe I could try to find any typed projects that might rely heavily on managers and mixins?

Works for me. Would be cool if cleaner solution. So far I haven't been able to figure one out 😟

JohnVillalovos:
@nejch Curious if you had time to think about this. My thought is that this works but isn't pretty. But sometimes that is what we have to do. And it is only two lines of code.

As the expression says, "Perfect is the enemy of good" 😊

Issue #1509, opened 2021-06-10
API docs missing for main Gitlab class


TheFriendlyCoder:
The latest version of the [API docs](https://python-gitlab.readthedocs.io/en/stable/api/gitlab.html) do not include docs for the main gitlab.Gitlab class. This prevents projects from leveraging intersphinx (the documentation plugin for Sphinx) to cross reference docs with this library. This should be fixed.

For testing purposes you can use the sphobjinv tool to query for the class in the object inventory, by doing something like this:

```
pip install sphobjinv
sphobjinv suggest -au https://python-gitlab.readthedocs.io/en/stable/objects.inv Gitlab | grep gitlab\.Gitlab
```

The expected output from this command should be something like:

```
:py:class:`gitlab.Gitlab`
```

nejch:
By latest version do you mean that it used to work and it's now broken or was that just a general comment?

Sorry can't check, typing from my phone here :)

PR #1516, opened 2021-06-13
docs: make Gitlab class and v4 objects usable for intersphinx

nejch:
Not especially proud of this one, but https://stackoverflow.com/questions/8649105/how-to-show-instance-attributes-in-sphinx-doc tells me you have to get into `inspect` hackery just to display the instance attributes for classes, so this seemed easier.

I found that we stopped generating most of the API reference docs after the `gitlab/client.py` and `gitlab/v4/objects/__init__.py` split, so this adds a few fixes for that.

Closes https://github.com/python-gitlab/python-gitlab/issues/1509.

Example of Gitlab, populated with instance attributes:
image


JohnVillalovos (approved):
Thanks @nejch

PR #1522, opened 2021-06-15
Support Release edition API

### LINKED ISSUE #1521: Deprecate tag.set_release_description and remove it from docs
It seems that the code we use to create a release from a tag [here](https://github.com/python-gitlab/python-gitlab/blob/8753add72061ea01c508a42d16a27388b1d92677/gitlab/v4/objects/tags.py#L19) has been deprecated and removed

We should update to use the new release API

I'll try to investigate a little later tonight as it is blocking my team for some automated release tools.

## Expected Behavior
- set_release_description() create or update a release

## Actual Behavior
- 404 error

## Specifications

  - python-gitlab version: 2.8 (and 2.5)

PPaques:
Following https://github.com/python-gitlab/python-gitlab/issues/1521: 

- Removal of the documentation for deprecated API
- Support the PUT function of the [release API](https://docs.gitlab.com/ee/api/releases/#update-a-release)
- Add documentation
- unit-test + integration test (well, I've tried)

Closes #1521

### DIFF
--- a/tests/functional/api/test_releases.py
+++ b/tests/functional/api/test_releases.py
@@ -23,6 +23,15 @@ def test_create_project_release(project, project_file):
     assert release.description == release_description
 
 
+def test_update_save_project_release(project, release):
+    updated_description = f"{release.description} updated"
+    release.description = updated_description
+    release.save()
+
+    release = project.releases.get(release.tag_name)
+    assert release.description == updated_description
+
+
 def test_delete_project_release(project, release):
     project.releases.delete(release.tag_name)
     assert release not in project.releases.list()
--- a/docs/gl_objects/projects.rst
+++ b/docs/gl_objects/projects.rst
@@ -412,10 +412,6 @@ Create a tag::
 
     tag = project.tags.create({'tag_name': '1.0', 'ref': 'master'})
 
-Set or update the release note for a tag::
-
-    tag.set_release_description('awesome v1.0 release')
-
 Delete a tag::
 
     project.tags.delete('1.0')
--- a/docs/gl_objects/releases.rst
+++ b/docs/gl_objects/releases.rst
@@ -27,6 +27,12 @@ Get a single release::
 
     release = project.releases.get('v1.2.3')
 
+Edit a release::
+
+    release.name = "Demo Release"
+    release.description = "release notes go here"
+    release.save()
+
 Create a release for a project tag::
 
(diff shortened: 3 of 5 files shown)


nejch:
Thanks a lot @PPaques, could you please run `tox -e black` to format the code, and  change your commit messages to use `test` type instead of `tests` :) a quick `git rebase -i <base SHA>` should do!

PPaques:
Done :)

I had to do a `git commit -n` for unit-test and integration parts.

The pre-commit hook `mypymypy` was complaining a lot about `tests/functional/api/test_releases.py:*: error: Function is missing a type annotation`

tests/functional/api/test_releases.py
```diff
@@ -23,6 +23,17 @@ def test_create_project_release(project, project_file):
...
 
+def test_update_save_project_release(project, release):
+
+    release = project.releases.get(release_tag_name)
+    release.description = release_description + "updated"
+    release.save()
+
+    assert project.releases.get(release_tag_name)
+    assert release.tag_name == release_tag_name
+    assert release.description == release_description + "updated"
```
nejch, line 34: Since we have a function-scoped fixture for this (https://github.com/python-gitlab/python-gitlab/blob/master/tests/functional/conftest.py#L318-L332), I think it's best to use it instead of reusing one from previous tests, as we then introduce a dependency between them.

You are already passing it in the test but override it. See how we also use the release fixture in subsequent tests. Maybe something like this, removing the initial `get()`:
```suggestion
    updated_description = f"{release.description} updated"
    release.description = updated_description
    release.save()

    release = project.releases.get(release.tag_name)
    assert release.description == updated_description
```

nejch (changes requested):
Thanks again @PPaques! Just a small note on the functional tests, if you could take a look. Sorry for the delay.

tests/functional/api/test_releases.py
PPaques: I'll try to look at it today.
Otherwise it will be in 2 weeks! Time for 🌊 after this long year!

nejch: Sure! If you can't make it this weekend I can merge this as is and tweak tests in a follow-up, it's not blocking.

Enjoy your holidays! :)

PPaques: Change done :)
Thanks for the review :)

nejch (approved):

PR #1533, opened 2021-06-25
feat(api): add group hooks

### LINKED ISSUE #1496: Group Hooks API not supported
The webhook API endpoints for groups appears to be missing from python-gitlab. Only system hooks and project hooks are implemented. Luckily the api is almost identical to that of project hooks, so implementation should be pretty straightforward.

Missing API: https://docs.gitlab.com/ee/api/groups.html#hooks
Project Hooks API counterpart: https://docs.gitlab.com/ee/api/projects.html#hooks 

## Expected Behavior
To be able to list, get, add, edit and delete group hooks.

## Actual Behavior
No API available.

## Specifications

  - python-gitlab version: v2.7.1
  - Gitlab server version (or gitlab.com): 13.11.3-ee

sugonyak:
Resolves #1496

### DIFF
--- a/docs/gl_objects/groups.rst
+++ b/docs/gl_objects/groups.rst
@@ -338,3 +338,43 @@ You can use the ``ldapgroups`` manager to list available LDAP groups::
 
     # list the groups for a specific LDAP provider
     ldap_groups = gl.ldapgroups.list(search='foo', provider='ldapmain')
+
+Groups hooks
+============
+
+Reference
+---------
+
+* v4 API:
+
+  + :class:`gitlab.v4.objects.GroupHook`
+  + :class:`gitlab.v4.objects.GroupHookManager`
+  + :attr:`gitlab.v4.objects.Group.hooks`
+
+* GitLab API: https://docs.gitlab.com/ce/api/groups.html#hooks
+
+Examples
+--------
+
+List the group hooks::
+
+    hooks = group.hooks.list()
+
+Get a group hook::
+
+    hook = group.hooks.get(hook_id)
+
+Create a group hook::
+
+    hook = group.hooks.create({'url': 'http://my/action/url', 'push_events': 1})
+
+Update a group hook::
+
+    hook.push_events = 0
+    hook.save()
+
+Delete a group hook::
+
+    group.hooks.delete(hook_id)
+    # or
+    hook.delete()
--- a/tests/functional/api/test_groups.py
+++ b/tests/functional/api/test_groups.py
@@ -211,6 +211,7 @@ def test_group_wiki(group):
     assert len(group.wikis.list()) == 0
 
 
+@pytest.mark.skip(reason="EE feature")
 def test_group_hooks(group):
     hook = group.hooks.create({"url": "http://hook.url"})
     assert len(group.hooks.list()) == 1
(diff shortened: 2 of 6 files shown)


sugonyak:
Not quite sure if I should change something in CLI part, could someone advise?

Also, tried to use pre-commit checks and mypy gives me a lot of errors in files not affected by my changes, is it normal?

nejch (commented):
Thanks a lot @sugonyak! Looks good. Just a few comments and to answer your questions:

> Not quite sure if I should change something in CLI part, could someone advise?

Since you only use standard methods, the CLI part should be auto-generated including docs and nothing extra is needed here.

> Also, tried to use pre-commit checks and mypy gives me a lot of errors in files not affected by my changes, is it normal?

Sorry about that, looks like the mypy pre-commit does not respect the file list from mypy's config, I need to check. If `tox -e mypy` is happy then for now just bypass it, I'll try to fix it asap.

> I see that tests for project hooks are not implemented, should I implement some for group hooks?

I've suggested factoring them out into a common file, then it should be easier to add tests and maybe reuse fixtures. But even if you just add group tests that's fine, just maybe really put them into `test_hooks.py` for easier reuse. Thanks!

docs/gl_objects/groups.rst
```diff
@@ -338,3 +338,43 @@ You can use the ``ldapgroups`` manager to list available LDAP groups::
...
+  + :class:`gitlab.v4.objects.GroupHook`
+  + :class:`gitlab.v4.objects.GroupHookManager`
+  + :attr:`gitlab.v4.objects.Group.hooks`
+
+* GitLab API: https://docs.gitlab.com/ce/api/groups.html#hooks
+
+Examples
+--------
+
+List the project hooks::
```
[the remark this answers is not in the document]
sugonyak: Missed it, sorry.

tests/unit/objects/test_groups.py
```diff
@@ -153,3 +153,28 @@ def test_refresh_group_import_status(group, resp_groups):
...
+
+
+@pytest.mark.skip(reason="missing test")
+def test_update_group_hook(gl):
+    pass
+
+
+@pytest.mark.skip(reason="missing test")
+def test_delete_group_hook(gl):
+    pass
```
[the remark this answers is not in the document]
sugonyak: Yeah, seems like a good idea.

tests/functional/api/test_groups.py
```diff
@@ -209,3 +209,15 @@ def test_group_wiki(group):
     wiki.save()
     wiki.delete()
     assert len(group.wikis.list()) == 0
+
+
+def test_group_hooks(group):
```
nejch, line 214: Unfortunately we don't have an EE license/docker container running for the functional tests so we will have to skip this for now to have them pass:

```suggestion
@pytest.mark.skip(reason="EE feature")
def test_group_hooks(group):
```

nejch (changes requested):
Awesome, thanks for adding all the unit tests even for other endpoints.

We'll just need to skip the functional test for now unfortunately as we only run them against the gitlab-ce container in CI.

nejch (approved):

tests/functional/api/test_groups.py
nejch: I went ahead and applied this so that it makes it into tomorrow's scheduled release, thanks again @sugonyak!

sugonyak: Awesome, thank you! I'm a bit AFK right now :)

Issue #1520, opened 2021-06-14
Breaking changes in 3.0.0


nejch:
The following is a curated list of upcoming breaking changes for 3.0.0 - we can pin the issue so people can see before/after the major release.
- [x] drop python3.6 support https://github.com/python-gitlab/python-gitlab/pull/1705
- [x] default to gitlab.com if no URL given (https://github.com/python-gitlab/python-gitlab/pull/1278)
- [x] allow options from CLI args or environment variables (https://github.com/python-gitlab/python-gitlab/pull/1296)
- [x] raise error if there is a 301/302 redirection (https://github.com/python-gitlab/python-gitlab/pull/1486) (Not likely a breaking changing, but doesn't hurt to wait until 3.0.0)
- [x] Clean up constants defined in `gitlab.v4.objects` and keep only `gitlab.const` (https://github.com/python-gitlab/python-gitlab/pull/1593)
- [x] remove `mr.pipelines()` method (we have `mr.pipelines.list()` now) https://github.com/python-gitlab/python-gitlab/pull/1592
- [x] remove `members.all()` method (we have `members_all.list()` now) https://github.com/python-gitlab/python-gitlab/pull/1594
- [x] remove `issuesstatistics` in favor of `issues_statistics` (slightly more readable) https://github.com/python-gitlab/python-gitlab/pull/1595
- [x] drop deprecated API endpoints that confuse users (`branch.protect`) https://github.com/python-gitlab/python-gitlab/pull/1598
- [x] drop deprecated release notes via tags API endpoint (https://github.com/python-gitlab/python-gitlab/pull/1596)
- [x] switch from `master` to `main` (https://github.com/python-gitlab/python-gitlab/pull/1629)
- [x] stop URL-encoding `.` in path segments (https://github.com/python-gitlab/python-gitlab/pull/1766)
- [x] do not require config-file to run CLI (https://github.com/python-gitlab/python-gitlab/pull/1743)
- [x] rename confusing `to_project_id` argument (https://github.com/python-gitlab/python-gitlab/pull/1599)
- [x] replace deprecated name_regex attribute in `delete_in_bulk()` (https://github.com/python-gitlab/python-gitlab/pull/1536)

Any more ideas? Just edit the description if anything's missing!

nejch:
> if implemented, #1529 would be a breaking change

Thanks @chris-peterson, I've opened a PR for this.

PR #1536, opened 2021-06-29
fix(api): replace deprecated `name_regex` attribute in `delete_in_bulk()`

antti-mikael:
The `name_regex` attribute has been deprecated in favor of `name_regex_delete`.
See: https://gitlab.com/gitlab-org/gitlab/-/commit/ce99813cf54


gitlab/v4/objects/container_registry.py
```diff
@@ -49,6 +49,6 @@ def delete_in_bulk(self, name_regex=".*", **kwargs):
             GitlabDeleteError: If the server cannot perform the request
         """
         valid_attrs = ["keep_n", "older_than"]
-        data = {"name_regex": name_regex}
+        data = {"name_regex_delete": name_regex}
```
nejch, line 52: We should probably keep our function arguments in sync with the API. I'd say either replace the `name_regex` variable to match the API (making this a breaking change) and we can add this to the 3.0.0 milestone (see also https://github.com/python-gitlab/python-gitlab/issues/1520), or fetch the old `name_regex` from kwargs if you want to keep backwards compatibility.

I'd prefer making a clean break with this if you're ok waiting for the 3.0.0 release for this to land.

nejch (changes requested):
Thanks a lot @antti-mikael! Just a small note on keeping the arguments in sync with the API.

gitlab/v4/objects/container_registry.py
antti-mikael: Yes, I agree. I was bit hesitant to make a breaking change. Third option would be to support both `name_regex` and `name_regex_delete` but that would just add complexity for very little gain.

I'll push the breaking change :+1:

antti-mikael: > We should probably keep our function arguments in sync with the API.

Do we want to keep the default values in sync as well? The old `name_regex` has default value of `.*` that is reflected in the function signature. However the new `name_regex_delete` does *not* have a default. Should the default be left as-is or should the argument be made required/non-optional?

The newly pushed commit (db6eed0) does the former.

nejch: Hey @antti-mikael, thanks for keeping this PR up to date.

We now have 3.0.0 coming with the next scheduled release, so I think we can go with the breaking change.

I was thinking that we could drop the default value to follow the new API and make it required.

For reference, in the GUI it is also required now - you also have to enter a value even though the frontend provides a "default" grayed-out text (which will likely also be removed: https://gitlab.com/gitlab-org/gitlab/-/issues/333905#note_604415045).

Would you agree @JohnVillalovos?

JohnVillalovos: @nejch Sounds good to me. Thanks.

antti-mikael: > thanks for keeping this PR up to date

No problem.

> I was thinking that we could drop the default value to follow the new API and make it required.

Agreed, and commit 1099e96 should implement that change.

PR #1537, opened 2021-06-29
feat(api): add `name_regex_keep` attribute in `delete_in_bulk()`

antti-mikael:


nejch (approved):
Thanks @antti-mikael!

Issue #1494, opened 2021-06-03
Object deletion from lazy get


vbisserie:
## Description of the problem, including code/CLI snippet
Lazy objects with name containing http encoded character (like "/") can not be deleted.

## Expected Behavior
```
p = gitlab.projects.get(1)
b = p.branches.get("foo/bar", lazy=True)  # The name shall contain http encoded char like "/"
b.delete()
```
Branch shall be deleted

## Actual Behavior
`b.delete()` will raise an exception 404

## Specifications

  - python-gitlab version: v2.7.1
  - API version you are using (v3/v4): v4

## Suggestion
https://github.com/python-gitlab/python-gitlab/blob/master/gitlab/utils.py#L60
`clean_str_id` seems to be called on an already encoded name.
"foo/bar" -> "foo%2Fbar" -> "foo%252Fbar"

nejch:
Related to https://github.com/python-gitlab/python-gitlab/issues/1356

PR #1541, opened 2021-07-02
fix: lazy id object

oux:
to align lazy id object and non-lazy id object and then make
ObjectDeleteMixin works with lazy object

Fixes python-gitlab/python-gitlab#1494

Signed-off-by: Sebastien MICHEL <sebastien.t.michel@renault.com>

### DIFF
--- a/gitlab/mixins.py
+++ b/gitlab/mixins.py
@@ -100,9 +100,10 @@ class GetMixin(_RestManagerBase):
             GitlabAuthenticationError: If authentication is not correct
             GitlabGetError: If the server cannot perform the request
         """
+        clean_id = id
         if not isinstance(id, int):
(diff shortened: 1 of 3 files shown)


nejch (changes requested):
Hi @oux, sorry for the very late response. This was part of a much larger URL-encoding issue and I was reluctant to merge a small, specific solution.

I believe we have now solved the underlying issue in a more general way that solves it for all endpoints and ensures a string is only ever URL-encoded once: https://github.com/python-gitlab/python-gitlab/pull/1819

However, I see you've added some tests here which may be useful. If you remove the `mixins.py` change and add only the tests, I'd be really happy to merge this as a test PR. Please let me know if you'd like to continue working on this PR. :)

PR #1549, opened 2021-07-16
feat(api): Add merge_ref for merge requests

mfocko:
Support merge_ref on merge requests that returns commit of attempted
merge of the MR.

Signed-off-by: Matej Focko <mfocko@redhat.com>

Needed to finish packit/ogr#584

### DIFF
--- a/docs/gl_objects/mrs.rst
+++ b/docs/gl_objects/mrs.rst
@@ -182,6 +182,11 @@ Attempt to rebase an MR::
 
     mr.rebase()
 
+Attempt to merge changes between source and target branch::
+
+    response = mr.merge_ref()
+    print(response['commit_id'])
+
 Merge Request Pipelines
 =======================
 
--- a/gitlab/v4/objects/merge_requests.py
+++ b/gitlab/v4/objects/merge_requests.py
@@ -297,6 +297,21 @@ class ProjectMergeRequest(
         data = {}
         return self.manager.gitlab.http_put(path, post_data=data, **kwargs)
 
+    @cli.register_custom_action("ProjectMergeRequest")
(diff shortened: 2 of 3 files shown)


JohnVillalovos:
Looks good to me. Just need to remove the capitalization in the git subject. line.

JohnVillalovos:
I now realize we should update the docs. This page: https://python-gitlab.readthedocs.io/en/stable/gl_objects/mrs.html

An example of calling the new function should be added. It is located in the file `docs/gl_objects/mrs.rst`

A "nice to have" would be a functional test case but I don't think it is required. Maybe @nejch can comment.

mfocko:
> I now realize we should update the docs. This page: https://python-gitlab.readthedocs.io/en/stable/gl_objects/mrs.html
> 
> An example of calling the new function should be added. It is located in the file `docs/gl_objects/mrs.rst`

will have a look :)

> A "nice to have" would be a functional test case but I don't think it is required. Maybe @nejch can comment.

I was actually thinking about it, but it would be still quite general since we cannot assume commit SHA that is returned, maybe we could test just getting hash if it can be merged and checking other two cases (conflicts, closed MR) for exception.

mfocko:
OK, so I fixed the tests, also looked at the API ([link](https://gitlab.com/gitlab-org/gitlab/-/blob/master/lib/api/merge_requests.rb#L508)), apparently the only way it can fail are conflicts (based on the API code)

JohnVillalovos (approved):
Thanks. Looks good to me. Would like to get either @nejch or @max-wittig  to also review before merging.

max-wittig (approved):

mfocko:
Thanks for your reviews! :)

PR #1552, opened 2021-07-21
docs(project): add example on getting a single project using name with namespace

Psycojoker:
Hello,

This is apparently not present in the documentation (I couldn't find it) while the API supports it and it's very useful.

I've tried to keep the terminology "name_with_namespace" from `project.attributes`.

Kind regards,


JohnVillalovos (approved):
Thanks for the PR! Looks good.

PR #1555, opened 2021-07-27
Do not require Release name for creation

iamed2:
This has not been required by the [GitLab API](https://docs.gitlab.com/ee/api/releases/#create-a-release) since v12.5, so should not be required here.

### DIFF
--- a/gitlab/v4/objects/releases.py
+++ b/gitlab/v4/objects/releases.py
@@ -19,7 +19,7 @@ class ProjectReleaseManager(CRUDMixin, RESTManager):
     _obj_cls = ProjectRelease
     _from_parent_attrs = {"project_id": "id"}
     _create_attrs = RequiredOptional(
-        required=("name", "tag_name", "description"), optional=("ref", "assets")
+        required=("tag_name", "description"), optional=("name", "ref", "assets")
     )
     _update_attrs = RequiredOptional(
         optional=("name", "description", "milestones", "released_at")
--- a/tests/functional/api/test_releases.py
+++ b/tests/functional/api/test_releases.py
@@ -23,6 +23,24 @@ def test_create_project_release(project, project_file):
     assert release.description == release_description
 
 
+def test_create_project_release_no_name(project, project_file):
+    unnamed_release_tag_name = "v2.3.4"
+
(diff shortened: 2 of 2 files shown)


JohnVillalovos:
> 
> 
> This has never been required by the [GitLab API](https://docs.gitlab.com/ee/api/releases/#create-a-release) so should not be required here.

Nit: It was required before. See: https://gitlab.com/gitlab-org/gitlab/-/commit/eb9dfa19fa32e84456b54db61152ad4359f34462

Please update the commit message.

morris25:
This will fix https://github.com/invenia/TagBotGitLab/issues/33

iamed2:
Sorry, my bad, I only checked the versions GitLab was still hosting docs for. Looks like the change was released in [GitLab 12.5](https://gitlab.com/gitlab-org/gitlab/-/blob/master/changelogs/archive-12.md#anchor-1250). 

I could not find a list of scopes for conventional-changelog, so I made my best guess based on commit history. 

Unrelated, I also found that the pre-commit mypy check raised a ton of unrelated failures and prevented me from committing, so I had to disable it locally. `tox -e mypy` still succeeds, and it was not fixed by upgrading the version of mypy in the pre-commit configuration.

JohnVillalovos:
Thanks. Overall this looks good.

But we need to test that it works. Can you add a new test to:
https://github.com/python-gitlab/python-gitlab/blob/a9924f48800f57fa8036e3ebdf89d1e04b9bf1a1/tests/functional/api/test_releases.py#L8

I suggest copying the `test_create_project_release` test and make it something like `test_create_project_release_no_name`

And then create the release without the `name` value and make sure it works.

Use a different `tag_name`, to avoid conflicting with the existing test.

iamed2:
Test added. Sorry for pushing without running tests, but for some reason many integration tests are broken for me, and they are also extremely slow on my machine.

JohnVillalovos:
> 
> 
> Test added. Sorry for pushing without running tests, but for some reason many integration tests are broken for me, and they are also extremely slow on my machine.

No worries. Yeah the functional tests are slow both on Github and my personal machine. My personal machine is a bit faster but not by that much.

JohnVillalovos (approved):

PR #1561, opened 2021-07-31
feat(api): support file format for repository archive

dAnjou:


JohnVillalovos (changes requested):
This will also need a documentation update in https://python-gitlab.readthedocs.io/en/stable/gl_objects/projects.html

See "Get the repository archive:"

And can it specify the possible options from https://docs.gitlab.com/ee/api/repositories.html#get-file-archive:
tar.gz, tar.bz2, tbz, tbz2, tb2, bz2, tar, and zip.

dAnjou:
Thanks, @JohnVillalovos! Kinda submitted this as a conversation starter because I also wasn't quite sure about whether and where to add tests.

Will make the changes you suggested.

dAnjou:
> And can it specify the possible options from https://docs.gitlab.com/ee/api/repositories.html#get-file-archive:
> tar.gz, tar.bz2, tbz, tbz2, tb2, bz2, tar, and zip.

I was hesitant to explicitly list them in the method documentation because GitLab might change them, so instead I added a note with a link to the example.

nejch:
> > And can it specify the possible options from https://docs.gitlab.com/ee/api/repositories.html#get-file-archive:
> > tar.gz, tar.bz2, tbz, tbz2, tb2, bz2, tar, and zip.
> 
> I was hesitant to explicitly list them in the method documentation because GitLab might change them, so instead I added a note with a link to the example.

That's a good point, maybe this is easier long-term WDYT @JohnVillalovos?

If you want to add tests - maybe extract these lines into `test_repository_archive()` and try adding a different format:

https://github.com/python-gitlab/python-gitlab/blob/90ecf2f91129ffa0cfb5db58300fbd11638d4ecc/tests/functional/api/test_repository.py#L51-L55

And maybe test that it really worked with e.g.
https://docs.python.org/3/library/tarfile.html#tarfile.is_tarfile
https://docs.python.org/3/library/zipfile.html#zipfile.is_zipfile

JohnVillalovos:
> > > And can it specify the possible options from https://docs.gitlab.com/ee/api/repositories.html#get-file-archive:
> > > tar.gz, tar.bz2, tbz, tbz2, tb2, bz2, tar, and zip.
> > 
> > 
> > I was hesitant to explicitly list them in the method documentation because GitLab might change them, so instead I added a note with a link to the example.
> 
> That's a good point, maybe this is easier long-term WDYT @JohnVillalovos?

Sounds reasonable to me. Could also add in the `note`

```
As of 2021-08 The format options are: tar.gz, tar.bz2, tbz, tbz2, tb2, bz2, tar, and zip.

For the formats available, refer to
https://docs.gitlab.com/ce/api/repositories.html#get-file-archive
```

I am perfectly fine with the current doc update with only the pointer to the Gitlab docs.

dAnjou:
I've added a test and ran it locally a few times. For some reason it always got a 404 for `tar.gz` and `tar.bz2`. No idea why, because I created a project on the GitLab instance that the tests use and getting these formats via the API worked perfectly fine.

nejch:
> I've added a test and ran it locally a few times. For some reason it always got a 404 for `tar.gz` and `tar.bz2`. No idea why, because I created a project on the GitLab instance that the tests use and getting these formats via the API worked perfectly fine.

@dAnjou Thanks, maybe it's something to do with URL encoding since they include `.` and we've had issues with this before. I'll take a look.

JohnVillalovos:
Thanks @dAnjou 

Can you rebase the patch rather than adding a merge commit?

JohnVillalovos:
> I've added a test and ran it locally a few times. For some reason it always got a 404 for `tar.gz` and `tar.bz2`. No idea why, because I created a project on the GitLab instance that the tests use and getting these formats via the API worked perfectly fine.

Strange. I wonder why that is happening. @nejch did you ever have a chance to investigate?

JohnVillalovos:
> Strange. I wonder why that is happening. @nejch did you ever have a chance to investigate?

I think I figured it out. See https://github.com/python-gitlab/python-gitlab/pull/1758 for my "fix" and can see the functional tests pass.

@nejch would like to get your feedback on this.

I really don't think that we should be converting periods to `%2F` as I believe we have other issues about that causing problems.

nejch:
Ok, so it _is_ encoding drama again :sweat_smile: 

> I really don't think that we should be converting periods to %2E as I believe we have other issues about that causing problems.

Could be, would love to get rid of more custom code. Do you have a link to the other issue? We just need to make sure that this is not needed anymore:

https://github.com/python-gitlab/python-gitlab/blob/3ee061c270de3e3becbcaccaed20ffeba833808e/gitlab/client.py#L596-L601

So maybe need to check again what https://github.com/python-gitlab/python-gitlab/commit/a221d7b35bc20da758e7467fe789e16613c54275 was trying to achieve. And possibly test this with paths that may include dots so we don't have a regression :thinking:

JohnVillalovos:
> Ok, so it _is_ encoding drama again 😅
> 
> > I really don't think that we should be converting periods to %2E as I believe we have other issues about that causing problems.
> 
> Could be, would love to get rid of more custom code. Do you have a link to the other issue? We just need to make sure that this is not needed anymore:

I created a PR https://github.com/python-gitlab/python-gitlab/pull/1766 to get rid of it and referenced the issues I found, though I believe I had seen other things related to this but didn't find them in my quick search.

> So maybe need to check again what [a221d7b](https://github.com/python-gitlab/python-gitlab/commit/a221d7b35bc20da758e7467fe789e16613c54275) was trying to achieve. And possibly test this with paths that may include dots so we don't have a regression 🤔

I thought we cleaned that up somewhat in d56a4345c1ae05823b553e386bfa393541117467 But I'm not exactly sure what you are asking for, probably because I'm a bit tired at the moment and not reading everything 😟

Not sure if you really meant this commit d7c79113a4dd4f23789ac8adb17add590929ae53 which first added that '.' -> '%2E' translation. Very little info there besides saying it didn't work with a default gitlab installation. But my PR above is passing all the functional tests without an issues.

nejch:
> Not sure if you really meant this commit [d7c7911](https://github.com/python-gitlab/python-gitlab/commit/d7c79113a4dd4f23789ac8adb17add590929ae53) which first added that '.' -> '%2E' translation. Very little info there besides saying it didn't work with a default gitlab installation. But my PR above is passing all the functional tests without an issues.

Sorry yes this is what I meant, I just did a quick blame there. :facepalm: 

Ahh interesting: it seems like the underlying issue was fixed in GitLab 13.2 by upgrading the Grape library. Previously GitLab needed `%2E` encoding for the API it seems, If I understand this MR and the MRs linked correctly: https://gitlab.com/gitlab-org/gitlab/-/merge_requests/36038. So looks like it's being tested for regressions upstream as well :+1:

JohnVillalovos:
> Ahh interesting: it seems like the underlying issue was fixed in GitLab 13.2 by upgrading the Grape library. Previously GitLab needed `%2E` encoding for the API it seems, If I understand this MR and the MRs linked correctly: https://gitlab.com/gitlab-org/gitlab/-/merge_requests/36038. So looks like it's being tested for regressions upstream as well 👍

Nice research!!! 👍

JohnVillalovos:
@dAnjou Would you be willing to rebase this? I think we fixed the issue that caused the functional tests to fail in a PR that got merged today PR #1766

JohnVillalovos:
> @dAnjou Would you be willing to rebase this? I think we fixed the issue that caused the functional tests to fail in a PR that got merged today PR #1766

I got impatient 😊So I did it myself.

JohnVillalovos (approved):

dAnjou:
Thank you, @JohnVillalovos! Glad the change made it 😊

Issue #1539, opened 2021-07-01
pre-commit hook for mypy fails


antti-mikael:
## Description of the problem, including code/CLI snippet

```sh
$ git status --short
M  gitlab/v4/objects/container_registry.py
$ git commit
black....................................................................Passed
flake8...................................................................Passed
isort....................................................................Passed
mypy.....................................................................Failed
- hook id: mypy
- exit code: 1

gitlab/v4/objects/ldap.py:20: error: Function is missing a type annotation
gitlab/v4/objects/repositories.py:15: error: Function is missing a type annotation
gitlab/v4/objects/repositories.py:38: error: Function is missing a type annotation
...
```

Maybe the mypy hook could be configured to ignore missing annotations or warn instead of erring -- at least until the annotations are in place:

```patch
diff --git a/.pre-commit-config.yaml b/.pre-commit-config.yaml
index 80c8fb8..c69e125 100644
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,3 +24,4 @@ repos:
     rev: v0.812
     hooks:
       - id: mypy
+        args: ["--allow-untyped-defs"]
```

## Expected Behavior

I'm able to make commits for files that don't have type annotations.

## Actual Behavior

mypy fails for missing function annotations.

## Specifications

  - python-gitlab version: master (85713bb)
  - API version you are using (v3/v4): N/A
  - Gitlab server version (or gitlab.com): N/A

nejch:
Thanks for opening the issue, we already had this reported but i forgot to open one!

Yeah, that would help although then people might get CI failures because pre-commit wouldn't catch issues in files where it already should check for it.

I think a better solution would be to replace the `files` list in our config with an `exclude` pattern because currently the pre-commit hook is probably overriding the default files list and hence fails.

https://mypy.readthedocs.io/en/stable/config_file.html#confval-files

https://github.com/python-gitlab/python-gitlab/blob/85713bbbecdcec577a72749d2e495f823791b00f/.mypy.ini#L1-L2

I will test this tonight but feel free to already open a PR if you get it to work :)

antti-mikael:
Sorry, I totally missed the `.mypy.ini`. I'll see if I have the time to fiddle with this.

antti-mikael:
> I think a better solution would be to replace the `files` list in our config with an `exclude` pattern because currently the pre-commit hook is probably overriding the default files list and hence fails.

[...]

PR #1562, opened 2021-08-01
chore: fix mypy pre-commit hook

nejch:
Closes https://github.com/python-gitlab/python-gitlab/issues/1539.

Additionally, since mypy 0.910 doesn't bundle stubs anymore we need to track them for pre-commit's own virtualenv :/ See https://github.com/pre-commit/pre-commit/issues/730

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,6 +3,28 @@ profile = "black"
 multi_line_output = 3
 order_by_type = false
 
+[tool.mypy]
+disallow_incomplete_defs = true
+disallow_untyped_defs = true
+
+[[tool.mypy.overrides]] # Overrides for currently untyped modules
+module = [
+    "docs.*",
+    "docs.ext.*",
+    "gitlab.v4.objects.*",
+    "setup",
+    "tests.functional.*",
+    "tests.functional.api.*",
+    "tests.unit.*"
+]
+ignore_errors = true
+
+[[tool.mypy.overrides]] # Overrides to negate above patterns
+module = [
+    "gitlab.v4.objects.projects"
+]
+ignore_errors = false
+
 [tool.semantic_release]
 version_variable = "gitlab/__version__.py:__version__"
 commit_subject = "chore: release v{version}"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,6 +6,7 @@ order_by_type = false
 [tool.mypy]
 disallow_incomplete_defs = true
 disallow_untyped_defs = true
+files = "."
 
 [[tool.mypy.overrides]] # Overrides for currently untyped modules
 module = [
--- a/tox.ini
+++ b/tox.ini
@@ -41,7 +41,7 @@ basepython = python3
 envdir={toxworkdir}/lint
 deps = -r{toxinidir}/requirements-lint.txt
 commands =
-  mypy {posargs} .
+  mypy {posargs}
 
 [testenv:twine-check]
 basepython = python3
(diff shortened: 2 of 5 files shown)


pyproject.toml
```diff
@@ -3,6 +3,28 @@ profile = "black"
 multi_line_output = 3
 order_by_type = false
 
+[tool.mypy]
```
JohnVillalovos, line 6: So we also have `.mypy.ini`

Why this change here?

nejch: Sorry, I forgot to mention. While upgrading to mypy 0.91 and looking at docs I found that it now supports `pyproject.toml` so I thought I'd consolidate more tool config in there. I've actually deleted `.mypy.ini` in this PR. But I just realized I left out the comments from the previous config, I'll amend that.

JohnVillalovos: Ah. I somehow missed noticing the deletion of the file.

pyproject.toml
```diff
@@ -3,6 +3,28 @@ profile = "black"
...
+    "tests.functional.*",
+    "tests.functional.api.*",
+    "tests.unit.*"
+]
+ignore_errors = true
+
+[[tool.mypy.overrides]] # Overrides to negate above patterns
+module = [
+    "gitlab.v4.objects.projects"
+]
```
antti-mikael, line 25: The overrides `module` listing does not match the previous `files` definition from `.mypy.ini`. Is this intentional?

nejch: Yes, since `pre-commit` explicitly passes files to mypy (including sometimes `docs/`, `tests/`, I'm unsure why), I had to add some files outside `gitlab/` to ignores as well. Since it only supports wildcards, I then had to negate the `gitlab.v4.objects.*` ignore pattern with this one, since it's already typed.

tox.ini
```diff
@@ -41,7 +41,7 @@ basepython = python3
 envdir={toxworkdir}/lint
 deps = -r{toxinidir}/requirements-lint.txt
 commands =
-  mypy {posargs}
+  mypy {posargs} .
```
javatarz, line 44: This might break when `tox -e mypy -- tests/unit/test_cli.py` is run for a single file/sub-dir check.

javatarz, line 44: The `.` causes failures when `tox` is used to run `mypy` on a single file with the command `tox -e mypy -- tests/unit/test_cli.py`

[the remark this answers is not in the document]
javatarz: I've looked at `mypy` docs and issues and there seems to be no way to address failures when the same file shows up twice. Don't think we can fix this one.

javatarz (approved):

tox.ini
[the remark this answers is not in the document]
nejch: Thanks for that! I can actually move this to just be `files = "."` inside `pyproject.toml` (`gitlab/` would be better but pre-commit wouldn't care anyway), and then it'll just override it and should work.

javatarz:
Can we please get this merged? The pre-commit hooks are broken and the `mypy` stage blocks commits every time.

The workaround I've been using is to disable `mypy` locally (via a commit) and then remove the commit before pushing. This is painful.

tox.ini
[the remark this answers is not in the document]
nejch: Done.

nejch:
> Can we please get this merged? The pre-commit hooks are broken and the `mypy` stage blocks commits every time.
> 
> The workaround I've been using is to disable `mypy` locally (via a commit) and then remove the commit before pushing. This is painful.

Sorry, I know this was really annoying!

@JohnVillalovos I'm back from the dead :laughing: would you be able to take a quick look/merge?

JohnVillalovos:
Thanks @nejch !

PR #1566, opened 2021-08-04
docs(mergequests): gl.mergequests.list documentation was missleading

Psycojoker:
Hello,

Quoting gitlab's official documentation: https://docs.gitlab.com/ce/api/merge_requests.html#list-merge-requests

> Get all merge requests the authenticated user has access to. By default it returns only merge requests created by the current user. To get all merge requests, use parameter scope=all. 

I've tried to adapt the documentation according to this fact.

Kind regards,

### DIFF
--- a/docs/gl_objects/mrs.rst
+++ b/docs/gl_objects/mrs.rst
@@ -30,9 +30,13 @@ Reference
 Examples
 --------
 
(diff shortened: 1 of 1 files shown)


nejch:
Thank you @Psycojoker!

PR #1567, opened 2021-08-06
fix(api): allow creating an approval rule without requiring 'name'

ptalbert:
The API allows to create an approval rule by referencing a project-level
approval rule ID. In this case, it is not necessary to pass a rule
'name'. Change the ProjectMergeRequestApprovalRuleManager create method
to use a different set of _create_attrs when an approval_project_rule_id
is passed in.

Signed-off-by: Patrick Talbert <ptalbert@redhat.com>


JohnVillalovos:
@ptalbert  Will you be updating this to fix the CI failures?

PR #1580, opened 2021-08-29
Improve helper example section with env var usage

glennie:
Hi,
I wanted to use an env var for the token.
I spend some time to figure out that I needed to escape '%' because of configparser.
Therefore, I made this PR to improve the current documentation hoping others will spend much less time than me.
Kind regards,

### DIFF
--- a/docs/cli-usage.rst
+++ b/docs/cli-usage.rst
@@ -170,6 +170,25 @@ In `/path/to/helper.sh`:
 
     #!/bin/bash
     pass show path/to/password | head -n 1
+    
+Example for an `environnement variable`:
+
+.. code-block:: ini
+
+   [global]
+   default = somewhere
+   ssl_verify = true
+   timeout = 5
+
+   [somewhere]
+   url = http://somewhe.re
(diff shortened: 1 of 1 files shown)


docs/cli-usage.rst
```diff
@@ -170,6 +170,25 @@ In `/path/to/helper.sh`:
...
+
+   [somewhere]
+   url = http://somewhe.re
+   private_token = helper: printf -- '%%s' ${GITLAB_TOKEN}
+   timeout = 1
+
+**Notice:**
+
+   * ``printf`` is the executable provided by coreutils package (on a debian based distribution).
+   * ``%`` must be escaped using an other `%` (in order to allow `configparser` load configuration properly.
```
nejch, line 191: Thank you for this contribution @glennie! Initially I wanted to say this would probably be superseded by https://github.com/python-gitlab/python-gitlab/pull/1296, but I like your idea because you can define multiple env vars (e.g. one per gitlab id entry).

Just one question: Wouldn't simply using `echo -n ${VARIABLE}` achieve the same? Have you tried this? `-n` should strip the newline.

Issue #1584, opened 2021-09-01
typing improvement


function2-llx:
Hello,

This is not a bug report, but a minor improvement suggestion. 

I'm suggesting the [type annotation here](https://github.com/python-gitlab/python-gitlab/blob/master/gitlab/v4/objects/repositories.py#L183) change to `bytes` instead of `str`, so that my PyCharm IDE wouldn't complain the type mismatch (see the figure below).

![图片](https://user-images.githubusercontent.com/38486514/131671072-07def0e7-c8ee-41cd-88f5-a2d669136a70.png)

Thanks in advance!

PR #1585, opened 2021-09-02
docs: correct documented return type

JohnVillalovos:
repository_archive() returns 'bytes' not 'str'

https://docs.gitlab.com/ee/api/repositories.html#get-file-archive

Fixes: #1584


nejch:
Thanks @JohnVillalovos :)

Issue #1586, opened 2021-09-02
are tests required to be packaged into the wheel ?


dzen:
## Description of the problem, including code/CLI snippet
Simple question here : our CI/CD is crashing because the wheel does package the tests, which clashes with our current code.
I think polluting the python path with such common package name is not recommended.

What is your thoughts about it ?

## Expected Behavior
A local application with tests does find the local tests, not the python-gitlab one's

## Actual Behavior
The current version ships a "tests" package which clashes with local tests module

## Specifications

  - python-gitlab version: 2.10.1
  - API version you are using (v3/v4): v4
  - Gitlab server version (or gitlab.com): N/A

nejch:
I see this issue was introduced with https://github.com/python-gitlab/python-gitlab/pull/1562/files which turned it into a package.

Originally the tests were bundled as some downstream package maintainers run tests against the installed dist and not the git repo, see https://github.com/freebsd/freebsd-ports/blob/main/devel/py-python-gitlab/Makefile#L55 and https://github.com/python-gitlab/python-gitlab/issues/907 for an example.

But with our move to the pytest structure this doesn't make sense anymore, true. Off the top of my head I don't know a clean way other than putting at least unit tests back in the gitlab namespace if we want to bundle them for package maintainers, but we should take a look at other options maybe.

dzen:
Great Thank you for your quick response!

PR #1587, opened 2021-09-02
fix(build): only package tests in source distribution

nejch:
Closes https://github.com/python-gitlab/python-gitlab/issues/1586.

This way tests are packaged in the source distribution (via includes in Manifest), but excluded from the wheel.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -25,6 +25,8 @@ jobs:
             toxenv: py38
           - python-version: 3.9
             toxenv: py39
+          - python-version: 3.9
+            toxenv: smoke
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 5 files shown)


JohnVillalovos (approved):
Looks good.

[sand] Would be cool if there was some way the CI could test the package to make sure that `tests` are excluded from the wheel and also check to make sure they are in the source distribution. But I'm fine merging as-is.

nejch:
> Looks good.
> 
> [sand] Would be cool if there was some way the CI could test the package to make sure that `tests` are excluded from the wheel and also check to make sure they are in the source distribution. But I'm fine merging as-is.

Why not :) added some quick smoke tests to run along the unit tests.

JohnVillalovos (approved):
Nice. Thanks!

PR #1606, opened 2021-09-22
fix: urlencode id part of path

Pro:
The `tag_name` part of the path must be urlencoded.

Found this issue when trying to update a release which has a slash in its title, e.g., "1.2.3/testing"

See also:
https://docs.gitlab.com/ee/api/releases/index.html#update-a-release

### DIFF
--- a/gitlab/mixins.py
+++ b/gitlab/mixins.py
@@ -15,6 +15,7 @@
 # You should have received a copy of the GNU Lesser General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import urllib.parse
 from types import ModuleType
 from typing import (
     Any,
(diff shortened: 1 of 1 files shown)


Pro:
@nejch this PR should be ready for review/merge

Pro:
I squashed all the commits. Currently I cannot invest more time into writing tests, maybe in a few weeks when there is more time, sorry.

nejch:
@Pro this might need a rebase now as we've switched to f-strings.

nejch:
Hi @Pro! This was part of a much larger URL-encoding issue and this PR provides a specific solution in `update()`.

I believe we have now solved the underlying issue in a more general way that solves it for all endpoints and ensures a string is only ever URL-encoded once: https://github.com/python-gitlab/python-gitlab/pull/1819.

So I'll close this PR as I think the issue is fixed and we have tests for this now. But thanks again for your contribution, it showed us how widespread the URL-encoding problem was. Feel free to report any more issues or PRs if you find anything else :)

Issue #1609, opened 2021-09-23
Cant query current ldap groups


andrewlarssen:
## Description of the problem, including code/CLI snippet
There are methods to both add and remove ldap links on groups but not to query the existing ldap links

## Expected Behavior
There should be a method on group to query the currently linked ldap groups hitting "https://gitlab.example.com/api/v4/groups/:id/ldap_group_links"

## Actual Behavior
No method exists

## Specifications

  - python-gitlab version: latest
  - API version you are using (v3/v4): 4
  - Gitlab server version (or gitlab.com): latest

max-wittig:
That's an EE feature. As I have no EE instance to test with, this is not really a priority for the project.

But we're very open to MRs, so if you want to contribute that I would be happy to review it.

andrewlarssen:
Thanks @max-wittig for the reply. Happy to create MR. I have already been having a play. There are a couple of things that may need to discuss like the inconsistency of the API. Other api calls return an empty list if there are no elements found where as this returns a 404 and a message. Once I have an MR I will add another comment

andrewlarssen:
@max-wittig created https://github.com/python-gitlab/python-gitlab/pull/1612

nejch:
Now done in https://github.com/python-gitlab/python-gitlab/pull/2371

PR #1612, opened 2021-09-24
Add support for query ldap groups links

andrewlarssen:
You can currently add or delete ldap links but you can't query existing links

### DIFF
--- a/gitlab/v4/objects/groups.py
+++ b/gitlab/v4/objects/groups.py
@@ -38,6 +38,8 @@ __all__ = [
     "GroupDescendantGroupManager",
     "GroupSubgroup",
     "GroupSubgroupManager",
+    "GroupLdapLinks",
+    "GroupLdapLinksManager",
 ]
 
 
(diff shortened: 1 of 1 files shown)


andrewlarssen:
For issue https://github.com/python-gitlab/python-gitlab/issues/1609

andrewlarssen:
Have tested this manually on a real gitlab instance. I could not see any automated tests for gitlab -> ldap features

andrewlarssen:
Fixed 2 linting issues. One was commit message so had to force push

nejch (changes requested):
Thanks a lot @andrewlarssen! Just a comment on naming consistency. If you're able to add some unit tests to this, would also be great! Here https://github.com/python-gitlab/python-gitlab/blob/master/tests/unit/objects/test_groups.py or you could create a new `test_ldap_links.py` if it's getting too crowded there. Unit tests mock the responses so we can test EE features there.

While reviewing this I noticed our existing method for deletion uses deprecated API endpoints: https://docs.gitlab.com/ee/api/groups.html#delete-ldap-group-link

So as a bonus if you find time, I think these methods actually could be removed and you can add the CreateMixin/DeleteMixin to the manager, and ObjectDeleteMixin to the object. Then we could remove the custom methods here:

https://github.com/python-gitlab/python-gitlab/blob/227607ca47c78e3958c6649edb644c7e26d55281/gitlab/v4/objects/groups.py#L112-L150)

This would be a breaking change, but we have a major release coming up anyway. Let me know if this makes sense or you need help. Also, you'll just need to amend your message again as commitlint does not like capitalized sentences :)

PR #1617, opened 2021-10-06
feat(build): officially support and test python 3.10

nejch:
Just added this as `feat` so it shows up in the changelog but I can change it as well.


max-wittig:
Nice!

Issue #1622, opened 2021-10-06
Switch from `master` to `main` for default branch


nejch:
It's been a little over a year since GitHub switched from master to main (October 1st 2020). Maybe we can do the same now before releasing 3.0.0.

https://github.blog/changelog/2020-10-01-the-default-branch-for-newly-created-repositories-is-now-main/

GitLab also switched to `main` as the default branch for new projects in 14.0, so this should make the docs more relevant now.

PR #1629, opened 2021-10-09
chore: rename `master` branch to `main`

nejch:
BREAKING CHANGE: As of python-gitlab 3.0.0, the default branch for development
has changed from `master` to `main`.

GitHub changed on 1-10-2020, GitLab changed with 14.0.

Closes https://github.com/python-gitlab/python-gitlab/issues/1622. ~~Will switch the branch in the project first before assigning this.~~ Switched to main and pushed master again to have an old branch in case some people relied on it.


JohnVillalovos:
Also a thought. If decide to keep `master` maybe put in a final commit which deletes everything and directs people to use the `main` branch. Then `master` branch exists but won't cause confusion for people.

PR #1648, opened 2021-10-20
Also retry HTTP-based transient errors

mitar:
So currently the library can automatically retry when there are some types of API errors. But sometimes the HTTP calls fails on a lower level (connection broken, etc.). I think that should count as a transient error as well and be retried if configured so. This PR adds this.

Closes https://github.com/python-gitlab/python-gitlab/issues/1885

### DIFF
--- a/gitlab/client.py
+++ b/gitlab/client.py
@@ -609,18 +609,29 @@ class Gitlab(object):
             prepped.url, {}, streamed, verify, None
         )
 
+        retry_transient_errors = kwargs.get(
+            "retry_transient_errors", self.retry_transient_errors
+        )
         cur_retries = 0
         while True:
-            result = self.session.send(prepped, timeout=timeout, **settings)
+            try:
+                result = self.session.send(prepped, timeout=timeout, **settings)
+            except requests.ConnectionError:
+                if retry_transient_errors and (
+                    max_retries == -1 or cur_retries < max_retries
+                ):
+                    wait_time = 2 ** cur_retries * 0.1
+                    cur_retries += 1
+                    time.sleep(wait_time)
+                    continue
+
+                raise
 
             self._check_redirects(result)
 
             if 200 <= result.status_code < 300:
                 return result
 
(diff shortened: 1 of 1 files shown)


nejch:
Thanks for this @mitar. As this grows (I believe you also added the initial retry flag?), I'm wondering if we should switch to using the https://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html#urllib3.util.Retry utility for this. It might reduce our duplication and even clean up our custom 429 handling.

Would you like to give it a try here? If not we can do this in a follow-up, but it'd be great if we could at least have tests for this so that we can see it still behaves properly later. One example of similar tests is here, I think:
https://github.com/python-gitlab/python-gitlab/blob/master/tests/unit/test_gitlab_http_methods.py#L58-L123

mitar:
Unstale.

nejch:
@mitar did you still want to take a look at urllib3's retry class or you want to proceed this way?

nejch:
@mitar as we're getting more requests on this I'm happy to merge this as is and we can work on the rest later, could you just rebase/resolve conflicts please? :bow:

Sineaggi:
@mitar Would you mind if someone else took over this pr?

mitar:
Hey, sorry, I got busy with other things. I can add anyone to the branch if you wan to rebase/fix conflicts. I will try to do it as well, but I do not know when I will be able to circle back to this.

mitar:
(But definitely sadly I do not have time to add more tests or to port it to urlib3. I would suggest that this gets merged and then followup work is done to get things further. Also, I have been running this code for some time now and it works great/stable.)

Sineaggi:
I've rebased @mitar's branch here https://github.com/python-gitlab/python-gitlab/pull/1904

Added tests and 52x http response range to retries.

nejch:
Closing this in favor of https://github.com/python-gitlab/python-gitlab/pull/1904, as I see git commit authorship was preserved. Thanks for the work here @mitar.

PR #1666, opened 2021-11-05
Update docs

gthieleb:
add description for ubuntu

### DIFF
--- a/docs/cli-usage.rst
+++ b/docs/cli-usage.rst
@@ -446,6 +446,12 @@ Bash
 .. code-block:: console
 
    eval "$(register-python-argcomplete gitlab)"
+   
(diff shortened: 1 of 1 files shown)


nejch:
Hi @gthieleb, thanks for this. Just wondering, is this really necessary? I just tried the following (Ubuntu 20.04, you can also spin up a quick container to test it instead of using a venv):

```shell
# if testing in venv
python3 -m venv .venv
. .venv/bin/activate

pip install python-gitlab[autocompletion]
eval "$(register-python-argcomplete gitlab)"
```

And I had no issues, working autocompletion. Otherwise if I'm missing something please let me know.
But this will need a small change in the commit message for CI to pass :) (see `CONTRIBUTING.rst`).

PR #1667, opened 2021-11-05
refactor: use new-style formatting for named placeholders

nejch:
Basically just https://pyformat.info/#named_placeholders with some sed. Similar to the f-strings, just getting rid of more python2 stuff. Although this one might have some performance penalty I'm not sure it's noticeable

### DIFF
--- a/gitlab/v4/objects/commits.py
+++ b/gitlab/v4/objects/commits.py
@@ -143,7 +143,7 @@ class ProjectCommit(RESTObject):
 
 
 class ProjectCommitManager(RetrieveMixin, CreateMixin, RESTManager):
-    _path = "/projects/%(project_id)s/repository/commits"
+    _path = "/projects/{project_id}/repository/commits"
     _obj_cls = ProjectCommit
     _from_parent_attrs = {"project_id": "id"}
     _create_attrs = RequiredOptional(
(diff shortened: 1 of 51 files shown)


tests/functional/api/test_merge_requests.py
```diff
@@ -27,9 +27,11 @@ def test_merge_requests(project):
             "commit_message": "New commit in new branch",
         }
     )
-    project.mergerequests.create(
+
+    mr = project.mergerequests.create(
         {"source_branch": "branch1", "target_branch": "main", "title": "MR readme2"}
     )
+    assert mr.id is not None
```
JohnVillalovos, line 34: What's this?

gitlab/v4/objects/commits.py
```diff
@@ -202,10 +202,10 @@ def create(
         # project_id and commit_id are in the data dict when using the CLI, but
         # they are missing when using only the API
         # See #511
-        base_path = "/projects/%(project_id)s/statuses/%(commit_id)s"
+        base_path = base_path = "/projects/{project_id}/statuses/{commit_id}"
```
JohnVillalovos, line 205: This looks wrong to me, or at least a redundant `=` sign.

JohnVillalovos (changes requested):
The one redundant `=` sign should be fixed. Also seems like unrelated test change.

tests/functional/api/test_merge_requests.py
nejch: Wow thanks for that, brainfart during rebase 😁 fixed.

gitlab/v4/objects/commits.py
nejch: Same as above 🤦 thanks

JohnVillalovos (approved):

PR #1668, opened 2021-11-06
chore: add type-hints to gitlab/v4/objects/groups.py

JohnVillalovos:
* Add type-hints to gitlab/v4/objects/groups.py
     * Add return value to share() function as GitLab docs show it returns
       a value.
     * Add 'get()' method so that type-checkers will understand that
       getting a group is of type Group.

### DIFF
--- a/gitlab/v4/objects/groups.py
+++ b/gitlab/v4/objects/groups.py
@@ -1,3 +1,8 @@
+from typing import Any, BinaryIO, cast, Dict, List, Optional, Type, TYPE_CHECKING, Union
+
+import requests
+
+import gitlab
 from gitlab import cli
 from gitlab import exceptions as exc
 from gitlab import types
@@ -74,7 +79,7 @@ class Group(SaveMixin, ObjectDeleteMixin, RESTObject):
 
     @cli.register_custom_action("Group", ("project_id",))
     @exc.on_http_error(exc.GitlabTransferProjectError)
-    def transfer_project(self, project_id, **kwargs):
+    def transfer_project(self, project_id: int, **kwargs: Any) -> None:
         """Transfer a project to this group.
 
         Args:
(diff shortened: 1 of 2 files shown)


gitlab/v4/objects/groups.py
```diff
@@ -184,11 +201,11 @@ def share(self, group_id, group_access, expires_at=None, **kwargs):
             "group_access": group_access,
             "expires_at": expires_at,
         }
-        self.manager.gitlab.http_post(path, post_data=data, **kwargs)
+        return self.manager.gitlab.http_post(path, post_data=data, **kwargs)
```
nejch, line 204: Should we also document this new return in the docstring?

I'm thinking in the future we should just remove the docstring typing since it duplicates the hints but just wondering.

nejch (commented):
Thanks again @JohnVillalovos just one little question :)

gitlab/v4/objects/groups.py
JohnVillalovos: Done!

gitlab/v4/objects/groups.py
```diff
@@ -177,18 +194,24 @@ def share(self, group_id, group_access, expires_at=None, **kwargs):
...
         data = {
             "group_id": group_id,
             "group_access": group_access,
             "expires_at": expires_at,
         }
-        self.manager.gitlab.http_post(path, post_data=data, **kwargs)
+        server_data = self.manager.gitlab.http_post(path, post_data=data, **kwargs)
+        if TYPE_CHECKING:
+            assert isinstance(server_data, dict)
+        return Group(self.manager, server_data)
```
JohnVillalovos, line 210: @nejch This I'm not 100% sure about.

Maybe it should be this instead:

```
manager = GroupManager(self.manager.gitlab, parent=self.manager._parent)
return Group(manager, server_data)
```

nejch: Since we already have the group, should this be enough?

```suggestion
        self._update_attrs(server_data)
```
The API docs are not super clear but it should be the same group details that are returned, not the other group I hope - that would be confusing :grin:

JohnVillalovos: Hmm. I'm not sure... 🤷‍♂️

I'm not sure which group details it returns. The group we are sharing or the group we shared it with?

nejch: I just checked with
```python
gl.http_post('/groups/123/share', group_id=456, group_access=30)
```

And the json response had the details from the group we are sharing :) So this should be ok.

JohnVillalovos: Cool and thanks! Updated it with your suggestion

PR #1673, opened 2021-11-07
chore: add type-hints to gitlab/v4/objects/merge_requests.py

JohnVillalovos:
* Add type-hints to gitlab/v4/objects/merge_requests.py
 * Add return value to cancel_merge_when_pipeline_succeeds() function
   as GitLab docs show it returns a value.
 * Add return value to approve() function as GitLab docs show it
   returns a value.
 * Add 'get()' method so that type-checkers will understand that
   getting a project merge request is of type ProjectMergeRequest.

### DIFF
--- a/gitlab/v4/objects/merge_requests.py
+++ b/gitlab/v4/objects/merge_requests.py
@@ -3,6 +3,11 @@ GitLab API:
 https://docs.gitlab.com/ee/api/merge_requests.html
 https://docs.gitlab.com/ee/api/merge_request_approvals.html
 """
+from typing import Any, cast, Dict, Optional, TYPE_CHECKING, Union
+
+import requests
+
+import gitlab
 from gitlab import cli
 from gitlab import exceptions as exc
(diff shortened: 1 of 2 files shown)


gitlab/v4/objects/merge_requests.py
```diff
@@ -428,6 +456,11 @@ class ProjectMergeRequestManager(CRUDMixin, RESTManager):
         "labels": types.ListAttribute,
     }
 
+    def get(
+        self, id: Union[str, int], lazy: bool = False, **kwargs: Any
+    ) -> ProjectMergeRequest:
+        return cast(ProjectMergeRequest, super().get(id=id, lazy=lazy, **kwargs))
```
nejch, line 462: Maybe for a follow-up it'd be good to document this in the Development docs so we can point people to this in case they are adding new managers and get hit by mypy errors?

nejch (approved):

PR #1677, opened 2021-11-07
chore(ci): add workflow to lock old issues

nejch:
See https://github.com/dessant/lock-threads for more info. The terraform repo also uses this action so.. must be ok :eyes:


JohnVillalovos (approved):
Looks good to me. Merge whenever you want.

Thanks.

PR #1681, opened 2021-11-08
Ensure get() methods have correct type-hints

JohnVillalovos:
Fix classes which don't have correct 'get()' methods for classes
derived from GetMixin.

Add a unit test which verifies that classes have the correct return
type in their 'get()' method.


tests/unit/objects/test_ensure_type_hints.py
```diff
@@ -0,0 +1,84 @@
...
+            if obj_cls != signature.return_annotation:
+                filename = inspect.getfile(class_value)
+                # print("Bad return annotation")
+                failed_messages.append(
+                    (
+                        f"class definition for {class_name!r} in file {filename!r} "
+                        f"must have defined a 'get' method with a return annotation of "
+                        f"{obj_cls} but found {signature.return_annotation}"
+                    )
+                )
```
nejch, line 62: The way I'd probably do this is have an assert per class, so instead of looping over classes in a single test case I'd provide them as parametrized fixtures or something like that.
Then we don't need to manually construct the error message. But maybe I'm just obsessed with pytest and it's not needed :)

nejch (commented):
I like the idea so that people get a helpful message to guide them, because this can get quite complex for new contributors I think.

For me this (and `test_mro.py`) are kind of almost meta tests, so I wonder if we should move them together with `test_dists.py` or so (all the other modules in `objects/` really test functions on gitlab resources).

tests/unit/objects/test_ensure_type_hints.py
JohnVillalovos: But then don't you only show one error and stop? At the moment we have like 100+ to fix. I like to be able to see every one that fails.

If there is a way to show every failure as an assert that would be great!

nejch: I think pytest should show all errors unless it's passed `--exitfirst`. But if it's too complicated to set up with fixtures maybe it's not worth it :)

JohnVillalovos: I think I figured it out!

JohnVillalovos:
> 
> 
> I like the idea so that people get a helpful message to guide them, because this can get quite complex for new contributors I think.
> 
> For me this (and `test_mro.py`) are kind of almost meta tests, so I wonder if we should move them together with `test_dists.py` or so (all the other modules in `objects/` really test functions on gitlab resources).

Moving them works for me. I had put them in objects because those were the only part of the code-base that they checked.

If we move them, where do you think we should move them.

nejch:
> > I like the idea so that people get a helpful message to guide them, because this can get quite complex for new contributors I think.
> > For me this (and `test_mro.py`) are kind of almost meta tests, so I wonder if we should move them together with `test_dists.py` or so (all the other modules in `objects/` really test functions on gitlab resources).
> 
> Moving them works for me. I had put them in `objects` because those were the only part of the code-base that they checked.
> 
> If we move them, where do you think we should move them.

I was thinking maybe https://github.com/python-gitlab/python-gitlab/tree/main/tests/smoke along with `test_dists.py` since these have more of a meta/linting style to me. WDYT? I think "sanity/" might be more appropriate than "smoke/" now, but that was my fault :grinning:

tests/unit/objects/test_ensure_type_hints.py
```diff
@@ -0,0 +1,84 @@
...
+    # Ignore any modules that we are ignoring in our pyproject.toml
+    excluded_modules = set()
+    with open("pyproject.toml", "r") as in_file:
+        pyproject = toml.load(in_file)
+    overrides = pyproject.get("tool", {}).get("mypy", {}).get("overrides", [])
+    for override in overrides:
+        if override.get("ignore_errors") is True:
+            for module in override.get("module", []):
+                if module.startswith("gitlab.v4.objects"):
+                    excluded_modules.add(module)
```
nejch, line 29: Would this work also?
```suggestion
    for override in overrides:
        if not override.get("ignore_errors"):
            continue
        for module in override.get("module", []):
            if module.startswith("gitlab.v4.objects"):
                excluded_modules.add(module)
```

tests/unit/objects/test_ensure_type_hints.py
```diff
@@ -0,0 +1,84 @@
...
+            f"must have defined a 'get' method with a return annotation of "
+            f"{obj_cls} but found {signature.return_annotation}\n"
+            f"Recommend adding the followinng method:\n"
+            f"def get(\n"
+            f"     self, id: Union[str, int], lazy: bool = False, **kwargs: Any\n"
+            f" ) -> {obj_cls.__name__}:\n"
+            f"     return cast({obj_cls.__name__}, super().get(id=id, lazy=lazy, "
+            f"**kwargs))\n"
+        )
+        assert obj_cls == signature.return_annotation, fail_message
```
nejch, line 84: Cool!

tests/unit/objects/test_ensure_type_hints.py
JohnVillalovos: Done!

JohnVillalovos:
> I was thinking maybe https://github.com/python-gitlab/python-gitlab/tree/main/tests/smoke along with `test_dists.py` since these have more of a meta/linting style to me. WDYT? I think "sanity/" might be more appropriate than "smoke/" now, but that was my fault 😀

I created a new `tests/meta/` directory and put `test_mro.py` and `test_ensure_type_hints.py` in there.

nejch (approved):

PR #1683, opened 2021-11-09
chore: add type-hints to setup.py and check with mypy

JohnVillalovos:


nejch:
Hehe I guess this won't hurt :) Just got a conflict now, sorry for the delay.

PR #1694, opened 2021-11-17
refactor: explicitly import gitlab.const values into top-level namespace

JohnVillalovos:
We are planning on adding enumerated constants into gitlab/const.py,
but if we do that than they will end up being added to the top-level
gitlab namespace. We really want to get users to start using
`gitlab.const.` to access the constant values in the future.

Explicitly add the current values defined in gitlab.const into the
top-level namespace and stop using the previous
'from gitlab.const import *' method.

### DIFF
--- a/gitlab/__init__.py
+++ b/gitlab/__init__.py
@@ -17,6 +17,7 @@
 """Wrapper for the GitLab API."""
 
 import warnings
+from typing import Any
 
 import gitlab.config  # noqa: F401
 from gitlab.__version__ import (  # noqa: F401
@@ -28,7 +29,22 @@ from gitlab.__version__ import (  # noqa: F401
     __version__,
 )
 from gitlab.client import Gitlab, GitlabList  # noqa: F401
-from gitlab.const import *  # noqa: F401,F403
 from gitlab.exceptions import *  # noqa: F401,F403
 
 warnings.filterwarnings("default", category=DeprecationWarning, module="^gitlab")
+
+
+# NOTE(jlvillal): We are deprecating access to the gitlab.const values which
+# were previously imported into this namespace by the
+# 'from gitlab.const import *' statement.
+def __getattr__(name: str) -> Any:
+    # Deprecate direct access to constants without namespace
+    if name in gitlab.const._DEPRECATED:
+        warnings.warn(
+            f"\nDirect access to 'gitlab.{name}' is deprecated and will be "
+            f"removed in a future major python-gitlab release. Please "
+            f"use 'gitlab.const.{name}' instead.",
+            DeprecationWarning,
+        )
+        return getattr(gitlab.const, name)
+    raise AttributeError(f"module {__name__} has no attribute {name}")
--- a/gitlab/const.py
+++ b/gitlab/const.py
@@ -17,6 +17,41 @@
 
 from gitlab.__version__ import __title__, __version__
 
+# NOTE(jlvillal): '_DEPRECATED' only affects users accessing constants via the
+# top-level gitlab.* namespace. See 'gitlab/__init__.py:__getattr__()' for the
+# consumer of '_DEPRECATED' For example 'x = gitlab.NO_ACCESS'.  We want users
+# to instead use constants by doing code like: gitlab.const.NO_ACCESS.
+_DEPRECATED = [
+    "DEFAULT_URL",
+    "DEVELOPER_ACCESS",
+    "GUEST_ACCESS",
+    "MAINTAINER_ACCESS",
+    "MINIMAL_ACCESS",
+    "NO_ACCESS",
+    "NOTIFICATION_LEVEL_CUSTOM",
+    "NOTIFICATION_LEVEL_DISABLED",
+    "NOTIFICATION_LEVEL_GLOBAL",
+    "NOTIFICATION_LEVEL_MENTION",
+    "NOTIFICATION_LEVEL_PARTICIPATING",
+    "NOTIFICATION_LEVEL_WATCH",
+    "OWNER_ACCESS",
+    "REPORTER_ACCESS",
+    "SEARCH_SCOPE_BLOBS",
+    "SEARCH_SCOPE_COMMITS",
+    "SEARCH_SCOPE_GLOBAL_SNIPPET_TITLES",
+    "SEARCH_SCOPE_ISSUES",
+    "SEARCH_SCOPE_MERGE_REQUESTS",
+    "SEARCH_SCOPE_MILESTONES",
+    "SEARCH_SCOPE_PROJECT_NOTES",
+    "SEARCH_SCOPE_PROJECTS",
+    "SEARCH_SCOPE_USERS",
+    "SEARCH_SCOPE_WIKI_BLOBS",
+    "USER_AGENT",
+    "VISIBILITY_INTERNAL",
+    "VISIBILITY_PRIVATE",
+    "VISIBILITY_PUBLIC",
+]
+
 DEFAULT_URL: str = "https://gitlab.com"
 
 NO_ACCESS: int = 0
--- a/docs/gl_objects/access_requests.rst
+++ b/docs/gl_objects/access_requests.rst
@@ -7,11 +7,11 @@ Users can request access to groups and projects.
 When access is granted the user should be given a numerical access level. The
 following constants are provided to represent the access levels:
 
-* ``gitlab.GUEST_ACCESS``: ``10``
-* ``gitlab.REPORTER_ACCESS``: ``20``
-* ``gitlab.DEVELOPER_ACCESS``: ``30``
-* ``gitlab.MAINTAINER_ACCESS``: ``40``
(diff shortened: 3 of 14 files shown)


gitlab/__init__.py
```diff
@@ -28,7 +28,36 @@
...
+    SEARCH_SCOPE_MILESTONES,
+    SEARCH_SCOPE_PROJECT_NOTES,
+    SEARCH_SCOPE_PROJECTS,
+    SEARCH_SCOPE_USERS,
+    SEARCH_SCOPE_WIKI_BLOBS,
+    USER_AGENT,
+    VISIBILITY_INTERNAL,
+    VISIBILITY_PRIVATE,
+    VISIBILITY_PUBLIC,
+)
```
nejch, line 60: I realize that star imports are awkward, but if this list keeps growing I wonder if they should just be controlled/auto-exported in const's `__all__` kind of how we do in https://github.com/python-gitlab/python-gitlab/blob/2cd15ac44d8a45fa2d0dcab80cc933e3871db7f8/gitlab/v4/objects/__init__.py#L78, but maybe with just specifics exported?

JohnVillalovos: So I want to block what gets pulled into the top level `gitlab` namespace. So that is why I am making it explicit.

If we do it in `gitlab.const` then they will still get pulled into the top-level `gitlab` namespace which is what we want to move away from. We want people to use `gitlab.const.FOO`, not `gitlab.FOO`.

The ultimate goal if are able to deprecate is that this will go away at some point.

Does that make sense?

JohnVillalovos: To clarify. The reason for this is so that when we add new constants in `gitlab.const` want to make the users access it via `gitlab.const`. So basically this will freeze the variables we bring in from `gitlab.const` into the top-level `gitlab` name-space. Since we would like to move people to access these const variables via `gitlab.const`.

Any new variables/Enums we add will need to be accessed via `gitlab.const`. If we just update the `__all__` value in `gitlab.const` then they will still be able to access them via the top level `gitlab` namespace.

JohnVillalovos: > I realize that star imports are awkward, but if this list keeps growing

That is the goal of this. To stop the list from growing. It won't get bigger than this and hopefully if we are able to deprecate it then it will go away in the future. This would be long-term like at least a year.

nejch: I think I got the idea I was just more wondering about explicitly importing vs explicitly exporting. If deprecating them frpm the main namespace is the goal it would also be easy with `__all__`, e.g. if you try this patch on your branch:

```diff
diff --git a/gitlab/__init__.py b/gitlab/__init__.py
index f1fb7da..e76ed6b 100644
--- a/gitlab/__init__.py
+++ b/gitlab/__init__.py
@@ -19,6 +19,7 @@
 import warnings
 
 import gitlab.config  # noqa: F401
+import gitlab.const
 from gitlab.__version__ import (  # noqa: F401
     __author__,
     __copyright__,
@@ -28,36 +29,18 @@ from gitlab.__version__ import (  # noqa: F401
     __version__,
 )
 from gitlab.client import Gitlab, GitlabList  # noqa: F401
-from gitlab.const import (  # noqa: F401
-    DEFAULT_URL,
-    DEVELOPER_ACCESS,
-    GUEST_ACCESS,
-    MAINTAINER_ACCESS,
-    MINIMAL_ACCESS,
-    NO_ACCESS,
-    NOTIFICATION_LEVEL_CUSTOM,
-    NOTIFICATION_LEVEL_DISABLED,
-    NOTIFICATION_LEVEL_GLOBAL,
-    NOTIFICATION_LEVEL_MENTION,
-    NOTIFICATION_LEVEL_PARTICIPATING,
-    NOTIFICATION_LEVEL_WATCH,
-    OWNER_ACCESS,
-    REPORTER_ACCESS,
-    SEARCH_SCOPE_BLOBS,
-    SEARCH_SCOPE_COMMITS,
-    SEARCH_SCOPE_GLOBAL_SNIPPET_TITLES,
-    SEARCH_SCOPE_ISSUES,
-    SEARCH_SCOPE_MERGE_REQUESTS,
-    SEARCH_SCOPE_MILESTONES,
-    SEARCH_SCOPE_PROJECT_NOTES,
-    SEARCH_SCOPE_PROJECTS,
-    SEARCH_SCOPE_USERS,
-    SEARCH_SCOPE_WIKI_BLOBS,
-    USER_AGENT,
-    VISIBILITY_INTERNAL,
-    VISIBILITY_PRIVATE,
-    VISIBILITY_PUBLIC,
-)
 from gitlab.exceptions import *  # noqa: F401,F403
 
 warnings.filterwarnings("default", category=DeprecationWarning, module="^gitlab")
+
+
+def __getattr__(name):
+    # Deprecate direct access to constants without namespace
+    if name in gitlab.const.__all__:
+        warnings.warn(
+            f"\nDirect access to 'gitlab.{name}' is deprecated and will be"
+            f"\nremoved in a future major python-gitlab release. Please"
+            f"\nuse 'gitlab.const.{name}' instead.",
+            DeprecationWarning)
+        return getattr(gitlab.const, name)
+    raise AttributeError(f"module {__name__} has no attribute {name}")
diff --git a/gitlab/const.py b/gitlab/const.py
index 12faf88..c45494b 100644
--- a/gitlab/const.py
+++ b/gitlab/const.py
@@ -56,3 +56,35 @@ SEARCH_SCOPE_GLOBAL_SNIPPET_TITLES: str = "snippet_titles"
 SEARCH_SCOPE_PROJECT_NOTES: str = "notes"
 
 USER_AGENT: str = f"{__title__}/{__version__}"
+
+
+__all__ = [
+    "DEFAULT_URL",
+    "DEVELOPER_ACCESS",
+    "GUEST_ACCESS",
+    "MAINTAINER_ACCESS",
+    "MINIMAL_ACCESS",
+    "NO_ACCESS",
+    "NOTIFICATION_LEVEL_CUSTOM",
+    "NOTIFICATION_LEVEL_DISABLED",
+    "NOTIFICATION_LEVEL_GLOBAL",
+    "NOTIFICATION_LEVEL_MENTION",
+    "NOTIFICATION_LEVEL_PARTICIPATING",
+    "NOTIFICATION_LEVEL_WATCH",
+    "OWNER_ACCESS",
+    "REPORTER_ACCESS",
+    "SEARCH_SCOPE_BLOBS",
+    "SEARCH_SCOPE_COMMITS",
+    "SEARCH_SCOPE_GLOBAL_SNIPPET_TITLES",
+    "SEARCH_SCOPE_ISSUES",
+    "SEARCH_SCOPE_MERGE_REQUESTS",
+    "SEARCH_SCOPE_MILESTONES",
+    "SEARCH_SCOPE_PROJECT_NOTES",
+    "SEARCH_SCOPE_PROJECTS",
+    "SEARCH_SCOPE_USERS",
+    "SEARCH_SCOPE_WIKI_BLOBS",
+    "USER_AGENT",
+    "VISIBILITY_INTERNAL",
+    "VISIBILITY_PRIVATE",
+    "VISIBILITY_PUBLIC",
+]

```

JohnVillalovos: @nejch Thanks. I did (basically) what you suggested. So should be good to go.

gitlab/const.py
```diff
@@ -17,6 +17,42 @@
 
 from gitlab.__version__ import __title__, __version__
 
+# NOTE(jlvillal): '__deprecated_all__' (our own custom name, not a Python
+# standard name) only affects users accessing constants via the top-level
+# namespace. See 'gitlab/__init__.py:__getattr__()' for the consumer of
+# '__deprecated_all__' For example 'x = gitlab.NO_ACCESS'.  We want users to
+# instead use constants by doing code like: gitlab.const.NO_ACCESS.
```
nejch, line 24: I'm sure a lot of people who mainly do scripting will always do `from gitlab.const import ...` into their own modules and access them directly, but agreed about polluting the top `gitlab.` module access. So I think the explicit namespace like that in examples makes sense but not everyone will do it :grinning:

gitlab/const.py
```diff
@@ -17,6 +17,42 @@
...
+__deprecated_all__ = [
```
nejch, line 25: Hmm I guess my idea for using `__all__` was to also potentially control `from gitlab.const import *` but if done like this, we then wouldn't use it like that. So in that case it can even just be `_DEPRECATED` and not mix it with the builtin magic?

JohnVillalovos: Done. Changed to _DEPRECATED.

If you want to prohibit consumers from doing `from gitlab.const import *` then I can go back to `__all__`.

But I guess if they want to do it we can just let them. I know I despise `from blah import *` but I guess I think we should let people do that if they want to. But yuck if they do! 🙂

gitlab/const.py
JohnVillalovos: I'm thinking nothing for me to do in regards to this comment. But re-open if I am wrong.

PR #1695, opened 2021-11-17
chore: add type-hints to remaining gitlab/v4/objects/*.py files

JohnVillalovos:
Fixed a bug in the auto-manager creation code if we added a type-hint for the `manager` attribute.

Add type-hints for the remainder of the code base.

Finally! 😊

### DIFF
--- a/gitlab/base.py
+++ b/gitlab/base.py
@@ -150,6 +150,10 @@ class RESTObject(object):
         # annotations. If an attribute is annotated as being a *Manager type
         # then we create the manager and assign it to the attribute.
         for attr, annotation in sorted(self.__annotations__.items()):
+            # We ignore creating a manager for the 'manager' attribute as that
+            # is done in the self.__init__() method
+            if attr in ("manager",):
+                continue
             if not isinstance(annotation, (type, str)):
                 continue
             if isinstance(annotation, type):
--- a/gitlab/v4/objects/epics.py
+++ b/gitlab/v4/objects/epics.py
@@ -1,3 +1,5 @@
+from typing import Any, cast, Dict, Optional, TYPE_CHECKING, Union
+
 from gitlab import exceptions as exc
 from gitlab import types
 from gitlab.base import RequiredOptional, RESTManager, RESTObject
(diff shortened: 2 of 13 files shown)


gitlab/v4/objects/epics.py
```diff
@@ -42,11 +44,17 @@ class GroupEpicManager(CRUDMixin, RESTManager):
...
 
+    def get(self, id: Union[str, int], lazy: bool = False, **kwargs: Any) -> GroupEpic:
+        return cast(GroupEpic, super().get(id=id, lazy=lazy, **kwargs))
+
 
 class GroupEpicIssue(ObjectDeleteMixin, SaveMixin, RESTObject):
     _id_attr = "epic_issue_id"
+    # Define type for 'manager' here So mypy won't complain about
+    # 'self.manager.update()' call in the 'save' method.
+    manager: "GroupEpicIssueManager"
```
nejch, line 55: We will hit this in a few more places it seems:

```
nejc@nuc:~/repos/python-gitlab$ grep -r 'self.manager.update' gitlab/v4
gitlab/v4/objects/labels.py:        server_data = self.manager.update(None, updated_data, **kwargs)
gitlab/v4/objects/labels.py:        server_data = self.manager.update(None, updated_data, **kwargs)
gitlab/v4/objects/epics.py:    # 'self.manager.update()' call in the 'save' method.
gitlab/v4/objects/epics.py:        self.manager.update(obj_id, updated_data, **kwargs)
```
Can we solve this with Protocols? Again, I haven't looked into it enough to give smart suggestions :grin: 
https://mypy.readthedocs.io/en/latest/more_types.html#advanced-uses-of-self-types

If we have to conditionally import `typing_extensions` I think it'd still be worth it if solves the issue.

JohnVillalovos: Protocols were added in Python 3.8. So I guess in another year we could do it after we deprecate Python 3.7. 😏

But to answer your question I am not sure if Protocols will or won't work. Since we are using a `Mixin` based design pattern I'm not sure.

gitlab/base.py
```diff
@@ -150,6 +150,10 @@ def _create_managers(self) -> None:
         # annotations. If an attribute is annotated as being a *Manager type
         # then we create the manager and assign it to the attribute.
         for attr, annotation in sorted(self.__annotations__.items()):
+            # We ignore creating a manager for the 'manager' attribute as that
+            # is done in the self.__init__() method
+            if attr in ("manager",):
+                continue
```
nejch, line 156: Just wondering is there's any specific reason this isn't just `if attr == "manager":`?

JohnVillalovos: For the future! Besides that no good reason.

gitlab/v4/objects/epics.py
nejch: I don't think 3.8 is strictly necessary as there is https://pypi.org/project/typing-extensions/, but agreed needs more exploration and I see you already added them 😁

gitlab/v4/objects/files.py
```diff
@@ -31,7 +36,11 @@ def decode(self) -> bytes:
         """
         return base64.b64decode(self.content)
 
-    def save(self, branch, commit_message, **kwargs):
+    # NOTE(jlvillal): Signature doesn't match SaveMixin.save() so ignore
+    # type error
+    def save(  # type: ignore
```
nejch, line 41: For future, I wonder if `@typing.overload` with all the possible signatures might help here. Disclaimer again this is just from reading I haven't tested that stuff 😀

Issue #1698, opened 2021-11-18
List merge requests using specific approver_ids regression


skycaptain:
## Description of the problem, including code/CLI snippet

There might be a possible regression of #1010 with the latest version. The issue is exactly the same:

```python
group = gl.groups.get(id=475)
group_mrs = group.mergerequests.list(
    all=True,
    state="opened",
    author_id=423,
    approver_ids=[423],
)
```

```log
[21:49:48] DEBUG    Starting new HTTPS connection (1): *redacted*:443                                connectionpool.py:971
           DEBUG    https://*redacted*:443 "GET /api/v4/user HTTP/1.1" 200 None                      connectionpool.py:452
[21:49:49] DEBUG    https://*redacted*:443 "GET /api/v4/groups/475 HTTP/1.1" 200 None                connectionpool.py:452
           DEBUG    https://*redacted*:443 "GET                                                      connectionpool.py:452
                    /api/v4/groups/475/merge_requests?state=opened&author_id=423&approver_ids=423                          
                    HTTP/1.1" 400 60                                                                                       
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/gitlab/exceptions.py", line 304, in wrapped_f
    return f(*args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/gitlab/mixins.py", line 242, in list
    obj = self.gitlab.http_list(path, **data)
  File "/usr/local/lib/python3.9/site-packages/gitlab/client.py", line 718, in http_list
    return list(GitlabList(self, url, query_data, **kwargs))
  File "/usr/local/lib/python3.9/site-packages/gitlab/client.py", line 881, in __init__
    self._query(url, query_data, **self._kwargs)
  File "/usr/local/lib/python3.9/site-packages/gitlab/client.py", line 891, in _query
    result = self._gl.http_request("get", url, query_data=query_data, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/gitlab/client.py", line 628, in http_request
    raise gitlab.exceptions.GitlabHttpError(
gitlab.exceptions.GitlabHttpError: 400: approver_ids should be an array, 'None' or 'Any'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "*redacted*.py", line 30, in <module>
    group_mrs = group.mergerequests.list(
  File "/usr/local/lib/python3.9/site-packages/gitlab/exceptions.py", line 306, in wrapped_f
[...]

Issue #806, opened 2019-06-20
Scope "bug"


romansvozil:
## Description of the problem, including code/CLI snippet
```py
    for job in project.jobs.list(scope=['success', 'failed']):
        print(job.attributes['status'])
```
It returns jobs of only last scope.
Does python-gitlab api supports multiple scope statement for finding jobs? Because gitlap api [yes](https://docs.gitlab.com/ee/api/jobs.html#list-project-jobs)

## Expected Behavior
something like:
success
success
success
failed
success

## Actual Behavior
failed
failed
failed

## Specifications

  - python-gitlab version:1.8.0
  - API version you are using (v3/v4): v4

EricTendian:
I was able to reproduce this. Turning on my logger to the debug level, it looks like this is the request that is sent:

```
https://gitlab.com:443 "GET /api/v4/projects/PROJECT_ID/pipelines/PIPELINE_ID/jobs?scope=success&scope=failed&page=1 HTTP/1.1" 200 1800
```

The proper request would be `GET /api/v4/projects/PROJECT_ID/pipelines/PIPELINE_ID/jobs?scope[]=success&scope[]=failed&page=1 HTTP/1.1` (note the `[]`)

I'm not confident this is the right way to fix it, but it seems like the `ListMixin` may need to be updated to properly handle the scope parameter.

duncz:
Was this issue ever resolved? I'm encountering the same thing. Or is there any workaround?

EricTendian:
@duncz don't know if it was resolved, I ended up working around this via not passing a scope but instead filtering the jobs I got in the response (a more expensive operation, but it works).

romansvozil:
@duncz there is the solution I used in my project:

```py
jobs = git_project.jobs.list(query_parameters={'scope[]': ['pending', 'running', 'failed', 'success']})
```

Should work.

JohnVillalovos:
Related issue: https://github.com/python-gitlab/python-gitlab/issues/1735

Issue #780, opened 2019-05-23
Can't create impersonation token from CLI


jgeorgeson:
## Description of the problem, including code/CLI snippet

Creation of user impersonation token fails from CLI with invalid scope error

```bash
$ gitlab -d -o json -g example user-impersonation-token create --user-id 168 --name test --scopes '["api"]'
```
It seems to be trying to pass the supplied value as a string instead of converting to an array (see the output in _Actual Behavior_). Have tried multiple variations of the scopes value such as `api`, `[api]`, `["api"]`, `"[api]"`, and `"['api']"`.

Using curl like this

```bash
$ curl -X POST -H "Content-type: application/json" -H "PRIVATE-TOKEN: 123412341234" https://gitlab.example.com/api/v4/users/2/impersonation_tokens -d '{"name": "test", "scopes": ["api"]}'
```
Successfully creates the token.

## Expected Behavior

Token is created.

## Actual Behavior

```bash
send: b'POST /api/v4/users/2/impersonation_tokens HTTP/1.1\r\nHost: gitlab.example.com\r\nUser-Agent: python-requests/2.21.0\r\nAccept-Encoding: gzip, deflate\r\nAccept: */*\r\nConnection: keep-alive\r\nPRIVATE-TOKEN: 1234123412341234\r\nContent-type: application/json\r\nContent-Length: 55\r\n\r\n'
send: b'{"user_id": "2", "name": "test", "scopes": "[\'api\']"}'
reply: 'HTTP/1.1 400 Bad Request\r\n'
DEBUG:urllib3.connectionpool:https://gitlab.example.com:443 "POST /api/v4/users/2/impersonation_tokens HTTP/1.1" 400 29
Impossible to create object (400: scopes is invalid)
header: Server header: Date header: Content-Type header: Content-Length header: Connection header: Cache-Control header: Vary header: X-Content-Type-Options header: X-Frame-Options header: X-Request-Id header: X-Runtime 
```

## Specifications

  - python-gitlab version: 1.8.0
  - API version you are using (v3/v4): v4
  - Gitlab server version (or gitlab.com): 11.4.4

ruzickap:
Any news about it? I have the same issue...

krzyslom:
Hi
I also struggle with a similar problem when using CLI - `gitlab project-deploy-token create`. Just like the author of the issue - I've tried multiple variations of the scopes value and using curl works.

python-gitlab 2.10.1
API v4
Gitlab 14.1.3

PR #1699, opened 2021-11-19
fix: use the [] after key names for array variables in `params`

JohnVillalovos:
1. If a value is of type ArrayAttribute then append '[]' to the name
   of the value.

This is step 3 in a series of steps of our goal to add full
support for the GitLab API data types[1]:
  * array
  * hash
  * array of hashes

Step one was: commit 5127b1594c00c7364e9af15e42d2e2f2d909449b
Step two was: commit a57334f1930752c70ea15847a39324fa94042460

Fixes: #1698
Related to https://github.com/python-gitlab/python-gitlab/issues/780
Closes https://github.com/python-gitlab/python-gitlab/issues/806

[1] https://docs.gitlab.com/ee/api/#encoding-api-parameters-of-array-and-hash-types

### DIFF
--- a/gitlab/mixins.py
+++ b/gitlab/mixins.py
@@ -238,7 +238,12 @@ class ListMixin(HeadMixin, _RestManagerBase):
             GitlabListError: If the server cannot perform the request
         """
 
-        data, _ = utils._transform_types(kwargs, self._types, transform_files=False)
+        data, _ = utils._transform_types(
+            data=kwargs,
+            custom_types=self._types,
+            transform_data=True,
+            transform_files=False,
+        )
 
         if self.gitlab.per_page:
             data.setdefault("per_page", self.gitlab.per_page)
@@ -303,7 +308,9 @@ class CreateMixin(_RestManagerBase):
             data = {}
 
         self._create_attrs.validate_attrs(data=data)
-        data, files = utils._transform_types(data, self._types)
+        data, files = utils._transform_types(
+            data=data, custom_types=self._types, transform_data=False
+        )
 
         # Handle specific URL for creation
         path = kwargs.pop("path", self.path)
@@ -370,7 +377,9 @@ class UpdateMixin(_RestManagerBase):
         if self._obj_cls is not None and self._obj_cls._id_attr is not None:
             excludes = [self._obj_cls._id_attr]
         self._update_attrs.validate_attrs(data=new_data, excludes=excludes)
-        new_data, files = utils._transform_types(new_data, self._types)
+        new_data, files = utils._transform_types(
+            data=new_data, custom_types=self._types, transform_data=False
+        )
 
         http_method = self._get_update_method()
         result = http_method(path, post_data=new_data, files=files, **kwargs)
--- a/gitlab/types.py
+++ b/gitlab/types.py
@@ -63,8 +63,8 @@ class GitlabAttribute:
     def set_from_cli(self, cli_value: Any) -> None:
         self._value = cli_value
 
-    def get_for_api(self) -> Any:
-        return self._value
+    def get_for_api(self, *, key: str) -> Tuple[str, Any]:
+        return (key, self._value)
 
 
 class _ListArrayAttribute(GitlabAttribute):
@@ -76,19 +76,27 @@ class _ListArrayAttribute(GitlabAttribute):
         else:
             self._value = [item.strip() for item in cli_value.split(",")]
 
-    def get_for_api(self) -> str:
+    def get_for_api(self, *, key: str) -> Tuple[str, str]:
         # Do not comma-split single value passed as string
         if isinstance(self._value, str):
-            return self._value
+            return (key, self._value)
 
         if TYPE_CHECKING:
             assert isinstance(self._value, list)
-        return ",".join([str(x) for x in self._value])
+        return (key, ",".join([str(x) for x in self._value]))
 
 
 class ArrayAttribute(_ListArrayAttribute):
     """To support `array` types as documented in
     https://docs.gitlab.com/ee/api/#array"""
+
+    def get_for_api(self, *, key: str) -> Tuple[str, Any]:
+        if isinstance(self._value, str):
+            return (f"{key}[]", self._value)
+
+        if TYPE_CHECKING:
+            assert isinstance(self._value, list)
+        return (f"{key}[]", self._value)
 
 
 class CommaSeparatedListAttribute(_ListArrayAttribute):
@@ -98,8 +106,8 @@ class CommaSeparatedListAttribute(_ListArrayAttribute):
 
 
 class LowercaseStringAttribute(GitlabAttribute):
-    def get_for_api(self) -> str:
-        return str(self._value).lower()
+    def get_for_api(self, *, key: str) -> Tuple[str, str]:
+        return (key, str(self._value).lower())
 
 
 class FileAttribute(GitlabAttribute):
--- a/gitlab/utils.py
+++ b/gitlab/utils.py
@@ -55,34 +55,53 @@ def response_content(
 
 
 def _transform_types(
-    data: Dict[str, Any], custom_types: dict, *, transform_files: Optional[bool] = True
+    data: Dict[str, Any],
+    custom_types: dict,
+    *,
+    transform_data: bool,
+    transform_files: Optional[bool] = True,
 ) -> Tuple[dict, dict]:
     """Copy the data dict with attributes that have custom types and transform them
     before being sent to the server.
 
-    If ``transform_files`` is ``True`` (default), also populates the ``files`` dict for
+    ``transform_files``: If ``True`` (default), also populates the ``files`` dict for
     FileAttribute types with tuples to prepare fields for requests' MultipartEncoder:
     https://toolbelt.readthedocs.io/en/latest/user.html#multipart-form-data-encoder
+
+    ``transform_data``: If ``True`` transforms the ``data`` dict with fields
+    suitable for encoding as query parameters for GitLab's API:
+    https://docs.gitlab.com/ee/api/#encoding-api-parameters-of-array-and-hash-types
 
     Returns:
         A tuple of the transformed data dict and files dict"""
 
     # Duplicate data to avoid messing with what the user sent us
     data = data.copy()
+    if not transform_files and not transform_data:
+        return data, {}
+
     files = {}
 
-    for attr_name, type_cls in custom_types.items():
+    for attr_name, attr_class in custom_types.items():
         if attr_name not in data:
             continue
 
-        type_obj = type_cls(data[attr_name])
+        gitlab_attribute = attr_class(data[attr_name])
 
-        # if the type if FileAttribute we need to pass the data as file
-        if transform_files and isinstance(type_obj, types.FileAttribute):
-            key = type_obj.get_file_name(attr_name)
+        # if the type is FileAttribute we need to pass the data as file
+        if isinstance(gitlab_attribute, types.FileAttribute) and transform_files:
+            key = gitlab_attribute.get_file_name(attr_name)
             files[attr_name] = (key, data.pop(attr_name))
-        else:
-            data[attr_name] = type_obj.get_for_api()
+            continue
+
+        if not transform_data:
+            continue
+
+        if isinstance(gitlab_attribute, types.GitlabAttribute):
+            key, value = gitlab_attribute.get_for_api(key=attr_name)
+            if key != attr_name:
+                del data[attr_name]
+            data[key] = value
 
     return data, files
 
@@ -94,6 +113,8 @@ def copy_dict(
 ) -> None:
     for k, v in src.items():
         if isinstance(v, dict):
+            # NOTE(jlvillal): This provides some support for the `hash` type
+            # https://docs.gitlab.com/ee/api/#hash
             # Transform dict values to new attributes. For example:
             # custom_attributes: {'foo', 'bar'} =>
             #   "custom_attributes['foo']": "bar"
--- a/tests/unit/mixins/test_mixin_methods.py
+++ b/tests/unit/mixins/test_mixin_methods.py
@@ -206,6 +206,25 @@ def test_list_mixin(gl):
 
 
 @responses.activate
+def test_list_mixin_with_attributes(gl):
+    class M(ListMixin, FakeManager):
+        _types = {"my_array": gl_types.ArrayAttribute}
+
+    url = "http://localhost/api/v4/tests"
+    responses.add(
+        method=responses.GET,
+        headers={},
+        url=url,
+        json=[],
+        status=200,
+        match=[responses.matchers.query_param_matcher({"my_array[]": ["1", "2", "3"]})],
+    )
+
+    mgr = M(gl)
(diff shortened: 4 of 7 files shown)


gitlab/mixins.py
```diff
@@ -231,7 +231,17 @@ def list(self, **kwargs: Any) -> Union[base.RESTObjectList, List[base.RESTObject
             for attr_name, type_cls in self._types.items():
                 if attr_name in data.keys():
                     type_obj = type_cls(data[attr_name])
-                    data[attr_name] = type_obj.get_for_api()
+                    if isinstance(type_obj, g_types.ArrayAttribute):
```
JohnVillalovos, line 234: I wonder if this checking should be done in some helper function where we pass in a dictionary and it returns back a List[Tuple[str, str]] data structure, which can then go to `requests.request`

I'm almost thinking we should defer all of this until we get to `gitlab/client.py:http_request()`. But the problem is that by that point we don't have the `self._types` structure 😟

JohnVillalovos:
So in my testing this does not break anything. It also allows `approver_ids` and `approved_by_ids` to work for **ONE** specified user (before it just failed) in listing merge requests. But it does not work for two or more, as it just returns the match for the first one.

I'm not a fan of the silent failure as that is likely worse than current situation where it dies.

JohnVillalovos:
@nejch  I think this works. Would like to get your feedback.

JohnVillalovos:
@nejch  This is ready for review now. Interested to hear your feedback. Thanks.

gitlab/mixins.py
JohnVillalovos: Done

gitlab/types.py
```diff
@@ -31,8 +31,43 @@ def set_from_cli(self, cli_value: Any) -> None:
...
+
+        if TYPE_CHECKING:
+            assert isinstance(self._value, list)
+        return [(f"{key}[]", str(value)) for value in self._value]
+
+
+class CsvStringAttribute(GitlabAttribute):
+    """For values which are sent to the server as a Comma Separated Values
+    (CSV) string.  We allow them to be specified as a list and we convert it
+    into a CSV"""
```
nejch, line 69: nitpick: since we changed `ListAttribute` to `ArrayAttribute`, could we now call this `CommaSeparatedListAttribute` to follow upstream API naming? CSV almost sounds closer to RFC 4180 / https://docs.python.org/3/library/csv.html which I associate more with file specs/parsing.

image

nejch (commented):
Thanks John. First just a nitpick question, I'll take a bit more time to take a look at this 😅

gitlab/types.py
JohnVillalovos: Good idea. Thanks and done!

nejch:
@JohnVillalovos do you think we could break this down into smaller PRs to bring down the diff so I can narrow down the focus? Just want to make sure I really get this right, as discussed offline we want to support array of hashes from the CLI later in a proper way.

I'm thinking:
1. > Rename ListAttribute to ArrayAttribute.

We can do this one first in a tiny PR, but `ListAttribute` -> `CommaSeparatedListAttribute` I think. This won't actually fix anything, it just more accurately describes what we are already doing iIf I'm not mistaken. Then we actually _add_ `ArrayAttribute` rather than rename, if I'm correct.

Then we could go from there. Sorry I'm just having some difficutly, the duplicated `get_as_tuple_list` methods currently confuse me so I'd like to look at that in isolation. If we know the signature we could maybe achieve that with overloading, or otherwise a Mixin that self-describes why it's done like that for a particular method style.

Thanks for your patience :bow:

JohnVillalovos:
> @JohnVillalovos do you think we could break this down into smaller PRs to bring down the diff so I can narrow down the focus? Just want to make sure I really get this right, as discussed offline we want to support array of hashes from the CLI later in a proper way.

Absolutely

> We can do this one first in a tiny PR, but `ListAttribute` -> `CommaSeparatedListAttribute` I think. 

Created a PR: https://github.com/python-gitlab/python-gitlab/pull/1858

nejch:
I finally had some more time to look at this in more detail! I have some thoughts to maybe simplify this.

We only need to do encoding when sending data as query parameters, not as a payload. So I think the encoding part is relevant for GET requests, as we can use json-encoded data as provided otherwise. I just tested this with `topics` on project creation, which is an array: 
`project = gl.projects.create({"name":"test-repo", "visibility": "private", "topics": ["gitlab", "python"]})`

GitLab's docs suggest this too:
> GET requests usually send a query string, while PUT or POST requests usually send the payload body

So https://docs.gitlab.com/ee/api/#encoding-api-parameters-of-array-and-hash-types is a bit misleading here because they don't use `application/json` as they do in the example above that section, and it should be possible for us to do that instead as long as we ensure it is sent as JSON and not as query params. See also https://github.com/python-gitlab/python-gitlab/issues/780#issuecomment-914993025 where an array is sent successfully via the API when created as a list. (Of course we still need to later have some basic support hash/array of hashes in order to parse them from the CLI).

And I'm thinking that complex data structures only make sense for POST/PUT requests, while arrays are still useful for GET for list filters, for example. So we only need to encode arrays properly for query params, IMO. This would really simplify things as all we need to do is transform e.g. `{"scopes": ["api", "read_user"}` into `{"scopes[]": ["api", "read_user"}`, because requests already does a urlencode there:
https://github.com/psf/requests/blob/8bce583b9547c7b82d44c8e97f37cf9a16cbe758/requests/models.py#L112

E.g.:
```python
>>> from urllib.parse import urlencode
>>> urlencode({"scopes": ["api", "read_user"]}, doseq=True)
'scopes=api&scopes=read_user'
>>> urlencode({"scopes[]": ["api", "read_user"]}, doseq=True)
'scopes%5B%5D=api&scopes%5B%5D=read_user'
```
I feel like we could avoid some of the conversions happening here if we do it like that. WDYT @JohnVillalovos?

nejch:
Should we try to revive this one @JohnVillalovos? Seems like an important fix. Probably fixes a few more issues down the line, will need to check those.

JohnVillalovos:
> Should we try to revive this one @JohnVillalovos? Seems like an important fix. Probably fixes a few more issues down the line, will need to check those.

Yeah, probably. Let me see if I get some time this weekend. Thanks for reminding me!

JohnVillalovos:
@nejch It is ready for review again. Thanks for your patience.

gitlab/utils.py
```diff
@@ -71,18 +71,22 @@ def _transform_types(
...
-            data[attr_name] = type_obj.get_for_api()
+        gitlab_attribute = attribute_class(data[attr_name])
+        if isinstance(gitlab_attribute, types.GitlabAttribute):
+            # if the type is FileAttribute we need to pass the data as file
+            if isinstance(gitlab_attribute, types.FileAttribute):
+                if transform_files:
+                    key = gitlab_attribute.get_file_name(attr_name)
+                    files[attr_name] = (key, data.pop(attr_name))
+                continue
+            key, value = gitlab_attribute.get_for_api(key=attr_name)
```
nejch, line 86: Haven't explored at all why, but pylance isn't happy here. :shrug: I guess it doesn't matter.
image

gitlab/types.py
```diff
@@ -63,8 +63,8 @@ def get(self) -> Any:
     def set_from_cli(self, cli_value: Any) -> None:
         self._value = cli_value
 
-    def get_for_api(self) -> Any:
-        return self._value
+    def get_for_api(self, *, key: str) -> Tuple[str, Any]:
+        return (key, self._value)
```
nejch, line 67: I've been thinking even earlier with the FileAttribute that it's kind of a roundabout way that we have to supply the key manually when we actually define it along with the type. We could probably refactor this later to just have the key on instantiation or so, and leave the method signatures as before. Then we also don't need to have conversations about keyword arguments :rofl:

tests/unit/mixins/test_mixin_methods.py
```diff
@@ -205,6 +205,25 @@ class M(ListMixin, FakeManager):
     assert responses.assert_call_count(url, 2) is True
 
 
+@responses.activate
+def test_list_mixin_with_attributes(gl):
+    class M(ListMixin, FakeManager):
+        _types = {"my_array": gl_types.ArrayAttribute}
+
+    url = "http://localhost/api/v4/tests?my_array%5B%5D=1&my_array%5B%5D=2&my_array%5B%5D=3"
```
nejch, line 213: I checked locally and the query params are not required here. They are checked by the `query_param_matcher` already. E.g. test it by changing the matcher to `my_error[]`:

```log
E           requests.exceptions.ConnectionError: Connection refused by Responses - the call doesn't match any registered mock.
E           
E           Request: 
E           - GET http://localhost/api/v4/tests?my_array%5B%5D=1&my_array%5B%5D=2&my_array%5B%5D=3
E           
E           Available matches:
E           - GET http://localhost/api/v4/tests Parameters do not match. {my_array[]: [1, 2, 3]} doesn't match {my_error[]: [1, 2, 3]}
```


```suggestion
    url = "http://localhost/api/v4/tests"
```

nejch (commented):
~~Thanks @JohnVillalovos looks super simple now. I have some concerns but I think I'll just merge this and use it to track and open a refactor based on my comments.~~

I have to take that back - seems like this also transforms the key for POST/PUT payloads which shouldn't happen. This maybe needs a test with `create()` to ensure we don't use `key[]` in payloads.

gitlab/utils.py
```diff
@@ -94,6 +98,8 @@ def copy_dict(
 ) -> None:
     for k, v in src.items():
         if isinstance(v, dict):
+            # NOTE(jlvillal): This provides some support for the `hash` type
```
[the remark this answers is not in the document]
JohnVillalovos: It is from my time working on OpenStack. As it is easier to read a line and see the nick of the person than to attempt to do a `git blame`.

https://github.com/openstack/ironic/blob/master/devstack/lib/ironic#L3188

Search for "NOTE(" and there are a lot of instances of it.

JohnVillalovos:
> I have to take that back - seems like this also transforms the key for POST/PUT payloads which shouldn't happen. This maybe needs a test with `create()` to ensure we don't use `key[]` in payloads.

I think we need to transform "parameters'. Whether it is a GET/POST/PUT. Now handling values in the `data` portion is a different thing we need to figure out.

nejch:
> > I have to take that back - seems like this also transforms the key for POST/PUT payloads which shouldn't happen. This maybe needs a test with `create()` to ensure we don't use `key[]` in payloads.
> 
> I think we need to transform "parameters'. Whether it is a GET/POST/PUT. Now handling values in the `data` portion is a different thing we need to figure out.

Yes, that's right :) sorry for the confusion. payload = no transform, query params = transform.

gitlab/utils.py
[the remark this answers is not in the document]
nejch: Ah ok I see. I guess I didn't know why the person adding the note was crucial for the comments :D (since vscode & friends have blame-on-hover also)

tests/unit/mixins/test_mixin_methods.py
JohnVillalovos: Nice. Thanks!

gitlab/utils.py
```diff
@@ -71,18 +71,22 @@ def _transform_types(
     data = data.copy()
     files = {}
 
-    for attr_name, type_cls in custom_types.items():
+    for attr_name, attribute_class in custom_types.items():
```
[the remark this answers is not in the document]
JohnVillalovos: Done.

nejch:
@JohnVillalovos I think this doesn't work quite right yet. Can you try this test in this patch?

```diff
diff --git a/tests/unit/mixins/test_mixin_methods.py b/tests/unit/mixins/test_mixin_methods.py
index 69cdb78..e8870b9 100644
--- a/tests/unit/mixins/test_mixin_methods.py
+++ b/tests/unit/mixins/test_mixin_methods.py
@@ -314,6 +314,26 @@ def test_create_mixin_custom_path(gl):
     assert responses.assert_call_count(url, 1) is True
 
 
+
+@responses.activate
+def test_create_mixin_with_attributes(gl):
+    class M(CreateMixin, FakeManager):
+        _types = {"my_array": gl_types.ArrayAttribute}
+
+    url = "http://localhost/api/v4/tests"
+    responses.add(
+        method=responses.POST,
+        headers={},
+        url=url,
+        json={},
+        status=200,
+        match=[responses.matchers.json_params_matcher({"my_array": ["1", "2", "3"]})],
+    )
+
+    mgr = M(gl)
+    mgr.create({"my_array": [1, 2, 3]})
+
+
 def test_update_mixin_missing_attrs(gl):
     class M(UpdateMixin, FakeManager):
         _update_attrs = gl_types.RequiredOptional(

```

The reason we might want this is, for example `scopes` as an array in create mixins for #780.

JohnVillalovos:
> @JohnVillalovos I think this doesn't work quite right yet. Can you try this test in this patch?

Thanks @nejch  I totally missed that the 'data` was not being sent as `params` in the POST/PUT methods 😔

Updated the code to only modify the data for the `list()` method.

gitlab/utils.py
```diff
@@ -55,7 +55,11 @@ def response_content(
 
 
 def _transform_types(
-    data: Dict[str, Any], custom_types: dict, *, transform_files: Optional[bool] = True
+    data: Dict[str, Any],
+    custom_types: dict,
+    *,
+    is_params: bool,
```
nejch, line 61: Would it make sense to have this as `transform_data` since we already have `data` in there? Just a thought for a follow-up, not needed now. Then it gives us a nice `if not transform_data and not transform_files:`.

It'd maybe also be cool to have this with a default value just like `transform_files` so we only pass it around when we have to override. But also non-blocking.

gitlab/utils.py
```diff
@@ -64,25 +68,38 @@ def _transform_types(
     FileAttribute types with tuples to prepare fields for requests' MultipartEncoder:
     https://toolbelt.readthedocs.io/en/latest/user.html#multipart-form-data-encoder
 
+    ``is_params``: data will be sent to ``requests`` in the ``params`` field.
```
nejch, line 71: I think the point here is that we prepare the data for that, not that we actually send it to params, right?
E.g. looking at the sentence above it and my comment before, it might go something like:
```suggestion
    If ``transform_data`` is ``True`` (default), also transforms the ``data`` dict 
    with fields suitable for encoding as query parameters for GitLab's API:
    https://docs.gitlab.com/ee/api/#encoding-api-parameters-of-array-and-hash-types
```

gitlab/utils.py
JohnVillalovos: > Would it make sense to have this as `transform_data` since we already have `data` in there? Just a thought for a follow-up, not needed now. Then it gives us a nice `if not transform_data and not transform_files:`.

Sure.

> It'd maybe also be cool to have this with a default value just like `transform_files` so we only pass it around when we have to override. But also non-blocking.

I don't like the idea of a default value. IMHO, default values are for things whose value will be the default for the large majority of the cases. I don't feel like this one is that case.

gitlab/utils.py
JohnVillalovos: Done

PR #1701, opened 2021-11-20
chore: correct test_groups.py test

JohnVillalovos:
The test was checking twice if the same group3 was not in the returned
list.  Should have been checking for group3 and group4.

Also added a test that only skipped one group and checked that the
group was not in the returned list and a non-skipped group was in the
list.


nejch:
Oops, that looks like I added that by mistake, thanks :)

PR #1702, opened 2021-11-20
chore: attempt to be more informative for missing attributes

### LINKED ISSUE #1138: `ProjectPipeline` from `.list(as_list=False)` has fewer information than the one from `.get(id)`
```
project = gl.projects.get('...')
all_pipelines = project.pipelines.list(as_list=False)
pipelines = itertools.islice(all_pipelines, search_limit)
print(pipelines[0].duration)
```

## Expected Behavior
Duration of the pipeline is returned.

## Actual Behavior
`AttributeError` is raised. Same happens with many other attributes. Only when using `refresh()` on a pipeline, the object has all of the expected information.

Objects fetched via `project.pipelines.get(id)` do not suffer from this problem.

```
KeyError                                  Traceback (most recent call last)
~/.virtualenvs/taf/lib/python3.7/site-packages/gitlab/base.py in __getattr__(self, name)
     55         try:
---> 56             return self.__dict__['_updated_attrs'][name]
     57         except KeyError:

KeyError: 'duration'

During handling of the above exception, another exception occurred:

KeyError                                  Traceback (most recent call last)
~/.virtualenvs/taf/lib/python3.7/site-packages/gitlab/base.py in __getattr__(self, name)
     58             try:
---> 59                 value = self.__dict__['_attrs'][name]
     60

KeyError: 'duration'

During handling of the above exception, another exception occurred:

KeyError                                  Traceback (most recent call last)
~/.virtualenvs/taf/lib/python3.7/site-packages/gitlab/base.py in __getattr__(self, name)
     78                 try:
---> 79                     return self.__dict__['_parent_attrs'][name]
     80                 except KeyError:

KeyError: 'duration'

During handling of the above exception, another exception occurred:

AttributeError                            Traceback (most recent call last)
<ipython-input-25-919ee3634190> in <module>
----> 1 gather_data(proj, 200, branch_name='master')

~/taf/taf.tools.pipeline_duration/app.py in gather_data(project, search_limit, branch_name, passed_only)
     40     return durations

~/.virtualenvs/taf/lib/python3.7/site-packages/gitlab/base.py in __getattr__(self, name)
     79                     return self.__dict__['_parent_attrs'][name]
     80                 except KeyError:
---> 81                     raise AttributeError(name)
     82
     83     def __setattr__(self, name, value):

AttributeError: duration
```

## Specifications

  - python-gitlab version: 1.8.0
  - Gitlab server version (or gitlab.com): 12.7.6

JohnVillalovos:
A commonly reported issue from users on Gitter is that they get an
AttributeError for an attribute that should be present. This is often
caused due to the fact that they used the `list()` method to retrieve
the object and objects retrieved this way often only have a subset of
the full data.

Add more details in the AttributeError message that explains the
situation to users. This will hopefully allow them to resolve the
issue.

Closes #1138

### DIFF
--- a/gitlab/base.py
+++ b/gitlab/base.py
@@ -16,9 +16,11 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import importlib
+import textwrap
 from types import ModuleType
 from typing import Any, Dict, Iterable, NamedTuple, Optional, Tuple, Type
 
+import gitlab
 from gitlab import types as g_types
 from gitlab.exceptions import GitlabParsingError
 
@@ -30,6 +32,12 @@ __all__ = [
     "RESTObjectList",
     "RESTManager",
 ]
+
+
+_URL_ATTRIBUTE_ERROR = (
+    f"https://python-gitlab.readthedocs.io/en/{gitlab.__version__}/"
+    f"faq.html#attribute-error-list"
+)
 
 
 class RESTObject(object):
@@ -45,13 +53,20 @@ class RESTObject(object):
 
     _id_attr: Optional[str] = "id"
     _attrs: Dict[str, Any]
+    _created_from_list: bool  # Indicates if object was created from a list() action
     _module: ModuleType
     _parent_attrs: Dict[str, Any]
     _short_print_attr: Optional[str] = None
     _updated_attrs: Dict[str, Any]
     manager: "RESTManager"
 
-    def __init__(self, manager: "RESTManager", attrs: Dict[str, Any]) -> None:
+    def __init__(
+        self,
+        manager: "RESTManager",
+        attrs: Dict[str, Any],
+        *,
+        created_from_list: bool = False,
+    ) -> None:
         if not isinstance(attrs, dict):
             raise GitlabParsingError(
                 "Attempted to initialize RESTObject with a non-dictionary value: "
@@ -64,6 +79,7 @@ class RESTObject(object):
                 "_attrs": attrs,
                 "_updated_attrs": {},
                 "_module": importlib.import_module(self.__module__),
+                "_created_from_list": created_from_list,
             }
         )
         self.__dict__["_parent_attrs"] = self.manager.parent_attrs
@@ -106,8 +122,22 @@ class RESTObject(object):
             except KeyError:
                 try:
                     return self.__dict__["_parent_attrs"][name]
-                except KeyError:
-                    raise AttributeError(name)
+                except KeyError as exc:
+                    message = (
+                        f"{type(self).__name__!r} object has no attribute {name!r}"
+                    )
+                    if self._created_from_list:
+                        message = (
+                            f"{message}\n\n"
+                            + textwrap.fill(
+                                f"{self.__class__!r} was created via a list() call and "
+                                f"only a subset of the data may be present. To ensure "
+                                f"all data is present get the object using a "
+                                f"get(object.id) call. For more details, see:"
+                            )
+                            + f"\n\n{_URL_ATTRIBUTE_ERROR}"
+                        )
+                    raise AttributeError(message) from exc
 
     def __setattr__(self, name: str, value: Any) -> None:
         self.__dict__["_updated_attrs"][name] = value
@@ -229,7 +259,7 @@ class RESTObjectList(object):
 
     def next(self) -> RESTObject:
         data = self._list.next()
-        return self._obj_cls(self.manager, data)
+        return self._obj_cls(self.manager, data, created_from_list=True)
 
     @property
     def current_page(self) -> int:
--- a/tests/unit/test_base.py
+++ b/tests/unit/test_base.py
@@ -90,6 +90,30 @@ class TestRESTObject:
         with pytest.raises(gitlab.exceptions.GitlabParsingError):
             FakeObject(fake_manager, ["a", "list", "fails"])
 
+    def test_missing_attribute_does_not_raise_custom(self, fake_gitlab, fake_manager):
+        """Ensure a missing attribute does not raise our custom error message
+        if the RESTObject was not created from a list"""
+        obj = FakeObject(manager=fake_manager, attrs={"foo": "bar"})
+        with pytest.raises(AttributeError) as excinfo:
+            obj.missing_attribute
+        exc_str = str(excinfo.value)
+        assert "missing_attribute" in exc_str
+        assert "was created via a list()" not in exc_str
+        assert base._URL_ATTRIBUTE_ERROR not in exc_str
+
+    def test_missing_attribute_from_list_raises_custom(self, fake_gitlab, fake_manager):
+        """Ensure a missing attribute raises our custom error message if the
+        RESTObject was created from a list"""
+        obj = FakeObject(
+            manager=fake_manager, attrs={"foo": "bar"}, created_from_list=True
+        )
+        with pytest.raises(AttributeError) as excinfo:
+            obj.missing_attribute
+        exc_str = str(excinfo.value)
+        assert "missing_attribute" in exc_str
+        assert "was created via a list()" in exc_str
+        assert base._URL_ATTRIBUTE_ERROR in exc_str
+
     def test_picklability(self, fake_manager):
         obj = FakeObject(fake_manager, {"foo": "bar"})
         original_obj_module = obj._module
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -16,6 +16,18 @@ I cannot edit the merge request / issue I've just retrieved
     See the :ref:`merge requests example <merge_requests_examples>` and the
     :ref:`issues examples <issues_examples>`.
 
+.. _attribute_error_list:
+
+I get an ``AttributeError`` when accessing attributes of an object retrieved via a ``list()`` call.
+    Fetching a list of objects, doesn’t always include all attributes in the
+    objects. To retrieve an object with all attributes use a ``get()`` call.
+
+    Example with projects::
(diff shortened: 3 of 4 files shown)


nejch:
This will also close https://github.com/python-gitlab/python-gitlab/issues/1138.

tests/unit/test_base.py
```diff
@@ -90,6 +90,30 @@ def test_instantiate_non_dict(self, fake_gitlab, fake_manager):
...
+        )
+        with pytest.raises(AttributeError) as excinfo:
+            obj.missing_attribute
+        assert "missing_attribute" in str(excinfo.value)
+        assert "was created via a list()" in str(excinfo.value)
+        assert (
+            "https://python-gitlab.readthedocs.io/en/stable/faq.html"
+            "#attribute-error-list"
+        ) in str(excinfo.value)
+
```
nejch, line 116: IMO if we split this into 2 exact tests we make sure we hit all asserts and can skip some comments.

```suggestion
    def test_missing_attribute_does_not_raise_custom(self, fake_gitlab, fake_manager):
        obj = FakeObject(manager=fake_manager, attrs={"foo": "bar"})
        with pytest.raises(AttributeError) as excinfo:
            obj.missing_attribute
        assert str(excinfo.value) == "missing_attribute"
        assert "was created via a list()" not in str(excinfo.value)

    def test_missing_attribute_from_list_raises_custom(self, fake_gitlab, fake_manager):
        obj = FakeObject(
            manager=fake_manager, attrs={"foo": "bar"}, list_created_object=True
        )
        with pytest.raises(AttributeError) as excinfo:
            obj.missing_attribute
        assert "missing_attribute" in str(excinfo.value)
        assert "was created via a list()" in str(excinfo.value)
```

Also should we omit the URL in case this gets moved around? I did in the suggestion. Otherwise maybe some constant/template reused from production code might help.

gitlab/base.py
```diff
@@ -45,13 +45,19 @@ class RESTObject(object):
 
     _id_attr: Optional[str] = "id"
     _attrs: Dict[str, Any]
+    _list_created_object: bool  # Indicates if object was created from a list() action
```
nejch, line 48: just a nitpick: to be concise (and similar to other existing params like `_from_parent_attrs`) would this also work? It would seem quite clear to me already.

```suggestion
    _from_list: bool
```

nejch (changes requested):
Thanks @JohnVillalovos, just a few small notes from me.

I was thinking something similar for missing methods in e.g. `Issue` vs `ProjectIssue` to guide users towards creating the right objects or even give a `make_editable()` method for those to convert them. So I really like it and would maybe build on this later as well. I think there might also be a few other cases where this could be extended later, such as after running save/update and so on.

gitlab/base.py
```diff
@@ -225,7 +243,7 @@ def __next__(self) -> RESTObject:
 
     def next(self) -> RESTObject:
         data = self._list.next()
-        return self._obj_cls(self.manager, data)
+        return self._obj_cls(manager=self.manager, attrs=data, list_created_object=True)
```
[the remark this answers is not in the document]
JohnVillalovos: Okay. I can remove that for now. I am a fan of calling things with keyword arguments as in my opinion it helps reduce the cognitive load of readers of the code (basically they don't have to think as hard to understand the code). So I was just fixing things up as I went along.

tests/unit/test_base.py
JohnVillalovos: I like that idea to split up the tests. I will do that.

I would like to keep the URL part. I use this program called `shellcheck` and one of the things I love about it is that it gives me URLs for errors. Now maybe we figure out a best URL to put stuff, but having the URL so someone can just click it to find information on how to resolve their issue is a big win in my opinion.

An example of what it does:
```
In foo.sh line 5:
echo $1
     ^-- SC2086: Double quote to prevent globbing and word splitting.

Did you mean:
echo "$1"

For more information:
  https://www.shellcheck.net/wiki/SC2086 -- Double quote to prevent globbing ...
```

For a first step let me put the template/URL of the message into the code and that way not duplicating as much.

gitlab/base.py
JohnVillalovos: I'm not sure. I feel like `_list_created_object` reduces the cognitive load for readers of the code more than `_from_list`. Also I like the comment to help people understand it, like me two months from now 😏

JohnVillalovos:
> Thanks @JohnVillalovos, just a few small notes from me.
> 
> I was thinking something similar for missing methods in e.g. `Issue` vs `ProjectIssue` to guide users towards creating the right objects or even give a `make_editable()` method for those to convert them. So I really like it and would maybe build on this later as well. I think there might also be a few other cases where this could be extended later, such as after running save/update and so on.

Thanks for the review! Will work on updating the PR.

I like the idea of catching the common issues that users of the library encounter and trying to point them to how to fix it so they file less issues 😊

tests/unit/test_base.py
nejch: Ah yes sorry I just meant in the test. I'd definitely keep it in the exception message, it's really helpful. So if it's templated that works for me!

gitlab/base.py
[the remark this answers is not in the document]
JohnVillalovos: Removed

tests/unit/test_base.py
JohnVillalovos: Done

gitlab/base.py
JohnVillalovos: Since you said it was a "nitpick" I left it as is. If you feel more strongly than that I can change it.

gitlab/base.py
```diff
@@ -32,6 +32,11 @@
 ]
 
 
+_URL_ATTRIBUTE_ERROR = (
+    "https://python-gitlab.readthedocs.io/en/stable/faq.html#attribute-error-list"
```
nejch, line 36: Just thinking, maybe we could use a permalink to make sure the user gets the exact link? Sometimes RTD is a bit odd about updating versions or maybe I missed something. I just checked and had to manually change the branch from master to main as well.

```suggestion
    f"https://python-gitlab.readthedocs.io/en/v{__version__}/faq.html#attribute-error-list"
```
(will need an extra import though)

gitlab/base.py
nejch: Ok :) Since we are already in `RESTObject` that's why I was thinking it was clear. Maybe `_created_from_list` would be somewhere in the middle

gitlab/base.py
```diff
@@ -107,7 +120,16 @@ def __getattr__(self, name: str) -> Any:
...
+                    message = name
+                    if self._list_created_object:
+                        message = (
+                            f"Unable to find attribute {name!r}. This object "
+                            f"({self.__class__}) was created via a list() call and "
+                            f"only a subset of the data may be present. To ensure all "
+                            f"data is present get the object using a get(object.id) "
+                            f"call. See {_URL_ATTRIBUTE_ERROR} for details."
+                        )
+                    raise AttributeError(message)
```
nejch, line 132: I tried this locally and I think we have a chance to improve the re-raises here and the message.
1. use the standard object AttributeError message `f"{type(self).__name__!r} object has no attribute {name!r}"`
2. raise `from` to clearly chain exceptions (but this was already missing before, just a nice to have)
3. format this a bit so it stands out for the user (otherwise it's just a long string). Maybe you'll find a better way, just an idea below.

Something like (sorry github wont let me add suggestions on deleted lines):

```python
            except KeyError:
                try:
                    return self.__dict__["_parent_attrs"][name]
                except KeyError as e:
                    message = f"{type(self).__name__!r} object has no attribute {name!r}"
                    if self._list_created_object:
                        message = (
                            f"{message}\n\n"
                            f"({self.__class__!r}) was created via a list() call\n"
                            f"and only a subset of the data may be present. To ensure\n"
                            f"all data is present get the object using a get(object.id)\n"
                            f"call. For more details, see:\n\n{_URL_ATTRIBUTE_ERROR}"
                        )
                    raise AttributeError(message) from e
```

This gives me:
```
...
During handling of the above exception, another exception occurred:

  File "/home/nejc/repos/python-gitlab/gitlab/base.py", line 121, in __getattr__
    return self.__dict__["_parent_attrs"][name]
KeyError: 'bla'

The above exception was the direct cause of the following exception:

  File "/home/nejc/repos/python-gitlab/gitlab/base.py", line 132, in __getattr__
    raise AttributeError(message) from e
AttributeError: 'Project' object has no attribute 'bla'

(<class 'gitlab.v4.objects.projects.Project'>) was created via a list() call
and only a subset of the data may be present. To ensure
all data is present get the object using a get(object.id)
call. For more details, see:

https://python-gitlab.readthedocs.io/en/stable/faq.html#attribute-error-list
>>> 
```

JohnVillalovos: Done. I did use the `textwrap` library for wrapping it to 70 characters.

gitlab/base.py

gitlab/base.py
```diff
@@ -45,13 +53,20 @@ class RESTObject(object):
...
     _short_print_attr: Optional[str] = None
     _updated_attrs: Dict[str, Any]
     manager: "RESTManager"
 
-    def __init__(self, manager: "RESTManager", attrs: Dict[str, Any]) -> None:
+    def __init__(
+        self,
+        manager: "RESTManager",
+        attrs: Dict[str, Any],
+        *,
```
nejch, line 67: We should maybe do a follow-up issue for this to discuss if we want to enforce/have this across the codebase, but probably not for user-facing code.

PR #1707, opened 2021-11-24
chore: remove duplicate/no-op tests from meta/test_ensure_type_hints

JohnVillalovos:
Before we were generating 725 tests for the
meta/test_ensure_type_hints.py tests.  Which isn't a huge concern as
it was fairly fast. But when we had a failure we would usually get two
failures for each problem as the same test was being run multiple
times.

Changed it so that:
  1. Don't add tests that are not for *Manager classes
  2. Use a set so that we don't have duplicate tests.

After doing that our generated test count in
meta/test_ensure_type_hints.py went from 725 to 178 tests.

Additionally removed the parsing of `pyproject.toml` to generate files
to ignore as we have finished adding type-hints to all files in
gitlab/v4/objects/. This also means we no longer use the toml library
so remove installation of `types-toml`.

To determine the test count the following command was run:
  $ tox -e py39 -- -k test_ensure_type_hints

### DIFF
--- a/tests/meta/test_ensure_type_hints.py
+++ b/tests/meta/test_ensure_type_hints.py
@@ -8,7 +8,6 @@ import inspect
 from typing import Tuple, Type
 
 import _pytest
-import toml
 
 import gitlab.mixins
 import gitlab.v4.objects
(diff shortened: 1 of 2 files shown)


tests/meta/test_ensure_type_hints.py
```diff
@@ -30,7 +30,7 @@ def pytest_generate_tests(metafunc: _pytest.python.Metafunc) -> None:
             if module.startswith("gitlab.v4.objects"):
                 excluded_modules.add(module)
```
nejch, line 31: This might be unrelated here but for more speed-up we can also remove the entire `toml` parsing section here now I think (I can't comment higher up in the diff).

tests/meta/test_ensure_type_hints.py
```diff
@@ -49,9 +49,10 @@ def pytest_generate_tests(metafunc: _pytest.python.Metafunc) -> None:
             if module_name == "gitlab.base":
                 continue
 
-            class_info_list.append((class_name, class_value))
-
-    metafunc.parametrize("class_info", class_info_list)
+            if not class_name.endswith("Manager"):
+                continue
+            class_info_set.add((class_name, class_value))
+    metafunc.parametrize("class_info", class_info_set)
```
nejch, line 55: Think we lost a few newlines here from before :D
```suggestion
            if not class_name.endswith("Manager"):
                continue

            class_info_set.add((class_name, class_value))

    metafunc.parametrize("class_info", class_info_set)
```

nejch (commented):
Didn't even realize before! Just a small note from me.

tests/meta/test_ensure_type_hints.py
JohnVillalovos: True. Good point. I'll do that too.

tests/meta/test_ensure_type_hints.py
JohnVillalovos: Done

PR #1709, opened 2021-11-24
docs: only use type annotations for documentation

nejch:
We probably still have a lot of incorrect types in the docstrings and this should just take care of it in the future.

| Before | After |
|--------|------|
| image | image |

Clashes with the internal extension though so needs more work

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -42,6 +42,7 @@ extensions = [
     "sphinx.ext.autodoc",
     "sphinx.ext.autosummary",
     "ext.docstrings",
+    "sphinx_autodoc_typehints",  # must be loaded after napoleon modules
     "sphinxcontrib.autoprogram",
 ]
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -42,9 +42,10 @@ extensions = [
     "sphinx.ext.autodoc",
     "sphinx.ext.autosummary",
     "ext.docstrings",
-    "sphinx_autodoc_typehints",  # must be loaded after napoleon modules
     "sphinxcontrib.autoprogram",
 ]
+
+autodoc_typehints = "both"
 
 # Add any paths that contain templates here, relative to this directory.
 templates_path = ["_templates"]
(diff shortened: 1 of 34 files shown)


JohnVillalovos:
Looks cool! Thanks for this. I'll review once you mark it ready for review.

nejch:
Turns out I was just using an outdated external module but it's now part of `autodoc` so it was just a matter of adding a flag. This looks a bit more like the original now, but pulls typehints from code. Should be ready now.

docs/conf.py
```diff
@@ -42,6 +42,7 @@
     "sphinx.ext.autodoc",
     "sphinx.ext.autosummary",
     "ext.docstrings",
+    "sphinx_autodoc_typehints",  # must be loaded after napoleon modules
```
JohnVillalovos: What is a "napoleon" module? Is it one of the ones above?

gitlab/client.py
```diff
@@ -269,7 +269,7 @@ def version(self) -> Tuple[str, str]:
         object.
 
         Returns:
-            tuple (str, str): The server version and server revision.
+            tuple: The server version and server revision.
                               ('unknown', 'unknwown') if the server doesn't
```
JohnVillalovos, line 273: While in here, maybe do:  s/unknwown/unknown/  to fix the mis-spelling.

JohnVillalovos (approved):
I'm okay to merge it as is.

Thanks for doing this.

docs/conf.py
nejch: yes, `ext.docstrings` was extending google's modules but this is no longer needed now as I didn't need to add the extra module in the end.

gitlab/client.py
nejch: Oh, thanks. Even the `tuple` here is not needed 🤔  seems like my search/replace missed a few of these, I'll check again.

nejch: https://github.com/python-gitlab/python-gitlab/pull/1709/commits/79e785e765f4219fe6001ef7044235b82c5e7754 also cleans up return types and a bit of formatting.

JohnVillalovos (approved):
Looks good to me.

Please merge whenever you want to.

PR #1710, opened 2021-11-24
chore: add get() methods for GetWithoutIdMixin based classes

JohnVillalovos:
Add the get() methods for the GetWithoutIdMixin based classes.

Update the tests/meta/test_ensure_type_hints.py tests to check to
ensure that the get methods are defined with the correct return type.

### DIFF
--- a/tests/meta/test_ensure_type_hints.py
+++ b/tests/meta/test_ensure_type_hints.py
@@ -4,13 +4,32 @@ Ensure type-hints are setup correctly and detect if missing functions.
 Original notes by John L. Villalovos
 
 """
+import dataclasses
+import functools
 import inspect
-from typing import Tuple, Type
+from typing import Optional, Type
 
 import _pytest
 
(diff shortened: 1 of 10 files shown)


tests/meta/test_ensure_type_hints.py
```diff
@@ -4,15 +4,34 @@
...
+
+    def __lt__(self, other: object) -> bool:
+        if not isinstance(other, ClassInfo):
+            return NotImplemented
+        return (self.type.__module__, self.name) < (other.type.__module__, other.name)
+
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, ClassInfo):
+            return NotImplemented
+        return (self.type.__module__, self.name) == (other.type.__module__, other.name)
```
nejch, line 32: TBH I hope we can at some point get rid of all this scaffolding by using more native mypy stuff if it's possible/becomes available 😅

PR #1714, opened 2021-11-25
chore: remove pytest-console-scripts specific config

### LINKED ISSUE #1713: Self-test failures on Fedora 35 (python 3.10)

rpmbuild of a python-gitlab package on Fedora 35 (python 3.10) fails, with the following:

```
+ /usr/bin/pytest gitlab
=========================================================================== test session starts ===========================================================================
platform linux -- Python 3.10.0, pytest-6.2.4, py-1.10.0, pluggy-0.13.1
rootdir: /home/jwilson/rpmbuild/BUILD/python-gitlab-2.9.0, configfile: tox.ini
collected 0 items

============================================================================ warnings summary =============================================================================
../../../../../usr/lib/python3.10/site-packages/_pytest/config/__init__.py:1233
  /usr/lib/python3.10/site-packages/_pytest/config/__init__.py:1233: PytestConfigWarning: Unknown config option: script_launch_mode

    self._warn_or_fail_if_strict(f"Unknown config option: {key}\n")

-- Docs: https://docs.pytest.org/en/stable/warnings.html
=========================================================================== 1 warning in 0.02s ============================================================================
error: Bad exit status from /var/tmp/rpm-tmp.0OnLs0 (%check)

```
Saw this initially trying to build a 2.10.1 package, but the same happens with a 2.9.0 package that built fine on Fedora 34 (python 3.9).

## Specifications

  - python-gitlab version: 2.9 through 2.10.1, and maybe earlier

JohnVillalovos:
Remove the pytest-console-scripts specific config from the global
'[pytest]' config section.

Move it to a command line option for the functional tests which use
the Docker container.

Closes #1713

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -84,13 +84,10 @@ exclude_lines =
   if TYPE_CHECKING:
   if debug:
 
-[pytest]
-script_launch_mode = subprocess
-
 [testenv:cli_func_v4]
 deps = -r{toxinidir}/requirements-docker.txt
 commands =
-  pytest --cov --cov-report xml tests/functional/cli {posargs}
+  pytest --script-launch-mode=subprocess --cov --cov-report xml tests/functional/cli {posargs}
 
 [testenv:py_func_v4]
 deps = -r{toxinidir}/requirements-docker.txt


tox.ini
```diff
@@ -84,18 +84,15 @@ exclude_lines =
...
 deps = -r{toxinidir}/requirements-docker.txt
 commands =
-  pytest --cov --cov-report xml tests/functional/cli {posargs}
+  pytest --script-launch-mode=subprocess --cov --cov-report xml tests/functional/cli {posargs}
 
 [testenv:py_func_v4]
 deps = -r{toxinidir}/requirements-docker.txt
 commands =
-  pytest --cov --cov-report xml tests/functional/api {posargs}
+  pytest --script-launch-mode=subprocess --cov --cov-report xml tests/functional/api --script-launch-mode-subprocess {posargs}
```
nejch, line 95: I think this is not needed here. I added this for the CLI tests initially with this arg as well (https://github.com/python-gitlab/python-gitlab/commit/79489c775141c4ddd1f7aecae90dae8061d541fe#diff-ef2cef9f88b4fe09ca3082140e67f5ad34fb65fb6e228f119d3812261ae51449L58)

```suggestion
  pytest --cov --cov-report xml tests/functional/api {posargs}
```

nejch (commented):
Thanks @JohnVillalovos just a quick note :)

JohnVillalovos:
@nejch Thanks for the info above. If you have any ideas why my change is failing it would be appreciated. I guess I could just drop using the decorator as it seems to work okay without it. But I think it is using "in-process" mode by default instead of "subprocess" mode. Not sure if that matters or not.

tox.ini
JohnVillalovos: Done. Thanks.

JohnVillalovos:
> @nejch Thanks for the info above. If you have any ideas why my change is failing it would be appreciated. I guess I could just drop using the decorator as it seems to work okay without it. But I think it is using "in-process" mode by default instead of "subprocess" mode. Not sure if that matters or not.

@nejch  I got it figured out. Ready for review. Thanks.

nejch:
> > @nejch Thanks for the info above. If you have any ideas why my change is failing it would be appreciated. I guess I could just drop using the decorator as it seems to work okay without it. But I think it is using "in-process" mode by default instead of "subprocess" mode. Not sure if that matters or not.
> 
> @nejch I got it figured out. Ready for review. Thanks.

Thanks @JohnVillalovos. I hope my comment wasn't too confusing. What you did in `tox.ini` would also work, I just wanted to say it was not needed for `py_func_v4` (hence the failure as it wasnt using the scripts plugin), but only `cli_func_v4`. What you did here also works but is maybe less scalable as we need to do it for every module instead of the entire suite. So I'll leave it up to you, I'd say you could still add: 

```ini
[testenv:cli_func_v4]
deps = -r{toxinidir}/requirements-docker.txt
commands =
  pytest --script-launch-mode=subprocess --cov --cov-report xml tests/functional/cli {posargs}
```

and it would work (but omit it for py_func_v4).

The reason I added subprocess mode initially I think was because it would more closely resemble the environment of a user running the CLI I believe.

JohnVillalovos:
> Thanks @JohnVillalovos. I hope my comment wasn't too confusing. What you did in `tox.ini` would also work, I just wanted to say it was not needed for `py_func_v4` (hence the failure as it wasnt using the scripts plugin), but only `cli_func_v4`. What you did here also works but is maybe less scalable as we need to do it for every module instead of the entire suite. So I'll leave it up to you, I'd say you could still add:

Thanks @nejch  When I got the error before when using the command line argument `--script-launch-mode=subprocess` I just went and worked around it. I didn't realize that the reason it failed in `py_func_v4` was because it wasn't using the plugin. I knew it was being installed so I assumed the command line argument should also work. I'm guessing that if it doesn't use the plugin then the command-line plugin arguments don't get added.

PR #1719, opened 2021-11-27
fix: transcoded special symbols in url path

### LINKED ISSUE #1510: The branch name containing "#" needs to be transcoded by urlencode
## Description of the problem
The "protect/unprotect" interface cannot be used when the branch name contains "#"

## Expected Behavior
`test.py`
```python
#!/usr/bin/env python

import sys
import gitlab
url = 'http://192.168.104.186/'
token = 'bA3inGgz77LBPNdLwCxp'

gl = gitlab.Gitlab(url, private_token=token)

project_id=28
project = gl.projects.get(project_id)

new_br = sys.argv[1]

project.branches.create({'branch': new_br, 'ref':'master'}).protect()
```
```bash
python test.py ohh#yes
```
The branch "ohh#yes" should be created and set to protect the branch

## Actual Behavior
But，get http error：
image

on project, branch created, but not set protect
image

## Specifications

  - python-gitlab version: 2.8.0
  - Gitlab server version (or gitlab.com): local version 12.10.12

## The solution
Modify the _build_url function in "gitlab/client.py"
replace "#" in path with "%23" :
image

## Conclusion
image

Get what I want：
image

So far, only "#" has been found to be problematic. Other special symbols, such as "!" "@" "&" "-" No problem
Can I submit a Merge Request？

XuHandsome:
Transcoded special symbols "#" in url path,
Closes #1510

### DIFF
--- a/gitlab/client.py
+++ b/gitlab/client.py
@@ -463,6 +463,7 @@ class Gitlab(object):
             str: The full URL
         """
         if path.startswith("http://") or path.startswith("https://"):
+            path = path.replace("#", "%23")
             return path
         else:
             return f"{self._url}{path}"


nejch:
Thanks @XuHandsome. I just want to say this goes in the opposite direction of https://github.com/python-gitlab/python-gitlab/issues/1356, which proposed a general approach to issues to URL-encoding.

If you want to tackle this, I think that adding tests to prove the fix would really help. Right now I'm not sure this should be in `_build_url`.

To explain your initial issue a bit more:
- there is no issue in the branch creation in your initial example, `project.branches.create({'branch': new_br, 'ref':'master'})` works fine as you noticed yourself. The branch is created.
- however, you were using a deprecated `branch.protect()` endpoint, which used the branch as part of the URL. Instead, if you use the `protected_branches` API, you should have no issues: https://python-gitlab.readthedocs.io/en/latest/gl_objects/protected_branches.html#examples

However there's still some insight here, there may be other endpoints where `#` may be needed as part of either POST data or the URL, and we should keep an eye on it, but solve it in a general way, IMO.

XuHandsome:
Yes, I re-used the new interface test and found no problem. I'm glad it was fixed

PR #1723, opened 2021-11-30
Close-out `master` branch

JohnVillalovos:
Until we delete the `master` branch, delete all content except a short
README that tells people to use the `main` branch.


nejch:
I'm almost reluctant to do this just because there's so much stuff people do with python-gitlab out in the wild (see e.g. patches from master being applied [here](https://gitweb.gentoo.org/repo/gentoo.git/tree/dev-vcs/python-gitlab/files/python-gitlab-2.10.1-do-not-package-tests.patch) post-release) that we're not aware of, I'm afraid it would break things completely for some people. For example some custom dependency management systems like yocto/bitbake that take deps from git refs (bad practice to do it from the HEAD of master, but still).

If we do it along with the 3.0.0 release I feel like it'd be a more justified change. Is there some other reason I might have missed for this?

JohnVillalovos:
I feel like leaving it around is going to confuse people. This way they find out quickly it isn't working. If they are using tip of `master` then they are probably expecting it to be moving along. So now whatever process they have is silently broken.

This helps prevent things from silently breaking. It doesn't break people who are using SHAs. The only people it would break is people who are using `master` and using the tip.

PR #1725, opened 2021-12-01
Document that branch listing defaults to only returning 20 by default.

vallamost:
The documentation does not mention that by default only 20 branches are returned in a response. You must pass in `all=True` if you want to receive all branches for a repo.

### DIFF
--- a/docs/gl_objects/branches.rst
+++ b/docs/gl_objects/branches.rst
@@ -15,10 +15,15 @@ References
 
 Examples
 --------
+Branch listing uses the base ListMixin class which provides pagination.
 
-Get the list of branches for a repository::
+Get a list of branches for a repository (Limited to 20 branches in a pagination response)::
 
     branches = project.branches.list()
+    
+Get a list of ALL branches for a repository::
+
+    branches = project.branches.list(all=True)    
 
 Get a single repository branch::
 


JohnVillalovos:
This is discussed in: https://python-gitlab.readthedocs.io/en/stable/api-usage.html#pagination

vallamost:
> This is discussed in: https://python-gitlab.readthedocs.io/en/stable/api-usage.html#pagination

Okay? The current documentation reads as that if branches.list() will list all branches in a repository. There's no mention of it defaulting to only 20 items in the documentation or in response object.

nejch:
> > This is discussed in: https://python-gitlab.readthedocs.io/en/stable/api-usage.html#pagination
> 
> Okay? The current list branch documentation reads as branches.list() will list all branches in a repository. There's no mention of it defaulting to only 20 items in the branch list documentation or in a branch list response object. That's what I am trying to clarify for people in the future :)

It's tricky then because this is true for absolutely all `list()` methods. This is why it's discussed at the beginning to cover all endpoints. Maybe an admonition at the top of each page for API examples (or at least `list()` examples) could have a reminder for the most common gotchas & options (pagination with `all`, `lazy`, attributes available via `list()` vs `get()` objects etc).

This could possibly be done in a template as well to avoid repetition. I'm not sure if you're familiar with sphinx templating though @vallamost?

JohnVillalovos:
Maybe could have a common bit of documentation that could be included where the `list()` calls are.

https://sublime-and-sphinx-guide.readthedocs.io/en/latest/reuse.html

Or @nejch  you could give more info on this template idea.

nejch:
Hey @vallamost thanks again for this pull request.

We believe we now cover this scenario with a runtime warning to the user that links to the pagination docs whenever any type of `list()` call returns only a portion of the available objects (see commit https://github.com/python-gitlab/python-gitlab/commit/1339d645ce58a2e1198b898b9549ba5917b1ff12).

I think we can close this for now as that commit should cover all `list` endpoints in a generic way. But please reopen if you think we should add this in specific docs (where `list` is used, not just in branches) :) Thanks!

PR #1726, opened 2021-12-02
chore: add running unit tests on windows/macos

JohnVillalovos:
Add running the unit tests on windows-latest and macos-latest with
Python 3.10.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,31 +14,39 @@ env:
 
 jobs:
   unit:
-    runs-on: ubuntu-20.04
+    runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        include:
-          - python-version: 3.7
+        os: [ubuntu-latest]
+        python:
+          - version: "3.7"
             toxenv: py37
-          - python-version: 3.8
+          - version: "3.8"
             toxenv: py38
-          - python-version: 3.9
+          - version: "3.9"
             toxenv: py39
-          - python-version: "3.10"
-            toxenv: py310
-          - python-version: "3.10"
-            toxenv: smoke
+          - version: "3.10"
+            toxenv: py310,smoke
+        include:
+          - os: macos-latest
+            python:
+              version: "3.10"
+              toxenv: py310,smoke
+          - os: windows-latest
+            python:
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -41,6 +41,26 @@ jobs:
...
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install dependencies
+        run: pip install tox pytest-github-actions-annotate-failures
+      - name: Run tests
+        env:
+          TOXENV: ${{ matrix.toxenv }}
+        run: tox
+
```
nejch, line 63: I think we could actually incorporate this into our existing matrix and use `runs-on: ${{ matrix.os }}`.

However we'd need to change our matrix and play a bit with `include:` elements to make sure we only test 3.10 for most things and add old versions for Linux only.

The documentation for this is here: https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#example-including-additional-values-into-combinations (there's a bit more if you go down on these specific filters).

Let me know if you want to try out yourself or if you want me to give it a shot if it's too tedious. It would be easier to extend later with macos or moreif needed.

JohnVillalovos: I did try that initially but it didn't work very well. The problem I had was that the tox environment doesn't match up with the Python version exactly. When I tried to work around that issue it got ugly. So I went with the simple way of mostly duplicating it.

Would be great if you want to give it a shot.

As far as MacOS I think we could add that here easily enough. As we are only running one tox invocation per OS.

JohnVillalovos: Here is an example of the ugliness: https://github.com/python-gitlab/python-gitlab/commit/bcd00343f16c7a7d94a9b57c809a380be8768e69

JohnVillalovos: As a note, I updated this to also run on `macos-latest`

JohnVillalovos: I figured it out! 😊

So resolved.

JohnVillalovos:
@nejch  As a note, this caused the name of the tests to change. So that will need to be updated somewhere in the project settings.

PR #1729, opened 2021-12-05
chore: add initial pylint check

JohnVillalovos:
Initial pylint check is added. A LONG list of disabled checks is also
added. In the future we should work through the list and resolve the
errors or disable them on a more granular level.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -35,3 +35,5 @@ jobs:
         run: tox -e mypy
       - name: Run isort import order checker (https://pycqa.github.io/isort/)
         run: tox -e isort -- --check
+      - name: Run pylint Python code static checker (https://www.pylint.org/)
+        run: tox -e pylint
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -20,6 +20,15 @@ repos:
     rev: 5.9.3
     hooks:
       - id: isort
+  - repo: https://github.com/pycqa/pylint
+    rev: v2.12.2
+    hooks:
+      - id: pylint
+        additional_dependencies:
+          - argcomplete==1.12.3
+          - requests==2.26.0
+          - requests-toolbelt==0.9.1
+        files: 'gitlab/'
   - repo: https://github.com/pre-commit/mirrors-mypy
     rev: v0.910
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -33,6 +33,7 @@ repos:
     rev: v0.910
     hooks:
       - id: mypy
+        args: []
         additional_dependencies:
           - types-PyYAML==6.0.1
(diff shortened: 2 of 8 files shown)


.github/workflows/lint.yml
```diff
@@ -35,3 +35,5 @@ jobs:
         run: tox -e mypy
       - name: Run isort import order checker (https://pycqa.github.io/isort/)
         run: tox -e isort -- --check
+      - name: Run pylint Python code static checker (https://www.pylint.org/)
+        run: tox -e pylint
```
nejch, line 39: Can we also add this to pre-commit? 
https://pylint.pycqa.org/en/latest/user_guide/pre-commit-integration.html

JohnVillalovos: Done! 😊

.pre-commit-config.yaml
```diff
@@ -20,6 +20,10 @@ repos:
     rev: 5.9.3
     hooks:
       - id: isort
+  - repo: https://github.com/pycqa/pylint
+    rev: v2.12.2
+    hooks:
+      - id: pylint
```
nejch, line 26: I think we'll need the `local:` approach documented in the docs (I linked it above) due to the extra deps we have.

JohnVillalovos: Thanks @nejch  That was the key and a few other things.

.pre-commit-config.yaml
```diff
@@ -20,6 +20,13 @@ repos:
     rev: 5.9.3
     hooks:
       - id: isort
+  - repo: local
+    hooks:
+      - id: pylint
+        name: pylint
+        entry: pylint
+        language: system
+        files: 'gitlab/'
```
nejch, line 29: Thanks! I'll try to run this locally but does it mean it'll always run on all source files?

JohnVillalovos: Not sure I completely understand but it will run on all files located under the `gitlab/' directory. Meaning it won't check the `tests/` directory or anything else not inside the `gitlab/` directory.

nejch: Sorry, just thought we'd have the same issue we had before where it would override what files get analyzed and ignore the linter's config. Seems to work ok!

However this is still a bit of headache now 🤦 on a clean system, it expects `pylint` installed which people might not have as we use tox for managing dependencies (see below), plus versions wouldn't match potentially. I see some people use the tox venv for this sometimes as a solution: https://github.com/pre-commit/pre-commit/issues/1309. Not sure I like that though, WDYT @JohnVillalovos should we just still add it or find a solution in a follow-up?

```
flake8...................................................................Passed
isort....................................................................Passed
pylint...................................................................Failed
- hook id: pylint
- exit code: 1

Executable `pylint` not found
```

JohnVillalovos: Fixed it. So it is working now.

PR #1733, opened 2021-12-06
feat: package file delete on package file object

simonisateur:


JohnVillalovos:
Thanks. Would you be willing to add a test case for this? Probably in `tests/functional/api/test_packages.py` would be good.

simonisateur:
@JohnVillalovos I've added a unit tests and classified the merge as a `feat`.

simonisateur:
@JohnVillalovos I've fixed my test. I like the setup of the project with commitlint.

.github/workflows/docs.yml
```diff
@@ -1,5 +1,11 @@
 name: Docs
 
+# If a pull-request is pushed then cancel all previously running jobs related
+# to that pull-request
```
JohnVillalovos, line 4: This looks wrong.

I think the MR needs to be rebased.

JohnVillalovos (changes requested):

.github/workflows/docs.yml
simonisateur: Yes it does. I fixed it

JohnVillalovos: Thanks. Can you squash it down to one commit? Instead of the current commit and then a "merge commit of main branch"

simonisateur: yes

JohnVillalovos: > yes

Awesome! Thank you.

simonisateur: All in one commit. My code and the merge

simonisateur: Is the message `feat: add delete on package_file` object ok ?

JohnVillalovos:
Thanks. As a note for the future. It can sometimes be simpler to do the tests in the `tests/functional/api/` directory as those actually run against a running Gitlab instance.

But thanks for the unit tests!

JohnVillalovos (approved):

simonisateur:
Thanks for your precious time. I will note that! As a fact I will revisit the functional test in the future to help me learn pytest and be more confident in the delete at the same time.

My delete is'nt a great test because it only test if it does'nt work, but not if it works.

nejch:
Thank you very much @simonisateur!

PR #1740, opened 2021-12-11
feat(api): add support for Topics API

nejch:


.renovaterc.json
```diff
@@ -7,7 +7,7 @@
   },
   "regexManagers": [
     {
-      "fileMatch": ["^tests/functional/fixtures/.env$"],
+      "fileMatch": ["^tests\\/functional\\/fixtures\\/.env$"],
```
JohnVillalovos, line 10: nit:  Use r-strings

```
"fileMatch": [r"^tests\/functional\/fixtures\/.env$"],
```

JohnVillalovos (approved):
I'm okay with it as is, but I think r-strings usage would be good. But not something to stop this from merging.

Thanks!

.renovaterc.json
nejch: Thanks @JohnVillalovos. This is a JSON file though, I don't think that's supported?

JohnVillalovos: Duh! I was just assuming Python 😟

PR #1741, opened 2021-12-11
docs: add the start of a contributor's guide

JohnVillalovos:
Create an initial contributor's guide that will hopefully be expanded
on in the future.

### DIFF
--- a/docs/contributor/contributing.rst
+++ b/docs/contributor/contributing.rst
@@ -1,0 +1,50 @@
+.. _code-contribution-guide:
+
+============================
+So You Want to Contribute...
+============================
+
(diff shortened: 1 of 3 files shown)


nejch:
Thanks a lot @JohnVillalovos! Just one initial concern I have in general is that this somewhat duplicates https://github.com/python-gitlab/python-gitlab/blob/main/CONTRIBUTING.rst. Can we combine these somehow? the root-level one should probably be more concise and link to more details I guess, but have enough info for a quickstart.

For example, I noticed the same kind of duplication with `README.rst` and our install instructions and intro etc, so I ended up just linking the repo file in the docs. Would that make sense to avoid having to update things twice?

PR #1742, opened 2021-12-12
chore: add Python 3.11 testing

JohnVillalovos:
Add a unit test for Python 3.11. This will use the latest version of
Python 3.11 that is available from
https://github.com/actions/python-versions/

At this time it is 3.11.0-alpha.2 but will move forward over time
until the final 3.11 release and updates. So 3.11.0, 3.11.1, ... will
be matched.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -33,6 +33,8 @@ jobs:
             toxenv: py39
           - version: "3.10"
             toxenv: py310,smoke
+          - version: '3.11.0-alpha - 3.11' # SemVer's version range syntax
+            toxenv: py311,smoke
         include:
           - os: macos-latest
             python:


JohnVillalovos:
We probably don't want this but I thought it would be interesting to test on the latest Python version available for GitHub actions.

.github/workflows/test.yml
```diff
@@ -42,6 +42,10 @@ jobs:
             python:
               version: "3.10"
               toxenv: py310,smoke
+          - os: ubuntu-latest
+            python:
+              version: '3.11.0-alpha - 3.11' # SemVer's version range syntax
+              toxenv: py311,smoke
```
nejch, line 48: Since it already works fine, can we have this above with the normal matrix entries as it's also on ubuntu?

nejch:
> 

I don't mind :grinning: we can always remove if it starts to get flaky. Just one question.

Also wow, just realized GH Actions does not support something like `allow_failure`. Huh. https://github.com/actions/toolkit/issues/399

PR #1743, opened 2021-12-12
feat(cli): do not require config file to run CLI

nejch:
BREAKING CHANGE: A config file is no longer needed to run
the CLI. python-gitlab will default to https://gitlab.com
with no authentication if there is no config file provided.
python-gitlab will now also only look for configuration
in the provided PYTHON_GITLAB_CFG path, instead of merging
it with user- and system-wide config files. If the
environment variable is defined and the file cannot be
opened, python-gitlab will now explicitly fail.

Needed for https://github.com/python-gitlab/python-gitlab/pull/1296

### DIFF
--- a/gitlab/config.py
+++ b/gitlab/config.py
@@ -20,25 +20,65 @@ import os
 import shlex
 import subprocess
 from os.path import expanduser, expandvars
+from pathlib import Path
 from typing import List, Optional, Union
 
-from gitlab.const import USER_AGENT
+from gitlab.const import DEFAULT_URL, USER_AGENT
 
-
-def _env_config() -> List[str]:
-    if "PYTHON_GITLAB_CFG" in os.environ:
-        return [os.environ["PYTHON_GITLAB_CFG"]]
-    return []
-
-
-_DEFAULT_FILES: List[str] = _env_config() + [
+_DEFAULT_FILES: List[str] = [
     "/etc/python-gitlab.cfg",
-    os.path.expanduser("~/.python-gitlab.cfg"),
+    str(Path.home() / ".python-gitlab.cfg"),
 ]
 
 HELPER_PREFIX = "helper:"
 
 HELPER_ATTRIBUTES = ["job_token", "http_password", "private_token", "oauth_token"]
+
+
+def _resolve_file(filepath: Union[Path, str]) -> str:
+    resolved = Path(filepath).resolve(strict=True)
+    return str(resolved)
+
+
+def _get_config_files(
+    config_files: Optional[List[str]] = None,
+) -> Union[str, List[str]]:
+    """
+    Return resolved path(s) to config files if they exist, with precedence:
+    1. Files passed in config_files
+    2. File defined in PYTHON_GITLAB_CFG
+    3. User- and system-wide config files
+    """
+    resolved_files = []
+
+    if config_files:
+        for config_file in config_files:
+            try:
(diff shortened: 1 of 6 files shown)


gitlab/config.py
```diff
@@ -20,27 +20,54 @@
 import shlex
 import subprocess
 from os.path import expanduser, expandvars
+from pathlib import Path
```
JohnVillalovos, line 23: nit: As we discussed in the relative import patch. I think this should be `import pathlib` to improve readability.

Heck the line above should just be `import os` in my opinion.

nejch: Hmm to me this is such a [standard use of pathlib](https://docs.python.org/3/library/pathlib.html#basic-use) I would keep it least for this PR. We have other instances of it in the repo that would make it inconsistent and many projects do it this way, so I'd maybe keep that for discussion in the other PR :)

JohnVillalovos: Yeah lets discuss in the other one. I think that documentation is showing it that way to make the documentation easier to parse.

But from the [Zen of Python](https://www.python.org/dev/peps/pep-0020/)

> Explicit is better than implicit.

As I said this is a "nit" for me. I just want to improve code readability.

gitlab/config.py
```diff
@@ -20,27 +20,54 @@
...
+        return config_files
+
+    try:
+        env_config = os.environ["PYTHON_GITLAB_CFG"]
+        return _resolve_file(env_config)
+    except KeyError:
+        pass
+    except OSError as e:
+        raise GitlabConfigMissingError(
+            f"Cannot read config from PYTHON_GITLAB_CFG in " f"{env_config}: {e}"
```
JohnVillalovos, line 56: extra quotes here.

Also how about `f"{env_config!r}"` the `!r` will surround it with quotes.

gitlab/config.py
```diff
@@ -66,155 +93,149 @@ def __init__(
...
-                self.gitlab_id = self._config.get("global", "default")
+                self.gitlab_id = _config.get("global", "default")
             except Exception as e:
                 raise GitlabIDError(
                     "Impossible to get the gitlab id (not specified in config file)"
                 ) from e
 
         try:
-            self.url = self._config.get(self.gitlab_id, "url")
+            self.url = _config.get(self.gitlab_id, "url")
```
JohnVillalovos, line 129: Does this work if there is no config file?

JohnVillalovos (changes requested):
Can functional tests be added for this. In particular a functional test that demonstrates it works without a config file? Though I guess that will try to contact `https://gitlab.com/` 😟

gitlab/config.py
nejch: Thanks! `black` did a sneaky auto-format there :grin: 

I decided to remove the variable altogether as it was redundant, we already get it from OSError:
`Cannot read config from PYTHON_GITLAB_CFG in 'non-existent': [Errno 2] No such file or directory: '/home/nejc/repos/python-gitlab/non-existent'`

gitlab/config.py
nejch: Thanks for that! Added some logic that more explicitly fails on CLI args. There's a functional test for that as well now.

JohnVillalovos (approved):
Looks good to me. I wasn't sure if you wanted to merge it or not.

Please merge whenever you are ready.

Thanks.

PR #1765, opened 2021-12-19
chore: fix unit test if config file exists locally

### LINKED ISSUE #1764: Unit test failing: tests/unit/test_config.py::test_env_config_not_defined_does_not_raise
Unit test appears to be picking up local file.

```_____________________________________________________________________________________________________________________ test_env_config_not_defined_does_not_raise ______________________________________________________________________________________________________________________

mock_clean_env = None

    def test_env_config_not_defined_does_not_raise(mock_clean_env):
>       assert config._get_config_files() == []
E       AssertionError: assert ['/home/jlvil...n-gitlab.cfg'] == []
E         Left contains one more item: '/home/jlvillal/.python-gitlab.cfg'
E         Full diff:
E         - []
E         + ['/home/jlvillal/.python-gitlab.cfg']

tests/unit/test_config.py:130: AssertionError
=============================================================================================================================== short test summary info ===============================================================================================================================
FAILED tests/unit/test_config.py::test_env_config_not_defined_does_not_raise - AssertionError: assert ['/home/jlvil...n-gitlab.cfg'] == []
====================================================================================================================== 1 failed, 687 passed, 33 skipped in 2.12s ======================================================================================================================
ERROR: InvocationError for command /home/jlvillal/sources/github-mine/python-gitlab/.tox/py310/bin/pytest tests/unit tests/meta -v (exited with code 1)
```

JohnVillalovos:
Closes #1764

### DIFF
--- a/tests/unit/test_config.py
+++ b/tests/unit/test_config.py
@@ -126,8 +126,10 @@ def test_env_config_missing_file_raises(monkeypatch):
         config._get_config_files()
 
 
-def test_env_config_not_defined_does_not_raise(mock_clean_env):
-    assert config._get_config_files() == []
+def test_env_config_not_defined_does_not_raise(mock_clean_env, monkeypatch):
+    with monkeypatch.context() as m:
+        m.setattr(config, "_DEFAULT_FILES", [])
+        assert config._get_config_files() == []
 
 
 def test_default_config(mock_clean_env, monkeypatch):


tests/unit/test_config.py
```diff
@@ -126,8 +126,10 @@ def test_env_config_missing_file_raises(monkeypatch):
         config._get_config_files()
 
 
-def test_env_config_not_defined_does_not_raise(mock_clean_env):
-    assert config._get_config_files() == []
+def test_env_config_not_defined_does_not_raise(mock_clean_env, monkeypatch):
+    with monkeypatch.context() as m:
+        m.setattr(config, "_DEFAULT_FILES", [])
+        assert config._get_config_files() == []
```
nejch, line 132: Since we're only setting the attribute and not mocking any library, I think the context manager here might not be needed, WDYT?

I'm happy to merge this as is though.

nejch:
Thanks @JohnVillalovos, missed that one. Just one quick question

tests/unit/test_config.py
JohnVillalovos: I like it as it is. My worry is that if we set the attribute it will also affect other tests which use that attribute. I guess I would rather be safe...

PR #1766, opened 2021-12-19
fix: stop encoding '.' to '%2E'

### LINKED ISSUE #1006: Creation project with "." in name works incorrectly.
I'm trying to create project with "." in both name and path, using:
```python
gl.projects.create({'namespace_id': group.id, 'name': projectName, 'path': projectName})
```

## Expected Behavior
Name and path are the same and equal to actual "projectName".
URL git_url/test_group/project.test
## Actual Behavior
Name is correct, but path has "." replaced with "-".
For example "project.test" has path "project-test".
URL: git_url/test_group/project-test

## Specifications

  - python-gitlab version: 1.12.1
  - Gitlab server version (or gitlab.com): 12.1.4-ee

JohnVillalovos:
Forcing the encoding of '.' to '%2E' causes issues. It also goes
against the RFC:
https://datatracker.ietf.org/doc/html/rfc3986.html#section-2.3

From the RFC:
  For consistency, percent-encoded octets in the ranges of ALPHA
  (%41-%5A and %61-%7A), DIGIT (%30-%39), hyphen (%2D), period (%2E),
  underscore (%5F), or tilde (%7E) should not be created by URI
  producers...

Closes #1006
Related #1356
Related #1561

BREAKING CHANGE: stop encoding '.' to '%2E'. This could potentially be
a breaking change for users who have incorrectly configured GitLab
servers which don't handle period '.' characters correctly.

### DIFF
--- a/gitlab/client.py
+++ b/gitlab/client.py
@@ -593,24 +593,19 @@ class Gitlab(object):
         json, data, content_type = self._prepare_send_data(files, post_data, raw)
         opts["headers"]["Content-type"] = content_type
 
-        # Requests assumes that `.` should not be encoded as %2E and will make
-        # changes to urls using this encoding. Using a prepped request we can
-        # get the desired behavior.
-        # The Requests behavior is right but it seems that web servers don't
-        # always agree with this decision (this is the case with a default
-        # gitlab installation)
-        req = requests.Request(verb, url, json=json, data=data, params=params, **opts)
-        prepped = self.session.prepare_request(req)
-        if TYPE_CHECKING:
-            assert prepped.url is not None
-        prepped.url = utils.sanitized_url(prepped.url)
-        settings = self.session.merge_environment_settings(
-            prepped.url, {}, streamed, verify, None
-        )
-
         cur_retries = 0
         while True:
-            result = self.session.send(prepped, timeout=timeout, **settings)
+            result = self.session.request(
+                method=verb,
+                url=url,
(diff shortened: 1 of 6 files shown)


nejch:
@JohnVillalovos could you make this a `fix` type so it shows in the release and also add the `stop encoding '.' to '%2E'` sentence at the beginning of the breaking change trailer so it's clear in the release notes in the breaking changes list?

JohnVillalovos:
> @JohnVillalovos could you make this a `fix` type so it shows in the release and also add the `stop encoding '.' to '%2E'` sentence at the beginning of the breaking change trailer so it's clear in the release notes in the breaking changes list?

Done I think. Thanks.

gitlab/client.py
```diff
@@ -593,24 +593,18 @@ def http_request(
...
+            result = self.session.request(
+                method=verb,
+                url=url,
+                json=json,
+                data=data,
+                params=params,
+                verify=verify,
+                stream=streamed,
+                **opts,
+            )
```
nejch, line 607: I think we are missing `timeout` here? It's not pulled in from `opts`, e.g.:

https://github.com/python-gitlab/python-gitlab/blob/39155df3ce3aa80df8ef8db5c6b11231c5a65cca/gitlab/client.py#L587-L590

JohnVillalovos: Nice catch! Like me fix that.

PR #1773, opened 2021-12-22
fix: handle situation where gitlab.com does not return values

### LINKED ISSUE #1686: int() argument must be a string, a bytes-like object or a real number, not 'NoneType' while accessing pagination related attributes
Accessing current_page , per_page , total , total_pages attributes , throwing the following error.
**int() argument must be a string, a bytes-like object or a real number, not 'NoneType'**

  
```
  gl = gitlab.Gitlab('https://gitlab.com/', private_token=developer_access_token,
                       pagination="keyset", order_by="id",
                       per_page=100)
    iterator_projects = gl.projects.list(as_list=False, owned=True)
```

## Expected Behavior

It should return the number of pages expected

## Actual Behavior

throwing exceptions.

## Specifications

  - python-gitlab version:2.10.1
  - Gitlab server version (or gitlab.com):gitlab.com

JohnVillalovos:
If a query returns more than 10,000 records than the following values
are NOT returned:
  x.total_pages
  x.total

Modify the code to allow no value to be set for these values. If there
is not a value returned the functions will now return None.

https://docs.gitlab.com/ee/user/gitlab_com/index.html#pagination-response-headers

Closes #1686

### DIFF
--- a/docs/api-usage.rst
+++ b/docs/api-usage.rst
@@ -265,8 +265,17 @@ The generator exposes extra listing information as received from the server:
 * ``prev_page``: if ``None`` the current page is the first one
 * ``next_page``: if ``None`` the current page is the last one
 * ``per_page``: number of items per page
-* ``total_pages``: total number of pages available
-* ``total``: total number of items in the list
+* ``total_pages``: total number of pages available. This may be a ``None`` value.
+* ``total``: total number of items in the list. This may be a ``None`` value.
+
+.. note::
+
+   For performance reasons, if a query returns more than 10,000 records, GitLab
+   does not return the ``total_pages`` or ``total`` headers.  In this case,
+   ``total_pages`` and ``total`` will have a value of ``None``.
+
+   For more information see:
+   https://docs.gitlab.com/ee/user/gitlab_com/index.html#pagination-response-headers
 
 Sudo
 ====
--- a/gitlab/client.py
+++ b/gitlab/client.py
@@ -917,14 +917,12 @@ class GitlabList(object):
             self._next_url = next_url
         except KeyError:
             self._next_url = None
-        self._current_page: Optional[Union[str, int]] = result.headers.get("X-Page")
-        self._prev_page: Optional[Union[str, int]] = result.headers.get("X-Prev-Page")
-        self._next_page: Optional[Union[str, int]] = result.headers.get("X-Next-Page")
-        self._per_page: Optional[Union[str, int]] = result.headers.get("X-Per-Page")
-        self._total_pages: Optional[Union[str, int]] = result.headers.get(
-            "X-Total-Pages"
-        )
-        self._total: Optional[Union[str, int]] = result.headers.get("X-Total")
+        self._current_page: Optional[str] = result.headers.get("X-Page")
+        self._prev_page: Optional[str] = result.headers.get("X-Prev-Page")
+        self._next_page: Optional[str] = result.headers.get("X-Next-Page")
+        self._per_page: Optional[str] = result.headers.get("X-Per-Page")
+        self._total_pages: Optional[str] = result.headers.get("X-Total-Pages")
+        self._total: Optional[str] = result.headers.get("X-Total")
 
         try:
             self._data: List[Dict[str, Any]] = result.json()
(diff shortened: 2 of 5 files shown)


docs/api-usage.rst
```diff
@@ -265,8 +265,17 @@ The generator exposes extra listing information as received from the server:
...
+* ``total``: total number of items in the list. This may be a ``None`` value.
+
+.. note::
+
+   ``total_pages`` and ``total`` may have a value of ``None``. For performance
+   reasons, if a query returns more than 10,000 records, GitLab doesn’t return
+   any value for ``total_pages`` or ``total``.
+
+   For more information see:
+   https://docs.gitlab.com/ee/user/gitlab_com/index.html#pagination-response-headers
```
nejch, line 278: I think not just the values but the headers are missing entirely, right?

```suggestion
   For performance reasons, if a query returns more than 10,000 records, GitLab
   does not return the ``total_pages`` or ``total`` headers.  In this case,
   ``total_pages`` and ``total`` will have a value of ``None``.

   For more information see:
   https://docs.gitlab.com/ee/user/gitlab_com/index.html#pagination-response-headers
```

gitlab/client.py
```diff
@@ -966,17 +966,23 @@ def per_page(self) -> int:
...
     @property
-    def total(self) -> int:
+    def total(self) -> Optional[int]:
         """The total number of items."""
-        if TYPE_CHECKING:
-            assert self._total is not None
+        if self._total is None:
+            # NOTE(jlvillal): When a query returns more than 10,000 items, GitLab
+            # doesn't return the header for this value. So we return None.
+            # https://docs.gitlab.com/ee/user/gitlab_com/index.html#pagination-response-headers
```
nejch, line 984: Since the 2 properties are in the code together maybe we can put the comment above them to not have it duplicated?

gitlab/client.py
```diff
@@ -966,17 +966,23 @@ def per_page(self) -> int:
...
+    def total_pages(self) -> Optional[int]:
         """The total number of pages."""
-        if TYPE_CHECKING:
-            assert self._total_pages is not None
+        if self._total_pages is None:
+            # NOTE(jlvillal): When a query returns more than 10,000 items, GitLab
+            # doesn't return the header for this value. So we return None.
+            # https://docs.gitlab.com/ee/user/gitlab_com/index.html#pagination-response-headers
+            return None
         return int(self._total_pages)
```
nejch, line 976: If the header being returned is the more common scenario, should we be inverting this into try/except?

```suggestion
        try:
            return int(self._total_pages)
        except TypeError:
            # NOTE(jlvillal): When a query returns more than 10,000 items, GitLab
            # doesn't return the header for this value. So we return None.
            # https://docs.gitlab.com/ee/user/gitlab_com/index.html#pagination-response-headers
            return None
```

gitlab/client.py
JohnVillalovos: Sure. Done.

docs/api-usage.rst
JohnVillalovos: Done

gitlab/client.py
JohnVillalovos: Not sure why but I'm not a fan of try/except for standard use cases. I'll switch it to a 'if blah is not None`.

nejch:
LGTM I'd just like to reproduce https://github.com/python-gitlab/python-gitlab/issues/1686 in some tests first with an xfail (e.g. I'll see if I can get responses to match the presence/absence of those headers). Are you ok to wait and rebase @JohnVillalovos (unless you want to take a look at it yourself but you already have a few PRs :D)

JohnVillalovos:
> LGTM I'd just like to reproduce #1686 in some tests first with an xfail (e.g. I'll see if I can get responses to match the presence/absence of those headers). Are you ok to wait and rebase @JohnVillalovos (unless you want to take a look at it yourself but you already have a few PRs :D)

Works for me. Thanks!

nejch:
I have https://github.com/python-gitlab/python-gitlab/pull/1788 ready for this and I also saw I was being overly pedantic as we're already doing the same with one-liners :sweat_smile: :

https://github.com/python-gitlab/python-gitlab/blob/c764bee191438fc4aa2e52d14717c136760d2f3f/gitlab/client.py#L945-L959

JohnVillalovos:
> I have #1788 ready for this and I also saw I was being overly pedantic as we're already doing the same with one-liners 😅 :

Thanks. I reviewed and merged #1788. Then rebased this PR. Was surprised that if an `xfail` marked test passes it doesn't cause things to fail 😕 Removed the `xfail`, and it is working as expected now.

JohnVillalovos:
> Was surprised that if an `xfail` marked test passes it doesn't cause things to fail 😕

I figured that out and added `xfail_strict` to the `pytest` configuration. Now if an `xfail` test passes it will cause the test run to fail.

Issue #1777, opened 2021-12-25
Documentation bug


hailianghe:
Hello. the content  below looks  error.
![image](https://user-images.githubusercontent.com/29939806/147384879-a46a23f2-e269-48df-b880-b9fb4c1c90e8.png)
note_id = discussion.attributes['note'][-1]['id'] shoud be 
note_id = discussion.attributes['notes'][-1]['id']
------
In page: gl_objects/discussions.rst

JohnVillalovos:
Thanks. Would you like to propose a PR with a suggested fix?

PR #1779, opened 2021-12-26
Fix Documentation bug (Issue #1777)

hailianghe:
Fix Documentation bug (Issue #1777)  :)


JohnVillalovos:
You probably want a commit message something like:

```
docs: correct documentation for updating discussion note

Closes #1777 
```

PR #1783, opened 2021-12-28
chore: ensure reset_gitlab() succeeds

JohnVillalovos:
Ensure reset_gitlab() succeeds by waiting to make sure everything has
been deleted as expected. If the timeout is exceeded fail the test.

Not using `wait_for_sidekiq` as it didn't work. During testing I
didn't see any sidekiq processes as being busy even though not
everything was deleted.

### DIFF
--- a/tests/functional/conftest.py
+++ b/tests/functional/conftest.py
@@ -7,6 +7,10 @@ from subprocess import check_output
 import pytest
 
 import gitlab
+import gitlab.base
+
+SLEEP_INTERVAL = 0.1
+TIMEOUT = 60  # seconds before timeout will occur
 
 
 @pytest.fixture(scope="session")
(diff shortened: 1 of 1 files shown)


tests/functional/conftest.py
```diff
@@ -30,6 +34,37 @@ def reset_gitlab(gl):
...
+    wait_for_list_empty(rest_manager=gl.variables, description="variables")
+
+    for _ in range(max_iterations):
+        if len(gl.users.list()) <= 1:
+            break
+        time.sleep(SLEEP_INTERVAL)
+    assert [user.username for user in gl.users.list()] == ["root"], (
+        f"Did not delete all users. "
+        f"elapsed_time: {time.perf_counter() - start_time}"
+    )
```
nejch, line 66: Thanks for this! Can we even just not care about the root user specifics and use the same helper for both cases e.g. something like this (not sure what to name it then :grinning: ):

(not sure if I missed something)

```suggestion
    def wait_for_list_empty(
        rest_manager: gitlab.base.RESTManager, description: str, length: int = 0
    ) -> None:
        """Wait for the list() to have required length or fail test if timeout is exceeded"""
        for _ in range(max_iterations):
            if len(rest_manager.list()) <= length:
                break
            time.sleep(SLEEP_INTERVAL)
        assert len(rest_manager.list()) <= length, (
            f"Did not delete all {description}. "
            f"Elapsed_time: {time.perf_counter() - start_time}"
        )

    wait_for_list_empty(rest_manager=gl.projects, description="projects")
    wait_for_list_empty(rest_manager=gl.groups, description="groups")
    wait_for_list_empty(rest_manager=gl.variables, description="variables")
    wait_for_list_empty(rest_manager=gl.users, description="users", length=1)
```

JohnVillalovos: Looks nice. I will do that. Now which PR do we think will land first? Because the 2nd PR to land will need to be rebased.

nejch: This seems like a smaller change so I'd go with this one unless you think it'll be a very tedious rebase on the other one :grin:

JohnVillalovos: > This seems like a smaller change so I'd go with this one unless you think it'll be a very tedious rebase on the other one 😁

Works for me. Order doesn't matter for difficulty I believe.

JohnVillalovos: Also I think this one is better to go first. Should make other one cleaner in not needing to touch `wait_for_sidekiq` code.

PR #1784, opened 2021-12-29
chore: fixtures: after delete() wait to verify deleted

JohnVillalovos:
In our fixtures that create:
  - groups
  - project merge requests
  - projects
  - users

They delete the created objects after use. Now wait to ensure the
objects are deleted before continuing as having unexpected objects
existing can impact some of our tests.

### DIFF
--- a/tests/functional/conftest.py
+++ b/tests/functional/conftest.py
@@ -9,9 +9,7 @@ import pytest
 
 import gitlab
 import gitlab.base
-
-SLEEP_INTERVAL = 0.5
-TIMEOUT = 60  # seconds before timeout will occur
+from tests.functional import helpers
 
 
 @pytest.fixture(scope="session")
@@ -49,8 +47,6 @@ def reset_gitlab(gl):
             logging.info(f"Marking for deletion user: {user.username!r}")
             user.delete(hard_delete=True)
 
-    max_iterations = int(TIMEOUT / SLEEP_INTERVAL)
-
     # Ensure everything has been reset
     start_time = time.perf_counter()
 
@@ -60,7 +56,7 @@ def reset_gitlab(gl):
         """Wait for the list() length to be no greater than expected maximum or fail
         test if timeout is exceeded"""
         logging.info(f"Checking {description!r} has no more than {max_length} items")
-        for count in range(max_iterations):
+        for count in range(helpers.MAX_ITERATIONS):
             items = rest_manager.list()
             if len(items) <= max_length:
                 break
@@ -68,7 +64,7 @@ def reset_gitlab(gl):
                 f"Iteration: {count} Waiting for {description!r} items to be deleted: "
                 f"{[x.name for x in items]}"
             )
-            time.sleep(SLEEP_INTERVAL)
+            time.sleep(helpers.SLEEP_INTERVAL)
 
         elapsed_time = time.perf_counter() - start_time
         error_message = (
@@ -280,10 +276,7 @@ def group(gl):
 
     yield group
 
-    try:
-        group.delete()
-    except gitlab.exceptions.GitlabDeleteError as e:
-        print(f"Group already deleted: {e}")
+    helpers.safe_delete(group)
 
 
 @pytest.fixture(scope="module")
@@ -296,10 +289,7 @@ def project(gl):
 
     yield project
 
-    try:
-        project.delete()
-    except gitlab.exceptions.GitlabDeleteError as e:
-        print(f"Project already deleted: {e}")
+    helpers.safe_delete(project)
 
 
 @pytest.fixture(scope="function")
@@ -327,7 +317,7 @@ def merge_request(project, wait_for_sidekiq):
         assert result is True, "sidekiq process should have terminated but did not"
 
         project.refresh()  # Gets us the current default branch
-        project.branches.create(
+        mr_branch = project.branches.create(
             {"branch": source_branch, "ref": project.default_branch}
         )
         # NOTE(jlvillal): Must create a commit in the new branch before we can
@@ -359,18 +349,13 @@ def merge_request(project, wait_for_sidekiq):
             time.sleep(0.5)
         assert mr.merge_status != "checking"
 
-        to_delete.append((mr.iid, source_branch))
+        to_delete.extend([mr, mr_branch])
         return mr
 
     yield _merge_request
 
-    for mr_iid, source_branch in to_delete:
-        project.mergerequests.delete(mr_iid)
-        try:
-            project.branches.delete(source_branch)
-        except gitlab.exceptions.GitlabDeleteError:
-            # Ignore if branch was already deleted
-            pass
+    for object in to_delete:
+        helpers.safe_delete(object)
 
 
 @pytest.fixture(scope="module")
@@ -434,11 +419,8 @@ def user(gl):
 
     yield user
 
-    try:
-        # Use `hard_delete=True` or a 'Ghost User' may be created.
-        user.delete(hard_delete=True)
-    except gitlab.exceptions.GitlabDeleteError as e:
-        print(f"User already deleted: {e}")
+    # Use `hard_delete=True` or a 'Ghost User' may be created.
+    helpers.safe_delete(user, hard_delete=True)
 
 
 @pytest.fixture(scope="module")
--- a/tests/functional/helpers.py
+++ b/tests/functional/helpers.py
@@ -1,0 +1,39 @@
+import logging
+import time
+
+import pytest
+
+import gitlab.base
+
+SLEEP_INTERVAL = 0.5
+TIMEOUT = 60  # seconds before timeout will occur
+MAX_ITERATIONS = int(TIMEOUT / SLEEP_INTERVAL)
+
+
+def safe_delete(
+    object: gitlab.base.RESTObject,
+    *,
+    hard_delete: bool = False,
+) -> None:
+    """Ensure the object specified can not be retrieved. If object still exists after
+    timeout period, fail the test"""
+    manager = object.manager
+    for index in range(MAX_ITERATIONS):
+        try:
+            object = manager.get(object.get_id())
+        except gitlab.exceptions.GitlabGetError:
+            return
+
+        if index:
+            logging.info(f"Attempt {index+1} to delete {object!r}.")
+        try:
+            if hard_delete:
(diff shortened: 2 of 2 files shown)


JohnVillalovos:
Example of error in tests that this is trying to fix:
https://github.com/python-gitlab/python-gitlab/runs/7123034880?check_suite_focus=true

```
        group1 = gl.groups.create({"name": "group1", "path": "group1"})
        group2 = gl.groups.create({"name": "group2", "path": "group2"})
        p_id = gl.groups.list(search="group2")[0].id
        group3 = gl.groups.create({"name": "group3", "path": "group3", "parent_id": p_id})
        group4 = gl.groups.create({"name": "group4", "path": "group4"})
>       assert len(gl.groups.list()) == 4
E       assert 5 == 4
E        +  where 5 = len([<Group id:35 name:group1>, <Group id:36 name:group2>, <Group id:37 name:group3>, <Group id:38 name:group4>, <Group id:6 name:test-group-45bec8b04d964c3fafd743afcb68f528>])
E        +    where [<Group id:35 name:group1>, <Group id:36 name:group2>, <Group id:37 name:group3>, <Group id:38 name:group4>, <Group id:6 name:test-group-45bec8b04d964c3fafd743afcb68f528>] = <bound method ListMixin.list of <gitlab.v4.objects.groups.GroupManager object at 0x7f1e1b486ce0>>()
E        +      where <bound method ListMixin.list of <gitlab.v4.objects.groups.GroupManager object at 0x7f1e1b486ce0>> = <gitlab.v4.objects.groups.GroupManager object at 0x7f1e1b486ce0>.list
E        +        where <gitlab.v4.objects.groups.GroupManager object at 0x7f1e1b486ce0> = <gitlab.client.Gitlab object at 0x7f1e1b485d80>.groups
tests/functional/api/test_groups.py:31: AssertionError
```

JohnVillalovos:
@nejch FYI: I'd like to get this in before trying to get #1778 in. Also #1785 should get in too.

nejch:
> @nejch FYI: I'd like to get this in before trying to get #1778 in. Also #1785 should get in too.

Thanks @JohnVillalovos I'll try to get back into the functional testing topic a bit.. I think we should make this reliable for sure, although it also shows we have some poor assertions in those legacy tests there.

IMO we should never be asserting `len()` on any lists, just asserting on membership of those objects in the lists. That way our tests also wouldn't be as affected by async gitlab operations. E.g. if we create/delete a resource we shouldn't care if there are 9 or 10 of them on the server, only if that one exists or not :) Anyway, just some ramblings before I forget this, probably belongs in a separate issue :D

tests/functional/conftest.py
```diff
@@ -191,6 +191,33 @@ def _wait(timeout=30, step=0.5):
     return _wait
 
 
+def wait_for_deleted(
+    *,
+    pg_manager: gitlab.base.RESTManager,
+    object_id: int,
+    description: str,
```
nejch, line 198: I think we could potentially just take the description from `object.__name__`, that way we also explicitly know we're deleting via `ProjectMergeRequest` not `MergeRequest` etc.

tests/functional/conftest.py
```diff
@@ -293,6 +323,7 @@ def project(gl):
     name = f"test-project-{_id}"
 
     project = gl.projects.create(name=name)
+    project_id = project.id
```
nejch, line 326: Are these changes to do with pass-by-assignment? Asking because a lot of our docs and tests have been happily passing `resource.id` directly to methods for a while it seems with no issue.

tests/functional/conftest.py
```diff
@@ -191,6 +191,33 @@ def _wait(timeout=30, step=0.5):
     return _wait
 
 
+def wait_for_deleted(
```
nejch, line 194: In general, this now duplicates the deletion logic in all our fixtures where we do delete and then call this. So I would remove those deletes and add this as the only finalizer and maybe call it e.g. `safe_delete` or so?

nejch (changes requested):

tests/functional/conftest.py
```diff
@@ -191,6 +191,33 @@ def _wait(timeout=30, step=0.5):
...
+            if hard_delete:
+                object.delete(hard_delete=True)
+            else:
+                object.delete()
+        except gitlab.exceptions.GitlabDeleteError:
+            pass
+        time.sleep(SLEEP_INTERVAL)
+    pytest.fail(f"{description} {object_id} was not deleted")
+
+
```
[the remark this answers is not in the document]
JohnVillalovos: That sounds good to me.

tests/functional/conftest.py
nejch: Sorry, I meant `object.__class__.__name__` here, purely to avoid having to define a description argument manually every time because it's nearly identical. Adding a description manually is a bit error prone to typos, e.g. as in https://github.com/python-gitlab/python-gitlab/pull/1784/files#diff-9810a028f51682f62bc400e8654365eca614e1905cc524507818b9a22329dca8R366.

Might also give us back those 1-line function signatures :grin:

tests/functional/conftest.py
```diff
@@ -277,13 +275,11 @@ def group(gl):
         "path": f"group-{_id}",
     }
     group = gl.groups.create(data)
+    group_id = group.id
```
nejch, line 278:
```suggestion
```

tests/functional/helpers.py
```diff
@@ -0,0 +1,44 @@
...
+    except gitlab.exceptions.GitlabDeleteError:
+        logging.info(f"{object!r} already deleted.")
+        pass
+
+
+def safe_delete(
+    *,
+    manager: gitlab.base.RESTManager,
+    object_id: int,
+    description: str,
```
nejch, line 31: I just realized we basically don't need to pass the manager, id and description actually. Each object has a `self.manager`, `self.id` and `self.__class__.__name__` so we could potentially just pass the object to this and handle it inside the delete function - could be less error prone.

tests/functional/conftest.py
```diff
@@ -431,14 +430,14 @@ def user(gl):
     password = "fakepassword"
 
     user = gl.users.create(email=email, username=username, name=name, password=password)
+    user_id = user.id
```
nejch, line 433:
```suggestion
```

JohnVillalovos: Done

tests/functional/conftest.py
JohnVillalovos: Done. I think. Only passing the `object` now.

tests/functional/conftest.py
```diff
@@ -359,18 +351,14 @@ def _merge_request(*, source_branch: str):
...
-    for mr_iid, source_branch in to_delete:
-        project.mergerequests.delete(mr_iid)
-        try:
-            project.branches.delete(source_branch)
-        except gitlab.exceptions.GitlabDeleteError:
-            # Ignore if branch was already deleted
-            pass
+    for mr, mr_branch in to_delete:
+        helpers.safe_delete(object=mr)
+        helpers.safe_delete(object=mr_branch)
```
nejch, line 361: I think we don't need a list of tuples even - the order of deletion shouldn't matter. I guess we could just have:

```python
        to_delete += mr, mr_branch
        return mr

    yield _merge_request

    for resource in to_delete:
        helpers.safe_delete(object=resource)
```

tests/functional/conftest.py
```diff
@@ -434,11 +422,8 @@ def user(gl):
 
     yield user
 
-    try:
-        # Use `hard_delete=True` or a 'Ghost User' may be created.
-        user.delete(hard_delete=True)
-    except gitlab.exceptions.GitlabDeleteError as e:
-        print(f"User already deleted: {e}")
+    # Use `hard_delete=True` or a 'Ghost User' may be created.
+    helpers.safe_delete(object=user, hard_delete=True)
```
nejch, line 426: I'm still not sure whether
`helpers.safe_delete(object=user, hard_delete=True)`
is clearer than
`helpers.safe_delete(user, hard_delete=True)`

I guess a good example of my thinking is Python's own `list.sort()` vs `sorted()` that do almost the same thing:
if you're passing an obvious object to it then that can be positional IMO, whereas options are enforced as keyword-only.
See their docs on how that's explicitly enforced here:
https://docs.python.org/3/library/functions.html#sorted
https://docs.python.org/3/library/stdtypes.html#list.sort

nejch:
Thanks again @JohnVillalovos, just a few more random thoughts as I like that we're potentially even reducing the codebase here :)

tests/functional/helpers.py
```diff
@@ -0,0 +1,43 @@
...
+    max_iterations = int(TIMEOUT / SLEEP_INTERVAL)
+    manager = object.manager
+    for index in range(max_iterations):
+        try:
+            object = manager.get(object.get_id())
+        except gitlab.exceptions.GitlabGetError:
+            return
+        delete_object(object=object, index=index, hard_delete=hard_delete)
+        time.sleep(SLEEP_INTERVAL)
+    pytest.fail(f"{object!r} was not deleted")
```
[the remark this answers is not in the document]
JohnVillalovos: Done. Not exactly like that. But one function.

tests/functional/conftest.py
```diff
@@ -359,18 +349,13 @@ def _merge_request(*, source_branch: str):
             time.sleep(0.5)
         assert mr.merge_status != "checking"
 
-        to_delete.append((mr.iid, source_branch))
+        to_delete.extend([mr, mr_branch])
```
JohnVillalovos, line 352: FYI: `to_delete += mr, mr_branch` doesn't work or at least flake8 complains because `to_delete` is defined outside the scope of this nested function.

tests/functional/conftest.py
nejch: Haha I guess that's my real question there if that's really always desired.

tests/functional/conftest.py
nejch: I think they can still be deleted after closing but looks good like this to me!

tests/functional/helpers.py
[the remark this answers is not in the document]
nejch: Cool! I still think we don't need to re-assign the manager or the object. We don't need the side effect of the new `get()`. But LGTM!

PR #1786, opened 2021-12-30
chore: add logging to `tests/functional/conftest.py`

JohnVillalovos:
I have found trying to debug issues in the functional tests can be
difficult. Especially when trying to figure out failures in the CI
running on Github.

Add logging to `tests/functional/conftest.py` to have a better
understanding of what is happening during a test run which is useful
when trying to troubleshoot issues in the CI.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -90,3 +90,10 @@ disable = [
 
 [tool.pytest.ini_options]
 xfail_strict = true
+
+# If 'log_cli=True' the following apply
+# NOTE: If set 'log_cli_level' to 'DEBUG' will show a log of all of the HTTP requests
+# made in functional tests.
+log_cli_level = "INFO"
+log_cli_format = "%(asctime)s.%(msecs)03d [%(levelname)8s] (%(filename)s:%(funcName)s:L%(lineno)s) %(message)s"
+log_cli_date_format = "%Y-%m-%d %H:%M:%S"
--- a/tests/functional/conftest.py
+++ b/tests/functional/conftest.py
@@ -1,3 +1,4 @@
+import logging
 import tempfile
 import time
 import uuid
@@ -9,7 +10,7 @@ import pytest
 import gitlab
 import gitlab.base
 
-SLEEP_INTERVAL = 0.1
+SLEEP_INTERVAL = 0.5
 TIMEOUT = 60  # seconds before timeout will occur
 
 
@@ -21,17 +22,29 @@ def fixture_dir(test_dir):
 def reset_gitlab(gl):
     # previously tools/reset_gitlab.py
     for project in gl.projects.list():
+        logging.info(f"Marking for deletion project: {project.path_with_namespace!r}")
         for deploy_token in project.deploytokens.list():
+            logging.info(
+                f"Marking for deletion token: {deploy_token.username!r} in "
+                f"project: {project.path_with_namespace!r}"
+            )
             deploy_token.delete()
         project.delete()
     for group in gl.groups.list():
+        logging.info(f"Marking for deletion group: {group.full_path!r}")
         for deploy_token in group.deploytokens.list():
+            logging.info(
+                f"Marking for deletion token: {deploy_token.username!r} in "
+                f"group: {group.path_with_namespace!r}"
+            )
             deploy_token.delete()
         group.delete()
     for variable in gl.variables.list():
+        logging.info(f"Marking for deletion variable: {variable.key!r}")
         variable.delete()
     for user in gl.users.list():
         if user.username != "root":
+            logging.info(f"Marking for deletion user: {user.username!r}")
             user.delete(hard_delete=True)
 
     max_iterations = int(TIMEOUT / SLEEP_INTERVAL)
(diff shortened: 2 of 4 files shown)


pyproject.toml
```diff
@@ -87,3 +87,14 @@ disable = [
...
+[tool.pytest.ini_options]
+log_cli = true
+# NOTE: If set 'log_cli_level' to 'DEBUG' will show a log of all of the HTTP requests
+# made in functional tests.
+log_cli_level = "INFO"
+log_cli_format = "%(asctime)s.%(msecs)03d [%(levelname)8s] (%(filename)s:%(funcName)s:L%(lineno)s) %(message)s"
+log_cli_date_format = "%Y-%m-%d %H:%M:%S"
+# Force pytest to output color. Without this there is no color when running in Github
+# CI. Even though PY_COLORS=1 is set. Also tried FORCE_COLORS=1.
+addopts = "--color=yes"
```
nejch, line 100: I think I just forgot to add `PY_COLORS` to tox's passenv when doing this, so I think if we don't need it if we add it here:
https://github.com/python-gitlab/python-gitlab/blob/f26bf7d3a86e4d5d1a43423476a46a381e62e8f9/tox.ini#L7

Happens because tox sanitizes the environment I guess, which actually might be very helpful at some point with things like `CI_JOB_TOKEN` ... :grin:

JohnVillalovos: Genius! Thanks 👍

tests/functional/conftest.py
```diff
@@ -21,17 +22,29 @@ def fixture_dir(test_dir):
...
+                f"group: {group.name!r}"
+            )
             deploy_token.delete()
         group.delete()
     for variable in gl.variables.list():
+        logging.info(f"Mark for deletion variable: {variable.name!r}")
         variable.delete()
     for user in gl.users.list():
         if user.username != "root":
+            logging.info(f"Mark for deletion user: {user.username!r}")
```
nejch, line 47: Just a suggestion for unique identifiers (also don't think variables have a `name` attribute) and style:

```suggestion
        logging.info(f"Marking for deletion project: {project.path_with_namespace!r}")
        for deploy_token in project.deploytokens.list():
            logging.info(
                f"Marking for deletion token: {deploy_token.username!r} in "
                f"project: {project.path_with_namespace!r}"
            )
            deploy_token.delete()
        project.delete()
    for group in gl.groups.list():
        logging.info(f"Mark for deletion group: {group.path_with_namespace!r}")
        for deploy_token in group.deploytokens.list():
            logging.info(
                f"Marking for deletion token: {deploy_token.username!r} in "
                f"group: {group.path_with_namespace!r}"
            )
            deploy_token.delete()
        group.delete()
    for variable in gl.variables.list():
        logging.info(f"Marking for deletion variable: {variable.key!r}")
        variable.delete()
    for user in gl.users.list():
        if user.username != "root":
            logging.info(f"Marking for deletion user: {user.username!r}")
```

tests/functional/conftest.py
```diff
@@ -44,13 +57,27 @@ def wait_for_maximum_list_length(
     ) -> None:
         """Wait for the list() length to be no greater than expected maximum or fail
         test if timeout is exceeded"""
-        for _ in range(max_iterations):
-            if len(rest_manager.list()) <= max_length:
+        logging.info(f"Checking {description!r} has no more than {max_length} items")
```
nejch, line 60: Can we then make this function name shorter as we get log lines quite a bit longer?
```
2022-01-04 05:49:51.067 [    INFO] (conftest.py:wait_for_maximum_list_length:L60) Checking 'projects' has no more than 0 items
```

tests/functional/conftest.py
```diff
@@ -44,13 +57,27 @@ def wait_for_maximum_list_length(
...
+        items = rest_manager.list()
+        elapsed_time = time.perf_counter() - start_time
+        if len(items) > max_length:
+            logging.error(
+                f"Too many {description!r} items still remaining and timeout "
+                f"({elapsed_time}) exceeded: {[x.name for x in items]}"
+            )
+        assert len(items) <= max_length, (
+            f"Did not delete required items for {description!r}. "
+            f"Elapsed_time: {elapsed_time}"
```
nejch, line 80: Looks like we're duplicating a bit here, but not sure if we can assert and log at the same time. Maybe just a plain assert is enough though.

tests/functional/conftest.py
```diff
@@ -44,13 +57,27 @@ def wait_for_maximum_list_length(
...
+            logging.info(
+                f"Iteration: {count} Waiting for {description!r} items to be deleted: "
+                f"{[x.name for x in items]}"
+            )
             time.sleep(SLEEP_INTERVAL)
-        assert len(rest_manager.list()) <= max_length, (
-            f"Did not delete required items for {description}. "
-            f"Elapsed_time: {time.perf_counter() - start_time}"
+
+        items = rest_manager.list()
```
nejch, line 71: Don't we already have it from inside the loop before the break?

nejch (commented):
Thanks @JohnVillalovos! Just a few comments for discussion from me.

> I have found trying to debug issues in the functional tests can be difficult. Especially when trying to figure out failures in the CI running on Github.
> 
> Add logging to `tests/functional/conftest.py` to have a better understanding of what is happening during a test run which is useful when trying to troubleshoot issues in the CI.

Just a note here, I prefer debugging locally as otherwise watchers can get spammed with notifications on many pushes. But I understand sometimes it's CI-only issues and then it's hard to debug so this makes sense.

tests/functional/conftest.py
```diff
@@ -163,9 +194,11 @@ def gitlab_config(check_is_alive, docker_ip, docker_services, temp_dir, fixture_
     config_file = temp_dir / "python-gitlab.cfg"
     port = docker_services.port_for("gitlab", 80)
 
+    logging.info("Waiting for GitLab container to become ready.")
     docker_services.wait_until_responsive(
         timeout=200, pause=5, check=lambda: check_is_alive("gitlab-test")
     )
+    logging.info("GitLab container is now ready.")
```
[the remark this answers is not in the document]
JohnVillalovos: Sounds good.

pyproject.toml
```diff
@@ -90,3 +90,10 @@ disable = [
 
 [tool.pytest.ini_options]
 xfail_strict = true
+
+log_cli = true
+# NOTE: If set 'log_cli_level' to 'DEBUG' will show a log of all of the HTTP requests
+# made in functional tests.
```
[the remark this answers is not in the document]
JohnVillalovos: Moved it to the workflow yaml file.

tox.ini
```diff
@@ -4,7 +4,9 @@ skipsdist = True
 envlist = py310,py39,py38,py37,pep8,black,twine-check,mypy,isort
 
 [testenv]
-passenv = GITLAB_IMAGE GITLAB_TAG
+# NOTE: PY_COLORS, NO_COLOR, and FORCE_COLOR are used by pytest, so pass them through if
+# set. https://docs.pytest.org/en/6.2.x/reference.html
```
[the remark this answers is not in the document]
JohnVillalovos: I was trying to explain where PY_COLORS is coming and why we are passing it. So future people will know why it is there and can hopefully make an informed decision on changing things. I know sometimes in code it can be worrisome to change things as we don't know why something is there and what it will break if it is changed.

But I can delete the comment too.

tests/functional/conftest.py
```diff
@@ -188,7 +221,9 @@ def gitlab_config(check_is_alive, docker_ip, docker_services, temp_dir, fixture_
 def gl(gitlab_config):
     """Helper instance to make fixtures and asserts directly via the API."""
 
+    logging.info("Create python-gitlab gitlab.Gitlab object")
```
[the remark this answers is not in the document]
JohnVillalovos: Done.

tests/functional/conftest.py
JohnVillalovos: Done

tests/functional/conftest.py
JohnVillalovos: Sure. I changed it to `wait_for_list_size`

tests/functional/conftest.py
JohnVillalovos: I made an error_message variable. Cleaner now.

tests/functional/conftest.py
JohnVillalovos: Good catch!

JohnVillalovos:
> Just a note here, I prefer debugging locally as otherwise watchers can get spammed with notifications on many pushes. But I understand sometimes it's CI-only issues and then it's hard to debug so this makes sense.

I personally have found this quite useful for local debugging too. I like having a progress indication for one. Even though it doesn't give me an ETA at least I know it is doing something.

tests/functional/conftest.py
JohnVillalovos: I guess 'groups' don't have `path_with_namespace` but they do have a `full_path` attribute. Using that.

tox.ini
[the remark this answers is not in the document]
nejch: I guess I meant that we might need to keep updating this then as we add more variables to pass, and that even adding it here it's possible it started changing for other tools not just pytest :) But I didn't check. Feel free to resolve this otherwise :)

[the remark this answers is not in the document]
JohnVillalovos: I went ahead and removed it.

PR #1787, opened 2021-12-30
test: drop httmock dependency in test_gitlab.py

nejch:
Explicitly registering the responses inside the test cases ended up actually being a little less verbose than writing dedicated fixtures, and also helps with header cleanup in  #1788 as well as reducing the code changes migrating from httmock.


JohnVillalovos (approved):
Thanks. Looks good.

PR #1788, opened 2021-12-30
test: reproduce missing pagination headers in tests 

nejch:
Built on top of https://github.com/python-gitlab/python-gitlab/pull/1787.

I tried parametrizing tests at first but this seemed cleaner in the end, it would get overly complicated because of https://github.com/pytest-dev/pytest/issues/349.

### DIFF
--- a/tests/unit/test_gitlab.py
+++ b/tests/unit/test_gitlab.py
@@ -18,6 +18,7 @@
 
 import pickle
 import warnings
+from copy import deepcopy
 
 import pytest
 import responses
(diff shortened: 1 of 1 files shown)


tests/unit/test_gitlab.py
```diff
@@ -18,6 +18,7 @@
 
 import pickle
 import warnings
+from copy import deepcopy
```
JohnVillalovos, line 21: nit: My personal taste is `import copy` and then use `copy.deepcopy()` as when reading the code quickly know where it comes from.

JohnVillalovos (approved):
Looks good to me.

Left a nit. But good to merge for me.

Issue #1649, opened 2021-10-22
Set up an action to lock very old closed issues


nejch:
Something like https://github.com/actions/stale, so there is no confusion with old issues (pre-2018) that had a completely different environment from now.

nejch:
> Yes please. I'd also vote for closing stale issues automatically. If no activity for X months or so just automatically close an issue.
> 
> X = ??? Not sure how long.

Yeah agreed that'd be great too, although I think having a label like `confirmed` would be good to prevent valid issues from getting closed. I don't think all of them should closed if we know we really need to fix them.

PR #1789, opened 2021-12-31
chore: add a stale workflow

JohnVillalovos:
Use the stale action to close issues and pull-requests with no
activity.

Issues: It will mark them as stale after 60 days and then close
them once they have been stale for 15 days.

Pull-Requests: It will mark pull-requests as stale after 90 days and then close
them once they have been stale for 15 days.

https://github.com/actions/stale

Closes: #1649

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,40 @@
+# https://github.com/actions/stale
+name: 'Close stale issues and PRs'
+on:
+  schedule:
+    - cron: '30 1 * * *'
+
+permissions:
+  issues: write
+  pull-requests: write
+
+concurrency:
+  group: lock
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v4
+        with:
+          any-of-labels: 'need info,Waiting for response'
+          stale-issue-message: >
+            This issue was marked stale because it has been open 60 days with no
+            activity. Please remove the stale label or comment on this issue. Otherwise,
+            it will be closed in 15 days.
+          days-before-issue-stale: 60
+          days-before-issue-close: 15
+          close-issue-message: >
+            This issue was closed because it has been marked stale for 15 days with no
+            activity. If this issue is still valid, please re-open.
+
+          stale-pr-message: >
+            This Pull Request (PR) was marked stale because it has been open 90 days
+            with no activity.  Please remove the stale label or comment on this PR.
+            Otherwise, it will be closed in 15 days.
+          days-before-pr-stale: 90
+          days-before-pr-close: 15
+          close-pr-message: >
+            This PR was closed because it has been marked stale for 15 days with no
+            activity. If this PR is still valid, please re-open.
+


nejch:
I like this a lot! I'm still thinking about my comment from https://github.com/python-gitlab/python-gitlab/issues/1649#issuecomment-962690690. I'd limit it to issues and PRs where we're getting stuck because of user inaction, with no clarification - and not those that are confirmed bugs/valid requests but we haven't done anything about them. Does that make sense?

For that we could maybe use `only-*` or `exempt-*` with some of our existing labels, or think of something new. We do use `awaiting feedback` quite a lot, that would be a good candidate for example.

I'd also make it a bit more lenient (maybe 90 days for stale, 15 days to close) in case people are on vacation? Sometimes even I don't respond to an issue within 30 days.. :smile_cat:

JohnVillalovos:
> I like this a lot! I'm still thinking about my comment from [#1649 (comment)](https://github.com/python-gitlab/python-gitlab/issues/1649#issuecomment-962690690). I'd limit it to issues and PRs where we're getting stuck because of user inaction, with no clarification - and not those that are confirmed bugs/valid requests but we haven't done anything about them. Does that make sense?

> For that we could maybe use `only-*` or `exempt-*` with some of our existing labels, or think of something new. We do use `awaiting feedback` quite a lot, that would be a good candidate for example.

That could work. Just need to figure out what we want to use. But I would use `exempt-*` as want this automated not something where we need to tag something before it will get automatically moved to stale.

> I'd also make it a bit more lenient (maybe 90 days for stale, 15 days to close) in case people are on vacation? Sometimes even I don't respond to an issue within 30 days.. 😸

90 days! Seems so long 😁 But yeah can do that.

.github/workflows/stale.yml
```diff
@@ -0,0 +1,28 @@
...
+
+concurrency:
+  group: lock
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v4
+        with:
```
nejch, line 19: Should we start like this? These are the most offending ones IMO, and then if it works we can do some cleanup and expand to more issues/PRs.

```suggestion
        with:
          any-of-labels: need info,Waiting for response
```

.github/workflows/stale.yml
```diff
@@ -0,0 +1,28 @@
...
+          stale-issue-message: 'This issue is stale because it has been open 60 days with no activity. Remove stale label or comment or this will be closed in 15 days.'
+          days-before-issue-stale: 60
+          days-before-issue-close: 15
+          close-issue-message: 'This issue was closed because it has been stalled for 15 days with no activity. If this issue is still valid please re-open'
+
+          stale-pr-message: 'This PR is stale because it has been open 90 days with no activity. Remove stale label or comment or this will be closed in 15 days.'
+          days-before-pr-stale: 90
+          days-before-pr-close: 15
+          close-pr-message: 'This PR was closed because it has been stalled for 15 days with no activity. If this PR is still valid please re-open'
```
nejch, line 28: Just a few style suggestions here but feel free to reformulate. Since they'll be proper comments I guess we have space for more text :)

```suggestion
          stale-issue-message: 'This issue was marked stale because it has been open 60 days with no activity. Please remove the stale label or comment on this issue. Otherwise, this will be closed in 15 days.'
          days-before-issue-stale: 60
          days-before-issue-close: 15
          close-issue-message: 'This issue was closed because it has been marked stale for 15 days with no activity. If this issue is still valid, please re-open.'

          stale-pr-message: 'This PR was marked stale because it has been open 90 days with no activity. Please remove the stale label or comment on this PR. Otherwise, this will be closed in 15 days.'
          days-before-pr-stale: 90
          days-before-pr-close: 15
          close-pr-message: 'This PR was closed because it has been marked stale for 15 days with no activity. If this PR is still valid, please re-open.'
```

nejch (commented):
I think we can go ahead with this with a few labels as suggested above @JohnVillalovos WDYT?

.github/workflows/stale.yml
JohnVillalovos: Okay. I wanted to go with `exempt-*` but I guess starting with a white list of what to mark stale works too 😊

.github/workflows/stale.yml
JohnVillalovos: Done. Thanks.

max-wittig (approved):

Issue #1498, opened 2021-06-03
Suggestion: url encode project paths automatically


chris-peterson:
## Description of the problem, including code/CLI snippet

Would be nice to not have to url-encode project paths when calling project APIs.  Feels pretty safe to replace `/` with `%2F` and not break folks

## Expected Behavior

Allow project paths to contain `/` without any escaping/encoding

## Actual Behavior

Results in 404s, project not found.  have to url encode and try again.

## Specifications

  - python-gitlab version: 2.7.1
  - API version you are using (v3/v4): v4
  - Gitlab server version (or gitlab.com): 13.10.3

nejch:
Hi @chris-peterson thanks, can you give some examples of endpoints that don't allow project paths to have `/`?

Things like `gl.projects.get("path/to/project")` should already work. Or did I misunderstand the issue? :)

chris-peterson:
@nejch -- I am grateful that the get function will accept non-encoded project ids.  the suggestion is for other APIs, e.g. updating a project.  here's an example:

works:
`gitlab project update --id 'mygroup%2Fmyproject' --tag-list 'some value'`

fails:
`gitlab project update --id 'mygroup/myproject' --tag-list 'some value'`

my workaround so far has been to either:
1) url encode project ids before calling APIs OR
2) always look up the integer ID first and pass that instead

nejch:
Thanks! I see, so this is mostly an issue with the CLI as with the API you would usually first get the object (even if lazily), but we should update this, it should be a few one-liners. Thanks for the report!

Related to #1356

Issue #783, opened 2019-05-28
inconsistency with project id option


glensc:
## Description of the problem, including code/CLI snippet

GitLab v3/v4 API support numeric project id and project path as arguments but seems the python-gitlab support only numeric argument with exception of the project command.

## Expected Behavior

Without having to create two roundtrips (or implement caching), I would expect the CLI API also support project name as project id argument.

Perhaps the `project_id` parameter is not properly [urlencoded](https://docs.gitlab.com/ce/api/#namespaced-path-encoding)?

## Actual Behavior

```
$ gitlab project get --id=group-name/project-name
id: 813
path: project-name

➔ gitlab project get --id=813
id: 813
path: project-name

➔ gitlab project-variable list --project-id=813
key: MYSQL_PASSWORD

➔ gitlab project-variable list --project-id=group-name/project-name
Impossible to list objects (404: 404 Not Found)
```

## Specifications

  - python-gitlab version: 1.8.0
  - API version you are using (v3/v4): v4
  - Gitlab server version (or gitlab.com): 11.10.4

glensc:
> Perhaps the `project_id` parameter is not properly [urlencoded](https://docs.gitlab.com/ce/api/#namespaced-path-encoding)?

yep. it's true:

```
➔ gitlab -d project-variable list --project-id=group-name/project-name
send: 'GET /api/v4/projects/group-name/project-name/variables?project_id=group-name%2Fproject-name HTTP/1.1\r...
```

the requested url should be:
```diff
-GET /api/v4/projects/group-name/project-name/variables?project_id=group-name%2Fproject-name HTTP/1.1
+GET /api/v4/projects/group-name%2Fproject-name/variables HTTP/1.1
````

also, the project_id parameter is superfluous, so I removed it in the example.

PR #1790, opened 2022-01-02
fix(cli): url-encode path components of the URL

JohnVillalovos:
In the CLI we need to make sure the components put into the path
portion of the URL are url-encoded. Otherwise they will be interpreted
as part of the path. For example can specify the project ID as a path,
but in the URL it must be url-encoded or it doesn't work.

Also stop adding the components of the path as query parameters in the
URL.

Closes: #783
Closes: #1498

### DIFF
--- a/gitlab/v4/cli.py
+++ b/gitlab/v4/cli.py
@@ -39,6 +39,7 @@ class GitlabCLI(object):
         self.action = action.lower()
         self.gl = gl
         self.args = args
+        self.parent_args: Dict[str, Any] = {}
         self.mgr_cls: Union[
             Type[gitlab.mixins.CreateMixin],
             Type[gitlab.mixins.DeleteMixin],
@@ -53,7 +54,10 @@ class GitlabCLI(object):
         # the class _path attribute, and replace the value with the result.
         if TYPE_CHECKING:
             assert self.mgr_cls._path is not None
-        self.mgr_cls._path = self.mgr_cls._path.format(**self.args)
+
+        self._process_from_parent_attrs()
+
+        self.mgr_cls._path = self.mgr_cls._path.format(**self.parent_args)
         self.mgr = self.mgr_cls(gl)
 
         if self.mgr_cls._types:
@@ -62,6 +66,18 @@ class GitlabCLI(object):
                     obj = type_cls()
                     obj.set_from_cli(self.args[attr_name])
                     self.args[attr_name] = obj.get()
+
+    def _process_from_parent_attrs(self) -> None:
+        """Items in the path need to be url-encoded. There is a 1:1 mapping from
+        mgr_cls._from_parent_attrs <--> mgr_cls._path. Those values must be url-encoded
+        as they may contain a slash '/'."""
+        for key in self.mgr_cls._from_parent_attrs:
+            if key not in self.args:
+                continue
+
+            self.parent_args[key] = gitlab.utils.clean_str_id(self.args[key])
+            # If we don't delete it then it will be added to the URL as a query-string
+            del self.args[key]
 
     def __call__(self) -> Any:
         # Check for a method that matches object + action
@@ -85,7 +101,7 @@ class GitlabCLI(object):
             data = {}
             if self.mgr._from_parent_attrs:
                 for k in self.mgr._from_parent_attrs:
-                    data[k] = self.args[k]
+                    data[k] = self.parent_args[k]
             if not issubclass(self.cls, gitlab.mixins.GetWithoutIdMixin):
                 if TYPE_CHECKING:
                     assert isinstance(self.cls._id_attr, str)
--- a/tests/functional/cli/test_cli_variables.py
+++ b/tests/functional/cli/test_cli_variables.py
@@ -1,3 +1,12 @@
+import copy
+
+import pytest
+import responses
+
+from gitlab import config
+from gitlab.const import DEFAULT_URL
+
+
 def test_list_instance_variables(gitlab_cli, gl):
     cmd = ["variable", "list"]
     ret = gitlab_cli(cmd)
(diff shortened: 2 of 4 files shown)


gitlab/v4/cli.py
```diff
@@ -53,7 +54,15 @@ def __init__(
...
-        self.mgr_cls._path = self.mgr_cls._path.format(**self.args)
+        # Items in the path need to be url-encoded
+        if self.mgr_cls._from_parent_attrs:
+            for k in self.mgr_cls._from_parent_attrs:
+                if k in self.args:
+                    self.parent_args[k] = gitlab.utils.clean_str_id(self.args[k])
+                    # If we don't delete it then it will be added to the URL as a
+                    # query-string
+                    del self.args[k]
+        self.mgr_cls._path = self.mgr_cls._path.format(**self.parent_args)
```
nejch, line 65: Cool! Maybe we could extract this into a helper method to keep `__init__` minimal. Then the comments can be a docstring instead.

tests/functional/cli/test_cli_variables.py
```diff
@@ -17,3 +17,10 @@ def test_list_project_variables(gitlab_cli, project):
     ret = gitlab_cli(cmd)
 
     assert ret.success
+
+
+def test_list_project_variables_with_path(gitlab_cli, project):
+    cmd = ["project-variable", "list", "--project-id", project.path_with_namespace]
+    ret = gitlab_cli(cmd)
+
+    assert ret.success
```
nejch, line 26: It's probably asking a bit much but if we're making sure we don't send extra query params, we could add a test that also matches the query params like I do with headers here:

https://github.com/python-gitlab/python-gitlab/blob/main/tests/functional/cli/test_cli.py

Look for `query_param_matcher` to see how I did the other one, I think.

We can do this in a follow-up also.

JohnVillalovos: Done

gitlab/v4/cli.py
JohnVillalovos: Done

gitlab/v4/cli.py
```diff
@@ -53,7 +54,10 @@ def __init__(
         # the class _path attribute, and replace the value with the result.
         if TYPE_CHECKING:
             assert self.mgr_cls._path is not None
-        self.mgr_cls._path = self.mgr_cls._path.format(**self.args)
+
+        self._process_from_parent_attributes()
```
nejch, line 58: suggestion: should we call it like the variable it processes? It will also keep it shorter then.
```suggestion
        self._process_from_parent_args()
```

tests/functional/cli/test_cli_variables.py
```diff
@@ -17,3 +26,59 @@ def test_list_project_variables(gitlab_cli, project):
...
+
+
+@pytest.mark.script_launch_mode("inprocess")
+@responses.activate
+def test_list_project_variables_with_path_url_check(
+    monkeypatch, script_runner, resp_get_project
+):
+    monkeypatch.setattr(config, "_DEFAULT_FILES", [])
+    resp_get_project_in_ci = copy.deepcopy(resp_get_project)
+    resp_get_project_in_ci.update(
```
nejch, line 56:
```suggestion
    resp_get_project_variables = copy.deepcopy(resp_get_project)
    resp_get_project_variables.update(
```

tests/functional/cli/test_cli_variables.py
```diff
@@ -17,3 +26,59 @@ def test_list_project_variables(gitlab_cli, project):
...
+
+@pytest.fixture
+def resp_get_project():
+    return {
+        "method": responses.GET,
+        "url": f"{DEFAULT_URL}/api/v4/projects/1",
+        "json": {"name": "name", "path": "test-path", "id": 1},
+        "content_type": "application/json",
+        "status": 200,
+    }
```
nejch, line 46: I think we can have this in `tests/functional/cli/conftest.py` and remove it from here and `tests/functional/cli/test_cli.py` as it will be made available for both of them. At least if it's exactly the same which I think it is.

gitlab/v4/cli.py
```diff
@@ -63,6 +67,17 @@ def __init__(
...
+    def _process_from_parent_attributes(self) -> None:
+        """Items in the path need to be url-encoded. There is a 1:1 mapping from
+        mgr_cls._from_parent_attrs <--> mgr_cls._path. Those values must be url-encoded
+        as they may contain a slash '/'."""
+        for key in self.mgr_cls._from_parent_attrs:
+            if key in self.args:
+                self.parent_args[key] = gitlab.utils.clean_str_id(self.args[key])
+                # If we don't delete it then it will be added to the URL as a
+                # query-string
+                del self.args[key]
```
nejch, line 79: I'm probably a bit annoying with this but I like to apply it even for small things, then the comment fits :yum: then the comment fits

> [Flat is better than nested.](https://www.python.org/dev/peps/pep-0020/)

```suggestion
        for key in self.mgr_cls._from_parent_attrs:
            if not key in self.args:
                continue

            self.parent_args[key] = gitlab.utils.clean_str_id(self.args[key])
            # If we don't delete it then it will be added to the URL as a query-string
            del self.args[key]
```

tests/functional/cli/test_cli_variables.py
```diff
@@ -17,3 +26,59 @@ def test_list_project_variables(gitlab_cli, project):
...
+            {
+                "variable_type": "env_var",
+                "key": "TEST_VARIABLE_2",
+                "value": "TEST_2",
+                "protected": False,
+                "masked": False,
+                "environment_scope": "*",
+            },
+        ]
+    )
```
nejch, line 78: I think that we might not even need to update this, since we only care about matching the URL being sent, not what we do with the data returned, right?

nejch (changes requested):
Thanks John. Just a few suggestions and could you maybe rename the commit message to say `fix(cli): ...` so that it's parsed as a scope?

gitlab/v4/cli.py
JohnVillalovos: Done. Though I used `self._process_from_parent_attrs` as the variable is `_from_parent_attrs`

tests/functional/cli/test_cli_variables.py
JohnVillalovos: Done

tests/functional/cli/test_cli_variables.py
JohnVillalovos: Done. Good idea.

gitlab/v4/cli.py
nejch: Ah sorry, was looking at the `self.parent_args` above, thanks!

gitlab/v4/cli.py
JohnVillalovos: I like it. Thanks. Done.

tests/functional/cli/test_cli_variables.py
JohnVillalovos: Well we need to make sure we return a list. I will make it an empty list.

nejch: ahh ok sure. ok, yeah that or just a subset of items is probably enough to keep the boilerplate down. Thanks!

JohnVillalovos:
> Thanks John. Just a few suggestions and could you maybe rename the commit message to say `fix(cli): ...` so that it's parsed as a scope?

Done!

PR #1793, opened 2022-01-02
chore: add authentication type to GitlabAuthenticationError

JohnVillalovos:
Add the type of authentication used to the GitlabAuthenticationError
exception. Hopefully this will make it easier to help user's debug
authentication issues they run into.

### DIFF
--- a/gitlab/client.py
+++ b/gitlab/client.py
@@ -96,6 +96,7 @@ class Gitlab(object):
         self.http_password = http_password
         self.oauth_token = oauth_token
         self.job_token = job_token
+        self.auth_type = ""
         self._set_auth_info()
 
         #: Create a session object for requests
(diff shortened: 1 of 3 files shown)


gitlab/client.py
```diff
@@ -722,6 +727,7 @@ def http_request(
                     response_code=result.status_code,
                     error_message=error_message,
                     response_body=result.content,
+                    auth_type=self.auth_type,
```
nejch, line 730: I think we could actually reuse `self.headers` and pass it (or just the dict keys) to the exception to infer the auth type, without adding custom variables here.

`http_username`/`http_password` is basically dead code as HTTP Basic auth has been out of GitLab since version 10 so we don't need to worry about that (I actually have a local draft that performs password-based OAuth login from that). So we're left with `Private-Token`, `Job-Token`, and `Authorization` (OAuth bearer) from headers. We could probably just pass the keys from the headers dict to not leak stuff accidentally.

gitlab/exceptions.py
```diff
@@ -52,7 +52,24 @@ def __str__(self) -> str:
...
+        super().__init__(
+            error_message=error_message,
+            response_code=response_code,
+            response_body=response_body,
+        )
+        self.auth_type = auth_type
+
+    def __str__(self) -> str:
+        if self.auth_type:
+            return f"{super().__str__()}: authentication_type: {self.auth_type}"
```
nejch, line 71: If we decide to infer the auth type automatically I'd maybe reformulate this a bit. Not sure exactly how, just `authentication_type` sounds like it's a variable name defined/passed somewhere. We can maybe check around how more verbose exceptions do it in cpython or some other libraries.

nejch (commented):
Thanks John! Just a few ideas from my side on this one :) At the beginning I wasn't quite sure about the use case for this, but for authentication errors might be handy with all the job token confusion. 

Especially with the job token we could add a link to the endpoints available or something (e.g. have a mapping for the error message).

PR #1796, opened 2022-01-02
docs: switch to Furo and refresh introduction pages

nejch:


docs/conf.py
```diff
@@ -57,11 +59,13 @@
...
 
 # The master toctree document.
-master_doc = "index"
+root_doc = "index"
 
 # General information about the project.
 project = "python-gitlab"
-copyright = "2013-2018, Gauvain Pocentek, Mika Mäenpää"
+copyright = (
+    f"2013-2018, Gauvain Pocentek, Mika Mäenpää.\n2018-{year}, python-gitlab team"
```
JohnVillalovos, line 67: Cool, adding the python-gitlab-team!

I saw an issue from Gauvain about doing this.

docs/index.rst
```diff
@@ -1,25 +1,18 @@
-.. python-gitlab documentation master file, created by
-   sphinx-quickstart on Mon Dec  8 15:17:39 2014.
-   You can adapt this file completely to your liking, but it should at least
-   contain the root `toctree` directive.
-
-Welcome to python-gitlab's documentation!
-=========================================
-
-Contents:
+.. include:: ../README.rst
```
nejch, line 1: This tries to deduplicate the intro a bit between our readme and our docs while making it look like a real website. I don't really like generating the whole TOC tree as a page if we already have it in the sidebar :thinking:

docs/conf.py
nejch: Hehe yeah I think we already did that somwhere, maybe setup.py. I was thinking in a follow-up I could clean this up a bit more.

nejch:
Ok, I'm not entirely sure about this yet, but it's ready for review. It does look nicer to me in general but I think we kind of need to restructure our docs more.

docs/cli-examples.rst
```diff
@@ -0,0 +1,168 @@
+############
```
JohnVillalovos, line 1: Probably would be good to have an example using `-d/--debug` but that can be a follow-up.

JohnVillalovos (approved):

JohnVillalovos:
Looks good. Thanks!

Issue #1715, opened 2021-11-26
Api return 404 when keyword contains point symbol


insoz:
if the branch/commit keyword contains point symbol `.` , the api will return 404 response

example:
`project.branches.get('stbale-1.x')`
`project.comits.get('stbale-1.x')`

------
In page: gl_objects/projects.rst

JohnVillalovos:
Can you run the following and provide the output?

```
#!/usr/bin/env python3

import gitlab
print("gitlab version:", gitlab.__version__)
print("gitlab location:", gitlab.__file__)
```

JohnVillalovos:
As a note, to help your debugging you can do the following, if you want:

```
import gitlab

gl = gitlab.Gitlab(... your values ...)
gl.enable_debug()
```

insoz:
> 

version: '2.10.1'
file: '/Users/insoz/code/ops-fast/venv/lib/python3.7/site-packages/gitlab/__init__.py'

JohnVillalovos:
Thanks. I have a suspicion it is because of:
https://github.com/python-gitlab/python-gitlab/blob/3225f2cfee740374ef36e5cd6796d2370d0e2344/gitlab/utils.py#L65

But not sure if we can just take that out or not. We really need to move to using `urllib.parse.quote()` for this stuff. Though `requests` does some it self and I haven't dug into exactly what we are responsible for and what `requests` will take care of.

JohnVillalovos:
Related issue: https://github.com/python-gitlab/python-gitlab/issues/1356

insoz:
Thanks! BTW. Do we have some plan to use `httpx` or something `asyncio` tool to asynchronous this lib ?

nejch:
> Thanks! BTW. Do we have some plan to use `httpx` or something `asyncio` tool to asynchronous this lib ?

Yes, `httpx` probably, though it still hasn't reached 1.0 and last I checked was missing some proxy or retry features or so, if I'm not mistaken, can't quite remember.
But there is an old PR here that already uses it if you really need it (likely missing lots of recent features): https://github.com/python-gitlab/python-gitlab/pull/1036. That'll need some major rework to update though once we're ready.

nejch:
@JohnVillalovos wasn't this fixed by https://github.com/python-gitlab/python-gitlab/commit/702e41dd0674e76b292d9ea4f559c86f0a99edfe?

JohnVillalovos:
[...]

PR #1800, opened 2022-01-04
chore: add test case to show branch name with period works

JohnVillalovos:
Add a test case to show that a branch name with a period can be
fetched with a `get()`

Closes: #1715

### DIFF
--- a/tests/functional/api/test_branches.py
+++ b/tests/functional/api/test_branches.py
@@ -1,0 +1,17 @@
+"""
+GitLab API:
+https://docs.gitlab.com/ee/api/branches.html
+"""
+
+
+def test_branch_name_with_period(project):
+    # Make sure we can create and get a branch name containing a period '.'
+    branch_name = "my.branch.name"
+    branch = project.branches.create({"branch": branch_name, "ref": "main"})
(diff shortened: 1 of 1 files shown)


tests/functional/api/test_branches.py
```diff
@@ -0,0 +1,17 @@
...
+    # Make sure we can create and get a branch name containing a period '.'
+    branch_name = "my.branch.name"
+    branch = project.branches.create({"branch": branch_name, "ref": "main"})
+    assert branch.name == branch_name
+
+    # Ensure we can get the branch
+    fetched_branch = project.branches.get(branch_name)
+    assert branch.id == fetched_branch.id
+
+    branch.delete()
```
nejch, line 17: I guess this isn't strictly needed as the project will be destroyed anyway and it wouldn't actually get deleted if an assert fails. But I'm ok with this as is :)

JohnVillalovos: On the plus side it helps prove that we can `branch.delete()`

nejch: Ah, that was the point of this :) ok, thanks

PR #1803, opened 2022-01-04
chore: add functional test of mergerequest.get()

JohnVillalovos:
Add a functional test of test mergerequest.get() and
mergerequest.get(..., lazy=True)

Closes: #1425

### DIFF
--- a/tests/functional/api/test_merge_requests.py
+++ b/tests/functional/api/test_merge_requests.py
@@ -32,6 +32,22 @@ def test_merge_requests(project):
     )
 
 
+def test_merge_requests_get(project, merge_request):
+    new_mr = merge_request(source_branch="test_get")
+    mr_iid = new_mr.iid
+    mr = project.mergerequests.get(mr_iid)
+    assert mr.iid == mr_iid
+    mr = project.mergerequests.get(str(mr_iid))
+    assert mr.iid == mr_iid
+
+
+def test_merge_requests_get_lazy(project, merge_request):
+    new_mr = merge_request(source_branch="test_get")
+    mr_iid = new_mr.iid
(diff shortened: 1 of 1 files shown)


tests/functional/api/test_merge_requests.py
```diff
@@ -32,6 +32,22 @@ def test_merge_requests(project):
     )
 
 
+def test_merge_requests_get(project, merge_request):
+    new_mr = merge_request(source_branch="test_get")
+    mr_iid = new_mr.iid
+    mr = project.mergerequests.get(mr_iid)
+    assert mr.iid == mr_iid
+    mr = project.mergerequests.get(str(mr_iid))
+    assert mr.iid == mr_iid
```
nejch, line 41: nitpick: I couldn't find in the linked issue why we're testing for `str(mr_iid)` as well, I'm fine with it though :)
But can we call it `test_merge_requests_get_with_str` and `test_merge_requests_get_with_lazy` then?

JohnVillalovos: Can do, but it is testing both with an integer and with a string...  That is why I just made it `test_merge_requests_get`

nejch: ok, and the string was specifically for the string from ID or did we want to test the path here?

JohnVillalovos: It was just me wanting to make sure we could pass either an `int` or a `str` as the value for `iid`. Why? I'm not sure, it seemed like a good idea six months ago when I first did this PR.

nejch: ok :)

PR #1807, opened 2022-01-05
Add support for Groups API method transfer.

sattlerc:


JohnVillalovos:
Thanks for the PR. Can you please squash the commits down to one commit. And make sure there is a commit message.

Also if possible unit and/or functional tests would be great.

Thanks again!

JohnVillalovos:
And updating the documentation on how to use it would also be useful.

sattlerc:
Note that the typo fix commit is unrelated to the feature commit. Should I remove it?

I looked at the documentation, but it seems to be autogenerated.

There are usage examples for a limited number of group methods in `docs/gl_objects/groups.rst`. The related group method `transfer_project` doesn't appear there, so I thought it appropriate not to add `transfer_group`.

gitlab/v4/objects/groups.py
```diff
@@ -93,12 +93,36 @@ def transfer_project(self, project_id: int, **kwargs: Any) -> None:
...
+            GitlabAuthenticationError: If authentication is not correct
+            GitlabTransferProjectError: If the group could not be transferred
+        """
+        path = f"/groups/{self.id}/transfer"
+
+        def data():
+            if group_id is not None:
+                yield ("group_id", group_id)
+
+        self.manager.gitlab.http_post(path, post_data=dict(data()), **kwargs)
```
JohnVillalovos, line 118: Does this work? I'm not sure why it is a function.

Also when `group_id` is `None` it returns `None`, but `dict(None)` doesn't work so it would fail on L118.

```suggestion
        post_data = {}
        if group_id is not None:
            post_data['group_id"] = group_id

        self.manager.gitlab.http_post(path, post_data=post_data, **kwargs)
```

Tests to validate this works would be great.

JohnVillalovos (changes requested):

gitlab/v4/objects/groups.py
sattlerc: It's functional style. When `group_id` is `None`, it returns the empty iterator, not `None`, so works correctly.

I'll use the imperative style.

JohnVillalovos: Thanks! Make it easier for our little brains on this side 😊

sattlerc:
I wasn't able to find tests for the related group method `transfer_project`. If you add tests for that one, I can base the ones for `transfer_group` on those.

JohnVillalovos (approved):

gitlab/v4/objects/groups.py
```diff
@@ -93,6 +93,28 @@ def transfer_project(self, project_id: int, **kwargs: Any) -> None:
         path = f"/groups/{self.id}/projects/{project_id}"
         self.manager.gitlab.http_post(path, **kwargs)
 
+    @cli.register_custom_action("Group", tuple(), ("group_id",))
+    @exc.on_http_error(exc.GitlabTransferProjectError)
+    def transfer_group(self, group_id: Optional[int] = None, **kwargs: Any) -> None:
+        """Transfer a project to this group.
```
nejch, line 99:
```suggestion
        """Transfer the group to a new parent group or turn the subgroup into
        a top-level group.
```

nejch (commented):
Thanks for this PR @sattlerc! Just another small comment from me.

thought: I also would almost prefer to just call it `transfer()` since the transfer direction is inverted from the `transfer_project()` in this same class, but unfortunately we have `transfer_project()` in the projects module already as well. Maybe that one could be deprecated and have both methods that do "transfer _this_ resource somewhere" called `transfer()`. Outside the scope of this PR but open to hear opinions here :)

And thanks for changing the functional code as this does look closer to the rest of the library e.g. further down in this module.
https://github.com/python-gitlab/python-gitlab/blob/9894b3580a7eb5c2e377c482820ff3210f913abe/gitlab/v4/objects/groups.py#L326-L331

nejch:
> I wasn't able to find tests for the related group method `transfer_project`. If you add tests for that one, I can base the ones for `transfer_group` on those.

I'll try and add some tests for `transfer_project` as well.

sattlerc:
I would prefer `transfer` as well. Should I change it?

gitlab/v4/objects/groups.py
sattlerc: Fixed.

nejch:
> I would prefer `transfer` as well (that's also the name of the API path). Should I change it?

Yeah, let's do that. This just looks nicer:

```python
group = gl.groups.get("gitlab-org/maintainers")
group.transfer("gitlab-com")
```

It also avoids us getting into an awkward situation if GitLab comes up with a "transfer a group into this group" endpoint at some point :sweat_smile: (as is the case with `transfer_project` here)

sattlerc:
Okay. I have have also renamed the exception from `GitlabTransferGroupError` to `GitlabGroupTransferError`.

nejch:
Thanks, LGTM @sattlerc, not sure if John has any more comments. I'll open an accompanying PR for project transfer and you can rebase after merge if you like, I'll ping here.

JohnVillalovos:
> not sure if John has any more comments. 

Only that tests would be nice. But I think you said you will work on setting up the test framework and then we can land this on top of that.

sattlerc:
Don't worry about it.

Issue #1805, opened 2022-01-05
Wrong return type annotation in ProjectCommit.refs()


robertschweizer:
## Description of the problem, including code/CLI snippet

`ProjectCommit.refs()` returns `List[Dict[str, Any]]`, but the type annotations say it returns `Union[Dict[str, Any], requests.Response]`.

## Specifications

  - python-gitlab version: 3.0.0
  - API version you are using (v3/v4): v4
  - Gitlab server version (or gitlab.com):  -

nejch:
Thanks for the report @robertschweizer. Looking at the endpoint, the method currently used is wrong anyway, because we might get a paginated response so we should be doing an `http_list()` instead. So we should revamp this a little.

nejch:
We probably also have the same issue with `ProjectCommit.diff()` and `ProjectCommit.merge_requests()` at least for typing, and not sure about pagination yet as it is not documented upstream.

JohnVillalovos:
Ah. I had assumed every API call returned a `Dict[str, Any]`. Okay will need to change things to handle this. Thanks for the report!

Issue #1231, opened 2020-11-10
Unable to get page count info when calling commit.diff()


hungluong5791:
## Description of the problem, including code/CLI snippet

Calling `commit.diff()` to get diff information for a commit only return `per_page` items or 20 by default. It seems `commit.diff()` does not take either `all` or `as_list` and only pagination works, but we don't know how many page there are except by manually checking the returned list size.

## Expected Behavior

Using `commit.diff()` should be able to give us all the diffs in a commit without having to manually checking the returned page size

## Actual Behavior

As described

## Specifications

  - python-gitlab version: 2.5.0
  - API version you are using (v3/v4): v4
  - Gitlab server version (or gitlab.com): gitlab.com

max-wittig:
Sounds like an upstream issue to me, as python-gitlab is just sending these query parameter. Have you tried it with curl as well?

hungluong5791:
@max-wittig Per this [reference](https://docs.gitlab.com/ee/api/commits.html) I don't think they have an `all` parameter for the diff endpoint. However since python-gitlab have its own generator mechanism (`as_list=False`) for paging convenience, I was thinking of adding it to the `Manager` for commit diffs as well?

I guess this is more like a feature request rather than a bug.

PR #1809, opened 2022-01-06
fix: change to `http_list` for some ProjectCommit methods

JohnVillalovos:
Fix the type-hints and use `http_list()` for the ProjectCommits methods:
   - diff()
   - merge_requests()
   - refs()

This will enable using the pagination support we have for lists.

Closes: #1805
Closes: #1231


nejch:
@JohnVillalovos I'd say this is a proper fix and not a chore especially as it fixes that older issue as well, could you change the commit message so we can trigger a fix release?

JohnVillalovos:
> @JohnVillalovos I'd say this is a proper fix and not a chore especially as it fixes that older issue as well, could you change the commit message so we can trigger a fix release?

Done!

PR #1812, opened 2022-01-07
chore: add `pprint()` and `pformat()` methods to RESTObject

JohnVillalovos:
This is useful in debugging and testing. As can easily print out the
values from an instance in a more human-readable form.

### DIFF
--- a/docs/api-usage.rst
+++ b/docs/api-usage.rst
@@ -179,6 +179,20 @@ resources. For example:
    project = gl.projects.get(1)
    project.star()
 
+You can print a Gitlab Object. For example:
+
+.. code-block:: python
+
+   project = gl.projects.get(1)
+   print(project)
+
+   # Or in a prettier format.
+   project.pprint()
+
+   # Or explicitly via `pformat()`. This is equivalent to the above.
+   print(project.pformat())
+
+
 Base types
 ==========
 
--- a/tests/unit/test_base.py
+++ b/tests/unit/test_base.py
@@ -201,3 +201,33 @@ class TestRESTObject:
         obj1 = FakeObject(fake_manager, {"attr1": "foo"})
         obj2 = FakeObject(fake_manager, {"attr1": "bar"})
         assert obj1 != obj2
+
+    def test_dunder_str(self, fake_manager):
+        fake_object = FakeObject(fake_manager, {"attr1": "foo"})
+        assert str(fake_object) == (
+            "<class 'tests.unit.test_base.FakeObject'> => {'attr1': 'foo'}"
+        )
+
+    def test_pformat(self, fake_manager):
+        fake_object = FakeObject(
+            fake_manager, {"attr1": "foo" * 10, "ham": "eggs" * 15}
+        )
+        assert fake_object.pformat() == (
+            "<class 'tests.unit.test_base.FakeObject'> => "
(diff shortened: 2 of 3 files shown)


tests/unit/test_base.py
```diff
@@ -201,3 +201,25 @@ def test_inequality_no_id(self, fake_manager):
...
+
+    def test_pprint(self, capfd, fake_manager):
+        obj1 = FakeObject(fake_manager, {"attr1": "foo"})
+        result = obj1.pprint()
+        assert result is None
+        stdout, stderr = capfd.readouterr()
+        assert stdout == (
+            "<class 'tests.unit.test_base.FakeObject'> => \n{'attr1': 'foo'}\n"
+        )
+        assert stderr == ""
```
nejch, line 225: I know we'll have to change it elsewhere later as well but I think you might like this suggestion to `fake_object` :grin:
```suggestion
    def test_dunder_str(self, fake_manager):
        fake_object = FakeObject(fake_manager, {"attr1": "foo"})
        assert str(fake_object) == (
            "<class 'tests.unit.test_base.FakeObject'> => {'attr1': 'foo'}"
        )

    def test_pformat(self, fake_manager):
        fake_object = FakeObject(fake_manager, {"attr1": "foo"})
        assert fake_object.pformat() == (
            "<class 'tests.unit.test_base.FakeObject'> => \n{'attr1': 'foo'}"
        )

    def test_pprint(self, capfd, fake_manager):
        fake_object = FakeObject(fake_manager, {"attr1": "foo"})
        result = fake_object.pprint()
        assert result is None
        stdout, stderr = capfd.readouterr()
        assert stdout == (
            "<class 'tests.unit.test_base.FakeObject'> => \n{'attr1': 'foo'}\n"
        )
        assert stderr == ""
```

Also was thinking if we should do `FakeObject(fake_manager, {"attr1": "foo", "attr2": "bar"})` to show it pretty-prints dicts with new-lines but not sure it's worth it. Up to you!

docs/api-usage.rst
```diff
@@ -179,6 +179,17 @@ resources. For example:
...
 
+You can print a Gitlab Object. For example:
+
+.. code-block:: python
+
+   project = gl.projects.get(1)
+   print(project)
+   # Or in a prettier format. These two are equivalent.
+   print(project.pformat())
+   project.pprint()
```
nejch, line 190: Just thinking if we should show them separately a little bit to emphasize it's 2 ways of doing the same thing rather than 2 steps.

```suggestion
   # Print in a prettier format.
   print(project.pformat())
   
   # Or explicitly via `pformat()`. This is equivalent to the above.
   project.pprint()
```

nejch (commented):
Just two tiny notes :)

tests/unit/test_base.py
JohnVillalovos: > I know we'll have to change it elsewhere later as well but I think you might like this suggestion to `fake_object` 😁

Works for me 😊

> Also was thinking if we should do `FakeObject(fake_manager, {"attr1": "foo", "attr2": "bar"})` to show it pretty-prints dicts with new-lines but not sure it's worth it. Up to you!

Done!

docs/api-usage.rst
JohnVillalovos: Done. Though I flipped `pprint` and `pformat` as I think that is what you meant.

PR #1813, opened 2022-01-08
chore(dist): add docs *.md files to sdist

derekschrock:
build_sphinx to fail due to setup.cfg warning-is-error

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,4 +1,4 @@
 include COPYING AUTHORS CHANGELOG.md requirements*.txt
 include tox.ini
 recursive-include tests *
-recursive-include docs *j2 *.py *.rst api/*.rst Makefile make.bat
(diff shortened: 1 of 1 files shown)


nejch:
Thanks for this @derekschrock! Does this come from https://github.com/freebsd/freebsd-ports/tree/main/devel/py-python-gitlab?

If so maybe I can add some tests to build docs from sdist to avoid this in the future. https://github.com/python-gitlab/python-gitlab/blob/main/tests/smoke/test_dists.py

derekschrock:
> Thanks for this @derekschrock! Does this come from https://github.com/freebsd/freebsd-ports/tree/main/devel/py-python-gitlab?

Yes, however it don't block [updating the port](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=261026) I can just add the file to the port until it's included in the sdist.

PR #1814, opened 2022-01-08
chore: fix missing comma

JohnVillalovos:
There was a missing comma which meant the strings were concatenated
instead of being two separate strings.


nejch:
Wow thanks for catching that! probably happened with `black` at some point.

JohnVillalovos:
> Wow thanks for catching that! probably happened with `black` at some point.

Yeah, I'm a genius on finding this stuff...

Or: https://codereview.doctor/python-gitlab/python-gitlab 😊

PR #1816, opened 2022-01-08
fix: remove custom URL encoding

JohnVillalovos:
We were using `str.replace()` calls to take care of URL encoding
issues.

Switch them to use our `utils._url_encode()` function which itself uses
`urllib.parse.quote()`

Closes: #1356

### DIFF
--- a/gitlab/utils.py
+++ b/gitlab/utils.py
@@ -15,8 +15,8 @@
 # You should have received a copy of the GNU Lesser General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import urllib.parse
 from typing import Any, Callable, Dict, Optional
-from urllib.parse import quote
 
 import requests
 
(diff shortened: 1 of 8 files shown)


gitlab/utils.py
```diff
@@ -15,8 +15,8 @@
 # You should have received a copy of the GNU Lesser General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import urllib.parse
```
nejch, line 18: :rofl: :+1:

gitlab/utils.py
```diff
@@ -57,7 +57,7 @@ def copy_dict(dest: Dict[str, Any], src: Dict[str, Any]) -> None:
 
 
 def clean_str_id(id: str) -> str:
-    return quote(id, safe="")
+    return urllib.parse.quote(id, safe="")
```
nejch, line 60: Just thinking: does it still make sense to call this `clean_str_id`? I think we could call it `urlencode()` or so? or maybe some other name that doesn't clash with https://docs.python.org/2/library/urllib.html#urllib.urlencode but applies more generally to what we do with it now.

gitlab/v4/objects/files.py
```diff
@@ -238,7 +239,7 @@ def raw(
         Returns:
             The file content
         """
-        file_path = file_path.replace("/", "%2F").replace(".", "%2E")
+        file_path = utils.clean_str_id(file_path)
```
nejch, line 242: Do we know for sure we don't need `%2E` from `.` in _all_ cases?

I guess this is similar to https://github.com/python-gitlab/python-gitlab/pull/1766 and the behavior was corrected in upstream gitlab. But I would feel a bit safer if we had a functional test for this to ensure no regressions. Maybe not for submodules below as that's a hassle, but for files, WDYT?

nejch (commented):
Thanks again John! Just a few questions.

gitlab/utils.py
JohnVillalovos: Done. changed to `url_encode`

gitlab/v4/objects/files.py
JohnVillalovos: I will work on adding a functional test to test this.

gitlab/utils.py
JohnVillalovos: 😊

gitlab/v4/objects/files.py
JohnVillalovos: So we are already testing a file with a period '.' here: https://github.com/python-gitlab/python-gitlab/blob/main/tests/functional/api/test_repository.py#L31

nejch:
As discussed let's call this a `fix` as I think it fixes a few issues including submodule updates for filepaths with dots, if I'm not mistaken.

gitlab/v4/objects/files.py
JohnVillalovos: Added a test for `raw()` using the filename `README.rst`. There are already tests for `create()`, `update()`, and `blame()` that use files with periods in them.

Issue #1750, opened 2021-12-15
`mergerequest.merge()` implicitly sets `should_remove_source_branch: false` since 2.7.0


carlfriedrich:
## Description of the problem, including code/CLI snippet

I just updated from python-gitlab 2.5.0 to 2.10.1 and noticed that my merge requests do not delete the source branch on merging anymore. I checked the debug output of all the versions in between and found out that the change was introduced in 2.7.0, which silently includes `should_remove_source_branch: false` in the `merge()` call.

This is undocumented and unexpected behavior. The function should not send a default value. If the value is not explicitly set in the function call, the source branch deletetion should be based on the according setting in the merge request:

![grafik](https://user-images.githubusercontent.com/45259958/146156801-83ec7d0a-b736-4131-b66d-8ec8efd12a4a.png)

This used to be the actual behavior in 2.6.0 and below.

Reproducible example:

```python
#!/usr/bin/env python3
from gitlab import Gitlab

gitlab = Gitlab("https://gitlab.com")
gitlab.enable_debug()

p = gitlab.projects.get(id=278964)
mr = p.mergerequests.get(id=1)
mr.merge()
```

## Expected Behavior

This is the request on python-gitlab 2.6.0:

```
send: b'PUT /api/v4/projects/278964/merge_requests/1/merge HTTP/1.1\r\nHost: gitlab.com\r\nUser-Agent: python-gitlab/2.6.0\r\nAccept-Encoding: gzip, deflate\r\nAccept: */*\r\nConnection: keep-alive\r\nContent-type: application/json\r\nContent-Length: 2\r\n\r\n'
send: b'{}'
```

## Actual Behavior

This is the request on python-gitlab >= 2.7.0

```
send: b'PUT /api/v4/projects/278964/merge_requests/1/merge HTTP/1.1\r\nHost: gitlab.com\r\nUser-Agent: python-gitlab/2.10.1\r\nAccept-Encoding: gzip, deflate\r\nAccept: */*\r\nConnection: keep-alive\r\nContent-type: application/json\r\nContent-Length: 38\r\n\r\n'
send: b'{"should_remove_source_branch": false}'
```

## Specifications

  - python-gitlab version: 2.70 - 2.10.1
  - API version you are using (v3/v4): v4
  - Gitlab server version (or gitlab.com): gitlab.com

nejch:
[...]

PR #1818, opened 2022-01-08
fix: remove default arguments for mergerequests.merge()

JohnVillalovos:
The arguments `should_remove_source_branch` and
`merge_when_pipeline_succeeds` are optional arguments. We should not
be setting any default value for them.

https://docs.gitlab.com/ee/api/merge_requests.html#accept-mr

Closes: #1750

### DIFF
--- a/gitlab/v4/objects/merge_requests.py
+++ b/gitlab/v4/objects/merge_requests.py
@@ -358,8 +358,8 @@ class ProjectMergeRequest(
     def merge(
         self,
         merge_commit_message: Optional[str] = None,
-        should_remove_source_branch: bool = False,
-        merge_when_pipeline_succeeds: bool = False,
+        should_remove_source_branch: Optional[bool] = None,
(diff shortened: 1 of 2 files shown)


nejch:
As we see in the functional tests this is _technically_ a breaking behavior. But only in the scenarios where
a) the project-level setting for "should remove source branch" is set to true (which I think has been the default for new projects for a while now)
b) the user does not supply the `should_remove_source_branch` param

In this case, `python-gitlab` used to override the project default due to its own default (which was `False`). But IMO this is wrong and unexpected behavior from `python-gitlab`, so I think we can get away with a `fix` here :+1:  For anyone reading this after the fact, the `merge()` default was added years before the project-level setting was introduced, so as soon as GitLab introduced the project-level setting our client started to have wrong behavior and it doesn't make sense anymore.

JohnVillalovos:
@nejch I agree. As it is very confusing if the project default is set to remove it but then we are by default not removing it.

Issue #1079, opened 2020-04-22
project.wikis.create with subpages results in 404 Wiki Page Not Found


kernelport:
## Description of the problem, including code/CLI snippet
page = project.wikis.create({'title': "title/subtitle", 'content': "test content"})   
does create a wiki subtitle in the directory title well.   
If I try to edit this page with:   
page.content = "update content"
page.title = "subtitle" # Attention: during update the page we have to set the title without the directory name
page.save()   
I get gitlab.exceptions.GitlabUpdateError: 404: 404 Wiki Page Not Found   

I can reproduce this by using curl:   

`curl --request PUT --data "content=update content" --header "PRIVATE-TOKEN: XXX"  https://gitlab.kernelport.com/api/v4/projects/231/wikis/title/subtitle   `

But when I encode the slug URL it does work with curl.   

`curl --request PUT --data "content=update content" --header "PRIVATE-TOKEN: XXX"  https://gitlab.kernelport.com/api/v4/projects/231/wikis/title%2Fsubtitle`

How can I handle this URL Encoding for slug in python-gitlab correct?

## Specifications

  - python-gitlab version: 2.0.1
  - API version you are using (v3/v4): V4
  - Gitlab server version (or gitlab.com): '12.8.1', 'd18b43a5f5a'

justinTM:
@kernelport I think I can replicate your url encoding fix as well.

this works:
```python
page = wiki.project.wikis.get('my/path/to/asdasd')
>>> page.save()
>>> page.slug
'my/path/to/asdasd'
>>> page.slug = 'my%2Fpath%2Fto%2Fasdasd'
>>> page.save()
>>> 
```

but this will 404 error:
```python
>>> page.slug = 'my%2Fpath%2Fto/asdasd'
>>> page.save()
Traceback (most recent call last):
  File "/Library/Python/3.8/site-packages/gitlab/exceptions.py", line 304, in wrapped_f
    return f(*args, **kwargs)
  File "/Library/Python/3.8/site-packages/gitlab/mixins.py", line 417, in update
    result = http_method(path, post_data=new_data, files=files, **kwargs)
  File "/Library/Python/3.8/site-packages/gitlab/client.py", line 807, in http_put
    result = self.http_request(
  File "/Library/Python/3.8/site-packages/gitlab/client.py", line 628, in http_request
    raise gitlab.exceptions.GitlabHttpError(
gitlab.exceptions.GitlabHttpError: 404: 404 Not Found

[...]

PR #1819, opened 2022-01-09
fix: use url-encoded ID in all paths

JohnVillalovos:
Make sure all usage of the ID in the URL path is encoded. Normally it
isn't an issue as most IDs are integers or strings which don't contain
a slash ('/'). But when the ID is a string with a slash character it
will break things.

Add a test case that shows this fixes wikis issue with subpages which
use the slash character.

Closes: #1079

### DIFF
--- a/gitlab/base.py
+++ b/gitlab/base.py
@@ -218,6 +218,15 @@ class RESTObject(object):
         return getattr(self, self._id_attr)
 
     @property
+    def encoded_id(self) -> Any:
+        """Ensure that the ID is url-encoded so that it can be safely used in a URL
+        path"""
+        obj_id = self.get_id()
+        if isinstance(obj_id, str):
+            obj_id = gitlab.utils._url_encode(obj_id)
+        return obj_id
+
+    @property
     def attributes(self) -> Dict[str, Any]:
         d = self.__dict__["_updated_attrs"].copy()
         d.update(self.__dict__["_attrs"])
--- a/gitlab/base.py
+++ b/gitlab/base.py
@@ -223,7 +223,7 @@ class RESTObject(object):
         path"""
         obj_id = self.get_id()
         if isinstance(obj_id, str):
-            obj_id = gitlab.utils._url_encode(obj_id)
+            obj_id = gitlab.utils.EncodedId(obj_id)
         return obj_id
 
     @property
--- a/gitlab/mixins.py
+++ b/gitlab/mixins.py
@@ -99,7 +99,8 @@ class GetMixin(_RestManagerBase):
             GitlabAuthenticationError: If authentication is not correct
             GitlabGetError: If the server cannot perform the request
         """
-        id = utils._url_encode(id)
+        if isinstance(id, str):
+            id = utils.EncodedId(id)
         path = f"{self.path}/{id}"
         if TYPE_CHECKING:
             assert self._obj_cls is not None
@@ -390,7 +391,7 @@ class UpdateMixin(_RestManagerBase):
         if id is None:
             path = self.path
         else:
-            path = f"{self.path}/{utils._url_encode(id)}"
+            path = f"{self.path}/{utils.EncodedId(id)}"
 
         self._check_missing_update_attrs(new_data)
         files = {}
@@ -443,7 +444,7 @@ class SetMixin(_RestManagerBase):
         Returns:
             The created/updated attribute
         """
-        path = f"{self.path}/{utils._url_encode(key)}"
+        path = f"{self.path}/{utils.EncodedId(key)}"
         data = {"value": value}
         server_data = self.gitlab.http_put(path, post_data=data, **kwargs)
         if TYPE_CHECKING:
@@ -476,7 +477,7 @@ class DeleteMixin(_RestManagerBase):
         if id is None:
             path = self.path
         else:
-            path = f"{self.path}/{utils._url_encode(id)}"
+            path = f"{self.path}/{utils.EncodedId(id)}"
         self.gitlab.http_delete(path, **kwargs)
 
 
--- a/gitlab/mixins.py
+++ b/gitlab/mixins.py
@@ -99,8 +99,7 @@ class GetMixin(_RestManagerBase):
             GitlabAuthenticationError: If authentication is not correct
             GitlabGetError: If the server cannot perform the request
         """
-        if not isinstance(id, int):
-            id = utils._url_encode(id)
+        id = utils._url_encode(id)
         path = f"{self.path}/{id}"
         if TYPE_CHECKING:
             assert self._obj_cls is not None
@@ -173,7 +172,7 @@ class RefreshMixin(_RestObjectBase):
             GitlabGetError: If the server cannot perform the request
         """
         if self._id_attr:
-            path = f"{self.manager.path}/{self.id}"
+            path = f"{self.manager.path}/{self.encoded_id}"
         else:
             if TYPE_CHECKING:
                 assert self.manager.path is not None
@@ -391,7 +390,7 @@ class UpdateMixin(_RestManagerBase):
         if id is None:
             path = self.path
         else:
-            path = f"{self.path}/{id}"
+            path = f"{self.path}/{utils._url_encode(id)}"
 
         self._check_missing_update_attrs(new_data)
         files = {}
@@ -477,9 +476,7 @@ class DeleteMixin(_RestManagerBase):
         if id is None:
             path = self.path
         else:
-            if not isinstance(id, int):
-                id = utils._url_encode(id)
-            path = f"{self.path}/{id}"
+            path = f"{self.path}/{utils._url_encode(id)}"
         self.gitlab.http_delete(path, **kwargs)
 
 
@@ -545,6 +542,7 @@ class SaveMixin(_RestObjectBase):
             return
 
         # call the manager
+        # Don't use `self.encoded_id` here as `self.manager.update()` will encode it.
         obj_id = self.get_id()
         if TYPE_CHECKING:
             assert isinstance(self.manager, UpdateMixin)
@@ -575,6 +573,7 @@ class ObjectDeleteMixin(_RestObjectBase):
         """
         if TYPE_CHECKING:
             assert isinstance(self.manager, DeleteMixin)
+        # Don't use `self.encoded_id` here as `self.manager.delete()` will encode it.
         self.manager.delete(self.get_id(), **kwargs)
 
 
@@ -598,7 +597,7 @@ class UserAgentDetailMixin(_RestObjectBase):
             GitlabAuthenticationError: If authentication is not correct
             GitlabGetError: If the server cannot perform the request
         """
-        path = f"{self.manager.path}/{self.get_id()}/user_agent_detail"
+        path = f"{self.manager.path}/{self.encoded_id}/user_agent_detail"
         result = self.manager.gitlab.http_get(path, **kwargs)
         if TYPE_CHECKING:
             assert not isinstance(result, requests.Response)
@@ -631,7 +630,7 @@ class AccessRequestMixin(_RestObjectBase):
             GitlabUpdateError: If the server fails to perform the request
         """
 
-        path = f"{self.manager.path}/{self.id}/approve"
+        path = f"{self.manager.path}/{self.encoded_id}/approve"
         data = {"access_level": access_level}
         server_data = self.manager.gitlab.http_put(path, post_data=data, **kwargs)
         if TYPE_CHECKING:
@@ -705,7 +704,7 @@ class SubscribableMixin(_RestObjectBase):
             GitlabAuthenticationError: If authentication is not correct
             GitlabSubscribeError: If the subscription cannot be done
         """
-        path = f"{self.manager.path}/{self.get_id()}/subscribe"
+        path = f"{self.manager.path}/{self.encoded_id}/subscribe"
         server_data = self.manager.gitlab.http_post(path, **kwargs)
         if TYPE_CHECKING:
             assert not isinstance(server_data, requests.Response)
@@ -725,7 +724,7 @@ class SubscribableMixin(_RestObjectBase):
             GitlabAuthenticationError: If authentication is not correct
             GitlabUnsubscribeError: If the unsubscription cannot be done
         """
-        path = f"{self.manager.path}/{self.get_id()}/unsubscribe"
+        path = f"{self.manager.path}/{self.encoded_id}/unsubscribe"
         server_data = self.manager.gitlab.http_post(path, **kwargs)
         if TYPE_CHECKING:
             assert not isinstance(server_data, requests.Response)
@@ -752,7 +751,7 @@ class TodoMixin(_RestObjectBase):
             GitlabAuthenticationError: If authentication is not correct
             GitlabTodoError: If the todo cannot be set
         """
-        path = f"{self.manager.path}/{self.get_id()}/todo"
+        path = f"{self.manager.path}/{self.encoded_id}/todo"
         self.manager.gitlab.http_post(path, **kwargs)
 
 
@@ -781,7 +780,7 @@ class TimeTrackingMixin(_RestObjectBase):
         if "time_stats" in self.attributes:
             return self.attributes["time_stats"]
 
-        path = f"{self.manager.path}/{self.get_id()}/time_stats"
+        path = f"{self.manager.path}/{self.encoded_id}/time_stats"
         result = self.manager.gitlab.http_get(path, **kwargs)
         if TYPE_CHECKING:
             assert not isinstance(result, requests.Response)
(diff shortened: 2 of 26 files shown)


nejch (changes requested):

gitlab/mixins.py
```diff
@@ -575,6 +575,8 @@ def delete(self, **kwargs: Any) -> None:
         """
         if TYPE_CHECKING:
             assert isinstance(self.manager, DeleteMixin)
+        # NOTE: Don't use `self.encoded_id` here as `self.manager.delete()` will encode
+        # it.
```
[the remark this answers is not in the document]
JohnVillalovos: So #1820 solves this issue. Interested to hear your thoughts on these two different approaches to solving the problem.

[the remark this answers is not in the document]
nejch: I like the simplicity of this one as an initial fix at least, but also like the idea of making it safer, if we don't complicate things too much though.

I'd merge this as is and work on that as a smaller change set on top of this if we decide to do it.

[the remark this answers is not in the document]
JohnVillalovos: Made comment one line 😊

I think I have cleaned up all the 'id' usage.

gitlab/mixins.py
```diff
@@ -99,8 +99,7 @@ def get(
             GitlabAuthenticationError: If authentication is not correct
             GitlabGetError: If the server cannot perform the request
         """
-        if not isinstance(id, int):
-            id = utils._url_encode(id)
+        id = utils._url_encode(id)
         path = f"{self.path}/{id}"
```
nejch, line 103: Just from a quick glance it would seem to me we could use the property here as well, or did I probably miss something?

```suggestion
        path = f"{self.path}/{self.encoded_id}"
```

gitlab/mixins.py
```diff
@@ -477,9 +476,7 @@ def delete(self, id: Union[str, int], **kwargs: Any) -> None:
         if id is None:
             path = self.path
         else:
-            if not isinstance(id, int):
-                id = utils._url_encode(id)
-            path = f"{self.path}/{id}"
+            path = f"{self.path}/{utils._url_encode(id)}"
```
nejch, line 479: Same question as above. Mind you I didn't check in detail :laughing:

gitlab/mixins.py
JohnVillalovos: No, because we need to store the encoded value below when doing lazy.

gitlab/mixins.py
JohnVillalovos: In this case `id` is not from `self.id`.

gitlab/mixins.py
JohnVillalovos: Oh and this is `id` not `self.id`

nejch: My bad, I'll look closer next time :laughing:

gitlab/mixins.py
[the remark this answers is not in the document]
JohnVillalovos: I updated `SaveMixin` with a comment on why not to encode. `UpdateMixin` should use encoded value as it calls post/put directly.

gitlab/mixins.py
```diff
@@ -173,7 +172,7 @@ def refresh(self, **kwargs: Any) -> None:
             GitlabGetError: If the server cannot perform the request
         """
         if self._id_attr:
-            path = f"{self.manager.path}/{self.id}"
+            path = f"{self.manager.path}/{self.encoded_id}"
```
nejch, line 175: Ok, maybe I'm starting to see having a strict "only encode once" mechanism will help.

Currently this breaks if the user creates a project lazily and then refreshes it, as it'll be encoded twice.

```
>>> import gitlab
>>> gl = gitlab.Gitlab()
>>> project = gl.projects.get("gitlab-org/gitlab", lazy=True)
>>> project.id
'gitlab-org%2Fgitlab'
>>> project.refresh()
  File "/home/nejc/repos/python-gitlab/gitlab/exceptions.py", line 307, in wrapped_f
    return f(*args, **kwargs)
  File "/home/nejc/repos/python-gitlab/gitlab/mixins.py", line 180, in refresh
    server_data = self.manager.gitlab.http_get(path, **kwargs)
  File "/home/nejc/repos/python-gitlab/gitlab/client.py", line 761, in http_get
    result = self.http_request(
  File "/home/nejc/repos/python-gitlab/gitlab/client.py", line 727, in http_request
    raise gitlab.exceptions.GitlabHttpError(
gitlab.exceptions.GitlabHttpError: 404: 404 Project Not Found
```

Can we continue with this PR and fix these specific paths, and then follow up with a robust check?

JohnVillalovos: I added a new commit to this PR to hopefully resolve this issue.

gitlab/mixins.py
```diff
@@ -173,7 +172,11 @@ def refresh(self, **kwargs: Any) -> None:
             GitlabGetError: If the server cannot perform the request
         """
         if self._id_attr:
-            path = f"{self.manager.path}/{self.id}"
+            if self._lazy:
+                id = self.get_id()
+            else:
+                id = self.encoded_id
+            path = f"{self.manager.path}/{id}"
```
nejch, line 179: Unfortunately I don't think the logic needs to be here :/

The way we've done this, maybe because `encoded_id`, it's basically for anything that happens after a lazy get, I think.

```
>>> gl = gitlab.Gitlab(private_token=pat)
>>> project = gl.projects.get("nejc/private-project", lazy=True)
>>> project.description = "My stuff"
>>> project.save()
...
```

Maybe we just don't need to url-encode for lazy objects at all with this fix, if we just ensure it's done for all real operations. It will also give prettier `my_object.id` for lazy objects.

JohnVillalovos: Okay. I believe I have fixed things by adding the `EncodedId` string class.

gitlab/utils.py
```diff
@@ -56,7 +56,88 @@ def copy_dict(dest: Dict[str, Any], src: Dict[str, Any]) -> None:
...
+
+    Path parameters that are required to be URL-encoded must be followed. If not, it
+    doesn’t match an API endpoint and responds with a 404. If there’s something in
+    front of the API (for example, Apache), ensure that it doesn’t decode the
+    URL-encoded path parameters.
+
+
+    When creating an EncodedId instance `__new__` will be called first and then
+    `__init__`.
+    """
```
nejch, line 84: Any chance we can get more of this on the screen/intellisense, e.g. by removing upstream text that might get outdated? And are we not documenting the default behavior of `__new__` and `__init__` here?

```suggestion
    """A custom `str` class that will return the URL-encoded value of the string:

      * Using it recursively will only url-encode the value once.
      * Can accept either `str` or `int` as input value.
      * Can be used in an f-string and output the URL-encoded string.

    See:

      * https://docs.gitlab.com/ee/api/index.html#namespaced-path-encoding
      * https://docs.gitlab.com/ee/api/index.html#path-parameters
    """
```

image

nejch (commented):
Thanks for the update again! Just a few questions from me :)

tests/functional/api/test_lazy_objects.py
```diff
@@ -0,0 +1,39 @@
...
+    assert lazy_project.id == project.id
+    assert lazy_project.description == "A new description"
+
+
+def test_delete_after_lazy_get_with_path(gl, group, wait_for_sidekiq):
+    project = gl.projects.create({"name": "lazy_project", "namespace_id": group.id})
+    result = wait_for_sidekiq(timeout=60)
+    assert result is True, "sidekiq process should have terminated but did not"
+    lazy_project = gl.projects.get(project.path_with_namespace, lazy=True)
+    lazy_project.delete()
```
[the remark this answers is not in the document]
JohnVillalovos: I wanted to use a fixture but the project is shared across the module. And so the project would be deleted and the next test would then fail 😢

tests/functional/api/test_lazy_objects.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def test_method_call(gl, lazy_project):
+    lazy_project.mergerequests.list()
```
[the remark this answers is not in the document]
JohnVillalovos: Sure, but I was just trying to imply that can call a child's method. Not just `list` but `list` was what I happened to use 😊

But can change the name!

gitlab/utils.py
```diff
@@ -56,7 +56,88 @@ def copy_dict(dest: Dict[str, Any], src: Dict[str, Any]) -> None:
...
+def _url_encode(id: int) -> int:
+    ...
+
+
+@overload
+def _url_encode(id: Union[str, EncodedId]) -> EncodedId:
+    ...
+
+
+def _url_encode(id: Union[int, str, EncodedId]) -> Union[int, EncodedId]:
```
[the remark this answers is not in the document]
JohnVillalovos: I don't think we still need this function, I guess I was trying to make the PR easier to review by also changing this. But I haven't looked to see how many extra changes that would be.

tests/functional/api/test_lazy_objects.py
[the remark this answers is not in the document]
nejch: Yes, I think we need to start adding `make_*` factories  that we call inside tests and our conventional fixtures. I will add a PR for this :)

gitlab/utils.py
JohnVillalovos: Sure I can work on that. Thanks.

tests/functional/api/test_lazy_objects.py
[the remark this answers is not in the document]
nejch: I guess we could also avoid all this by using function-scoped fixtures but then I'm sure we would get to 15 minute test suites and a lot more sidekiq waiting :/

tests/functional/api/test_wikis.py
```diff
@@ -0,0 +1,15 @@
+"""
+GitLab API:
+https://docs.gitlab.com/ee/api/wikis.html
+"""
+
+
+def test_wikis(project):
+
```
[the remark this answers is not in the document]
JohnVillalovos: Done

gitlab/utils.py
JohnVillalovos: > Any chance we can get more of this on the screen/intellisense, e.g. by removing upstream text that might get outdated? 

Sure.

> And are we not documenting the default behavior of `__new__` and `__init__` here?

You're a better Pythonista than me if you already knew this 😊  But I guess we can let me people who don't know that go figure it out.

JohnVillalovos: Done.

tests/functional/api/test_lazy_objects.py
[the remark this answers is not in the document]
JohnVillalovos: Done.

gitlab/utils.py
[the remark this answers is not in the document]
JohnVillalovos: I have removed it and put that change into a new commit to hopefully make reviewing easier.

tests/functional/api/test_lazy_objects.py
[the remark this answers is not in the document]
JohnVillalovos: Yeah. I'm not sure how much of an impact that would have for the test time. Could also maybe have a specific fixture that is function-scoped.  `project_for_function` ???

gitlab/base.py
```diff
@@ -217,6 +217,15 @@ def get_id(self) -> Any:
             return None
         return getattr(self, self._id_attr)
 
+    @property
+    def encoded_id(self) -> Any:
+        """Ensure that the ID is url-encoded so that it can be safely used in a URL
+        path"""
+        obj_id = self.get_id()
+        if isinstance(obj_id, str):
+            obj_id = gitlab.utils.EncodedId(obj_id)
```
nejch, line 226: question: do we need this check here if `EncodedId` will handle it anyway? I guess there's maybe a performance penalty when we receive `int` but not sure it's noticeable.

Same for others below, and seems like you did remove the check in some of them, right?

tests/functional/api/test_lazy_objects.py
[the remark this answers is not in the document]
nejch: Let's do that in a follow-up ;)

tests/unit/test_utils.py
```diff
@@ -15,23 +15,66 @@
...
+    def test_json_serializable(self):
+        obj = utils.EncodedId("someone")
+        assert '"someone"' == json.dumps(obj)
 
-    # periods/dots should not be modified
-    src = "docs/README.md"
-    dest = "docs%2FREADME.md"
-    assert dest == utils._url_encode(src)
+        obj = utils.EncodedId("we got/a/path")
+        assert '"we%20got%2Fa%2Fpath"' == json.dumps(obj)
```
nejch, line 80: Do we have a use case for this in our library? If not I'd maybe skip this part even if it's cool knowing it works :grinning:

JohnVillalovos: There was things that failed because the object wasn't JSON serializable. When I first tried doing it as simple class. One of the reasons I switched to doing it this way where it derives from the `str` class.

gitlab/utils.py
```diff
@@ -56,25 +56,64 @@ def copy_dict(dest: Dict[str, Any], src: Dict[str, Any]) -> None:
...
+            value = str(value)
+        # Make sure isinstance() for `EncodedId` comes before check for `str` as
+        # `EncodedId` is an instance of `str` and would pass that check.
+        elif isinstance(value, EncodedId):
+            # We use the original string value to URL-encode
+            value = value.original_str
+        elif isinstance(value, str):
+            pass
+        else:
+            raise ValueError(f"Unsupported type received: {type(value)}")
```
nejch, line 91: I'd say we raise TypeError here
```suggestion
            raise TypeError(f"Unsupported type received: {type(value)}")
```
E.g. https://github.com/python/cpython/blob/f4c03484da59049eb62a9bf7777b963e2267d187/Lib/plistlib.py#L359

gitlab/base.py
nejch: I see. Could we maybe just return int in `__new__` if we receive int, essentially making it a no-op? It would also simplify `__init__` then, as it would never get called for ints. Like discussed here.

https://stackoverflow.com/a/2491881

I feel like this way we'd also stick the closest to the current behavior, possibly avoiding some issues.

gitlab/utils.py
```diff
@@ -56,25 +56,64 @@ def copy_dict(dest: Dict[str, Any], src: Dict[str, Any]) -> None:
...
+        # Make sure isinstance() for `EncodedId` comes before check for `str` as
+        # `EncodedId` is an instance of `str` and would pass that check.
+        elif isinstance(value, EncodedId):
+            # This is the key part as we are always keeping the original string even
+            # through multiple recursions.
+            value = value.original_str
+        elif isinstance(value, str):
+            pass
+        else:
+            raise ValueError(f"Unsupported type received: {type(value)}")
```
nejch, line 114: Same as above
```suggestion
            raise TypeError(f"Unsupported type received: {type(value)}")
```

gitlab/utils.py
JohnVillalovos: Done

gitlab/utils.py
JohnVillalovos: Done

gitlab/utils.py
```diff
@@ -56,25 +56,68 @@ def copy_dict(dest: Dict[str, Any], src: Dict[str, Any]) -> None:
...
+    ) -> Union[int, "EncodedId"]:
+        # __new__() gets called before __init__()
+        if isinstance(value, int):
+            return value
+
+        # Make sure isinstance() for `EncodedId` comes before check for `str` as
+        # `EncodedId` is an instance of `str` and would pass that check.
+        if isinstance(value, EncodedId):
+            # We use the original string value to URL-encode
+            value = value.original_str
```
nejch, line 91: Can we do the same as with `int` here? In fact, this will even be safer than int and a more valid case, mypy wouldn't complain with this one.

```suggestion
        if isinstance(value, EncodedId):
            return value
```

JohnVillalovos: Good idea. Done but combined it with the `int` one.

gitlab/utils.py
```diff
@@ -56,25 +56,38 @@ def copy_dict(dest: Dict[str, Any], src: Dict[str, Any]) -> None:
...
+        https://docs.gitlab.com/ee/api/index.html#namespaced-path-encoding
+        https://docs.gitlab.com/ee/api/index.html#path-parameters
     """
-    return urllib.parse.quote(id, safe="")
+
+    # `original_str` will contain the original string value that was used to create the
+    # first instance of EncodedId. We will use this original value to generate the
+    # URL-encoded value each time.
+    original_str: str
+
```
nejch, line 78: I think we can now remove this as it's not used anywhere:

```suggestion
```

JohnVillalovos: Done

gitlab/utils.py
```diff
@@ -56,25 +56,33 @@ def copy_dict(dest: Dict[str, Any], src: Dict[str, Any]) -> None:
...
+    # mypy complains if return type other than the class type. So we ignore issue.
+    def __new__(  # type: ignore
+        cls, value: Union[str, int, "EncodedId"]
+    ) -> Union[int, "EncodedId"]:
+        if isinstance(value, (int, EncodedId)):
+            return value
+
+        if not isinstance(value, str):
+            raise TypeError(f"Unsupported type received: {type(value)}")
+        # Set the value our string will return
```
nejch, line 83: I feel like without the `__init__` this is now extremely obvious and we don't need the comment, WDYT?
```suggestion

```

PR #1820, opened 2022-01-09
fix: use url-encoded ID in all paths (alternate method)

JohnVillalovos:
An alternative to https://github.com/python-gitlab/python-gitlab/pull/1819

Make sure all usage of the ID in the URL path is encoded. Normally it
isn't an issue as most IDs are integers or strings which don't contain
a slash ('/'). But when the ID is a string with a slash character it
will break things.

Add a test case that shows this fixes wikis issue with subpages which
use the slash character.

Closes: #1079

### DIFF
--- a/gitlab/base.py
+++ b/gitlab/base.py
@@ -209,6 +209,15 @@ class RESTObject(object):
         return getattr(self, self._id_attr)
 
     @property
+    def encoded_id(self) -> Any:
+        """Ensure that the ID is url-encoded so that it can be safely used in a URL
+        path"""
+        obj_id = self.get_id()
+        if isinstance(obj_id, str):
+            obj_id = gitlab.utils.EncodedId(obj_id)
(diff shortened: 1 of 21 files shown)


JohnVillalovos:
I think this is cool 😎 

But I am thinking this is maybe a bit too magic. Also when things get json encoded they lose the url-encoding 😢

JohnVillalovos:
The `EncodedId` class is kind of cool 😎 in my opinion... But also kind of complicated. It was really interesting figuring this out.

Interested to hear what you think @nejch

nejch:
I guess we can close this one @JohnVillalovos as the changes were essentially incorporated into the original one right?

JohnVillalovos:
> I guess we can close this one @JohnVillalovos as the changes were essentially incorporated into the original one right?

Yep!

PR #1823, opened 2022-01-09
fix: broken URL for FAQ about attribute-error-list

JohnVillalovos:
```
The URL was missing a 'v' before the version number and thus the page
did not exist.

Previously the URL for python-gitlab 3.0.0 was:
https://python-gitlab.readthedocs.io/en/3.0.0/faq.html#attribute-error-list

Which does not exist.

Change it to:
https://python-gitlab.readthedocs.io/en/v3.0.0/faq.html#attribute-error-list
  add the 'v' --------------------------^
```

### DIFF
--- a/gitlab/base.py
+++ b/gitlab/base.py
@@ -36,7 +36,7 @@ __all__ = [
 
 
 _URL_ATTRIBUTE_ERROR = (
-    f"https://python-gitlab.readthedocs.io/en/{gitlab.__version__}/"
+    f"https://python-gitlab.readthedocs.io/en/v{gitlab.__version__}/"
     f"faq.html#attribute-error-list"
 )
 


nejch:
@JohnVillalovos thanks for the catch!

I still see value in using versioned docs for this, in case we change things in the future.

It was just an oversight on our part with a missing `v`, the link is actually available here:
https://python-gitlab.readthedocs.io/en/v3.0.0/faq.html#attribute-error-list

WDYT?

We can also solve the existing issue on RTD with redirects either in the GUI (https://docs.readthedocs.io/en/stable/user-defined-redirects.html) or in sphinx directly for config as code (something like https://github.com/sphinx-contrib/redirects).

I plan on renaming `gl_objects/` to `resources/` to match upstream, so maybe I can add that then, as well, after I do some testing.

JohnVillalovos:
Oh sure. I can update to add the v. Sounds good!

PR #1824, opened 2022-01-10
test(functional): check mixin behavior with lazy objects

nejch:


nejch:
Closing as we added these tests in https://github.com/python-gitlab/python-gitlab/pull/1819. I'll cherry-pick the gitignore :)

Issue #1825, opened 2022-01-11
ProjectMember.save() does not work because 'ProjectMemberAllManager' object has no attribute 'update'


theoturner:
Description of the problem, including code/CLI snippet

ProjectMember.save() does not work because 'ProjectMemberAllManager' object has no attribute 'update'.

```
project = GitlabClient().projects.get('my_proj_name')
member = project.members.list()[0]
member.access_level = 30
member.save()
```
Expected Behavior

Member saves with new properties.
Actual Behavior

```
Traceback (most recent call last):
  File "gitlab_client.py", line 85, in update_project_permissions
    member.save()
  File "/Users/theoturner/Library/Python/3.8/lib/python/site-packages/gitlab/mixins.py", line 551, in save
    server_data = self.manager.update(obj_id, updated_data, **kwargs)
AttributeError: 'ProjectMemberAllManager' object has no attribute 'update'
```
N.B. type(member) is <class 'gitlab.v4.objects.members.ProjectMember'> so I haven't accidentally got a ProjectMemberAllManager.
Specifications

python-gitlab version: 3.0.0, can be replicated in 2.8 and above
API version you are using (v3/v4): v4

Note: John Villalovos accidentally edited this and now attempting to return to original state...

max-wittig:
This is a bit of a conceptional issue. Basically `python-gitlab` tries to be a small as possible layer on top of the GitLab API as possible. So there is less automatic logic that you might expect, as this reduces complexity of the library a lot

To update a member, you would need to get the exact member and update this object. You cannot update a list, as python-gitlab in it's principle cannot know what you're trying to update.

Basically you need to do:

```python
project = GitlabClient().projects.get('my_proj_name')
member = project.members.list()[0]
member = project.members.get(member.id) # add this line
member.access_level = 30
member.save()
```

Does that clarify the usage?

JohnVillalovos:
@max-wittig  @nejch 

This does make me wonder, why can't `ProjectMemberAllManager' use a `CRUDMixin`?

https://github.com/python-gitlab/python-gitlab/blob/4a000b6c41f0a7ef6121c62a4c598edc20973799/gitlab/v4/objects/members.py#L106

JohnVillalovos:
Well we should have a `get` method based on:

https://docs.gitlab.com/ee/api/members.html#get-a-member-of-a-group-or-project-including-inherited-and-invited-members

[...]

Issue #848, opened 2019-07-25
/groups/:id/members/all for cli does not contain correct parameters


dgsmith:
## Description of the problem, including code/CLI snippet
When using the CLI command: `gitlab group-member all [...]` the accepted parameters are incorrect. The `all` "endpoint" should operate just like the `list`, allowing paging, etc. instead, it seems to have some other specific commands that aren't necessary.

## Expected Behavior
Being able to use commands such as,
- `gitlab group-member all --group-id 212 --page 2`
- `gitlab group-member all --group-id 212 --per-page 2 --page 10`
- etc.

## Actual Behavior
`gitlab group-member all` only accepts:
```
 -h, --help           show this help message and exit
  --group-id GROUP_ID
  --sudo SUDO
  --id ID
```

as specified in the [gitlab docs](https://docs.gitlab.com/ce/api/members.html#list-all-members-of-a-group-or-project-including-inherited-members) `--id` is not necessary and you should actually **be able to search** (probably something to add for the normal `list` as well.

## Specifications

  - python-gitlab version: 1.10.0
  - API version you are using (v3/v4): v4
  - Gitlab server version (or gitlab.com): 12.0.3-ee

chrisoldwood:
I believe this is related to #593 about the `runner all` verb as that behaves in the same way.

I used the group member code to work out why the `runner all` JSON support was broken but couldn't address the spurious `--id` argument. I added some thoughts on ticket #593, e.g. an `all-runners list` alternative object type, and a pointer about where in the CLI handling it might be making the assumption around why the `--id` argument is mandatory, but know too little at present to make a serious proposal 🙂.

I can envisage a possible "hack" where a custom action named `all` would bypass the logic around looking for an object's `id` attribute. This would potentially fix both cases but maybe there is some cleaner way of signalling that `all` is more like `list` and acts on an entire collection instead of a single item?

pvshingala1997:
Can you assign this to me?

PR #1827, opened 2022-01-12
fix: members: use new *All objects for *AllManager managers

JohnVillalovos:
Change it so that:

  GroupMemberAllManager uses GroupMemberAll object
  ProjectMemberAllManager uses ProjectMemberAll object

Create GroupMemberAll and ProjectMemberAll objects that do not support
any Mixin type methods. Previously we were using GroupMember and
ProjectMember which support the `save()` and `delete()` methods but
those methods will not work with objects retrieved using the
`/members/all/` API calls.

`list()` API calls:
  GET /groups/:id/members/all
  GET /projects/:id/members/all

`get()` API calls:
  GET /groups/:id/members/all/:user_id
  GET /projects/:id/members/all/:user_id

Closes: #1825
Closes: #848

[1] https://docs.gitlab.com/ee/api/members.html#list-all-members-of-a-group-or-project-including-inherited-and-invited-members
[2] https://docs.gitlab.com/ee/api/members.html#get-a-member-of-a-group-or-project-including-inherited-and-invited-members

### DIFF
--- a/tests/functional/cli/test_cli_v4.py
+++ b/tests/functional/cli/test_cli_v4.py
@@ -701,6 +701,31 @@ def test_delete_group_deploy_token(gitlab_cli, group_deploy_token):
     # TODO assert not in list
 
 
+def test_project_member_all(gitlab_cli, project):
+    cmd = [
+        "project-member-all",
+        "list",
+        "--project-id",
+        project.id,
+    ]
+    ret = gitlab_cli(cmd)
+
+    assert ret.success
+
+
+def test_group_member_all(gitlab_cli, group):
+    cmd = [
+        "group-member-all",
+        "list",
+        "--group-id",
+        group.id,
+    ]
+    ret = gitlab_cli(cmd)
+
+    assert ret.success
+
+
+# Deleting the project and group. Add your tests above here.
 def test_delete_project(gitlab_cli, project):
     cmd = ["project", "delete", "--id", project.id]
     ret = gitlab_cli(cmd)
@@ -713,3 +738,6 @@ def test_delete_group(gitlab_cli, group):
     ret = gitlab_cli(cmd)
 
     assert ret.success
+
+
+# Don't add tests below here as the group and project have been deleted
--- a/gitlab/v4/objects/members.py
+++ b/gitlab/v4/objects/members.py
@@ -17,9 +17,11 @@ __all__ = [
     "GroupBillableMemberMembership",
     "GroupBillableMemberMembershipManager",
     "GroupMember",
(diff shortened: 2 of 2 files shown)


nejch (commented):
Thanks @JohnVillalovos! At the beginning I was happy about the initial implementation because it wasn't duplicating things, but looks like it was kind of broken that way.

A good side-effect of this is that it will also work on the CLI now I'm pretty sure.

JohnVillalovos:
Yeah, because of that `members/all` we have no idea where they are a member from. It could be anywhere in the path above them. So we can't delete them in the `members/all` as have go up the path to figure out where they are actually getting their access from.

nejch:
@JohnVillalovos I'm pretty sure this also closes https://github.com/python-gitlab/python-gitlab/issues/848 by adding a new CLI command:

```
(.venv) nejc@zbook-fury:~/repos/python-gitlab$ gitlab project-member-all list --help
usage: gitlab project-member-all list [-h] [--sudo SUDO] --project-id PROJECT_ID [--page PAGE] [--per-page PER_PAGE] [--all]

optional arguments:
  -h, --help            show this help message and exit
  --sudo SUDO
  --project-id PROJECT_ID
  --page PAGE
  --per-page PER_PAGE
  --all
```
would you mind adding a quick test for that?

JohnVillalovos:
> @JohnVillalovos I'm pretty sure this also closes #848 by adding a new CLI command:
> 
> ```
> (.venv) nejc@zbook-fury:~/repos/python-gitlab$ gitlab project-member-all list --help
> usage: gitlab project-member-all list [-h] [--sudo SUDO] --project-id PROJECT_ID [--page PAGE] [--per-page PER_PAGE] [--all]
> 
> optional arguments:
>   -h, --help            show this help message and exit
>   --sudo SUDO
>   --project-id PROJECT_ID
>   --page PAGE
>   --per-page PER_PAGE
>   --all
> ```
> 
> would you mind adding a quick test for that?
> 
> Edit: and just one more nit, `fix(members):` 🙇

Done. And found bug with this test. Thanks!

tests/functional/cli/test_cli_v4.py
```diff
@@ -713,3 +738,6 @@ def test_delete_group(gitlab_cli, group):
     ret = gitlab_cli(cmd)
 
     assert ret.success
+
+
+# Don't add tests below here as the group and project have been deleted
```
nejch, line 743: Hehe. Will work on those fixtures :)

nejch (approved):

Issue #1828, opened 2022-01-13
Lazy creation of service object still fails with 404 error


goeranu:
## Creating a service lazily
The following code is meant to create a service object, even if the project in question does not yet have such an integration (i.e. a `services.get` without `lazy` would fail with a 404 error).

    gl = gitlab.Gitlab(
        gitlab_url,
        private_token=gitlab_private_token
    )
    projects = gl.projects.list(search="devops4397", simple=True)
    service = projects[0].services.get('jira', lazy=True)

## Expected Behavior
`service` should be a service object, and the following code should be executed.

## Actual Behavior
The call fails with an exception:

    Traceback (most recent call last):
      File "/home/atlas/src/services/gitlab-group/server/scripts/.venv/lib64/python3.10/site-packages/gitlab/exceptions.py", line 307, in wrapped_f
        return f(*args, **kwargs)
      File "/home/atlas/src/services/gitlab-group/server/scripts/.venv/lib64/python3.10/site-packages/gitlab/mixins.py", line 111, in get
        server_data = self.gitlab.http_get(path, **kwargs)
      File "/home/atlas/src/services/gitlab-group/server/scripts/.venv/lib64/python3.10/site-packages/gitlab/client.py", line 761, in http_get
        result = self.http_request(
      File "/home/atlas/src/services/gitlab-group/server/scripts/.venv/lib64/python3.10/site-packages/gitlab/client.py", line 727, in http_request
        raise gitlab.exceptions.GitlabHttpError(
    gitlab.exceptions.GitlabHttpError: 404: 404 Integration Not Found
    
    The above exception was the direct cause of the following exception:
    
    Traceback (most recent call last):
      File "/home/atlas/src/services/gitlab-group/server/scripts/testkod", line 19, in <module>
        service = projects[0].services.get('jira', lazy=True)
      File "/home/atlas/src/services/gitlab-group/server/scripts/.venv/lib64/python3.10/site-packages/gitlab/v4/objects/services.py", line 277, in get
        obj = cast(ProjectService, super(ProjectServiceManager, self).get(id, **kwargs))
      File "/home/atlas/src/services/gitlab-group/server/scripts/.venv/lib64/python3.10/site-packages/gitlab/exceptions.py", line 309, in wrapped_f
        raise error(e.error_message, e.response_code, e.response_body) from e
    gitlab.exceptions.GitlabGetError: 404: 404 Integration Not Found

## Specifications

[...]

PR #1829, opened 2022-01-13
fix(api): services: add missing `lazy` parameter

JohnVillalovos:
Commit 8da0b758c589f608a6ae4eeb74b3f306609ba36d added the `lazy`
parameter to the services `get()` method but missed then using the
`lazy` parameter when it called `super(...).get(...)`

Closes: #1828


nejch (approved):
Thanks for the quick action @JohnVillalovos! Seems like we have two scopes in the commit message somehow :grin: but all good I think

Issue #1081, opened 2020-04-22
Feature Request: project.wikis page.save() should return the updated page 


kernelport:
## Description of the problem, including code/CLI snippet
In case of change the page title,  the slug is updated.
Sometimes we need the new slug to update this page again.

max-wittig:
Could you please elaborate a bit on this? Have a code sample, write a failing test?

The content should actually be returned by GitLab.

PR #1832, opened 2022-01-13
feat(api): return result from `SaveMixin.save()`

JohnVillalovos:
Return the new object data when calling `SaveMixin.save()`.

Also remove check for `None` value when calling
`self.manager.update()` as that method only returns a dictionary.

Closes https://github.com/python-gitlab/python-gitlab/issues/1081.

### DIFF
--- a/gitlab/mixins.py
+++ b/gitlab/mixins.py
@@ -525,13 +525,16 @@ class SaveMixin(_RestObjectBase):
 
         return updated_data
 
-    def save(self, **kwargs: Any) -> None:
+    def save(self, **kwargs: Any) -> Optional[Dict[str, Any]]:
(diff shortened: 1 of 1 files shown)


nejch (approved):

nejch:
@JohnVillalovos I think we want this in the changelog so maybe even `feat` :)

JohnVillalovos:
> @JohnVillalovos I think we want this in the changelog so maybe even `feat` :)

Sure! Done.

nejch:
@kernelport I've marked this as closing https://github.com/python-gitlab/python-gitlab/issues/1081 since it implements the rest of your PR. Feel free to reopen the issue if it's not solved after the release. You'll be credited as the author of the commit here.

PR #1833, opened 2022-01-13
feat(api): add `project.transfer()` and deprecate `transfer_project()`

nejch:
Needed to bring things in line with https://github.com/python-gitlab/python-gitlab/pull/1807.

`transfer_project` was very confusing already as the direction of transfer can differ depending on the endpoint, see other PR for discussion.

### DIFF
--- a/gitlab/v4/objects/projects.py
+++ b/gitlab/v4/objects/projects.py
@@ -1,3 +1,4 @@
+import warnings
 from typing import Any, Callable, cast, Dict, List, Optional, TYPE_CHECKING, Union
 
 import requests
@@ -526,7 +527,7 @@ class Project(RefreshMixin, SaveMixin, ObjectDeleteMixin, RepositoryMixin, RESTO
 
     @cli.register_custom_action("Project", ("to_namespace",))
     @exc.on_http_error(exc.GitlabTransferProjectError)
-    def transfer_project(self, to_namespace: str, **kwargs: Any) -> None:
+    def transfer(self, to_namespace: str, **kwargs: Any) -> None:
         """Transfer a project to the given namespace ID
 
         Args:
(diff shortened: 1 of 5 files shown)


gitlab/v4/objects/projects.py
```diff
@@ -526,7 +527,7 @@ def mirror_pull(self, **kwargs: Any) -> None:
 
     @cli.register_custom_action("Project", ("to_namespace",))
     @exc.on_http_error(exc.GitlabTransferProjectError)
-    def transfer_project(self, to_namespace: str, **kwargs: Any) -> None:
+    def transfer(self, to_namespace: str, **kwargs: Any) -> None:
```
JohnVillalovos, line 530:
```suggestion
    def transfer(self, to_namespace: Union[int, str], **kwargs: Any) -> None:
```

gitlab/v4/objects/projects.py
```diff
@@ -543,6 +544,15 @@ def transfer_project(self, to_namespace: str, **kwargs: Any) -> None:
             path, post_data={"namespace": to_namespace}, **kwargs
         )
 
+    @cli.register_custom_action("Project", ("to_namespace",))
+    def transfer_project(self, *args: Any, **kwargs: Any) -> None:
```
JohnVillalovos, line 548: Why change the signature from `id` to `*args`?

Not a big deal, but curious.

nejch: Just wanted to keep the signature completely generic and pass everything to the real function in case things change, probably doesn't make sense here though as we know what to expect. I can change it

JohnVillalovos: No need to change it. Just curious. I don't think it hurts anything. Heck making the intellisense not work is maybe a good idea 😁

JohnVillalovos (approved):
Thanks. Looks good to me.

PR #1836, opened 2022-01-14
chore(objects): use `self.encoded_id` where applicable

JohnVillalovos:
Updated a few remaining usages of `self.id` to use `self.encoded_id`
as for the most part we shouldn't be using `self.id`

There are now only a few (4 lines of code) remaining uses of
`self.id`, most of which seem that they should stay that way.


JohnVillalovos:
@nejch  Last bit of them...

PR #1838, opened 2022-01-14
chore: rename `gitlab/__version__.py` -> `gitlab/_version.py`

JohnVillalovos:
It is confusing to have a `gitlab/__version__.py` because we also
create a variable `gitlab.__version__` which can conflict with
`gitlab/__version__.py`.

For example in `gitlab/const.py` we have to know that
`gitlab.__version__` is a module and not the variable due to the
ordering of imports. But in most other usage `gitlab.__version__` is a
version string.

To reduce confusion make the name of the version file
`gitlab/_version.py`.


nejch:
I don't think it's _that_ confusing as it's quite a common pattern (e.g. https://github.com/mtkennerly/poetry-dynamic-versioning/blob/4454e5082bf1fc2f4a39fa523a9b665eda5d638f/poetry_dynamic_versioning/__init__.py#L86-L89 among others).

`version.py` could also be confusing as we have https://docs.gitlab.com/ee/api/version.html, so maybe we could do `_version.py` then, to signify it's not anything to do with it.

JohnVillalovos:
> I don't think it's _that_ confusing as it's quite a common pattern (e.g. https://github.com/mtkennerly/poetry-dynamic-versioning/blob/4454e5082bf1fc2f4a39fa523a9b665eda5d638f/poetry_dynamic_versioning/__init__.py#L86-L89 among others).

I'm as big a fan of cargo-culting as the next person. But in our case due to circular imports it is confusing. Because at some points in the code `__version__` is the version string, but in other locations it is a module.  Which I have found confusing when trying to work on the code as I need to understand where in the import cycle we are to know which value it is.

I did update the commit message to point out one actual situation of confusion.
 
> `version.py` could also be confusing as we have https://docs.gitlab.com/ee/api/version.html, so maybe we could do `_version.py` then, to signify it's not anything to do with it?

Done!

nejch:
Ahh ok, that makes sense. Thanks!

PR #1839, opened 2022-01-14
chore(tests): use method `projects.transfer()`

JohnVillalovos:
When doing the functional tests use the new function
`projects.transfer` instead of the deprecated function
`projects.transfer_project()`

### DIFF
--- a/tests/functional/api/test_projects.py
+++ b/tests/functional/api/test_projects.py
@@ -333,13 +333,13 @@ def test_project_groups_list(gl, group):
 
 def test_project_transfer(gl, project, group):
     assert project.namespace["path"] != group.full_path
-    project.transfer_project(group.id)
+    project.transfer(group.id)
 
     project = gl.projects.get(project.id)
     assert project.namespace["path"] == group.full_path
 
     gl.auth()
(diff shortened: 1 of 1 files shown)


tests/functional/api/test_projects.py
```diff
@@ -333,13 +334,17 @@ def test_project_groups_list(gl, group):
...
+        project.transfer_project(group.id)
 
     project = gl.projects.get(project.id)
     assert project.namespace["path"] == group.full_path
 
     gl.auth()
-    project.transfer_project(gl.user.username)
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore")
+        project.transfer_project(gl.user.username)
```
nejch, line 347: What if we just use `transfer()` instead which will be the new method?

JohnVillalovos: Can do. I thought you were intentionally testing the deprecated version.

But yeah works for me to just use the new one.

nejch: I was thinking of that initially but ended up writing a unit test to test the deprecation instead. Thanks ! :)

PR #1842, opened 2022-01-15
chore: make `Gitlab.http_request()` a private method

JohnVillalovos:
Convert `Gitlab.http_request()` to `Gitlab._http_request()` to signify
it is a private/protected method so that users of the library know
they should not use the method and we make no API stability promises
for using it.

Add a `Gitlab.http_request()` method which will issue a Deprecation
warning when called. It will pass the call onto
`Gitlab._http_request()`

Also, in the interest of improving code read-ability, require keyword
arg usage when calling `Gitlab._http_request()`

### DIFF
--- a/gitlab/client.py
+++ b/gitlab/client.py
@@ -18,6 +18,7 @@
 
 import os
 import time
+import warnings
 from typing import Any, cast, Dict, List, Optional, Tuple, TYPE_CHECKING, Union
 
 import requests
@@ -549,7 +550,7 @@ class Gitlab(object):
 
     def _check_redirects(self, result: requests.Response) -> None:
         # Check the requests history to detect 301/302 redirections.
-        # If the initial verb is POST or PUT, the redirected request will use a
+        # If the initial method is POST or PUT, the redirected request will use a
         # GET request, leading to unwanted behaviour.
         # If we detect a redirection with a POST or a PUT request, we
         # raise an exception with a useful error message.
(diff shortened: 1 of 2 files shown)


gitlab/client.py
```diff
@@ -617,11 +618,45 @@ def http_request(
         obey_rate_limit: bool = True,
         max_retries: int = 10,
         **kwargs: Any,
+    ) -> requests.Response:
+        warnings.warn(
+            "The Gitlab.http_request() method is deprecated and will be removed in a "
+            "future version. This is a private method and should not be used.",
+            DeprecationWarning,
```
nejch, line 625: One thing about this method is that it provides a way for users to do stuff we haven't implemented (or perhaps even won't, if it's super specific). So I would maybe not be so strict about "should not be used". I would maybe say that API stability can't be guaranteed or something similar.

Edit: I've now checked a bit, because I was thinking the same that we should hide a bit more of the interface. But when looking at the original author's initial implementation, I think this might have been a deliberate choice and the distinction was more of a higher vs. lower-level API. At least what I can see from the commit message and how it was done with other private methods that were not meant to be consumed, see: https://github.com/python-gitlab/python-gitlab/commit/c5ad54062ad767c0d2882f64381ad15c034e8872.

So I've hashed out some docs on some use cases here: https://github.com/python-gitlab/python-gitlab/pull/1846
It's a rough draft just for discussion on this, the wording and terminology is probably off and obviously "high/mid/low" can be changed to include "internal" or so.. but I think at least we should not take something away completely, unless we can cover those use cases in other ways. :)

Some use cases I vaguely remember from issues include using HEAD to check file existence/size, getting headers from the server's response, and so on. I documented some of that.

max-wittig: I agree with @nejch here. It's also useful for new functionality in the GitLab API.

So let's tone the message down a bit.

PR #1843, opened 2022-01-15
test: use 'responses' in test_mixins_methods.py

JohnVillalovos:
Convert from httmock to responses in test_mixins_methods.py

This leaves only one file left to convert

### DIFF
--- a/tests/unit/mixins/test_mixin_methods.py
+++ b/tests/unit/mixins/test_mixin_methods.py
@@ -1,5 +1,5 @@
 import pytest
-from httmock import HTTMock, response, urlmatch  # noqa
+import responses
 
 from gitlab import base
 from gitlab.mixins import (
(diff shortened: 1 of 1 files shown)


tests/unit/mixins/test_mixin_methods.py
```diff
@@ -24,108 +24,127 @@ class FakeManager(base.RESTManager):
...
-        obj = mgr.get(42)
-        assert isinstance(obj, FakeObject)
-        assert obj.foo == "bar"
-        assert obj.id == 42
+    mgr = M(gl)
+    obj = mgr.get(42)
+    assert isinstance(obj, FakeObject)
+    assert obj.foo == "bar"
+    assert obj.id == 42
+    assert responses.assert_call_count(url, 1) is True
```
nejch, line 46: As we aren't sharing any fixtures here, was the assert for 1 call intended for all methods or mostly for the one with 2 calls?

Maybe we could split `test_list_mixin` into `test_list_mixin` and `test_list_mixin_as_generator` then to really only test a single scenario per case, and use responses' own assert.

nejch: we can do that in a follow-up, we still have some httmock anyway.

PR #1846, opened 2022-01-16
docs(api): describe use of lower-level methods

nejch:
Rough draft, based on users' reports and for discussion in https://github.com/python-gitlab/python-gitlab/pull/1842.

Also to let people know they have a way to fetch resources even if not yet implemented here.


nejch:
Closing this to reduce the number of drafts open - I'll keep the branch and reopen when ready.

PR #1847, opened 2022-01-16
fix(objects): make resource access tokens and repos available in CLI

nejch:
This was a bit of an oversight from me a while ago and it's carried over to newer PRs.

### DIFF
--- a/gitlab/v4/objects/__init__.py
+++ b/gitlab/v4/objects/__init__.py
@@ -39,6 +39,7 @@ from .export_import import *
 from .features import *
 from .files import *
 from .geo_nodes import *
+from .group_access_tokens import *
 from .groups import *
 from .hooks import *
 from .issues import *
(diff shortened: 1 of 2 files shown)


JohnVillalovos (approved):
Thanks for catching this!

JohnVillalovos:
So this looks good to me. Only thing would be nice a simple smoke test in the CLI to show that can call these now. Could be a `gitlab some_object --help`

nejch:
> So this looks good to me. Only thing would be nice a simple smoke test in the CLI to show that can call these now. Could be a `gitlab some_object --help`

Just found out this won't work properly until we merge https://github.com/python-gitlab/python-gitlab/pull/1699 as scopes are required. So I'll add a proper test after that :)

nejch:
> > So this looks good to me. Only thing would be nice a simple smoke test in the CLI to show that can call these now. Could be a `gitlab some_object --help`
> 
> Just found out this won't work properly until we merge #1699 as scopes are required. So I'll add a proper test after that :)

Actually, we can go ahead with at least the list endpoint for projects. Added some basic tests and when 14.7 is out we can enable more.

PR #1855, opened 2022-01-21
fix(cli): allow custom methods in managers

### LINKED ISSUE #849: AttributeError: 'ProjectRegistryTagManager' object has no attribute 'delete-in-bulk'
 When I try to use `delete-in-bulk` option, using `gitlab project-registry-tag delete-in-bulk --project-id <id>--repository-id <id> --name ".*" --name-regex ".*" --older-than 1week` it fails with error 

> AttributeError: 'ProjectRegistryTagManager' object has no attribute 'delete-in-bulk'

## Expected Behavior

delete in bulk api gets called

## Actual Behavior

I'm getting error

## Specifications

  - python-gitlab version: 1.10.0
  - Gitlab server version (or gitlab.com): 12.1.1 (f9abaa7d833)

nejch:
Closes #849 

@JohnVillalovos let me know if you're not available these days and I can re-assign :)

This is still a bit dirty - we need a `DeleteWithoutIdMixin` actually, hence the note in the test. But that's out of scope for this PR I'd say. Also I'm mocking the response in the functional CLI test because of the scaffolding required to get a full docker registry repo up for tests. It still tests the actual CLI not just units though. Maybe one day we add :)

### DIFF
--- a/gitlab/v4/cli.py
+++ b/gitlab/v4/cli.py
@@ -94,6 +94,7 @@ class GitlabCLI(object):
         return self.do_custom()
 
     def do_custom(self) -> Any:
+        class_instance: Union[gitlab.base.RESTManager, gitlab.base.RESTObject]
         in_obj = cli.custom_actions[self.cls_name][self.action][2]
 
         # Get the object (lazy), then act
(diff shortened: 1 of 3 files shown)


gitlab/v4/cli.py
```diff
@@ -106,11 +106,12 @@ def do_custom(self) -> Any:
                 if TYPE_CHECKING:
                     assert isinstance(self.cls._id_attr, str)
                 data[self.cls._id_attr] = self.args.pop(self.cls._id_attr)
-            obj = self.cls(self.mgr, data)
-            method_name = self.action.replace("-", "_")
-            return getattr(obj, method_name)(**self.args)
+            class_instance = self.cls(self.mgr, data)
```
nejch, line 109: I gotta be honest I spent 15 minutes thinking about what to name this thing. But I've had enough of `_obj` and `_cls` in here :sweat_smile:

JohnVillalovos (approved):

Issue #1852, opened 2022-01-20
Humble proposal: replace alpine base image by python:slim


rassie:
## Description of the problem, including code/CLI snippet

`python-gitlab` Docker image currently uses `alpine` as its base. There are good arguments to not use Alpine, but a slim Python image instead.

## Expected Behavior

Current example: I expect DNS to work properly in the image, since I want to contact internal and external Gitlabs with `python-gitlab`.

## Actual Behavior

Currently, there is an [issue]( https://github.com/alpinelinux/docker-alpine/issues/155) preventing DNS resolution for names which don't have an AAAA DNS entry (IPv6 seems preferred). This kind of issues occurs on a regular basis with Alpine, since it's using `musl` as it's `libc`, which leads to all kind of incompatibilities with software expecting a `glibc`-like behaviour. 

Also, at least FastAPI is explicitely warning against Alpine-based images for Python projects: https://github.com/tiangolo/uvicorn-gunicorn-fastapi-docker#-alpine-python-warning. The particular FastAPI problem is probably less important for `python-gitlab`, but if choosing Alpine has been driven only by image size, it might provide additional leverage for a move to `python:slim`.

On the whole, this is not a definitive recommendation, but rather a pledge to consider the alternatives.

nejch:
Thanks for the report! The `musl`/`glibc` saga alone might make it worth adding that, especially for running on various host distros in CI. But since we're deploying to GitLab's registry I see no issue of just adding another image instead of replacing it. Just have to decide on how to alias the default one, maybe `-alpine` and `-slim` with `-alpine` aliasing the default one as otherwise it'd be weird if the default was slimmer than `slim` :)

In our case, it seems our image is 24.70 MB and base slim is 41MB (if GitLab and DockerHub show compressed size consistently), and we only bundle requests and pyyaml, so FastAPI's concerns might not yet be an issue for our existing image atm. But that makes sense.

We are also currently building with kaniko which does no compression IIRC, so that could also improve in the future.

rassie:
@nejch Thanks! Will test next week, when I get the chance!

sanmai-NL:
You mention an incompatibility, but in my experience the defect is outside musl libc, Alpine Linux, etc. but in other tools. [Like with this defect](https://gitlab.alpinelinux.org/alpine/aports/-/issues/11879#note_111445).

PR #1856, opened 2022-01-22
feat(docker): provide a Debian-based slim image

nejch:
Closes https://github.com/python-gitlab/python-gitlab/issues/1852.

Tested in https://gitlab.com/nejc/python-gitlab/-/pipelines/453883269 with a fake tag, resulting images:

https://gitlab.com/nejc/python-gitlab/container_registry/2676067

@rassie hopefully this covers your use case.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -2,26 +2,30 @@ image: python:3.10
 
 stages:
   - deploy
-  - deploy-latest
+  - promote
 
-deploy_image:
+deploy-images:
   stage: deploy
   image:
     name: gcr.io/kaniko-project/executor:debug
     entrypoint: [""]
   script:
     - echo "{\"auths\":{\"$CI_REGISTRY\":{\"username\":\"$CI_REGISTRY_USER\",\"password\":\"$CI_REGISTRY_PASSWORD\"}}}" > /kaniko/.docker/config.json
-    - /kaniko/executor --context $CI_PROJECT_DIR --dockerfile $CI_PROJECT_DIR/Dockerfile --destination $CI_REGISTRY_IMAGE:$CI_COMMIT_TAG
-  only:
-    - tags
+    - executor --context $CI_PROJECT_DIR --dockerfile $CI_PROJECT_DIR/Dockerfile --destination $CI_REGISTRY_IMAGE:$CI_COMMIT_TAG-alpine
+    - executor --context $CI_PROJECT_DIR --dockerfile $CI_PROJECT_DIR/Dockerfile --destination $CI_REGISTRY_IMAGE:$CI_COMMIT_TAG-slim-bullseye --build-arg PYTHON_FLAVOR=slim-bullseye
+  rules:
+    - if: $CI_COMMIT_TAG
 
-deploy-latest:
-  stage: deploy-latest
+tag-latest:
+  stage: promote
   image:
     name: gcr.io/go-containerregistry/crane:debug
     entrypoint: [""]
   script:
-    - mkdir /root/.docker && echo "{\"auths\":{\"$CI_REGISTRY\":{\"username\":\"$CI_REGISTRY_USER\",\"password\":\"$CI_REGISTRY_PASSWORD\"}}}" > /root/.docker/config.json
-    - /ko-app/crane cp $CI_REGISTRY_IMAGE:$CI_COMMIT_TAG $CI_REGISTRY_IMAGE:latest
-  only:
-    - tags
+    - crane auth login -u $CI_REGISTRY_USER -p $CI_REGISTRY_PASSWORD $CI_REGISTRY
+    - crane tag $CI_REGISTRY_IMAGE:$CI_COMMIT_TAG-alpine ${CI_COMMIT_TAG}       # /python-gitlab:v1.2.3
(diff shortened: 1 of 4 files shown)


README.rst
```diff
@@ -53,8 +53,23 @@ From GitLab:
...
-----------------------
+Using the docker images
+-----------------------
+
+``python-gitlab`` provides Docker images in two flavors, based on the ``-alpine`` and ``-slim``
+python `base images <https://hub.docker.com/_/python>`__. The default tag is alpine, but you
+can explicitly use the alias (see below).
+
+The alpine is image is smaller, but you may want to use the Debian-based ``-slim`` tag if you
+are running into issues or need a more complete environment with a bash shell, such as in CI.
```
JohnVillalovos, line 64:
```suggestion
The alpine image is smaller, but you may want to use the Debian-based ``-slim`` tag if you
are running into issues or need a more complete environment with a bash shell, such as in a CI.
```

JohnVillalovos (changes requested):

JohnVillalovos (approved):

.gitlab-ci.yml
```diff
@@ -2,26 +2,30 @@ image: python:3.10
...
   script:
-    - mkdir /root/.docker && echo "{\"auths\":{\"$CI_REGISTRY\":{\"username\":\"$CI_REGISTRY_USER\",\"password\":\"$CI_REGISTRY_PASSWORD\"}}}" > /root/.docker/config.json
-    - /ko-app/crane cp $CI_REGISTRY_IMAGE:$CI_COMMIT_TAG $CI_REGISTRY_IMAGE:latest
-  only:
-    - tags
+    - crane auth login -u $CI_REGISTRY_USER -p $CI_REGISTRY_PASSWORD $CI_REGISTRY
+    - crane tag $CI_REGISTRY_IMAGE:$CI_COMMIT_TAG-alpine ${CI_COMMIT_TAG} # /python-gitlab:v1.2.3
+    - crane tag $CI_REGISTRY_IMAGE:$CI_COMMIT_TAG-alpine latest           # /python-gitlab:latest
+    - crane tag $CI_REGISTRY_IMAGE:$CI_COMMIT_TAG-alpine alpine           # /python-gitlab:alpine
+    - crane tag $CI_REGISTRY_IMAGE:$CI_COMMIT_TAG-slim slim               # /python-gitlab:slim
```
JohnVillalovos, line 29: Probably dumb question. I see how it gets tagged as for example: `python-gitlab:v3.1.0-slim` but how does it get tagged as `python-gitlab:slim` ?

JohnVillalovos: My assumption is that line 29 here tags it as `python-gitlab:v3.1.0-slim`

JohnVillalovos: Ah. Line 15 is what does that tag. And line 29 is tagging it as `python-gitlab:slim`

nejch: So kaniko with a `--destination` flag will automatically push to the registry with what's defined there that's where we tag it with the two `:${CI_COMMIT_TAG}-<flavor>` tags. After the first stage, we have these 2 in the registry.

So by the time we are calling `crane tag`, those version-tagged images are already in the registry and `crane tag` just re-applies these tags in this job remotely (more efficient than doing it locally).

kaniko docs:
https://github.com/GoogleContainerTools/kaniko

crane tag:
https://github.com/google/go-containerregistry/blob/main/cmd/crane/doc/crane_tag.md#crane-tag

README.rst
```diff
@@ -53,8 +53,24 @@ From GitLab:
...
-Using the docker image
-----------------------
+Using the docker images
+-----------------------
+
+``python-gitlab`` provides Docker images in two flavors, based on the ``-alpine`` and ``-slim``
+python `base images <https://hub.docker.com/_/python>`__. The default tag is alpine, but you
+can explicitly use the alias (see below).
+
+The alpine image is smaller, but you may want to use the Debian-based ``-slim`` tag if you are
```
max-wittig, line 63: That is confusing

max-wittig: We have a smaller image, which is not called `slim` and the bigger image is called `slim` 🤔

nejch: Yes, I was also pondering this in the issue https://github.com/python-gitlab/python-gitlab/issues/1852#issuecomment-1017393648 :( will try to come up with a better wording

nejch: It's common to have `:alpine` and `:slim` with alpine being smaller, so maybe I can just add that `latest` is the same as `alpine` for backward compatibility?

nejch: @max-wittig or I could simple move `:slim` to `:slim-bullseye` and then it's clear that "this is a slim debian image"?

https://github.com/docker-library/python/blob/d7fa897e3d1c57cf144b6253ffb08a11b7a7511c/3.10/bullseye/slim/Dockerfile

nejch: @max-wittig switched to `slim-bullseye` to be more explicit about the base image.

nejch: Looking at this again I don't really like it as people will get stuck on old images then, I'll take another look

nejch: So now we just follow the python flavor tags, I've documented that. And we can upgrade the debian dist every few years manually.

JohnVillalovos (approved):

PR #1864, opened 2022-01-28
docs: revert temporary changelog link for v3

wadimklincov:
Super minor change which allows accessing the changelog from the warning at the top from anywhere in the documentation.

Right now it's e.g. `https://python-gitlab.readthedocs.io/en/stable/api/changelog.html#breaking` instead of `https://python-gitlab.readthedocs.io/en/stable/changelog.html#breaking` when accessing it from the `API reference` page. This should fix that.

I simply tested it with the builtin Python server, not sure if the behavior is the same when deployed:

```bash
tox -e docs
cd build/sphinx/html
python -m http.server
```

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -121,10 +121,7 @@ html_theme = "furo"
 # Theme options are theme-specific and customize the look and feel of a theme
 # further.  For a list of options available for each theme, see the
 # documentation.
-html_theme_options = {
-    "announcement": "⚠ python-gitlab 3.0.0 has been released with several "
-    "<a href='changelog.html#breaking'>breaking changes</a>.",
-}
+# html_theme_options = {}
 
 # Add any paths that contain custom themes here, relative to this directory.
 # html_theme_path = []


docs/conf.py
```diff
@@ -123,7 +123,7 @@
 # documentation.
 html_theme_options = {
     "announcement": "⚠ python-gitlab 3.0.0 has been released with several "
-    "<a href='changelog.html#breaking'>breaking changes</a>.",
+    "<a href='/changelog.html#breaking'>breaking changes</a>.",
 }
```
nejch, line 127: Thanks for the catch @wadimklincov! It's served on RTD mainly so this should work.

But I'm thinking v3 has been out for a month and we follow semver so people should know by now, this banner was always meant to be temporary. Maybe we could remove it? So we could just revert https://github.com/python-gitlab/python-gitlab/commit/a349793307e3a975bb51f864b48e5e9825f70182 :)

wadimklincov: Sure @nejch, sounds good to me!

wadimklincov: @nejch Replaced it with the reverted commit now. Should I update the PR message? Not sure how it works on Github!

nejch: Thanks for that @wadimklincov. I edited the title but the important part is just the commit message as we use it for changelogs, so I edited it a bit to make sure `python-semantic-release` can parse it correctly.

BTW: if you add your Siemens email to your github account it'll link your contributions there under your account, currently I think it shows as unverified :)

wadimklincov: Ah, thanks for updating it @nejch, totally missed that!

Interesting, I'll try that, thank you!

PR #1865, opened 2022-01-29
refactor(objects): provide a complete artifacts manager

nejch:
Prep work for https://docs.gitlab.com/ee/api/job_artifacts.html#delete-project-artifacts so we can just do `project.artifacts.delete()`.

This is becoming a bigger refactor now and also deprecates:
- [x] `project.artifacts()` in favor of `project.artifacts.download()`
- [x] `project.artifact()` in favor of `project.artifacts.raw()`
- [x] `job.artifacts()` in favor of `job.artifacts.download()`
- [x] `job.artifact()` in favor of `job.artifacts.raw()`
- [x] `job.delete_artifacts()` in favor of `job.artifacts.delete()`
- [x] `job.keep_artifacts()` in favor of `job.artifacts.keep()`

Todo:
- [ ] De-duplicate methods, use Mixins where possible or introduce `ObjectDownloadMixin` vs `DownloadMixin`
- [ ] Decide on method names, as GitLab API has inconsistent endpoint naming again. (maybe see generic packages for inspiration)
- [ ] Ensure encoded IDs/paths where needed
- [ ] Cleanup and test all CLI scenarios for custom actions

This also has some fixes, e.g. `**kwargs` had no effect in some of these custom methods. So we should always discourage use of custom methods.

### DIFF
--- a/gitlab/v4/objects/__init__.py
+++ b/gitlab/v4/objects/__init__.py
@@ -18,6 +18,7 @@
 from .access_requests import *
 from .appearance import *
 from .applications import *
+from .artifacts import *
 from .audit_events import *
 from .award_emojis import *
 from .badges import *
--- a/gitlab/v4/objects/artifacts.py
+++ b/gitlab/v4/objects/artifacts.py
@@ -1,0 +1,250 @@
+"""
+GitLab API:
+https://docs.gitlab.com/ee/api/job_artifacts.html
+"""
+import warnings
+from typing import Any, Callable, Optional, TYPE_CHECKING
+
+import requests
+
+from gitlab import cli
+from gitlab import exceptions as exc
+from gitlab import utils
+from gitlab.base import RESTManager
+
+__all__ = [
+    "ProjectArtifactManager",
+    "ProjectJobArtifactManager",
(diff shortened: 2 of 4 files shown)


JohnVillalovos:
What would you think about breaking it up into smaller more easier to review commits?

nejch:
> What would you think about breaking it up into smaller more easier to review commits?

Sounds good, I'll start with the ProjectArtifactManager and do only download/raw and then work from there, if that makes sense. But having all of this here helped me get an overview of what needs to be done, could be transferred to an issue though:)

JohnVillalovos:
> > What would you think about breaking it up into smaller more easier to review commits?
> 
> Sounds good, I'll start with the ProjectArtifactManager and do only download/raw and then work from there, if that makes sense. But having all of this here helped me get an overview of what needs to be done, could be transferred to an issue though:)

Thanks. Makes sense to me.

nejch:
Split into https://github.com/python-gitlab/python-gitlab/pull/1869, https://github.com/python-gitlab/python-gitlab/pull/1879 and follow-ups.

PR #1866, opened 2022-01-30
chore: create new ArrayAttribute class

JohnVillalovos:
Create a new ArrayAttribute class. This is to indicate types which are
sent to the GitLab server as arrays
https://docs.gitlab.com/ee/api/#array

At this stage it is identical to the CommaSeparatedListAttribute class
but will be used later to support the array types sent to GitLab.

This is the second step in a series of steps of our goal to add full
support for the GitLab API data types[1]:
  * array
  * hash
  * array of hashes

Step one was: commit 5127b1594c00c7364e9af15e42d2e2f2d909449b

[1] https://docs.gitlab.com/ee/api/#encoding-api-parameters-of-array-and-hash-types

Related: #1698

### DIFF
--- a/gitlab/v4/objects/issues.py
+++ b/gitlab/v4/objects/issues.py
@@ -65,10 +65,7 @@ class IssueManager(RetrieveMixin, RESTManager):
         "updated_after",
         "updated_before",
     )
-    _types = {
-        "iids": types.CommaSeparatedListAttribute,
-        "labels": types.CommaSeparatedListAttribute,
-    }
+    _types = {"iids": types.ArrayAttribute, "labels": types.CommaSeparatedListAttribute}
 
     def get(self, id: Union[str, int], lazy: bool = False, **kwargs: Any) -> Issue:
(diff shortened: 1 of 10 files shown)


gitlab/v4/objects/issues.py
```diff
@@ -98,10 +95,7 @@ class GroupIssueManager(ListMixin, RESTManager):
         "updated_after",
         "updated_before",
     )
-    _types = {
-        "iids": types.CommaSeparatedListAttribute,
-        "labels": types.CommaSeparatedListAttribute,
-    }
+    _types = {"iids": types.ArrayAttribute, "labels": types.CommaSeparatedListAttribute}
```
nejch, line 98: lol, this is infuriating:
https://docs.gitlab.com/ee/api/issues.html#list-issues
```
GET /issues?iids[]=42&iids[]=43
GET /issues?labels=foo,bar
```

:grin: thanks for going through all the endpoints. I hope the API docs are accurate.

Issue #1867, opened 2022-02-01
Deleting a label fails with 404


just-doit:
## Description of the problem, including code/CLI snippet
I'm trying to delete all labels from a given project using this code:

```python
# connect to Gitlab
gl = gitlab.Gitlab(GITLAB_URL, private_token=GITLAB_TOKEN)
gl.auth()

# Get GitLab project
gl_project = gl.projects.get(GITLAB_PROJECT_ID)

# delete ALL labels
for gl_label in gl_project.labels.list(all=True):
    gl_label.delete()
```

The token works since I've used it successfully to do some other API calls before. The same user is able to delete labels manually via Web-UI.

## Expected Behavior
I would expect the code not to throw a `gitlab.exceptions.GitlabDeleteError: 404: 404 Label Not Found'

## Actual Behavior
The `delete` call fails with a `gitlab.exceptions.GitlabDeleteError: 404: 404 Label Not Found`.

## Specifications

  - python-gitlab version: 3.1.1
  - API version you are using (v3/v4): v4
  - Gitlab server version (or gitlab.com): GitLab Enterprise Edition 14.5.0-ee

nejch:
Thanks @just-doit for the report, could you maybe add `gl.enable_debug()` and show the output here, just redact the token if it appears anywhere.

just-doit:
Hi,
this is the complete output:

```txt
DEBUG:urllib3.connectionpool:Starting new HTTPS connection (1): <server>:443
send: b'GET /api/v4/user HTTP/1.1\r\nHost: <server>\r\nUser-Agent: python-gitlab/3.1.1\r\nAccept-Encoding: gzip, deflate\r\nAccept: */*\r\nConnection: keep-alive\r\nPRIVATE-TOKEN: <secret>\r\nContent-type: application/json\r\n\r\n'
DEBUG:urllib3.connectionpool:https://<server>:443 "GET /api/v4/user HTTP/1.1" 200 None
reply: 'HTTP/1.1 200 OK\r\n'
header: Server: nginx
header: Date: Tue, 01 Feb 2022 14:41:20 GMT
header: Content-Type: application/json
header: Transfer-Encoding: chunked
header: Connection: keep-alive
header: Vary: Accept-Encoding
header: Cache-Control: max-age=0, private, must-revalidate
header: Etag: W/"ed942cb0878d870e37b48a0aeaebf227"
header: Vary: Origin
header: X-Content-Type-Options: nosniff
[...]

PR #1868, opened 2022-02-01
fix: remove custom `delete` method for labels

JohnVillalovos:
The usage of deleting was incorrect according to the current API.
Remove custom `delete()` method as not needed.

Add tests to show it works with labels needing to be encoded.

Also enable the test_group_labels() test function. Previously it was
disabled.

Add ability to do a `get()` for group labels.

Closes: #1867

### DIFF
--- a/tests/functional/api/test_groups.py
+++ b/tests/functional/api/test_groups.py
@@ -104,7 +104,6 @@ def test_groups(gl):
     group2.members.delete(gl.user.id)
 
 
-@pytest.mark.skip(reason="Commented out in legacy test")
 def test_group_labels(group):
     group.labels.create({"name": "foo", "description": "bar", "color": "#112233"})
     label = group.labels.get("foo")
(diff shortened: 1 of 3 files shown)


tests/functional/api/test_groups.py
```diff
@@ -116,6 +115,12 @@ def test_group_labels(group):
     assert label.description == "baz"
     assert len(group.labels.list()) == 1
 
+    label.new_name = "comp:Flat Export"
+    label.save()
+    assert label.name == "comp:Flat Export"
+    label = group.labels.get("comp:Flat Export")
+    assert label.name == "comp:Flat Export"
```
nejch, line 122: I'm ok to merge this as is, but maybe we could use our own example here instead to indicate what we're testing.

Maybe `"scope::Complex Label"` or something like that? I'll leave it up to you :)

nejch (approved):

tests/functional/api/test_groups.py
JohnVillalovos: Sure I can do that.

JohnVillalovos: Done

Issue #191, opened 2016-12-06
Why does project-service require id on the CLI?


jurgenhaas:
Using `gitlab project-service get --project-id 111 --service-name=slack` returns with an error that `--id` is a required parameter but when using that parameter with the correct ID, GitLab responds with a 404.

When using the library and calling `g.project_services.get(service_name='slack', project_id=project.id)` there is no id required and we get the correct result. Any ideas why?

gpocentek:
ProjectService is a bit special and should probably be handled differently in the CLI code. I'll look into it.

Thanks for the report.

PR #1870, opened 2022-02-02
fix(services): use slug for id_attr instead of custom methods

nejch:
Closes https://github.com/python-gitlab/python-gitlab/issues/191


JohnVillalovos:
We probably need some functional API stuff for this. The services stuff is weird...

JohnVillalovos (approved):
Thanks. Looks nicer 😊

PR #1871, opened 2022-02-02
chore: require kwargs for `utils.copy_dict()`

JohnVillalovos:
The non-keyword arguments were a tiny bit confusing as the destination was
first and the source was second.

Change the order and require key-word only arguments to ensure we
don't silently break anyone.

### DIFF
--- a/gitlab/client.py
+++ b/gitlab/client.py
@@ -647,7 +647,7 @@ class Gitlab:
         url = self._build_url(path)
 
         params: Dict[str, Any] = {}
-        utils.copy_dict(params, query_data)
+        utils.copy_dict(src=query_data, dest=params)
 
         # Deal with kwargs: by default a user uses kwargs to send data to the
         # gitlab server, but this generates problems (python keyword conflicts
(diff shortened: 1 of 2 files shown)


nejch:
> Non-keyword arguments are a tiny bit confusing as the destination is first and the source is second.

Should we switch them everywhere? If someone uses this undocumented function outside the library I would be slightly confused :D

JohnVillalovos:
> > Non-keyword arguments are a tiny bit confusing as the destination is first and the source is second.
> 
> Should we switch them everywhere? If someone uses this undocumented function outside the library I would be slightly confused :D

Sorry I don't understand? Switch what everywhere?

Maybe my description should have been:

The non-keyword argument order in `utils.copy_dict()` is confusing as the destination is first and the source is second.

nejch:
I understand, but since we're already making a breaking change by introducing required keyword arguments, we can also switch them to be src first, dest, second - both in the definition and when they're called. Would that make sense?

JohnVillalovos:
I guess I don't think of it as a breaking change. But maybe that means we need to define what is considered a breaking change.

What part of the API do we consider covered for it be a breaking change? We probably need to document that. I don't think an internal utility library function is part of our API.

nejch:
> I guess I don't think of it as a breaking change. But maybe that means we need to define what is considered a breaking change.
> 
> What part of the API do we consider covered for it be a breaking change? We probably need to document that. I don't think an internal utility library function is part of our API.

Sorry, I wasn't super clear there, I didn't mean something user-facing or to add to the changelog, just that the way the function can be used now has changed. So if we're already doing that, we can also correct the order and switch the args :)

JohnVillalovos:
> Sorry, I wasn't super clear there, I didn't mean something user-facing or to add to the changelog, just that the way the function can be used now has changed. So if we're already doing that, we can also correct the order and switch the args :)

Ah! That makes total sense to me 👍

Issue #1116, opened 2020-06-12
ProjetcLabel object are not JSON serializable


MatthieuPERIN:
## Description of the problem, including code/CLI snippet

Idea is to use Gitlab-Python to manage Labels in numerous project in a more scallable way 😅 

MWE
```python
# setup gl-py
gl = gitlab.Gitlab(GITLAB_ADRESS, private_token=TOKEN, ssl_verify=False)
gl.auth()

#get the project
src_prj = gl.projects.get(SRC_PRJ_ID)

# try to serialize the first Label Project
with open('data.txt', 'w') as outfile:
    json.dump(src_prj.labels.list(all=True)[0], outfile)

```

Get the following 
```python
Traceback (most recent call last):
  File "C:/Users/MPERIN/Git/labelautomation/labelmodel.py", line 79, in <module>
    json.dump(src_prj.labels.list(all=True)[0], outfile)
  File "C:\Users\MPERIN\AppData\Local\Continuum\miniconda3\envs\labelautomation\lib\json\__init__.py", line 179, in dump
    for chunk in iterable:
  File "C:\Users\MPERIN\AppData\Local\Continuum\miniconda3\envs\labelautomation\lib\json\encoder.py", line 438, in _iterencode
    o = _default(o)
  File "C:\Users\MPERIN\AppData\Local\Continuum\miniconda3\envs\labelautomation\lib\json\encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type ProjectLabel is not JSON serializable
```

## Expected Behavior
The Gilabl-python object are well serialized into JSON

## Actual Behavior
The serialization crashes 😒 

## Specifications

  - python-gitlab version: 2.2 (pip)
  - API version you are using (v3/v4): v4
  - Gitlab server version (or gitlab.com): 11.10.0

max-wittig:
You can always request the `.attributes` of an objects. That should be json serializable, but I guess it would make sense for all projects to be serializable.

Btw. your GitLab version is not supported by python-gitlab and it's dangerous to use. No security whatsoever, all vulnerabilities of it are public. You should upgrade to 13.0.0

MatthieuPERIN:
@max-wittig Thanks for the solution on `.attributes` and yes a "builtin" solution would be nicer

Agree for the GitLab version but this is a corporate internal-only use forge pending upgrade to 13.X pending (long-due) hardware migration (always the same music it seems).

[...]

PR #1872, opened 2022-02-02
feat: add a `as_dict()` method to the Gitlab Objects

JohnVillalovos:
Add an `as_dict()` method that returns a dictionary representation copy
of the Gitlab Object. This is a copy and changes made to it will have
no impact on the Gitlab Object.


nejch:
The more I compare this to `.attributes`, the more I think it's doing something very very similar and the property does have more info available.

What if we add a method like `to_json()`  that reuses our existing property to address https://github.com/python-gitlab/python-gitlab/issues/1116? That's another common pattern ([see pandas](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_json.html)), and we could potentially pass args to `dumps` for things like indenting etc.

I think this would make the example in the issue quite pretty as well:
```python
with open('data.txt', 'w') as f:
    f.write(src_prj.labels.list(all=True)[0].to_json())
```

nejch:
> The more I compare this to `.attributes`, the more I think it's doing something very very similar and the property does have more info available.
> 
> What if we add a method like `to_json()` that reuses our existing property to address #1116? That's another common pattern ([see pandas](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_json.html)), and we could potentially pass args to `dumps` for things like indenting etc.
> 
> I think this would make the example in the issue quite pretty as well:
> 
> ```python
> with open('data.txt', 'w') as f:
>     f.write(src_prj.labels.list(all=True)[0].to_json())
> ```

Going back again, seems like the library supported both json and dict method for v3 with a custom encoder back in 2017. So maybe we could just re-add both:
https://github.com/python-gitlab/python-gitlab/blob/701169441194bf0441cee13f2ab5784ffad7a207/gitlab/base.py#L525-L536

JohnVillalovos:
> Going back again, seems like the library supported both json and dict method for v3 with a custom encoder back in 2017. So maybe we could just re-add both:

Sounds good to me. I've got `asdict` in there. Let me see about `to_json`

tests/unit/test_base.py
```diff
@@ -249,3 +249,66 @@ def test_pprint(self, capfd, fake_manager):
             " 'ham': 'eggseggseggseggseggseggseggseggseggseggseggseggseggseggseggs'}\n"
         )
         assert stderr == ""
+
+    def test_asdict(self, fake_manager):
+        fake_object = FakeObject(fake_manager, {"attr1": "foo", "alist": [1, 2, 3]})
+        assert fake_object.attr1 == "foo"
+        result = fake_object.asdict()
+        assert result == {"attr1": "foo", "alist": [1, 2, 3]}
+        # Demonstrate modifying the dictionary does not modify the object
```
nejch, line 258: If it's feasible I'd change each comment into a separate test function with a descriptive name so it's visible when that part fails. That might need turning the fake object into a function scoped fixture though.

gitlab/base.py
```diff
@@ -144,15 +145,17 @@ def __getattr__(self, name: str) -> Any:
     def __setattr__(self, name: str, value: Any) -> None:
         self.__dict__["_updated_attrs"][name] = value
 
+    def asdict(self) -> Dict[str, Any]:
+        data = {}
+        data.update(copy.deepcopy(self._attrs))
+        data.update(copy.deepcopy(self._updated_attrs))
+        return data
```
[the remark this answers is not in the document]
JohnVillalovos: Done. Slightly different than suggestion above.

tests/unit/test_base.py
JohnVillalovos: Done.

max-wittig:
Very cool!

PR #1875, opened 2022-02-03
feat: emit a warning when using a `list()` method returns max

JohnVillalovos:
A common cause of issues filed and questions raised is that a user
will call a `list()` method and only get 20 items. As this is the
default maximum of items that will be returned from a `list()` method.

To help with this we now emit a warning when the result from a
`list()` method is greater-than or equal to 20 (or the specified
`per_page` value) and the user is not using either `all=True`,
`all=False`, `as_list=False`, or `page=X`.

### DIFF
--- a/gitlab/client.py
+++ b/gitlab/client.py
@@ -24,6 +24,7 @@ import requests
 import requests.utils
 from requests_toolbelt.multipart.encoder import MultipartEncoder  # type: ignore
 
+import gitlab
 import gitlab.config
 import gitlab.const
 import gitlab.exceptions
@@ -36,6 +37,12 @@ REDIRECT_MSG = (
 )
 
 RETRYABLE_TRANSIENT_ERROR_CODES = [500, 502, 503, 504] + list(range(520, 531))
+
+# https://docs.gitlab.com/ee/api/#offset-based-pagination
+_PAGINATION_URL = (
+    f"https://python-gitlab.readthedocs.io/en/v{gitlab.__version__}/"
+    f"api-usage.html#pagination"
+)
 
 
 class Gitlab:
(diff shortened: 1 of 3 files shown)


JohnVillalovos:
Example program using this:
```
$ cat show-it.py
#!/usr/bin/python3 -ttu
import gitlab



gl = gitlab.Gitlab(url="https://gitlab.com")
project = gl.projects.get("gitlab-org/gitlab")

mrs = project.mergerequests.list(state="opened")
```

Run of program:
```
./show-it.py
python-gitlab/gitlab/client.py:839: UserWarning: Calling a `list()` method without specifying `all=True` or `as_list=False` will return a maximum of 20 items. Your query returned 20 of 1280 items. See https://python-gitlab.readthedocs.io/en/v3.1.1/api-usage.html#pagination for more details
  warnings.warn(
```

gitlab/client.py
```diff
@@ -818,7 +825,28 @@ def http_list(
 
         if page or as_list is True:
             # pagination requested, we return a list
-            return list(GitlabList(self, url, query_data, get_next=False, **kwargs))
+            gl_list = GitlabList(self, url, query_data, get_next=False, **kwargs)
+            items = list(gl_list)
+            if page is None and get_all is None and gl_list.per_page is not None:
```
nejch, line 830: We're first checking for `page` and then checking if it is None - maybe an early return for at least that branch (or more scenarios) would make the if expressions a bit simpler here?

Better yet, invert the logic to flatten the entire thing:
1. if as_list is false, return generator
2. if page, return list
3. ...maybe more
3. otherwise do the complex logic below

gitlab/client.py
```diff
@@ -818,7 +825,28 @@ def http_list(
...
+                if len(items) >= gl_list.per_page and (
+                    gl_list.total is None or len(items) < gl_list.total
+                ):
+                    total_items = "10,000+" if gl_list.total is None else gl_list.total
```
nejch, line 834: 10k is specific only to the (current) GitLab.com deployment, might be better to only provide the numbers we really know. IIRC it might be even different depending on the endpoint (e.g. commit list).

gitlab/client.py
JohnVillalovos: Thanks. Good idea. I split it up.

gitlab/client.py
JohnVillalovos: Okay. I changed it to `many`

nejch:
Finally had a look around and here's some more valid use cases where we don't want to emit warnings on this - essentially whenever any kind of list filter is applied, I'd say. In those cases users mostly only care about the first item.

1. Getting a user by username (can't use the GET endpoint): https://github.com/ansible-collections/community.general/blob/21ee4c84b7da0f5f540882b6d035c47da473db1f/plugins/modules/source_control/gitlab/gitlab_user.py#L521-L525
2. Get group/project via loose name search: https://github.com/ansible-collections/community.general/blob/21ee4c84b7da0f5f540882b6d035c47da473db1f/plugins/modules/source_control/gitlab/gitlab_group_members.py#L180-L190
3. Checking for existence of items: https://github.com/ansible-collections/community.general/blob/21ee4c84b7da0f5f540882b6d035c47da473db1f/plugins/modules/source_control/gitlab/gitlab_group.py#L282-L284

gitlab/client.py
```diff
@@ -807,25 +814,60 @@ def http_list(
...
+        if get_all is False:
+            return items
+        if page is not None:
+            return items
+        if gl_list.per_page is None:
+            return items
+        if len(items) < gl_list.per_page:
+            return items
+        if gl_list.total is not None and len(items) >= gl_list.total:
+            return items
```
nejch, line 854: Since we're getting that pylint complaint, now that it's nice and flat I think this would still look ok WDYT?
```suggestion
        if (
            get_all is False
            or page is not None
            or gl_list.per_page is None
            or (len(items) < gl_list.per_page))
            or (gl_list.total is not None and len(items) >= gl_list.total)
        ):
            return items
```

Not sure if there's any easy way to check for kwargs against items in the manager's `_list_filters` to further reduce noise and address https://github.com/python-gitlab/python-gitlab/pull/1875#issuecomment-1086635196, but that would be cool to add here :)

JohnVillalovos:
Answering without doing a lot of research on these:

> Finally had a look around and here's some more valid use cases where we don't want to emit warnings on this - essentially whenever any kind of list filter is applied, I'd say. In those cases users mostly only care about the first item.
> 
>     1. Getting a user by username (can't use the GET endpoint): https://github.com/ansible-collections/community.general/blob/21ee4c84b7da0f5f540882b6d035c47da473db1f/plugins/modules/source_control/gitlab/gitlab_user.py#L521-L525

Warning will not be issued unless for some reason 20 items are returned by the `list()` call. In that case it is likely highlighting a real error. I have submitted a PR that adds a `all=True` for this.

>     2. Get group/project via loose name search: https://github.com/ansible-collections/community.general/blob/21ee4c84b7da0f5f540882b6d035c47da473db1f/plugins/modules/source_control/gitlab/gitlab_group_members.py#L180-L190

Warning will not be issued unless for some reason 20 items are returned by the `list()` call. In that case it is likely highlighting a real error. I have submitted a PR that adds a `all=True` for this.

>     3. Checking for existence of items: https://github.com/ansible-collections/community.general/blob/21ee4c84b7da0f5f540882b6d035c47da473db1f/plugins/modules/source_control/gitlab/gitlab_group.py#L282-L284

Yes a warning has a good chance of being produced for this one. I have submitted a PR that adds a `all=False` for this.

https://github.com/ansible-collections/community.general/pull/4491

gitlab/client.py
JohnVillalovos: I removed the pylint ignore by using a nested function.

nejch (approved):
Thanks for this @JohnVillalovos! Took a while as I am really uneasy about eagerly sending warnings but this will definitely reduce the issues filed.

JohnVillalovos:
> Thanks for this @JohnVillalovos! Took a while as I am really uneasy about eagerly sending warnings but this will definitely reduce the issues filed.

Thanks @nejch !  I'm interested to see feedback on it once it starts getting used by people. I know when I tried it on my code I found spots where I was like "whoops!".

PR #1876, opened 2022-02-04
docs: add transient errors retry info

emirot:
Looking through code, If I understand correctly it looks like it won't retry forever but stops after max_retries count is reached.
I that correct, I would like to add that in doc.

### DIFF
--- a/docs/api-usage.rst
+++ b/docs/api-usage.rst
@@ -421,9 +421,9 @@ GitLab server can sometimes return a transient HTTP error.
 python-gitlab can automatically retry in such case, when
 ``retry_transient_errors`` argument is set to ``True``.  When enabled,
 HTTP error codes 500 (Internal Server Error), 502 (502 Bad Gateway),
-503 (Service Unavailable), and 504 (Gateway Timeout) are retried.  By
-default an exception is raised for these errors.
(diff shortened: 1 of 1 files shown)


docs/api-usage.rst
```diff
@@ -423,6 +423,7 @@ python-gitlab can automatically retry in such case, when
 HTTP error codes 500 (Internal Server Error), 502 (502 Bad Gateway),
 503 (Service Unavailable), and 504 (Gateway Timeout) are retried.  By
 default an exception is raised for these errors.
+It will retry until reaching `max_retries` value.
```
nejch, line 426: Thanks fa lot or this @emirot! I think it might be a bit confusing if we state the retry behavior after we mention the default (no retry) behavior - maybe your PR is a chance to reword this a bit. Feel free to also shuffle it around and put the default behavior wording higher up if you feel it's more appropriate.

```suggestion
503 (Service Unavailable), and 504 (Gateway Timeout) are retried. It will retry until reaching
the `max_retries` value. By default, `retry_transient_errors` is set to `False` and an exception
is raised for these errors.
```

nejch (changes requested):

nejch (approved):
Thank you @emirot!

PR #1880, opened 2022-02-05
chore: correct type-hints for per_page attrbute

JohnVillalovos:
There are occasions where a GitLab `list()` call does not return the
`x-per-page` header. For example the listing of custom attributes.

Update the type-hints to reflect that.

### DIFF
--- a/gitlab/client.py
+++ b/gitlab/client.py
@@ -1039,11 +1039,9 @@ class GitlabList:
         return int(self._next_page) if self._next_page else None
 
     @property
-    def per_page(self) -> int:
+    def per_page(self) -> Optional[int]:
         """The number of items per page."""
-        if TYPE_CHECKING:
-            assert self._per_page is not None
-        return int(self._per_page)
+        return int(self._per_page) if self._per_page is not None else None
 
     # NOTE(jlvillal): When a query returns more than 10,000 items, GitLab doesn't return
     # the headers 'x-total-pages' and 'x-total'. In those cases we return None.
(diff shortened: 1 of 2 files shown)


gitlab/client.py
```diff
@@ -1039,10 +1039,10 @@ def next_page(self) -> Optional[int]:
         return int(self._next_page) if self._next_page else None
 
     @property
-    def per_page(self) -> int:
+    def per_page(self) -> Optional[int]:
         """The number of items per page."""
-        if TYPE_CHECKING:
-            assert self._per_page is not None
+        if self._per_page is None:
+            return None
```
nejch, line 1045: We're doing this slightly differently just above for `prev_page` and `next_page` to also handle empty headers, not sure how likely that is, just a note.

nejch (commented):
Huh, interesting, another API quirk I guess. Should we be testing this somewhere? I guess it doesn't make sense to spend time on it.

gitlab/client.py
JohnVillalovos: Above is the same but just a single line. I will update this to match.

JohnVillalovos:
> Huh, interesting, another API quirk I guess. Should we be testing this somewhere? I guess it doesn't make sense to spend time on it.

I'm not sure how much value in testing for this, but I can if you want.

nejch (approved):

nejch:
> > Huh, interesting, another API quirk I guess. Should we be testing this somewhere? I guess it doesn't make sense to spend time on it.
> 
> I'm not sure how much value in testing for this, but I can if you want.

Probably not :) just another weird one!

PR #1882, opened 2022-02-05
chore: create a custom `warnings.warn` wrapper

JohnVillalovos:
Create a custom `warnings.warn` wrapper that will walk the stack trace
to find the first frame outside of the `gitlab/` path to print the
warning against. This will make it easier for users to find where in
their code the error is generated from

### DIFF
--- a/gitlab/__init__.py
+++ b/gitlab/__init__.py
@@ -20,6 +20,7 @@ import warnings
 from typing import Any
 
 import gitlab.config  # noqa: F401
+from gitlab import utils as _utils
 from gitlab._version import (  # noqa: F401
     __author__,
     __copyright__,
@@ -40,11 +41,13 @@ warnings.filterwarnings("default", category=DeprecationWarning, module="^gitlab"
 def __getattr__(name: str) -> Any:
     # Deprecate direct access to constants without namespace
     if name in gitlab.const._DEPRECATED:
-        warnings.warn(
-            f"\nDirect access to 'gitlab.{name}' is deprecated and will be "
-            f"removed in a future major python-gitlab release. Please "
-            f"use 'gitlab.const.{name}' instead.",
-            DeprecationWarning,
+        _utils.warn(
+            message=(
+                f"\nDirect access to 'gitlab.{name}' is deprecated and will be "
+                f"removed in a future major python-gitlab release. Please "
+                f"use 'gitlab.const.{name}' instead."
+            ),
+            category=DeprecationWarning,
         )
         return getattr(gitlab.const, name)
     raise AttributeError(f"module {__name__} has no attribute {name}")
--- a/gitlab/utils.py
+++ b/gitlab/utils.py
@@ -15,8 +15,11 @@
 # You should have received a copy of the GNU Lesser General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import pathlib
+import traceback
 import urllib.parse
-from typing import Any, Callable, Dict, Optional, Union
+import warnings
+from typing import Any, Callable, Dict, Optional, Type, Union
 
 import requests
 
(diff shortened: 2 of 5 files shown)


JohnVillalovos:
Test script:
```
$ cat show-warning.py
#!/usr/bin/python3 -ttu
import gitlab

x = gitlab.OWNER_ACCESS
```

With this PR, warning shows against the script:
```
/home/python-gitlab/./show-warning.py:4: DeprecationWarning:
Direct access to 'gitlab.OWNER_ACCESS' is deprecated and will be removed in a future major python-gitlab release. Please use 'gitlab.const.OWNER_ACCESS' instead.
  x = gitlab.OWNER_ACCESS
```


Without this PR, warning shows against `gitlab/__init__.py`:
```./show-warning.py
/home/python-gitlab/gitlab/__init__.py:43: DeprecationWarning:
Direct access to 'gitlab.OWNER_ACCESS' is deprecated and will be removed in a future major python-gitlab release. Please use 'gitlab.const.OWNER_ACCESS' instead.
  warnings.warn(
```

nejch:
Thanks @JohnVillalovos. If I understand correctly you want to show the caller and not the source in our library when showing warnings. Wouldn't this achieve the goal without custom wrappers?

```python
        warnings.warn(
            f"\nDirect access to 'gitlab.{name}' is deprecated and will be "
            f"removed in a future major python-gitlab release. Please "
            f"use 'gitlab.const.{name}' instead.",
            DeprecationWarning,
            stacklevel=2,
        )
```

Let me know if that works.

JohnVillalovos:
> Thanks @JohnVillalovos. If I understand correctly you want to show the caller and not the source in our library when showing warnings. Wouldn't this achieve the goal without custom wrappers?
> 
> ```python
>         warnings.warn(
>             f"\nDirect access to 'gitlab.{name}' is deprecated and will be "
>             f"removed in a future major python-gitlab release. Please "
>             f"use 'gitlab.const.{name}' instead.",
>             DeprecationWarning,
>             stacklevel=2,
>         )
> ```
> 
> Let me know if that works.

It works but have to determine what `stacklevel` is for each function. And if somehow the function got moved it would change where in the stack it is and then need to update that value. That is why I went with the wrapper which determines it automatically. I figure better to let the computer keep track of it than us manually doing it.

Some of these require `stacklevel=2` and some require `stacklevel=3`.

nejch:
> It works but have to determine what `stacklevel` is for each function. And if somehow the function got moved it would change where in the stack it is and then need to update that value. That is why I went with the wrapper which determines it automatically. I figure better to let the computer keep track of it than us manually doing it.
> 
> Some of these require `stacklevel=2` and some require `stacklevel=3`.

Ahh I see, ok, that wasn't clear to me at first glance. I was hoping it was a simpler one-time thing and we'd avoid having a heated debate about required keyword-only arguments but now we'll have to have it :sweat_smile:

gitlab/__init__.py
```diff
@@ -20,6 +20,7 @@
 from typing import Any
 
 import gitlab.config  # noqa: F401
+from gitlab import utils as _utils
```
nejch, line 23: Not sure I fully understand the context here for `_utils` compared to how it's imported in the rest of the library. If it's needed, maybe we just go with `_utils.py` then to avoid `import thing as another_thing` and be consistent? Would still be very pep8-ish:

> Even with `__all__` set appropriately, internal interfaces (packages, modules, classes, functions, attributes or other names) should still be prefixed with a single leading underscore.

gitlab/utils.py
```diff
@@ -90,3 +93,30 @@ def __new__(  # type: ignore
...
 def remove_none_from_dict(data: Dict[str, Any]) -> Dict[str, Any]:
     return {k: v for k, v in data.items() if v is not None}
+
+
+def warn(
+    *,
+    message: str,
+    category: Optional[Type] = None,
+    source: Optional[Any] = None,
+) -> None:
```
nejch, line 103: IMO we could have the signature the same as what's being wrapped, e.g. `message` is a positional. I think with one positional argument the intention is quite clear and a good middle ground in terms of explicitness/verbosity, as is done in stdlib in https://docs.python.org/3/library/warnings.html#warnings.warn.

gitlab/utils.py
```diff
@@ -90,3 +93,30 @@ def __new__(  # type: ignore
...
+    """This `warnings.warn` wrapper function attempts to show the location causing the
+    warning in the user code that called the library.
+
+    It does this by walking up the stack trace to find the first frame located outside
+    the `gitlab/` directory. This is helpful to users as it shows them their code that
+    is causing the warning.
+    """
+    # Get `stacklevel` for user code so we indicate where issue is in
+    # their code.
+    pg_dir = os.path.abspath(os.path.dirname(__file__))
```
nejch, line 113: Apart from some legacy config files I believe we've mostly already migrated everything to `pathlib`. 
This would be something like `Path(__file__).parent.resolve()` I think. Then you can use `is_relative_to()` later on.

```suggestion
    gitlab_dir = Path(__file__).parent.resolve()
```

nejch (commented):
Just had some comments @JohnVillalovos after the explanation now :)

gitlab/__init__.py
JohnVillalovos: I did that as I didn't want to possibly interfere with usage of `gitlab.utils`. Not positive this is necessary but was doing it to be safe.

gitlab/utils.py
JohnVillalovos: Looks nice. Will do.

gitlab/utils.py
JohnVillalovos: Changed to `message` being a positional argument.

gitlab/utils.py
JohnVillalovos: Done

JohnVillalovos: > Apart from some legacy config files I believe we've mostly already migrated everything to `pathlib`. This would be something like `Path(__file__).parent.resolve()` I think. Then you can use `is_relative_to()` later on.

Darn. No `is_relative_to()` until Python 3.9. 😟

PR #1884, opened 2022-02-06
docs: use `as_list=False` or `all=True` in Getting started

JohnVillalovos:
In the "Getting started with the API" section of the documentation,
use either `as_list=False` or `all=True` in the example usages of the
`list()` method.

Also add a warning about the fact that `list()` by default does not
return all items.

### DIFF
--- a/docs/api-usage.rst
+++ b/docs/api-usage.rst
@@ -93,19 +93,25 @@ Examples:
 .. code-block:: python
 
    # list all the projects
-   projects = gl.projects.list()
+   projects = gl.projects.list(as_list=False)
    for project in projects:
        print(project)
 
    # get the group with id == 2
    group = gl.groups.get(2)
-   for project in group.projects.list():
+   for project in group.projects.list(as_list=False):
        print(project)
 
    # create a new user
    user_data = {'email': 'jen@foo.com', 'username': 'jen', 'name': 'Jen'}
    user = gl.users.create(user_data)
    print(user)
+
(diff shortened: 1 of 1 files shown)


docs/api-usage.rst
```diff
@@ -93,20 +93,26 @@ Examples:
...
    # list all the projects
-   projects = gl.projects.list()
+   projects = gl.projects.list(as_list=False)
    for project in projects:
        print(project)
 
    # get the group with id == 2
    group = gl.groups.get(2)
-   for project in group.projects.list():
+   for project in group.projects.list(as_list=False):
```
nejch, line 102: Maybe we can just have `all=True` everywhere, seems like we mix it now a bit. And just explain `as_list=False` for when they want an efficient generator (if we don't already below).

This is the intro page for the API and `list(as_list=False)` might look funny when it's there before it's explained.

JohnVillalovos: The problem with this example and `all=True` is if they run it on `gitlab.com` it will take a LONG time.

nejch: Ah that's true, but we have the same with then endlessly printing over the iterator. Maybe it's clearer if we give an example each without, with all=True and with as_list=False explicitly, like we do with the token examples?

nejch (approved):

JohnVillalovos:
@nejch Since you approved this I rebased it and also removed the "Draft" status.

Issue #1631, opened 2021-10-11
Delete runner using `runner_token`


Silvanoc:
## Description of the problem, including code/CLI snippet

`python-gitlab` should support deleting a Runner registration by providing only it's authentication token (`runner-token`), the same way the GitLab API supports it.

## Expected Behavior

`gl.runners.delete(token=runner_token)` deletes the GitLab Runner that authenticates with the token `runner_token`.

## Actual Behavior

`gl.runners.delete(token=runner_token)` fails with following error message:
```
Traceback (most recent call last):
  File "<...>", line <...>, in <module>
    gl.runners.delete(token=runner_token)
  File "/home/silvano/.local/lib/python3.9/site-packages/gitlab/exceptions.py", line 304, in wrapped_f
    return f(*args, **kwargs)

TypeError: delete() missing 1 required positional argument: 'id'
```

Corresponding API specification: https://docs.gitlab.com/ee/api/runners.html#delete-a-runner-by-authentication-token

## Specifications

  - python-gitlab version: 2.10.1
  - API version you are using (v3/v4): v4
  - Gitlab server version (or gitlab.com): 14.3.1

PR #1887, opened 2022-02-10
feat(mixins): allow deleting resources without IDs

nejch:
Closes https://github.com/python-gitlab/python-gitlab/issues/1631

I started out this PR hoping we would have a generic DeleteMixin for both runners and artifacts, but of course the artifacts endpoint is inconsistent with the other ones. So again I had to add a tiny custom method :sweat_smile:


JohnVillalovos (approved):
Thanks. Looks good.

Issue #1889, opened 2022-02-10
Default value for max_retries is too short to "clear" all rate limit windows


swarren:
## Description of the problem, including code/CLI snippet

The default value of max_retries is insufficient to "clear" any rate limiting limit, for responses where the Retry-After header is missing. The default is 10, which yields a max cur_retries value of 9, which yields a maximum sleep time of ~50 seconds. At least some rate limits are computed over a 10 minute window (600 seconds) (e.g. the recently added limit on /users/:id), so ideally we'd have a default max_retries that would (cumulatively) sleep at least that long.

## Expected Behavior

Default max_retries value is large enough to cause the library to sleep long enough to pass any rate limit window.

## Actual Behavior

The library doesn't sleep long enough, so a exception is thrown when the rate limit is reached, despite retrying for a while.

## Specifications

  - python-gitlab version: 1.15.0 (but I checked the latest code in git and it appears identical)
  - API version you are using (v3/v4): Whatever the library default is.
  - Gitlab server version (or gitlab.com): gitlab.com

JohnVillalovos:
I notice this on the `user` endpoint

I removed the non `RateLimit` headers.

```
{'RateLimit-Limit': '2000',
 'RateLimit-Observed': '20',
 'RateLimit-Remaining': '1980',
 'RateLimit-Reset': '1645057248',
 'RateLimit-ResetTime': 'Thu, 17 Feb 2022 00:20:48 GMT'}
```

JohnVillalovos:
Looks like we need to add support for these new headers: https://docs.gitlab.com/ee/user/admin_area/settings/user_and_ip_rate_limits.html#response-headers

swarren:
@JohnVillalovos I don't believe that change will solve my original issue. There are no rate-limit or retry headers at all when the rate limit is hit on `/users/:id`:

`for i in $(seq 305); do curl --verbose --header "Authorization: Bearer xxx" "https://gitlab.com/api/v4/users/5400102"; done 2>&1 | tee log.txt
`

eventually yields:

`
HTTP/2 429 
< date: Thu, 10 Mar 2022 04:08:28 GMT
< content-type: application/json
< content-length: 89
< cache-control: no-cache
< vary: Origin
< x-content-type-options: nosniff
< x-frame-options: SAMEORIGIN
< x-request-id: 01FXS0728WE6KDCRW14TR5RE9P
< x-runtime: 0.026642
< strict-transport-security: max-age=31536000
[...]

PR #1895, opened 2022-02-17
fix: support RateLimit-Reset header

JohnVillalovos:
Some endpoints are not returning the `Retry-After` header when
rate-limiting occurrs. In those cases use the `RateLimit-Reset` [1]
header, if available.

Closes: #1889

[1] https://docs.gitlab.com/ee/user/admin_area/settings/user_and_ip_rate_limits.html#response-headers

### DIFF
--- a/gitlab/client.py
+++ b/gitlab/client.py
@@ -700,10 +700,14 @@ class Gitlab:
             if (429 == result.status_code and obey_rate_limit) or (
                 result.status_code in [500, 502, 503, 504] and retry_transient_errors
             ):
+                # Response headers documentation:
+                # https://docs.gitlab.com/ee/user/admin_area/settings/user_and_ip_rate_limits.html#response-headers
                 if max_retries == -1 or cur_retries < max_retries:
                     wait_time = 2 ** cur_retries * 0.1
                     if "Retry-After" in result.headers:
                         wait_time = int(result.headers["Retry-After"])
+                    elif "RateLimit-Reset" in result.headers:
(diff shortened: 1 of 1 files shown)


nejch:
Thanks @JohnVillalovos! Would it make sense to add another case to https://github.com/python-gitlab/python-gitlab/blob/4cb7d9224fb24e4a13fdff8271de5ce083ad7757/tests/functional/api/test_gitlab.py#L160-L183 with endpoints that don't return `retry-after`, e.g. `user` as mentioned in the issue? Although this might need 14.8 and not sure if it's enabled by default. Just wondering. https://docs.gitlab.com/ee/user/admin_area/settings/rate_limit_on_users_api.html#rate-limits-on-users-api

(If we do this, I would maybe turn the settings into a fixture to use in both tests, and do the cleanup after `yield` to make it more reusable, e.g.)
```python
@pytest.fixture
def rate_limit_settings(gl):
    settings = gl.settings.get()
    settings.throttle_authenticated_api_enabled = True
    settings.throttle_authenticated_api_requests_per_period = 1
    settings.throttle_authenticated_api_period_in_seconds = 3
    settings.save()

    yield settings

    settings.throttle_authenticated_api_enabled = False
    settings.save()
```

PS: Interesting, looks like these headers are also useful to avoid 429s altogether, looking at go-gitlab with some fancy rate-limiting code: https://github.com/xanzy/go-gitlab/commit/d8bb0b4522eed51a265ffa0f44a291403ed4d85d

nejch (approved):

PR #1896, opened 2022-02-19
feat: add Project CI Lint support

JohnVillalovos:
Add support for validating a project's CI configuration [1]

[1] https://docs.gitlab.com/ee/api/lint.html

### DIFF
--- a/gitlab/v4/objects/projects.py
+++ b/gitlab/v4/objects/projects.py
@@ -1,3 +1,8 @@
+"""
+GitLab API:
+https://docs.gitlab.com/ee/api/projects.html
+https://docs.gitlab.com/ee/api/lint.html
+"""
 from typing import (
     Any,
     Callable,
@@ -97,6 +102,8 @@ __all__ = [
     "ProjectRemoteMirrorManager",
     "ProjectStorage",
     "ProjectStorageManager",
+    "ProjectCiLint",
+    "ProjectCiLintManager",
 ]
 
 
@@ -158,6 +165,7 @@ class Project(RefreshMixin, SaveMixin, ObjectDeleteMixin, RepositoryMixin, RESTO
     badges: ProjectBadgeManager
     boards: ProjectBoardManager
     branches: ProjectBranchManager
+    ci_lint: "ProjectCiLintManager"
     clusters: ProjectClusterManager
     commits: ProjectCommitManager
     customattributes: ProjectCustomAttributeManager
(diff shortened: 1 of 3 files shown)


nejch:
Huh, this is an odd situation as we have POST and GET to the same endpoint doing almost the same thing, plus the POST on the instance level (https://github.com/python-gitlab/python-gitlab/blob/main/gitlab/client.py#L374-L393). 

I would expect just POST without `content` to default to using the latest HEAD instead of doing GET/POST, interesting. Might be a topic for the GitLab API call as well.

nejch:
Huh, this is an odd situation as we have POST and GET to the same endpoint doing almost the same thing, plus the POST on the instance level (https://github.com/python-gitlab/python-gitlab/blob/main/gitlab/client.py#L374-L393). 

I would expect just POST without `content` to default to using the latest HEAD instead of doing GET/POST, interesting. Might be a topic for the GitLab API call as well.

GabDug:
Hello! I hope this PR could be merged, thanks for the work :)

To add my two cents,
- About the best approach considering the different API calls. I think the lib could implement them, as I don't expect client libs to try and solve "weirdnesses" in the server API.
- Nevertheless, the `GET` endpoint could probably be included as a first step, as it provides value not included in the instance level call (get merged_yaml without providing any context other than the project id).

JohnVillalovos:
@nejch and @GabDug  I finally got back to this. Now has docs and tests. So it is finally ready for review.

gitlab/v4/objects/projects.py
```diff
@@ -1055,3 +1063,17 @@ class ProjectStorageManager(GetWithoutIdMixin, RESTManager):
...
+
+class ProjectCiLint(RESTObject):
+    pass
+
+
+class ProjectCiLintManager(GetWithoutIdMixin, CreateMixin, RESTManager):
+    _path = "/projects/{project_id}/ci/lint"
+    _obj_cls = ProjectCiLint
+    _from_parent_attrs = {"project_id": "id"}
+    # https://docs.gitlab.com/ee/api/lint.html#validate-a-projects-ci-configuration
```
nejch, line 1076: I think the section before this one is relevant as we cover both GET/POST, so from my perspective we could just link them all above in the module docstrings.

https://docs.gitlab.com/ee/api/lint.html#validate-a-ci-yaml-configuration-with-a-namespace

JohnVillalovos: I removed the anchor tags from both. Forgot to update them when I added support for `create` (`POST`)

JohnVillalovos: I do like to have them with this class as it is the one class that is not documented in https://docs.gitlab.com/ee/api/projects.html  So makes it easier for people to realize that when looking at the code.

nejch: Ah I see. I just prefer docstrings (even if in classes) over comments in general, but no blocker :)

JohnVillalovos: Oh. I can make it a doc-string...

nejch (approved):

PR #1904, opened 2022-03-02
Retry additional http transient errors

Sineaggi:
Based off of https://github.com/python-gitlab/python-gitlab/pull/1648

This adds additional retries for the 52x caused by cloudflare.

This adds tests for the 52x range and the `requests.ConnectionError` exception

### DIFF
--- a/gitlab/client.py
+++ b/gitlab/client.py
@@ -35,6 +35,8 @@ REDIRECT_MSG = (
     "{source!r} to {target!r}"
 )
 
+RETRYABLE_TRANSIENT_ERROR_CODES = [500, 502, 503, 504] + list(range(520, 531))
+
 
 class Gitlab:
     """Represents a GitLab server connection.
@@ -694,9 +696,9 @@ class Gitlab:
                 )
             except requests.ConnectionError:
                 if retry_transient_errors and (
-                        max_retries == -1 or cur_retries < max_retries
+                    max_retries == -1 or cur_retries < max_retries
                 ):
-                    wait_time = 2 ** cur_retries * 0.1
+                    wait_time = 2**cur_retries * 0.1
                     cur_retries += 1
                     time.sleep(wait_time)
                     continue
@@ -712,7 +714,8 @@ class Gitlab:
                 "retry_transient_errors", self.retry_transient_errors
             )
             if (429 == result.status_code and obey_rate_limit) or (
-                result.status_code in [500, 502, 503, 504] and retry_transient_errors
+                result.status_code in RETRYABLE_TRANSIENT_ERROR_CODES
+                and retry_transient_errors
             ):
                 # Response headers documentation:
                 # https://docs.gitlab.com/ee/user/admin_area/settings/user_and_ip_rate_limits.html#response-headers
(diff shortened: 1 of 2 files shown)


f9n:
Running perfectly 👍 

```
# On requirements.txt
git+https://github.com/Sineaggi/python-gitlab@retry-additional-http-transient-errors#egg=python-gitlab
```

gitlab/client.py
```diff
@@ -675,19 +675,33 @@ def http_request(
...
+                    timeout=timeout,
+                    verify=verify,
+                    stream=streamed,
+                    **opts,
+                )
+            except requests.ConnectionError:
+                if retry_transient_errors and (
+                    max_retries == -1 or cur_retries < max_retries
+                ):
+                    wait_time = 2 ** cur_retries * 0.1
```
nejch, line 699: We've recently updated black to v22 and I think it'll complain here after merge. Could you maybe rebase this PR and check that formatting is ok?

gitlab/client.py
```diff
@@ -698,7 +712,11 @@ def http_request(
                 "retry_transient_errors", self.retry_transient_errors
             )
             if (429 == result.status_code and obey_rate_limit) or (
-                result.status_code in [500, 502, 503, 504] and retry_transient_errors
+                (
+                    result.status_code in [500, 502, 503, 504]
+                    or 520 <= result.status_code <= 530
+                )
```
nejch, line 718: As this list grows I'm thinking we might get more arbitrary codes in this list - would it make sense to put it in a helper or constant? Then we can easily document/comment where these status codes come from.

nejch (commented):
Thanks for taking over @Sineaggi. I just have 2 quick questions here.

gitlab/client.py
Sineaggi: I've rebased and reformatted.

gitlab/client.py
Sineaggi: I've added `RETRYABLE_TRANSIENT_ERROR_CODES`

Sineaggi:
@nejch Can you re-review?

gitlab/client.py
```diff
@@ -698,7 +714,8 @@ def http_request(
                 "retry_transient_errors", self.retry_transient_errors
             )
```
nejch, line 715: Could we have this block removed as you've now added it above? I think it might just be a rebase issue.

nejch (changes requested):
Thanks again @Sineaggi, I was on holiday so there was a bit of delay there. I just made one last pass and just noticed some duplicate code. Could you clean that up?

We could probably also make the test code a bit shorter but that's for later perhaps. Thanks!

I'll make sure this makes it into the next release so I can also pause the scheduled release for the 28th until this is merged, if needed.

gitlab/client.py
nejch: Done in https://github.com/python-gitlab/python-gitlab/pull/1904/commits/5cbbf26e6f6f3ce4e59cba735050e3b7f9328388.

nejch (approved):

PR #1905, opened 2022-03-03
docs(chore): Include docs .js files in sdist

derekschrock:


nejch:
Thanks for the catch yet again @derekschrock :)

nejch:
Thanks for the catch yet again @derekschrock :)

PR #1951, opened 2022-03-28
docs:Small docs fixup for application scopes

wacuuu:
Hi all,

I found a small hickup in the docs that took me a few minutes to understand. Looking at [API docs](https://docs.gitlab.com/ee/api/applications.html) and the code I have found out that the way we try to push the application scopes doesn't match the API. I'd like to leave an appropriate example in docs so that people don't get stuck on this

### DIFF
--- a/docs/gl_objects/applications.rst
+++ b/docs/gl_objects/applications.rst
@@ -22,7 +22,7 @@ List all OAuth applications::
 
 Create an application::
 
-    gl.applications.create({'name': 'your_app', 'redirect_uri': 'http://application.url', 'scopes': ['api']})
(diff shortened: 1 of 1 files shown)


nejch:
Thanks for the catch @wacuuu! Surprised to see they're using space-delimited strings here, in most other places they use arrays for `scopes` which is why it was documented like that.

Could you please rebase again as there was a recent broken version of black that breaks our CI? https://github.com/psf/black/issues/2964.

wacuuu:
Oh, so that is why linter broke. I couldn't understand what is wrong.

Issue #1955, opened 2022-03-30
Is there a way to get the response iterator directly when downloading artifacts?


TCatshoek:
## Description of the problem, including code/CLI snippet

The artifact downloads support a streaming mode which wraps the iterator provided by the `requests` library  and allows the user to provide an `action` callable that is called with the chunks provided by the iterator as arguments.

I would like to be able to access the `response.iter_content()` iterator directly. 

## Expected Behavior

My use case would be something like this:

```python
iter_response = project.artifacts.download(
    ref_name="main", job="build", streamed=True, action='iterator'
)

do_things_with_iterator(iter_response)
```

( In my real project `do_things_with_iterator` is actually a fastapi `StreamingResponse` which takes an iterator or generator. This would allow downloads of artifacts without first having to download the entire file server side and then forwarding it to the client)

## Actual Behavior
As far as I can tell this is currently not possible.

I thought this would be relatively simple to implement, so I adapted the `response_content` function in `utils.py` to 

```python
def response_content(
    response: requests.Response,
    streamed: bool,
    action: Optional[Union[Callable, Literal["iterator"]]],
    chunk_size: int,
) -> Optional[Union[bytes, Iterator[Any]]]:
    if streamed is False:
        return response.content

    if action is None:
        action = _StdoutStream()

    if action == "iterator":
        return response.iter_content(chunk_size=chunk_size)

    for chunk in response.iter_content(chunk_size=chunk_size):
        if chunk:
            action(chunk)
    return None
```

But I ran into issues adapting the type annotations in other places, as some functions are not designed to handle iterators and will only handle bytes. I did not have a lot of time to spend on figuring this out though so if it seems worth the effort I can look into this further and create a pull request later.

## Specifications

  - python-gitlab version: 3.2.0
  - API version you are using (v3/v4): v4
  - Gitlab server version (or gitlab.com): gitlab.com

TCatshoek:
Created a pull request https://github.com/python-gitlab/python-gitlab/pull/1956

Please close if there's a better way to do this that I missed.

TCatshoek:
In the meantime I found a (slightly ugly) workaround to get the desired behavior from python-gitlab 3.2.0 and fastapi. It needs a supporting class that python-gitlab can write the chunks to in a different thread, which gets wrapped in a generator that fastapi can use.

```python
class BytesLoop:
    def __init__(self, s=b''):
        self.buffer = s
        self.lock = threading.Lock()

[...]

PR #1956, opened 2022-03-30
feat(downloads): allow streaming downloads access to response iterator

TCatshoek:
Allow access to the underlying response iterator when downloading in
streaming mode by specifying action="iterator"

Also update type annotations to support this change


nejch:
Thank you for all the work here @TCatshoek!

For this to work with typing on 3.7, we'll need to add a conditional dependency on https://pypi.org/project/typing-extensions/.

It would also be great to document this a little and if you could maybe add a specific use case there (e.g. from your issue) a bit to give context :) I think doing this with `action="iterator"` is fine but I can also poke around to see how other people might do this.

As a bonus, if you have the energy, it'd be great if we could maybe also do a functional test for this. I know we don't have any targeted tests for this but perhaps down here:
https://github.com/python-gitlab/python-gitlab/blob/5a1678f43184bd459132102cc13cf8426fe0449d/tests/functional/api/test_packages.py#L38-L62

TCatshoek:
@nejch Awesome! I'll work on tests and docs as soon as I have some time. 

In the meantime I managed to find a suitable, albeit kind of ugly workaround to get the behavior I want. I'll describe it in the issue I opened at https://github.com/python-gitlab/python-gitlab/issues/1955

TCatshoek:
I added the conditional dependency on typing_extensions for python 3.7. Had to shuffle some imports around to make the linter happy, I hope it is ok like this!

TCatshoek:
Added tests too!

docs/gl_objects/pipelines_and_jobs.rst
```diff
@@ -274,6 +274,18 @@ You can also directly stream the output into a file, and unzip it afterwards::
     subprocess.run(["unzip", "-bo", zipfn])
     os.unlink(zipfn)
 
+It is also possible to use the underlying iterator directly::
```
nejch, line 277:
```suggestion
Or, you can also use the underlying response iterator directly::
```

docs/gl_objects/pipelines_and_jobs.rst
```diff
@@ -274,6 +274,18 @@ You can also directly stream the output into a file, and unzip it afterwards::
...
+
+    artifact_bytes_iterator = build_or_job.artifacts(streamed=True, action='iterator')
+
+This can be used with FastAPI/Starlette StreamingResponse to forward a download from gitlab without having to download
+the entire file server side first::
```
nejch, line 282: Maybe this would sound slightly more generic?

```suggestion
This can be used with frameworks that expect an iterator (such as FastAPI/Starlette's
``StreamingResponse``) to forward a download from GitLab without having to download
the entire content server-side first::
```

requirements.txt
```diff
@@ -1,2 +1,3 @@
 requests==2.27.1
 requests-toolbelt==0.9.1
+typing_extensions; python_version<"3.8"
```
nejch, line 3: This works in testing, but I assume we'd also need this in `setup.py` for the installed package for users.

However, let me just take another look because if we're only doing this for `Literal`, it might make more sense to add a separate argument in addition to `action`, to keep typing clearer. (Also we'd need to change the docstrings otherwise)

gitlab/v4/objects/snippets.py
```diff
@@ -12,6 +13,11 @@
 from .discussions import ProjectSnippetDiscussionManager  # noqa: F401
 from .notes import ProjectSnippetNoteManager  # noqa: F401
 
+if sys.version_info >= (3, 8):
+    from typing import Literal
+else:
+    from typing_extensions import Literal
+
```
nejch, line 20: I'd likely do this in a try/except block as it used to be done for most py2/3 imports, but see my other comment about potentially getting rid of it - sorry and thanks for your patience :)

nejch (commented):
Just a few suggestions for the docs @TCatshoek and I'll just check if it makes more sense to add another argument, I'll get back to you ASAP. Thanks :)

docs/gl_objects/pipelines_and_jobs.rst
TCatshoek: Agreed,  that sounds better

docs/gl_objects/pipelines_and_jobs.rst
TCatshoek: Same here!

requirements.txt
TCatshoek: Ah yes, you're right. Let me know what you decide and I will update things accordingly!

gitlab/v4/objects/snippets.py
TCatshoek: I tried this using a try/except block at first but mypy complained:
```
error: Incompatible import of "Literal" (imported name has type "typing_extensions._SpecialForm", local name has type "typing._SpecialForm")
```
Then I found this in the mypy docs: https://mypy.readthedocs.io/en/stable/runtime_troubles.html#using-new-additions-to-the-typing-module which seems to work ok. 

If you know a way to get it to play nice with try/except I'd be happy to do that too of course.

TCatshoek:
Sorry, had to force push because the commit messages from accepting the suggested changes through the web interface were not accepted.

requirements.txt
TCatshoek: @nejch Any news on this?

nejch: Thanks for your patience @TCatshoek!

I just took another look at this and I feel we should maybe leave `action` a Callable only because we do this later:
https://github.com/python-gitlab/python-gitlab/blob/792cee939843d8df4c87bb8068be147ec97fabac/gitlab/utils.py#L44-L46

So I think that might be a bit confusing. How about an explicit `iterator: bool = False` argument that would follow the `streamed` logic a bit more? This would also get rid of the typing dependency we discussed as I see some people don't like having even the tiniest extra dependencies.. :)

Thanks again for working on this! Let me know what you think.

nejch: Just had some discussion with @JohnVillalovos and we're thinking of intruducing an `iterator=True` for our `list()` calls where we currently use `as_list=False` (which is kind of unclear in what it really does currently), so this would then make it a consistent interfaces as well. So I think a bool with `iterator` is the way to go,

nejch:
Would you still be interested in adapting this for `iterator=True` @TCatshoek? If so we could merge it before the 28th when the next release is scheduled. :)

TCatshoek:
Yeah! I think I should have some time this weekend, I'll try to get it done before the 28th

nejch (approved):

nejch:
Thanks for all the work here @TCatshoek!

TCatshoek:
@nejch Glad I could help! :D

Issue #987, opened 2020-01-08
Additional arguments are passed on to the API


mitar:
## Description of the problem, including code/CLI snippet

I want to control the logic for retrying, so I use:

```
project.files.raw(file_path='README.md', ref='master', max_retries=10, retry_transient_errors=True)
```

## Expected Behavior

The request is made to gitlab.com which looks like:

```
/api/v4/projects/123/repository/files/README%2Emd/raw?ref=master
```

And if request fails, it is retried.

## Actual Behavior

But the actual request is:

```
/api/v4/projects/123/repository/files/README%2Emd/raw?ref=master&max_retries=10&retry_transient_errors=True
```

## Specifications

  - python-gitlab version: python-gitlab==1.15.0
  - API version you are using (v3/v4): v4
  - Gitlab server version (or gitlab.com): gitlab.com

PR #1965, opened 2022-04-04
fix: avoid passing redundant arguments to API

nejch:
Closes https://github.com/python-gitlab/python-gitlab/issues/987.

This was already fixed in https://github.com/python-gitlab/python-gitlab/commit/b9ecc9a8c5d958bd7247946c4e8d29c18163c578 but partially reintroduced with the global `retry_transient_errors` feature.

### DIFF
--- a/gitlab/client.py
+++ b/gitlab/client.py
@@ -56,8 +56,8 @@ class Gitlab:
         pagination: Can be set to 'keyset' to use keyset pagination
         order_by: Set order_by globally
         user_agent: A custom user agent to use for making HTTP requests.
-        retry_transient_errors: Whether to retry after 500, 502, 503, or
-            504 responses. Defaults to False.
+        retry_transient_errors: Whether to retry after 500, 502, 503, 504
+            or 52x responses. Defaults to False.
     """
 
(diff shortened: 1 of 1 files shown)


max-wittig:
@nejch Mhmm but where is that `retry_transient_errors` is actually removed. I just see that it's set as a self attribute.

nejch:
> @nejch Mhmm but where is that `retry_transient_errors` is actually removed. I just see that it's set as a self attribute.

@max-wittig previously it would just get picked from the method's `**kwargs` via `.get()`:
https://github.com/python-gitlab/python-gitlab/blob/3b0806ed5ce135935d0362068400d41874e3f4a9/gitlab/client.py#L680-L682

but all kwargs are passed as query parameters to the API because they're copied to params earlier:
https://github.com/python-gitlab/python-gitlab/blob/34318871347b9c563d01a13796431c83b3b1d58c/gitlab/client.py#L669

I added it as an explicit argument to avoid that. I could have done`.pop()` from kwargs instead, but it would have to be done before that dict is copied, so I thought it'd be cleaner like this.

PR #1995, opened 2022-04-28
chore(renovate): set schedule to reduce noise

nejch:
We have mostly dev- and typing dependencies getting updated by renovate, and I don't see that much value in getting 3 updates per day for `types-requests` etc. Maybe this will also help people who follow the repo, as they won't be spammed with frequent pushes and PRs.

### DIFF
--- a/.renovaterc.json
+++ b/.renovaterc.json
@@ -1,7 +1,8 @@
 {
   "extends": [
     "config:base",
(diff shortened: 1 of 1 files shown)


JohnVillalovos (approved):

JohnVillalovos:
Sounds good to me. Thanks!

PR #1996, opened 2022-04-29
feat(ux): display project.name_with_namespace on project repr

Psycojoker:
This change the repr from:

```
$ gitlab.projects.get(id=some_id)
<Project id:some_id>
```

To:

```
$ gitlab.projects.get(id=some_id)
<Project id:some_id name_with_namespace:"group_name / project_name">
```

This is especially useful when working on random projects or listing of
projects since users generally don't remember projects ids.

----

On a side note I've encountered this error when trying to do a commit:

```
An unexpected error has occurred: CalledProcessError: command: ('/home/psycojoker/.cache/pre-commit/repox1_osb0o/node_env-default/bin/node', '/home/psycojoker/.cache/pre-commit/repox1_osb0o/node_env-default/bin/npm', 'install', '--dev', '--prod', '--ignore-prepublish', '--no-progress', '--no-save')
return code: 1
expected return code: 0
stdout: (none)
stderr:
    /home/psycojoker/.cache/pre-commit/repox1_osb0o/node_env-default/bin/node: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.28' not found (required by /home/psycojoker/.cache/pre-commit/repox1_osb0o/node_env-default/bin/node)
    
Check the log at /home/psycojoker/.cache/pre-commit/pre-commit.log
```

I have to admit that, while I totally understand the importance of linters and good commit messages, having a hard dependency on a specific version of nodejs and glibc just do to a commit on a full python feels a bit ... intense? :sweat_smile: 

Anyway, thanks a lot for this project :heart:

PR #1998, opened 2022-04-30
chore(ci): replace commitlint with commitizen

nejch:
See https://github.com/python-gitlab/python-gitlab/pull/1996#issue-1221171591

This also needs a small change in required checks temporarily

### DIFF
--- a/requirements-lint.txt
+++ b/requirements-lint.txt
@@ -1,5 +1,6 @@
-argcomplete==2.0.0
+argcomplete<=2.0.0
 black==22.3.0
+commitizen==2.24.0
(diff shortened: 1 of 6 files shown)


requirements-lint.txt
```diff
@@ -1,5 +1,6 @@
-argcomplete==2.0.0
+argcomplete<=2.0.0
```
nejch, line 1: I'll open a PR in commitizen to bump to v2 so we can pin this again later.

JohnVillalovos (approved):
Looks good. Thanks!

PR #1985, opened 2022-04-26
fix(merge_requests): collect payload arguments from custom action for `merge()`

nejch:
Closes #1982

Issue #1982, opened 2022-04-21
Options related to squash are not available when merging the merge request


Prakhar225:
## Description of the problem, including code/CLI snippet
Options related to squash are not available when merging the merge request

## Expected Behavior
All options available in the API should be available as parameters

## Actual Behavior
Only merge_commit_message, should_remove_source_branch and merge_when_pipeline_succeeds are available. The options sha, squash_commit_message and squash are missing.

## Specifications

  - python-gitlab version: 3.3.0
  - API version you are using (v3/v4): v4
  - Gitlab server version (or gitlab.com): gitlab.com

I checked the code and in the `gitlab.v4.objects.ProjectMergeRequest.merge` method, the kwargs are ignored. Only 3 hardcoded options are being used.

Prakhar225:
Please let me know if any additional information is required.

nejch:
Thanks for the report.

To keep this maintainable I guess we should move to always using `kwargs` in combination with `RequiredOptional` tuples even for custom methods, instead of manually parsing them.

PR #1999, opened 2022-05-01
refactor: decouple CLI from custom method arguments

nejch:
A more generic approach for https://github.com/python-gitlab/python-gitlab/pull/1985

This helps towards https://github.com/python-gitlab/python-gitlab/issues/1982. This way, instead of handling parameters in every individual custom method, we can add them in a more declarative way in the decorators and just pull payload in POST/PUT methods from that.

It will also help us completely decouple the CLI from the API client hopefully - making it easier to switch libraries or approaches if we can untangle this.

### DIFF
--- a/gitlab/base.py
+++ b/gitlab/base.py
@@ -19,11 +19,12 @@ import importlib
 import pprint
 import textwrap
 from types import ModuleType
-from typing import Any, Dict, Iterable, Optional, Type, Union
+from typing import Any, Callable, Dict, Iterable, Optional, Type, Union
 
 import gitlab
 from gitlab import types as g_types
 from gitlab.exceptions import GitlabParsingError
+from gitlab.types import F
 
 from .client import Gitlab, GitlabList
 
@@ -326,6 +327,28 @@ class RESTObjectList:
     def total(self) -> Optional[int]:
         """The total number of items."""
         return self._list.total
+
+
+def custom_attrs(
+    required: tuple = (), optional: tuple = (), exclusive: tuple = ()
+) -> Callable[[F], F]:
+    """Decorates a custom method to add a RequiredOptional attribute.
+
+    Args:
+        required: A tuple of API attributes required in the custom method
+        optional: A tuple of API attributes optional in the custom method
+        exclusive: A tuple of mutually exclusive API attributes in the custom method
+    """
+
+    def decorator(func: F) -> F:
+        setattr(
+            func,
+            "_custom_attrs",
+            g_types.RequiredOptional(required, optional, exclusive),
+        )
+        return func
+
+    return decorator
 
 
 class RESTManager:
--- a/gitlab/mixins.py
+++ b/gitlab/mixins.py
@@ -547,9 +547,8 @@ class AccessRequestMixin(_RestObjectBase):
     _updated_attrs: Dict[str, Any]
     manager: base.RESTManager
 
-    @cli.register_custom_action(
-        ("ProjectAccessRequest", "GroupAccessRequest"), (), ("access_level",)
-    )
+    @cli.register_custom_action(("ProjectAccessRequest", "GroupAccessRequest"))
+    @base.custom_attrs(optional=("access_level",))
     @exc.on_http_error(exc.GitlabUpdateError)
     def approve(
         self, access_level: int = gitlab.const.DEVELOPER_ACCESS, **kwargs: Any
@@ -721,7 +720,8 @@ class TimeTrackingMixin(_RestObjectBase):
             assert not isinstance(result, requests.Response)
         return result
 
-    @cli.register_custom_action(("ProjectIssue", "ProjectMergeRequest"), ("duration",))
+    @cli.register_custom_action(("ProjectIssue", "ProjectMergeRequest"))
+    @base.custom_attrs(required=("duration",))
     @exc.on_http_error(exc.GitlabTimeTrackingError)
     def time_estimate(self, duration: str, **kwargs: Any) -> Dict[str, Any]:
         """Set an estimated time of work for the object.
@@ -759,7 +759,8 @@ class TimeTrackingMixin(_RestObjectBase):
             assert not isinstance(result, requests.Response)
         return result
 
-    @cli.register_custom_action(("ProjectIssue", "ProjectMergeRequest"), ("duration",))
+    @cli.register_custom_action(("ProjectIssue", "ProjectMergeRequest"))
+    @base.custom_attrs(required=("duration",))
     @exc.on_http_error(exc.GitlabTimeTrackingError)
     def add_spent_time(self, duration: str, **kwargs: Any) -> Dict[str, Any]:
         """Add time spent working on the object.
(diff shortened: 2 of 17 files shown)


nejch:
Incidentally this makes a few of the decorators a bit more readable.. and should be much more readable when I get to removing the `@cli.register_custom_action` somehow, to completely get rid of the CLI from the API client itself.

gitlab/base.py
```diff
@@ -323,6 +324,21 @@ class RequiredOptional:
     optional: Tuple[str, ...] = ()
 
 
+def custom_attrs(required: tuple = (), optional: tuple = ()) -> Callable[[F], F]:
```
JohnVillalovos, line 327: Any reason not to call it `cli_attributes` or something like that?

gitlab/base.py
```diff
@@ -323,6 +324,21 @@ class RequiredOptional:
...
+    """Decorates a custom method to add a RequiredOptional attribute.
+
+    Args:
+        required: A tuple of API attributes required in the custom method
+        optional: A tuple of API attributes optional in the custom method
+    """
+
+    def decorator(func: F) -> F:
+        setattr(func, "_custom_attrs", RequiredOptional(required, optional))
```
JohnVillalovos, line 336: Can it be called `_required_optional` instead of `_custom_attrs`?

JohnVillalovos (commented):
Sorry for the short review. I haven't yet read the whole PR but wanted to leave a couple comments.

gitlab/base.py
nejch: My idea is to not use them only in the CLI but also in the API client itself, by inspecting these tuples to construct the request payload instead of sending them as query parameters (which will be ignored in POST/PUT as  seen in https://github.com/python-gitlab/python-gitlab/issues/1982).

So I wanted to make them more similar to `_create_attrs` and `_update_attrs` class attributes, which currently have the exact same structure. WDYT?

gitlab/base.py
nejch: Hehe I actually started with exactly that :) But I changed my mind later, see above:

> So I wanted to make them more similar to `_create_attrs` and `_update_attrs` class attributes, which currently have the exact same structure. WDYT?

Hope that makes sense.

gitlab/mixins.py
```diff
@@ -618,9 +618,8 @@ class AccessRequestMixin(_RestObjectBase):
     _updated_attrs: Dict[str, Any]
     manager: base.RESTManager
 
-    @cli.register_custom_action(
-        ("ProjectAccessRequest", "GroupAccessRequest"), (), ("access_level",)
-    )
+    @cli.register_custom_action(("ProjectAccessRequest", "GroupAccessRequest"))
+    @base.custom_attrs(optional=("access_level",))
```
JohnVillalovos, line 622: Have we considered having the wrapper figuring these out automatically? Is it possible using introspection to determine the function signature?

nejch: This might be a lean way to handle it, but we'd lose two things:
* no way to know what args should go into the payload vs. passed on to `http_*` method kwargs. This was the main reason I want to them out from function code into a decorator e.g. for required args.
* we'd lose the args in `--help` for that specific method on the CLI :blush: Not a big deal since I realize I'm claiming here this is being decoupled, but we'd need to be able to pass arbitrary args to the CLI (e.g. https://stackoverflow.com/questions/37367331/is-it-possible-to-use-argparse-to-capture-an-arbitrary-set-of-optional-arguments, though I tried quickly and it'd need more work for us).

nejch: Hmm, I played around with allowing arbitrary CLI args, and it would have to be added to the actual class subparsers which seems a bit convoluted. So I think I need to drop that idea until we use a better CLI library :)

JohnVillalovos: > Hmm, I played around with allowing arbitrary CLI args, and it would have to be added to the actual class subparsers which seems a bit convoluted. So I think I need to drop that idea until we use a better CLI library :)

Sorry, I'm a bit confused. Drop which idea? This whole PR or my idea about a wrapper that figures out the arguments or some other idea?

nejch: The idea to automatically figure out arguments, yes. We still need these currently for the CLI to work, because `argparse` does not allow arbitrary arguments (without hacking too much) - so we have to store them somewhere.

Also I thought it would be useful in the API portion for when we need to distinguish which arguments in the function we should send as data or not, e.g. to replace this kind of custom logic:
https://github.com/python-gitlab/python-gitlab/blob/387a14028b809538530f56f136436c783667d0f1/gitlab/v4/objects/merge_requests.py#L381-L386

gitlab/base.py
nejch: So the idea is that custom methods would also reuse them not just the CLI, the way that `create()` and `update()` currently already reuse `_update_attrs` and `_create_attrs`.

nejch:
@JohnVillalovos this should be ready for another round I think. If the value of this PR isn't really clear I can add a draft PR built on top of it, where I start removing custom argument parsing from some of the custom methods to start demoing that, similar to what I did in https://github.com/python-gitlab/python-gitlab/pull/1985/files.

JohnVillalovos:
Thanks @nejch 

Some quick comments

Can the commit message be improved? It seems to only have a subject line.

Am I correct in thinking (based on looking at https://github.com/python-gitlab/python-gitlab/pull/1985/files) that the `_custom_attrs` will be used to determine which values should be sent in the `POST` body versus in the URL string? If I am correct, would it make more sense to figure out what can not be in the `POST` body and put everything that can be in the `POST` body by default?

gitlab/mixins.py
```diff
@@ -721,7 +720,8 @@ def time_stats(self, **kwargs: Any) -> Dict[str, Any]:
             assert not isinstance(result, requests.Response)
         return result
 
-    @cli.register_custom_action(("ProjectIssue", "ProjectMergeRequest"), ("duration",))
+    @cli.register_custom_action(("ProjectIssue", "ProjectMergeRequest"))
```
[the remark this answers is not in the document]
nejch: Good question.. the thing is that, as I said in the comment above https://github.com/python-gitlab/python-gitlab/pull/1999#issuecomment-1114315465, `@cli.register_custom_action` is the very next thing we'd want to get rid of if we want to fully decouple the CLI from the client. :( So not sure we should invest time in that and then remove it right after. :thinking:

---

My idea for that was, instead of manually assigning custom methods to classes at each method definition, simply collect all non-standard (create,..) public methods for each Manager and RESTObject in the CLI module, kind of how we already collect the classes for the CLI:  
https://github.com/python-gitlab/python-gitlab/blob/main/gitlab/cli.py#L97

Let me check if I can do that in a clean way in a separate commit or follow-up PR, I'll mark this a draft until then.

Maybe it makes sense I open an issue with a mini roadmap to untangling the CLI from the API so we have clear steps on how to get there.

nejch:
Closing this to reduce the number of drafts open - I'll keep the branch and reopen when ready.

PR #2008, opened 2022-05-05
chore: exclude `build/` directory from mypy check

JohnVillalovos:
The `build/` directory is created by the tox environment
`twine-check`. When the `build/` directory exists `mypy` will have an
error.


nejch:
Thanks! much needed, surprised no one else has complained before this :grimacing:

PR #2009, opened 2022-05-05
chore: add `cz` to default tox environment list and skip_missing_interpreters

JohnVillalovos:
Add the `cz` (`comittizen`) check by default.

Set skip_missing_interpreters = True so that when a user runs tox and
doesn't have a specific version of Python it doesn't mark it as an
error.


nejch:
Makes sense like this :) thanks @JohnVillalovos

PR #2010, opened 2022-05-07
feat: display human-readable attribute in `repr()` if present

nejch:
This supersedes https://github.com/python-gitlab/python-gitlab/pull/1996 by reusing existing scaffolding we already have for many other objects.

As an aside this should have the added benefit of further decoupling the CLI because we actually use `_repr_attr` in the API itself which was not yet the case with `_short_print_attr` and it was a bit out of place in `base.py` IMO. Now, the CLi simply reuses it.

Note: @Psycojoker this replaces `name_with_namespace` in your PR with `path` because we already had it and it'd otherwise be a breaking change on the CLI. Let me know if you want to override that, otherwise we'll use `with_namespace` in the next major release which I hope we can have soon with some other features. In any case I think the least confusing option would be `path_with_namespace` because people can `name` the group and project anything they like, so `my-evil-group/my-evil-project` can be named to display as `GitLab.org / GitLab` if using names.

### DIFF
--- a/gitlab/base.py
+++ b/gitlab/base.py
@@ -49,8 +49,12 @@ class RESTObject:
     another. This allows smart updates, if the object allows it.
 
     You can redefine ``_id_attr`` in child classes to specify which attribute
-    must be used as uniq ID. ``None`` means that the object can be updated
+    must be used as the unique ID. ``None`` means that the object can be updated
     without ID in the url.
+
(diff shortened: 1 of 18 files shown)


JohnVillalovos (approved):

JohnVillalovos:
Thanks @nejch

Issue #2002, opened 2022-05-02
"OverflowError: string longer than 2147483647 bytes" when upload file > 2GB to generic package


thcao-ampere:
## Description of the problem, including code/CLI snippet

I used  project.generic_packages.upload  api and got this issue when upload file > 2GB
```
project = gl.projects.get(1, lazy=True)
package = project.generic_packages.upload(
    package_name="hello-world",
    package_version="v1.0.0",
    file_name="hello.tar.gz",
    path="/path/to/local/hello.tar.gz"
)
```

## Specifications

  - python-gitlab version: 2.10.1
  - API version you are using (v3/v4): v4
  - Gitlab server version (or gitlab.com): gitlab.com

nejch:
Thanks for the report @thcao-ampere.

Looks like a repeat of https://github.com/python-gitlab/python-gitlab/issues/646, which was fixed by multipart uploads (https://github.com/python-gitlab/python-gitlab/pull/1252).

We probably forgot to do it for https://github.com/python-gitlab/python-gitlab/blob/main/gitlab/v4/objects/packages.py#L44.

thcao-ampere:
Please help fix in 2.10.1 too.. becausae I'm using python 3.6 on CentOS 7.9

Thanks!

nejch:
@thcao-ampere we generally don't backport fixes as it'd be too difficult to maintain a compatibility matrix with older GitLab version. I suggest after this is fixed you use something like https://github.com/pyenv/pyenv to manage multiple python environments without affecting your system python.

But you can probably patch your old library based on our fix here if that's not possible, as a last resort.

thcao-ampere:
Was this issue fixed ?

nejch:
Not yet @thcao-ampere - but the backstage work is finished (https://github.com/python-gitlab/python-gitlab/pull/2011) so I think I can start looking at this fix now.

mattysan:
Just started using python-gitlab and ran into this issue; really would be helpful if we had a fix.

PR #2011, opened 2022-05-07
refactor(mixins): extract custom type transforms into utils

nejch:
I'm trying to add support for multipart uploads to the generic package upload (see #2002), but unfortunately the logic for that is currently done directly in the `create()`/`update()` methods (processing the `data`/`files` dict for the `MultipartEncoder`).

This extracts that into a helper and handles `transform_files` so we can also use it for `list()` which doesn't need it. This PR preps the code so the generic package one should be smaller.

In the future I think we can simplify multipart encoding more and get rid of `files` altogether, and just process everything in `data` - but wanted to keep this PR simple.

### DIFF
--- a/gitlab/utils.py
+++ b/gitlab/utils.py
@@ -19,9 +19,11 @@ import pathlib
 import traceback
 import urllib.parse
 import warnings
-from typing import Any, Callable, Dict, Optional, Type, Union
+from typing import Any, Callable, Dict, Optional, Tuple, Type, Union
 
 import requests
+
+from gitlab import types
 
 
 class _StdoutStream:
@@ -45,6 +47,39 @@ def response_content(
         if chunk:
             action(chunk)
     return None
+
+
+def _transform_types(
+    data: Dict[str, Any], custom_types: dict, *, transform_files: Optional[bool] = True
+) -> Tuple[dict, dict]:
+    """Copy the data dict with attributes that have custom types and transform them
+    before being sent to the server.
+
+    If ``transform_files`` is ``True`` (default), also populates the ``files`` dict for
+    FileAttribute types with tuples to prepare fields for requests' MultipartEncoder:
+    https://toolbelt.readthedocs.io/en/latest/user.html#multipart-form-data-encoder
+
+    Returns:
+        A tuple of the transformed data dict and files dict"""
+
+    # Duplicate data to avoid messing with what the user sent us
+    data = data.copy()
+    files = {}
+
+    for attr_name, type_cls in custom_types.items():
+        if attr_name not in data:
+            continue
+
+        type_obj = type_cls(data[attr_name])
+
+        # if the type if FileAttribute we need to pass the data as file
+        if transform_files and isinstance(type_obj, types.FileAttribute):
+            key = type_obj.get_file_name(attr_name)
+            files[attr_name] = (key, data.pop(attr_name))
+        else:
+            data[attr_name] = type_obj.get_for_api()
+
+    return data, files
 
 
 def copy_dict(
--- a/tests/unit/test_utils.py
+++ b/tests/unit/test_utils.py
@@ -18,7 +18,7 @@
 import json
 import warnings
 
-from gitlab import utils
+from gitlab import types, utils
 
 
 class TestEncodedId:
@@ -95,3 +95,30 @@ class TestWarningsWrapper:
         assert warn_message in str(warning.message)
         assert __file__ in str(warning.message)
         assert warn_source == warning.source
+
+
+def test_transform_types_copies_data_with_empty_files():
+    data = {"attr": "spam"}
+    new_data, files = utils._transform_types(data, {})
+
+    assert new_data is not data
+    assert new_data == data
+    assert files == {}
+
+
+def test_transform_types_with_transform_files_populates_files():
+    custom_types = {"attr": types.FileAttribute}
+    data = {"attr": "spam"}
+    new_data, files = utils._transform_types(data, custom_types)
+
+    assert new_data == {}
+    assert files["attr"] == ("attr", "spam")
+
+
+def test_transform_types_without_transform_files_populates_data_with_empty_files():
+    custom_types = {"attr": types.FileAttribute}
+    data = {"attr": "spam"}
+    new_data, files = utils._transform_types(data, custom_types, transform_files=False)
+
+    assert new_data == {"attr": "spam"}
+    assert files == {}
--- a/gitlab/mixins.py
+++ b/gitlab/mixins.py
@@ -33,7 +33,6 @@ import requests
 import gitlab
 from gitlab import base, cli
 from gitlab import exceptions as exc
-from gitlab import types as g_types
 from gitlab import utils
 
 __all__ = [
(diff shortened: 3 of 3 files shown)


gitlab/utils.py
```diff
@@ -47,6 +49,39 @@ def response_content(
     return None
 
 
+def _transform_types(
+    data: Dict[str, Any], custom_types: dict, transform_files: Optional[bool] = True
+) -> Tuple[dict, dict]:
```
JohnVillalovos, line 54: What do you think about using a dataclass for the return value?

```
@dataclasses.dataclass
class FileInfoOrSomeBetterName:
   filename: str
   data: bytes

@dataclasses.dataclass
class SomeSuperAwesomeName:
   data: Dict[str, Any]
   files: Dict[str, FileInfoOrSomeBetterName]
```

Not required.

gitlab/utils.py
```diff
@@ -47,6 +49,39 @@ def response_content(
...
+
+    If ``transform_files`` is ``True`` (default), also populates the ``files`` dict for
+    FileAttribute types with tuples to prepare fields for requests' MultipartEncoder:
+    https://toolbelt.readthedocs.io/en/latest/user.html#multipart-form-data-encoder
+
+    Returns:
+        A tuple of the transformed data dict and files dict"""
+
+    # Duplicate data to avoid messing with what the user sent us
+    data = data.copy()
```
JohnVillalovos, line 66: Should this use a `copy.deepcopy()` to be safer?

tests/unit/test_utils.py
```diff
@@ -95,3 +95,28 @@ def test_warn(self):
         assert warn_message in str(warning.message)
         assert __file__ in str(warning.message)
         assert warn_source == warning.source
+
+
+def test_transform_types_copies_data():
+    data = {"attr": "spam"}
+    new_data, _ = utils._transform_types(data, {})
+
+    assert new_data is not data
```
JohnVillalovos, line 104:
```suggestion
    new_data, files = utils._transform_types(data, {})

    assert new_data is not data
    assert new_data == data
    assert files == {}
```

gitlab/utils.py
```diff
@@ -47,6 +49,39 @@ def response_content(
...
+    files = {}
+
+    for attr_name, type_cls in custom_types.items():
+        if attr_name not in data.keys():
```
JohnVillalovos, line 70:
```suggestion
        if attr_name not in data:
```

gitlab/utils.py
```diff
@@ -47,6 +49,39 @@ def response_content(
...
+            continue
+
+        type_obj = type_cls(data[attr_name])
+
+        # if the type if FileAttribute we need to pass the data as file
+        if transform_files and isinstance(type_obj, types.FileAttribute):
+            key = type_obj.get_file_name(attr_name)
+            files[attr_name] = (key, data.pop(attr_name))
```
JohnVillalovos, line 78: Using a `dataclass` (or even a `NamedTuple`) for the value, instead of a Tuple would really improve readability in my opinion. I have an example above called `FileInfoOrSomeBetterName`

What do you think?

gitlab/utils.py
```diff
@@ -47,6 +49,39 @@ def response_content(
     return None
 
 
+def _transform_types(
+    data: Dict[str, Any], custom_types: dict, transform_files: Optional[bool] = True
```
JohnVillalovos, line 53: Forcing keyword arguments would improve readability, in my opinion.

```suggestion
    *, data: Dict[str, Any], custom_types: dict, transform_files: Optional[bool] = True
```

gitlab/utils.py
nejch: This was mostly a copy/paste refactor to extract the duplicate code to a common helper, but if I do more changes I'd like to get rid of `files` completely and handle everything in the payload without this extremely roundabout way we have now (args -> FileAttribute -> files -> data).

I'll give that a shot here, but if it gets messy I'll add the dataclasses as suggested :)

gitlab/utils.py
nejch: Might make sense although with this new helper we'd be calling it on most of the API calls so maybe we should check the implications (https://stackoverflow.com/questions/24756712/deepcopy-is-extremely-slow).

Seems like this code has been in create/update/list for a really long time though with no issues. Maybe I could also refactor to try to avoid running this helper every time unless needed.

JohnVillalovos: No problem keeping it as copy. If it has been working as is then it is probably fine.

I doubt there would be a noticeable performance impact from using `copy.deepcopy()`. Compared to the amount of time the network request takes. I did a simple test of doing a `copy.deepcopy()` 100,000 times and it took 1.15 seconds on my system. Which works out to 0.0000114471 seconds per copy. Now a plain `copy` was much faster at 0.028097620233893394 seconds for 100,000 iterations. Which is 0.0000002810 seconds per iteration.

But compared to how long a network request is then I feel like it will be just noise.

I was copying: `a = {"1": [1, 2, 3, 4, 5], "2": {"hello": "there"}, "3": False}`

nejch: Ah right, we probably will be copying larger dicts than that with projects/groups, but you're right we'll probably also hit rate limits before this starts to make a difference.

JohnVillalovos (approved):
If you want to merge this as-is I am okay with that. We can always make it better in the future.

gitlab/utils.py
nejch: I've added deepcopy as we said IO is probably slower than this. If people complain it's slow we can revert :)

tests/unit/test_utils.py
nejch: Added this but renamed the tests a bit as this changes what is being tested.

gitlab/utils.py
nejch: I forced keyword arguments on the optionals after required args, to have some middle ground between clarity and brevity if that makes sense :)

JohnVillalovos (approved):

gitlab/utils.py
nejch: Hmm I've had to revert the `deepcopy()` for now, because it uses pickle internally, and users might be providing some unpicklable objects in the dict:
https://github.com/python-gitlab/python-gitlab/runs/6643492246?check_suite_focus=true#step:5:282

nejch:
@JohnVillalovos I've made some changes roughly as you suggested but I never got to refactoring the `files` part, I still would like to do that in a follow-up so I haven't touched dataclasses yet as I think it would be only a temporary change :)

gitlab/utils.py
JohnVillalovos: > Hmm I've had to revert the `deepcopy()` for now, because it uses pickle internally, and users might be providing some unpicklable objects in the dict: https://github.com/python-gitlab/python-gitlab/runs/6643492246?check_suite_focus=true#step:5:282

Makes sense. Thanks.

JohnVillalovos:
> @JohnVillalovos I've made some changes roughly as you suggested but I never got to refactoring the `files` part, I still would like to do that in a follow-up so I haven't touched dataclasses yet as I think it would be only a temporary change :)

Sounds good. Approved. Not sure if you want to squash it into one commit or not. Or merge it as-is.

nejch:
> > @JohnVillalovos I've made some changes roughly as you suggested but I never got to refactoring the `files` part, I still would like to do that in a follow-up so I haven't touched dataclasses yet as I think it would be only a temporary change :)
> 
> Sounds good. Approved. Not sure if you want to squash it into one commit or not. Or merge it as-is.

Ah right makes sense, I've squashed it :) just didn't feel comfortable self-merging for this project.

JohnVillalovos (approved):

PR #2012, opened 2022-05-08
test: increase projects coverage

rnoberger:


tests/unit/objects/test_projects.py
```diff
@@ -28,6 +52,106 @@ def resp_get_project():
...
+@pytest.fixture
+def resp_user_projects():
+    with responses.RequestsMock() as rsps:
+        rsps.add(
+            method=responses.GET,
+            url="http://localhost/api/v4/users/1",
+            json=user_content,
+            content_type="application/json",
+            status=200,
+        )
```
nejch, line 64: We can make use of the `lazy=True` argument to create a lazy object (for both user and project) without the first GET call, so we only need to mock the second call.

See docs at https://python-gitlab.readthedocs.io/en/stable/api-usage.html#lazy-objects In fact, we already have fixtures for a lot of these in unit tests for this reason:
https://github.com/python-gitlab/python-gitlab/blob/main/tests/unit/conftest.py#L62-L89 (just check that the `id` in the lazy fixture matches your next calls, otherwise you might need to tweak your reponses, but I think it's ok)

So let's pass these to the tests and we can remove the extra mocked reponses:

```suggestion
```

I realize some of the existing tests do not do that, maybe you were trying to be consistent here. So I probably need to look at cleaning up existing code as well ;)

tests/unit/objects/test_projects.py
```diff
@@ -28,6 +52,106 @@ def resp_get_project():
...
+@pytest.fixture
+def resp_starred_projects():
+    with responses.RequestsMock() as rsps:
+        rsps.add(
+            method=responses.GET,
+            url="http://localhost/api/v4/users/1",
+            json=user_content,
+            content_type="application/json",
+            status=200,
+        )
```
nejch, line 84: Same as above, we'll use the lazy user fixture:
```suggestion
```

tests/unit/objects/test_projects.py
```diff
@@ -28,6 +52,106 @@ def resp_get_project():
...
+@pytest.fixture
+def resp_list_users():
+    with responses.RequestsMock() as rsps:
+        rsps.add(
+            method=responses.GET,
+            url="http://localhost/api/v4/projects/1",
+            json=project_content,
+            content_type="application/json",
+            status=200,
+        )
```
nejch, line 104: Same as above, we'll use the lazy project fixture:
```suggestion
```

tests/unit/objects/test_projects.py
```diff
@@ -28,6 +52,106 @@ def resp_get_project():
...
+@pytest.fixture
+def resp_list_forks():
+    with responses.RequestsMock() as rsps:
+        rsps.add(
+            method=responses.GET,
+            url="http://localhost/api/v4/projects/1",
+            json=project_content,
+            content_type="application/json",
+            status=200,
+        )
```
nejch, line 124: Same as above, we'll use the lazy project fixture:
```suggestion
```

tests/unit/objects/test_projects.py
```diff
@@ -28,6 +52,106 @@ def resp_get_project():
...
+@pytest.fixture
+def resp_list_languages():
+    with responses.RequestsMock() as rsps:
+        rsps.add(
+            method=responses.GET,
+            url="http://localhost/api/v4/projects/1",
+            json=project_content,
+            content_type="application/json",
+            status=200,
+        )
```
nejch, line 144: Same as above, we'll use the lazy project fixture:
```suggestion
```

tests/unit/objects/test_projects.py
```diff
@@ -98,19 +222,26 @@ def test_import_bitbucket_server(gl, resp_import_bitbucket_server):
     assert res["import_status"] == "scheduled"
 
 
-@pytest.mark.skip(reason="missing test")
-def test_list_user_projects(gl):
-    pass
+def test_list_user_projects(gl, resp_user_projects):
+    user_project = gl.users.get(1).projects.list()[0]
```
nejch, line 226: We use the `user` fixture from `conftest.py` that I linked above by passing it in the test. Then we just call additional on the already lazily created user object. We can also remove the `gl` fixture from the test because we already used in the `user` fixture itself:

```suggestion
def test_list_user_projects(user, resp_user_projects):
    user_project = user.projects.list()[0]
```

tests/unit/objects/test_projects.py
```diff
@@ -98,19 +222,26 @@ def test_import_bitbucket_server(gl, resp_import_bitbucket_server):
...
+    assert isinstance(user_project, UserProject)
+    assert user_project.name == "name"
+    assert user_project.id == 1
 
 
-@pytest.mark.skip(reason="missing test")
-def test_list_user_starred_projects(gl):
-    pass
+def test_list_user_starred_projects(gl, resp_starred_projects):
+    starred_projects = gl.users.get(1).starred_projects.list()[0]
```
nejch, line 233: Same as above:

```suggestion
def test_list_user_starred_projects(user, resp_starred_projects):
    starred_projects = user.starred_projects.list()[0]
```

tests/unit/objects/test_projects.py
```diff
@@ -98,19 +222,26 @@ def test_import_bitbucket_server(gl, resp_import_bitbucket_server):
...
+    assert isinstance(starred_projects, StarredProject)
+    assert starred_projects.name == "name"
+    assert starred_projects.id == 1
 
 
-@pytest.mark.skip(reason="missing test")
-def test_list_project_users(gl):
-    pass
+def test_list_project_users(gl, resp_list_users):
+    user = gl.projects.get(1).users.list()[0]
```
nejch, line 240: Same as above, but for `project` fixture:

```suggestion
def test_list_project_users(project, resp_list_users):
    user = project.users.list()[0]
```

tests/unit/objects/test_projects.py
```diff
@@ -133,9 +264,11 @@ def test_fork_project(gl):
     pass
 
 
-@pytest.mark.skip(reason="missing test")
-def test_list_project_forks(gl):
-    pass
+def test_list_project_forks(gl, resp_list_forks):
+    forks = gl.projects.get(1).forks.list()
```
nejch, line 268:
```suggestion
def test_list_project_forks(project, resp_list_forks):
    forks = project.forks.list()
```

tests/unit/objects/test_projects.py
```diff
@@ -153,9 +286,13 @@ def test_list_project_starrers(gl):
     pass
 
 
-@pytest.mark.skip(reason="missing test")
-def test_get_project_languages(gl):
-    pass
+def test_get_project_languages(gl, resp_list_languages):
+    python = gl.projects.get(1).languages().get("python")
+    ruby = gl.projects.get(1).languages().get("ruby")
+    coffee_script = gl.projects.get(1).languages().get("CoffeeScript")
```
nejch, line 292:
```suggestion
def test_get_project_languages(project, resp_list_languages):
    python = project.languages().get("python")
    ruby = project.languages().get("ruby")
    coffee_script = project.languages().get("CoffeeScript")
```

nejch (changes requested):
Thanks a lot for the PR @rnoberger. I've added the comments on how to reduce the fixture boilerplate as discussed, let me know if anything's unclear.

If applying the suggestions in the UI just make sure to change the commit message on GitHub or amend it later to match our convention :) Thanks!

tests/unit/objects/test_projects.py
```diff
@@ -5,9 +5,33 @@
...
+    "name": "first",
+    "id": 1,
+    "state": "active",
+}
+forks_content = [
+    {
+        "id": 1,
+    },
+    {
+        "id": "python-gitlab%2Fpython-gitlab",
```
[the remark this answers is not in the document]
rnoberger: I had a look at the documentation and the description of the id is

> The ID or [URL-encoded path of the project](https://docs.gitlab.com/ee/api/index.html#namespaced-path-encoding).

So I thought it might be good to catch errors if the URL-encoded path of the project makes some problems anywhere. But if the Server don't response with this type of information, maybe its better to delete this part. Or I make a comment on this. I will look at it at the start of the week and I will handle it how my feelings will guide me ;)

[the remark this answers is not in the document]
nejch: Yes, as you already guessed, this part of the documentation is for making requests (`If using namespaced API requests...`) rather than receiving responses. The server should consistently respond with an `id` of int type, I think this would be a breaking change on GitLab's side otherwise.

I believe we already test the request behavior in other places because this was an important part of making the CLI work with GET requests.

No rush at all, thanks again for the PR!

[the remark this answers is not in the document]
rnoberger: I implemented it now. There is just one thing I wonder about. Is it just the intelligence of my IDE or doesn't I have to write those small fixtures for `user` and `project`? I added them for clarification but the tests run well also without explicit define them.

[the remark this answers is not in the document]
nejch: Looks good now @rnoberger thanks again!

Not sure I fully understand your last question. pytest does a bit of magic to automatically discover fixtures by going up the directory tree looking at each `conftest.py`. This is why IDEs usually don't know what to do with pytest fixtures in test function signatures.

In our case, the `user` and `project` fixtures are already written in https://github.com/python-gitlab/python-gitlab/blob/main/tests/unit/conftest.py#L67-L89 and pytest will automatically use them in tests. Some IDEs might detect that and some won't, depending on plugins installed I assume. :)

nejch (approved):
Thank you for working on this @rnoberger, let us know in case you want to continue working on other parts and need any help :)

PR #1957, opened 2022-03-31
Fix `--job` argument interpreted as `--job-token`

meteozond:
## Description of the problem, including code/CLI snippet

`--job` argument interpreted as `--job-token` arugment and conflicts with `  --private-token`

Executing:

```
gitlab \
  --private-token ${GITLAB_API_KEY} -o json \
  project-artifact download \
  --project-id mdm/entities \
  --job ${JOB} \
  --ref-name ${REF} \
```

## Expected Behavior

Artifact downloaded

## Actual Behavior

`gitlab: error: argument --job-token: not allowed with argument --private-token`

## Specifications

  - python-gitlab version: 3.2.0
  - API version you are using (v3/v4): doesn't meter
  - Gitlab server version (or gitlab.com): doesn't meter

## PS

Looks like it happens because of default `argparse.ArgumentParser`s `allow_abbrev` behaviour - both `--job` and `--job-token` starts with `--job` chars. 

Could be fixed [here](https://github.com/python-gitlab/python-gitlab/blob/8e241e483bb8b316a831f5831e6db733dd04a08f/gitlab/cli.py#L108) with:

```diff
    parser = argparse.ArgumentParser(
        add_help=add_help, description="GitLab API Command Line Interface",
+       allow_abbrev=False
    )
```

PR #2013, opened 2022-05-08
fix(cli): changed default `allow_abbrev` value to fix arguments collision problem

JohnVillalovos:
The `--job` argument could be interpreted as `--job-token`

    Example failure:

      $ gitlab \
        --private-token ${GITLAB_API_KEY} -o json \
        project-artifact download \
        --project-id mdm/entities \
        --job ${JOB} \
        --ref-name ${REF} \

      gitlab: error: argument --job-token: not allowed with argument
      --private-token

### DIFF
--- a/gitlab/cli.py
+++ b/gitlab/cli.py
@@ -106,7 +106,9 @@ def cls_to_what(cls: RESTObject) -> str:
 
 def _get_base_parser(add_help: bool = True) -> argparse.ArgumentParser:
     parser = argparse.ArgumentParser(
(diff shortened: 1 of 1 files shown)


JohnVillalovos:
@nejch This is the same as: https://github.com/python-gitlab/python-gitlab/pull/1957 but ran `black` and updated the commit message.

nejch:
Thanks again @meteozond for the fix and @JohnVillalovos for the green CI :)

nejch (approved):

PR #2014, opened 2022-05-08
fix: duplicate subparsers being added to argparse

JohnVillalovos:
Python 3.11 added an additional check in the argparse libary which
detected duplicate subparsers being added. We had duplicate subparsers
being added.

Make sure we don't add duplicate subparsers.
Closes: #2015

### DIFF
--- a/gitlab/v4/cli.py
+++ b/gitlab/v4/cli.py
@@ -200,11 +200,15 @@ def _populate_sub_parser_by_class(
     mgr_cls_name = f"{cls.__name__}Manager"
     mgr_cls = getattr(gitlab.v4.objects, mgr_cls_name)
 
+    action_parsers: Dict[str, argparse.ArgumentParser] = {}
     for action_name in ["list", "get", "create", "update", "delete"]:
         if not hasattr(mgr_cls, action_name):
             continue
 
-        sub_parser_action = sub_parser.add_parser(action_name)
+        sub_parser_action = sub_parser.add_parser(
+            action_name, conflict_handler="resolve"
+        )
+        action_parsers[action_name] = sub_parser_action
         sub_parser_action.add_argument("--sudo", required=False)
         if mgr_cls._from_parent_attrs:
             for x in mgr_cls._from_parent_attrs:
@@ -268,7 +272,11 @@ def _populate_sub_parser_by_class(
     if cls.__name__ in cli.custom_actions:
         name = cls.__name__
         for action_name in cli.custom_actions[name]:
-            sub_parser_action = sub_parser.add_parser(action_name)
+            # NOTE(jlvillal): If we put a function for the `default` value of
+            # the `get` it will always get called, which will break things.
+            sub_parser_action = action_parsers.get(action_name)
+            if sub_parser_action is None:
+                sub_parser_action = sub_parser.add_parser(action_name)
             # Get the attributes for URL/path construction
             if mgr_cls._from_parent_attrs:
                 for x in mgr_cls._from_parent_attrs:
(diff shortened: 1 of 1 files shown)


JohnVillalovos:
@nejch Seems like Python 3.11.0-beta.1 doesn't like what we are doing...

JohnVillalovos:
I think this is the upstream commit:

https://github.com/python/cpython/commit/ad5e8520f3e117f45481513014548a39879d30d2

```
commit ad5e8520f3e117f45481513014548a39879d30d2
Author: Antony Lee <anntzer.lee@gmail.com>
Date:   Sun May 1 08:04:50 2022 +0200

    bpo-39716: Raise on conflicting subparser names. (GH-18605)



    Raise an ArgumentError when the same subparser name is added twice to an
    ArgumentParser.  This is consistent with the (default) behavior when the
    same option string is added twice to an ArgumentParser.

    (Support for `conflict_handler="resolve"` could be considered as a
    followup feature, although real use cases seem even rarer than
    "resolve"ing option-strings.)

    Automerge-Triggered-By: GH:rhettinger
```

JohnVillalovos:
The PR upstream: https://github.com/python/cpython/pull/18605

JohnVillalovos:
Probably should create an issue for this instead of my dummy PR 😊

gitlab/v4/cli.py
```diff
@@ -268,6 +270,9 @@ def _populate_sub_parser_by_class(
     if cls.__name__ in cli.custom_actions:
         name = cls.__name__
         for action_name in cli.custom_actions[name]:
+            if action_name in action_parsers:
+                continue
```
nejch, line 274: As we discussed on Signal this seems to have a side effect and ignores overrides https://pastebin.com/sAGVNCxm

Example: https://github.com/python-gitlab/python-gitlab/blob/main/gitlab/v4/objects/files.py#L98-L101

I wonder if if having a `conflict_handler="resolve"` in our ArgumentParser would help here:
https://docs.python.org/3/library/argparse.html#conflict-handler

> Sometimes (e.g. when using [parents](https://docs.python.org/3/library/argparse.html#parents)) it may be useful to simply override any older arguments with the same option string. To get this behavior, the value 'resolve' can be supplied to the conflict_handler= argument of [ArgumentParser](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser):

JohnVillalovos: Thanks for the tip on `conflict_handler` that seems to solve the issue.

gitlab/v4/cli.py
```diff
@@ -298,13 +304,17 @@ def _populate_sub_parser_by_class(
...
+                sub_parser_action = sub_parser.add_parser(action_name)
             if mgr_cls._from_parent_attrs:
                 for x in mgr_cls._from_parent_attrs:
                     sub_parser_action.add_argument(
                         f"--{x.replace('_', '-')}", required=True
                     )
-                sub_parser_action.add_argument("--sudo", required=False)
+                sub_parser_action.sub_parser_action.add_argument(
+                    "--sudo", required=False
+                )
```
nejch, line 317: Is this `sub_parser_action.sub_parser_action` intentional here?

gitlab/v4/cli.py
```diff
@@ -298,13 +304,17 @@ def _populate_sub_parser_by_class(
     if mgr_cls.__name__ in cli.custom_actions:
         name = mgr_cls.__name__
         for action_name in cli.custom_actions[name]:
-            sub_parser_action = sub_parser.add_parser(action_name)
+            sub_parser_action = action_parsers.get(action_name)
+            if sub_parser_action is None:
+                sub_parser_action = sub_parser.add_parser(action_name)
```
nejch, line 309: I'm thinking the `get()` can already handle `None` values on its own and might be more efficient, though not sure how black will behave here:

```suggestion
            sub_parser_action = action_parsers.get(action_name, default=sub_parser.add_parser(action_name))
```
(Same with the other one)

gitlab/v4/cli.py
JohnVillalovos: That was a mistake! Thanks for catching it.

gitlab/v4/cli.py
JohnVillalovos: I thought that same thing. But unfortunately the function in the "default" value always gets called even if the value is not used. This breaks things 😢

```
$ cat test.py
#!/usr/bin/python3 -ttu
def hello():
    print("hello")
a = {"1": "test"}
print(a.get("1", hello()))
```

Run it

```
$ python test.py
hello
test
```

Issue #2015, opened 2022-05-08
Python 3.11 unit tests are failing due to argparse error


JohnVillalovos:
Python 3.11 unit tests are failing due to an argparse error.

Appears to be because of this PR in cpython: https://github.com/python/cpython/pull/18605

Example of the error is:
```
______________________________ test_v4_parse_args ______________________________

    def test_v4_parse_args():
>       parser = cli._get_parser()

tests/unit/test_cli.py:126: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
gitlab/cli.py:254: in _get_parser
    return gitlab.v4.cli.extend_parser(parser)
gitlab/v4/cli.py:340: in extend_parser
    object_group = subparsers.add_parser(arg_name)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = _SubParsersAction(option_strings=[], dest='what', nargs='A...', const=None, default=None, type=None, choices={'applica....HelpFormatter'>, conflict_handler='error', add_help=True)}, required=True, help='Object to manipulate.', metavar=None)
name = 'project-audit-event', kwargs = {'prog': 'pytest project-audit-event'}
aliases = ()

    def add_parser(self, name, **kwargs):
        # set prog from the existing prefix
        if kwargs.get('prog') is None:
            kwargs['prog'] = '%s %s' % (self._prog_prefix, name)
    
        aliases = kwargs.pop('aliases', ())
    
        if name in self._name_parser_map:
>           raise ArgumentError(self, _('conflicting subparser: %s') % name)
E           argparse.ArgumentError: argument what: conflicting subparser: project-audit-event

/opt/hostedtoolcache/Python/3.11.0-beta.1/x64/lib/python3.11/argparse.py:1185: ArgumentError
```

PR #2016, opened 2022-05-08
chore: allow Python 3.11 tests to fail

JohnVillalovos:
This is due to https://github.com/python-gitlab/python-gitlab/issues/2015

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,23 +27,30 @@ jobs:
         python:
           - version: "3.7"
             toxenv: py37
+            experimental: false
           - version: "3.8"
             toxenv: py38
+            experimental: false
           - version: "3.9"
             toxenv: py39
+            experimental: false
           - version: "3.10"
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -27,23 +27,30 @@ jobs:
         python:
           - version: "3.7"
             toxenv: py37
+            experimental: false
```
nejch, line 30: Would GitHub Actions work if we just skip defining these for falsy values to avoid them in each single job? Now they show up like this:

image

Or alternatively define it at top-level only once via `matrix.experimental` and then only add an include where this is positive? Like they do here https://michaelheap.com/testing-experimental-versions-github-actions/#testing-multiple-versions?

I haven't tried it myself so not sure would need to play with this a bit.

nejch (commented):
Thanks for the catch @JohnVillalovos just one small question :)

JohnVillalovos:
Have a fix coming in #2014

PR #2019, opened 2022-05-10
Some improvements to our tox environment running

JohnVillalovos:
* Rename `pep8` environment to `flake8`
  * Run the `pylint` test by default


nejch (approved):
Ah thanks, totally missed we haven't been running that one env by default:)

PR #2025, opened 2022-05-23
doc: API Doc Updates

MichaelSweikata:
Updates the API docs to provide examples

Clarifies https://github.com/python-gitlab/python-gitlab/discussions/2021


nejch:
@MichaelSweikata thanks for the rework, the commit convention is quite specific so this failed again. The original PR already has some review by @JohnVillalovos so I think it's useful to keep history there, I've force-pushed a squashed commit to your branch and let's continue there anyway ;)

PR #2032, opened 2022-05-29
feat(client): introduce `iterator=True` and deprecate `as_list=False` in `list()`

JohnVillalovos:
`as_list=False` is confusing as it doesn't explain what is being
returned. Replace it with `iterator=True` which more clearly explains
to the user that an iterator/generator will be returned.

This maintains backward compatibility with `as_list` but does issue a
DeprecationWarning if `as_list` is set.

### DIFF
--- a/docs/api-usage.rst
+++ b/docs/api-usage.rst
@@ -93,13 +93,13 @@ Examples:
 .. code-block:: python
 
    # list all the projects
-   projects = gl.projects.list(as_list=False)
+   projects = gl.projects.list(iterator=True)
    for project in projects:
        print(project)
 
    # get the group with id == 2
    group = gl.groups.get(2)
-   for project in group.projects.list(as_list=False):
+   for project in group.projects.list(iterator=True):
        print(project)
 
    # create a new user
@@ -109,7 +109,7 @@ Examples:
 
 .. warning::
    Calling ``list()`` without any arguments will by default not return the complete list
-   of items. Use either the ``all=True`` or ``as_list=False`` parameters to get all the
+   of items. Use either the ``all=True`` or ``iterator=True`` parameters to get all the
    items when using listing methods. See the :ref:`pagination` section for more
    information.
 
@@ -156,9 +156,9 @@ conflict with python or python-gitlab when using them as kwargs:
 
 .. code-block:: python
 
-   gl.user_activities.list(from='2019-01-01', as_list=False)  ## invalid
+   gl.user_activities.list(from='2019-01-01', iterator=True)  ## invalid
 
-   gl.user_activities.list(query_parameters={'from': '2019-01-01'}, as_list=False)  # OK
+   gl.user_activities.list(query_parameters={'from': '2019-01-01'}, iterator=True)  # OK
 
 Gitlab Objects
 ==============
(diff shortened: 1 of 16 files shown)


gitlab/client.py
```diff
@@ -832,15 +835,33 @@ def http_list(
...
         query_data = query_data or {}
 
-        # In case we want to change the default behavior at some point
-        as_list = True if as_list is None else as_list
+        # Don't allow both `as_list` and `iterator` to be set.
+        if as_list is not None and iterator is not None:
+            raise gitlab.exceptions.GitlabError(
+                error_message=(
+                    "`as_list` and `iterator` are mutually exclusive. Only one "
+                    "argument can be set. Recommend to only set `iterator` argument"
```
nejch, line 843: No strong opinion on this, but if we stay with the messaging in the comment above and our other deprecation warnings, it would be something like:
```suggestion
                    "Only one of `as_list` and `iterator` should be defined."
                    "Use `iterator` instead of `as_list`"
```

gitlab/client.py
```diff
@@ -832,15 +835,33 @@ def http_list(
...
+                )
+            )
+
+        if as_list is not None:
+            iterator = not as_list
+            utils.warn(
+                message=(
+                    f"`as_list={as_list}` has been deprecated. Please use "
+                    f"`iterator={iterator}` instead"
```
nejch, line 852: Similar here, we have this:
https://github.com/python-gitlab/python-gitlab/blob/8867ee59884ae81d6457ad6e561a0573017cf6b2/gitlab/v4/objects/projects.py#L555-L573

```suggestion
                    f"`as_list={as_list}` is deprecated and will be removed in a future version.
                    f"Use `iterator={iterator}` instead."
```

gitlab/client.py
```diff
@@ -832,15 +835,33 @@ def http_list(
...
+                ),
+                category=DeprecationWarning,
+            )
+        if iterator is None:
+            iterator = False
```
nejch, line 857: IMO in this case this is redundant - we have `if iterator:` a few lines down from here so I think it's quite idiomatic without it as there isn't any user code messing with this in between.

```suggestion
```

nejch (changes requested):
Looks good @JohnVillalovos just a few minor comments.

Context for future readers:
https://github.com/python-gitlab/python-gitlab/pull/1956/files#r884302316

gitlab/client.py
```diff
@@ -832,15 +835,33 @@ def http_list(
         """
         query_data = query_data or {}
 
-        # In case we want to change the default behavior at some point
-        as_list = True if as_list is None else as_list
+        # Don't allow both `as_list` and `iterator` to be set.
+        if as_list is not None and iterator is not None:
+            raise gitlab.exceptions.GitlabError(
```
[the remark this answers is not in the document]
JohnVillalovos: Done.

gitlab/client.py
JohnVillalovos: Done

gitlab/client.py
JohnVillalovos: Done

docs/api-usage.rst
```diff
@@ -310,6 +310,10 @@ The generator exposes extra listing information as received from the server:
    For more information see:
    https://docs.gitlab.com/ee/user/gitlab_com/index.html#pagination-response-headers
 
+.. note::
+   Prior to python-gitlab 3.6.0 the argument ``as_list`` was used instead of
```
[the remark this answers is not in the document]
JohnVillalovos: Done. Made it `feat(client): ...`

gitlab/client.py
```diff
@@ -807,7 +807,9 @@ def http_list(
         self,
         path: str,
         query_data: Optional[Dict[str, Any]] = None,
-        as_list: Optional[bool] = None,
+        *,
+        as_list: Optional[bool] = None,  # Deprecated in favor of `iterator`
+        iterator: Optional[bool] = None,
```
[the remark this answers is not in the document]
JohnVillalovos: I doubt that is happening.

gitlab/client.py
JohnVillalovos: Done.

nejch (approved):
Thanks @JohnVillalovos, makes sense to me, also makes it consistent with the iterator arg on the upcoming streaming responses PR then.

PR #2036, opened 2022-05-30
chore: update type-hints return signature for GetWithoutIdMixin methods

JohnVillalovos:
Commit f0152dc3cc9a42aa4dc3c0014b4c29381e9b39d6 removed situation
where `get()` in a `GetWithoutIdMixin` based class could return `None`

Update the type-hints to no longer return `Optional` AKA `None`

### DIFF
--- a/tests/meta/test_ensure_type_hints.py
+++ b/tests/meta/test_ensure_type_hints.py
@@ -72,8 +72,8 @@ from typing import Any, cast, Union"
 GET_WITHOUT_ID_METHOD_TEMPLATE = """
 def get(
     self, id: Optional[Union[int, str]] = None, **kwargs: Any
-) -> Optional[{obj_cls.__name__}]:
-    return cast(Optional[{obj_cls.__name__}], super().get(id=id, **kwargs))
+) -> {obj_cls.__name__}:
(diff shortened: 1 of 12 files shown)


tests/meta/test_ensure_type_hints.py
```diff
@@ -102,7 +102,7 @@ def test_check_get_without_id_function_type_hints(
             base_type=gitlab.mixins.GetWithoutIdMixin,
             class_info=class_info,
             method_template=GET_WITHOUT_ID_METHOD_TEMPLATE,
-            optional_return=True,
+            optional_return=False,
```
nejch, line 105: question: do you think this argument is still needed? I don't think we'll be returning `None` from `get()` at any point from now on.

nejch (commented):
Thanks @JohnVillalovos lgtm, just a quick question. There's another quirk with GetWithoutIdMixin I'll open a follow-up for.

tests/meta/test_ensure_type_hints.py
nejch: not too important :)

nejch (approved):

PR #2040, opened 2022-05-31
chore: have `EncodedId` creation always return `EncodedId`

JohnVillalovos:
There is no reason to return an `int` as we can always return a `str`
version of the `int`

Change `EncodedId` to always return an `EncodedId`. This removes the
need to have `mypy` ignore the error raised.

### DIFF
--- a/gitlab/utils.py
+++ b/gitlab/utils.py
@@ -113,16 +113,14 @@ class EncodedId(str):
         https://docs.gitlab.com/ee/api/index.html#path-parameters
     """
 
-    # mypy complains if return type other than the class type. So we ignore issue.
-    def __new__(  # type: ignore
-        cls, value: Union[str, int, "EncodedId"]
-    ) -> Union[int, "EncodedId"]:
-        if isinstance(value, (int, EncodedId)):
+    def __new__(cls, value: Union[str, int, "EncodedId"]) -> "EncodedId":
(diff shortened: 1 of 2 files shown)


gitlab/utils.py
```diff
@@ -113,17 +113,17 @@ class EncodedId(str):
...
-        if not isinstance(value, str):
+        if isinstance(value, int):
+            return_value = str(value)
+        elif isinstance(value, str):
+            return_value = urllib.parse.quote(value, safe="")
+        else:
             raise TypeError(f"Unsupported type received: {type(value)}")
-        value = urllib.parse.quote(value, safe="")
-        return super().__new__(cls, value)
+        return super().__new__(cls, return_value)
```
nejch, line 126: Would you be against just redefining `value` here? Then all returns are the same, seems less visual load for me :)

nejch (commented):
LGTM just one question :)

gitlab/utils.py
JohnVillalovos: Mypy will complain if that is done. As redefining the type of the int

JohnVillalovos: While driving around I realized that the `str` `__new__()` method will happily take an `int` value. So it simplified the code.

nejch: Nice!

nejch (approved):

PR #2042, opened 2022-06-01
Clean-up the `validate_attrs` method/function

JohnVillalovos:
See the commit messages for more details.

- Use an f-string instead of % formatting
- Move `validate_attrs` to be part of the `RequiredOptional` class.

### DIFF
--- a/gitlab/types.py
+++ b/gitlab/types.py
@@ -16,7 +16,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import dataclasses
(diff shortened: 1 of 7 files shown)


gitlab/types.py
```diff
@@ -25,6 +25,33 @@ class RequiredOptional:
     optional: Tuple[str, ...] = ()
     exclusive: Tuple[str, ...] = ()
 
+    def validate_attrs(
+        self,
+        *,
+        data: Dict[str, Any],
```
nejch, line 31: :laughing:

PR #2049, opened 2022-06-03
test: move back to using latest Python 3.11 version

JohnVillalovos:


nejch (approved):
I'd prefer explicitly auto-bumping each pre-release to avoid more of these mystery failures, but works for me :) I'll try to find a way.

PR #2050, opened 2022-06-03
chore: enable 'consider-using-sys-exit' pylint check

JohnVillalovos:
Enable the 'consider-using-sys-exit' pylint check and fix errors
raised.


nejch (approved):
Thanks, not sure why we weren't doing this already :grin:

PR #2051, opened 2022-06-03
chore: enable more pylint checks

JohnVillalovos:


JohnVillalovos:
@nejch Let me know if you would prefer this split into multiple PRs.

nejch:
> @nejch Let me know if you would prefer this split into multiple PRs.

It's ok, LGTM! pylint complaining about `attribute-defined-outside-init` for rest objects IMO is a bit dubious in our case, as we already abuse instance attributes for API attributes. That will always be a moving target, but let's see :)

nejch (approved):

PR #2052, opened 2022-06-04
chore: rename `__call__()` to `run()` in GitlabCLI

JohnVillalovos:
Less confusing to have it be a normal method.


nejch (approved):
Works for me :)

PR #2053, opened 2022-06-04
chore: rename `what` to `gitlab_resource`

JohnVillalovos:
Naming a variable `what` makes it difficult to understand what it is
used for.

Rename it to `gitlab_resource` as that is what is being stored.

The Gitlab documentation talks about them being resources:
https://docs.gitlab.com/ee/api/api_resources.html

This will improve code readability.


nejch (commented):
I guess `whaction` should also be called `resource_action` then to match, if it doesn't get too verbose. But we can do that later

Issue #1604, opened 2021-09-17
Add ban/unban and approve/reject methods to User


sathieu:
## Description of the problem, including code/CLI snippet

API has endpoints to manage the state of a user:

- activate/deactivate: :heavy_check_mark: already done
- block/unblock: :heavy_check_mark: already done
- ban/unban: :construction: TODO
- approve/reject: :construction: TODO

nejch:
@antoineauger as discussed if you want a break from k8s and work on this :grin: 

There's a PR that implements approve/reject endpoints for this issue in https://github.com/python-gitlab/python-gitlab/pull/2061 - you could do the same for ban/unban:
https://docs.gitlab.com/ee/api/users.html#ban-user
https://docs.gitlab.com/ee/api/users.html#unban-user

See my comments in the other PR for a bit of guidance, although ideally we'd also have 1 response per mock and 1 assert per test so we can remove `assert_all_requests_are_fired=False`.

In the case of ban/unban I think we could also add a quick functional test like for block/unblock. https://github.com/python-gitlab/python-gitlab/blob/main/tests/functional/api/test_users.py#L30-L37.

With both PRs in we could then close this issue. Of course if there's any hiccups along the way I'm happy to look at docs PRs as well if you feel like it :grinning:

PR #2061, opened 2022-06-09
feat(users): add approve and reject methods to User

bgamari:
Closes https://github.com/python-gitlab/python-gitlab/issues/1604.

Follow-up to https://github.com/python-gitlab/python-gitlab/pull/2064.

### DIFF
--- a/docs/gl_objects/users.rst
+++ b/docs/gl_objects/users.rst
@@ -111,6 +111,13 @@ List a user's starred projects
 
     user.starred_projects.list()
 
+If the GitLab instance has new user account approval enabled some users may
+have ``user.state == 'blocked_pending_approval'``. Administrators can approve
+and reject such users::
+
+    user.approve()
+    user.reject()
+
 User custom attributes
 ======================
 
--- a/gitlab/v4/objects/users.py
+++ b/gitlab/v4/objects/users.py
@@ -284,6 +284,42 @@ class User(SaveMixin, ObjectDeleteMixin, RESTObject):
         return server_data
 
     @cli.register_custom_action("User")
+    @exc.on_http_error(exc.GitlabUserApproveError)
+    def approve(self, **kwargs: Any) -> Union[Dict[str, Any], requests.Response]:
+        """Approve a user creation request.
+
+        Args:
+            **kwargs: Extra options to send to the server (e.g. sudo)
+
+        Raises:
+            GitlabAuthenticationError: If authentication is not correct
+            GitlabUserApproveError: If the user could not be activated
+
+        Returns:
+            The new object data (*not* a RESTObject)
+        """
+        path = f"/users/{self.encoded_id}/approve"
+        return self.manager.gitlab.http_post(path, **kwargs)
+
+    @cli.register_custom_action("User")
+    @exc.on_http_error(exc.GitlabUserRejectError)
+    def reject(self, **kwargs: Any) -> Union[Dict[str, Any], requests.Response]:
+        """Reject a user creation request.
+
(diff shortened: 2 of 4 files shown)


bgamari:
I have amended the commit message as suggested.

Unfortunately, it's not clear how to test this robustly given that the GitLab instance the tests are being run against must be configured to have new account approvals enabled. However, I am using this patch locally in managing spam account creation on https://gitlab.haskell.org/

nejch:
> I have amended the commit message as suggested.
> 
> Unfortunately, it's not clear how to test this robustly given that the GitLab instance the tests are being run against must be configured to have new account approvals enabled. However, I am using this patch locally in managing spam account creation on https://gitlab.haskell.org/

Thanks for the quick response @bgamari! 

We run integration tests against a gitlab container, but for unit tests, we just mock the responses (see file linked above). We usually just use the example responses from the API docs for the mocked response. I guess we should document this in our contributing guide a bit more, thanks for the feedback! :grin: 

In this case, this would just be "Examples responses" in https://docs.gitlab.com/ee/api/users.html#approve-user - we only really need to test the success case.

It's not really the best example, but you could check https://github.com/python-gitlab/python-gitlab/blob/main/tests/unit/objects/test_users.py#L63-L72 to see how the fixtures are used because it's so similar (though ideally we'd have an individual response as a fixture). Let me know if you're happy to work on that, otherwise we can merge and do this after. Thanks again :)

nejch (changes requested):

bgamari:
Thanks for the feedback, @nejch. I'll add a few unit tests.

docs/gl_objects/users.rst
```diff
@@ -106,6 +106,12 @@ List a user's starred projects
 
     user.starred_projects.list()
 
+If the GitLab instance has new user account approval enabled some users may
+have ``user.state == 'blocked_pending_approval'``. Administrators can approve
+and reject such users::
+
+    user.approve()
```
JohnVillalovos, line 113: How do we "reject" someone?

bgamari: I've added a mention of `reject()` as well.

gitlab/v4/objects/users.py
```diff
@@ -283,6 +283,42 @@ def activate(self, **kwargs: Any) -> Union[Dict[str, Any], requests.Response]:
...
+    @exc.on_http_error(exc.GitlabUserRejectError)
+    def reject(self, **kwargs: Any) -> Union[Dict[str, Any], requests.Response]:
+        """Reject a user creation request.
+
+        Args:
+            **kwargs: Extra options to send to the server (e.g. sudo)
+
+        Raises:
+            GitlabAuthenticationError: If authentication is not correct
+            GitlabUserReject: If the user could not be rejected
```
JohnVillalovos, line 314: Minor fix:

```suggestion
            GitlabUserRejectError: If the user could not be rejected
```

JohnVillalovos (changes requested):

nejch (approved):

nejch:
@bgamari thanks again for this, I rebased your branch and squashed the commit co-authored by John to make pre-commit happy. @JohnVillalovos LGTM any other concerns on your side?

JohnVillalovos (approved):

JohnVillalovos:
> @bgamari thanks again for this, I rebased your branch and squashed the commit co-authored by John to make pre-commit happy. @JohnVillalovos LGTM any other concerns on your side?

@nejch LGTM. I'm good to merge if you are.

Issue #2065, opened 2022-06-10
gitlab project-merge-request-approval-rule list throws a KeyError "approval_rule_id" error.


MaximeCheramy:
## Description of the problem, including code/CLI snippet

gitlab project-merge-request-approval-rule list --mr-iid 1 --project-id foo/bar

## Expected Behavior

A list of approval rules

## Actual Behavior

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/dist-packages/gitlab/base.py", line 107, in __getattr__
    return self.__dict__["_updated_attrs"][name]
KeyError: 'approval_rule_id'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.9/dist-packages/gitlab/base.py", line 110, in __getattr__
    value = self.__dict__["_attrs"][name]
KeyError: 'approval_rule_id'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.9/dist-packages/gitlab/base.py", line 130, in __getattr__
    return self.__dict__["_parent_attrs"][name]
KeyError: 'approval_rule_id'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/local/bin/gitlab", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.9/dist-packages/gitlab/cli.py", line 362, in main
    gitlab.v4.cli.run(gl, what, action, args_dict, verbose, output, fields)
  File "/usr/local/lib/python3.9/dist-packages/gitlab/v4/cli.py", line 515, in run
    printer.display_list(data, fields, verbose=verbose)
  File "/usr/local/lib/python3.9/dist-packages/gitlab/v4/cli.py", line 483, in display_list
    self.display(get_dict(obj, fields), verbose=verbose, obj=obj)
  File "/usr/local/lib/python3.9/dist-packages/gitlab/v4/cli.py", line 461, in display
    id = getattr(obj, obj._id_attr)
  File "/usr/local/lib/python3.9/dist-packages/gitlab/base.py", line 146, in __getattr__
    raise AttributeError(message) from exc
AttributeError: 'ProjectMergeRequestApprovalRule' object has no attribute 'approval_rule_id'

<class 'gitlab.v4.objects.merge_request_approvals.ProjectMergeRequestA
pprovalRule'> was created via a list() call and only a subset of the
data may be present. To ensure all data is present get the object
[...]

PR #2066, opened 2022-06-11
fix(cli): project-merge-request-approval-rule

JohnVillalovos:
Using the CLI the command:
   gitlab project-merge-request-approval-rule list --mr-iid 1 --project-id foo/bar

Would raise an exception. This was due to the fact that `_id_attr` and
`_repr_attr` were set for keys which are not returned in the response.

Closes: #2065


tests/unit/objects/test_project_merge_request_approvals.py
```diff
@@ -168,6 +168,7 @@ def test_list_merge_request_approval_rules(project, resp_mr_approval_rules):
     assert len(approval_rules) == 1
     assert approval_rules[0].name == approval_rule_name
     assert approval_rules[0].id == approval_rule_id
+    repr(approval_rules)
```
nejch, line 171: Is this a leftover from testing? :)

gitlab/v4/objects/merge_request_approvals.py
```diff
@@ -165,8 +165,6 @@ def set_approvers(
 
 
 class ProjectMergeRequestApprovalRule(SaveMixin, ObjectDeleteMixin, RESTObject):
-    _id_attr = "approval_rule_id"
-    _repr_attr = "approval_rule"
     id: int
     approval_rule_id: int
     merge_request_iid: int
```
nejch, line 170: Seems like we could actually use `name` for the `_repr_attr` here from what i can tell based on the get&list response:
https://docs.gitlab.com/ee/api/merge_request_approvals.html#get-merge-request-level-rules

I see some of this is used later in the `save()` method to create a custom ID because of the mismatch, so I'd just like to take a quick look at what's happening there before I merge :)

nejch (commented):
Thanks @JohnVillalovos! Just a quick question there, and I'll just check a bit later today what that other method is doing with the custom ID.

tests/unit/objects/test_project_merge_request_approvals.py
JohnVillalovos: It is not. Before this would raise an exception. Let me add a comment to it.

I did mention it in the commit message.

gitlab/v4/objects/merge_request_approvals.py
JohnVillalovos: Maybe, but I don't think `name` could be used in a `get()` (which we do need to implement). I guess I'm not entirely sure when to use `_repr_attr` 😔

nejch: I tried to document this a bit in https://github.com/python-gitlab/python-gitlab/pull/2068 (WIP).
`_repr_attr` will be used in repr methods if it's present in the object, e.g. after returning the response from the server. So I assume it will not be present on lazily created objects (and repr will show only id), but will be available after get/create/list etc. It is not used to actually fetch the resource.

JohnVillalovos: > Seems like we could actually use `name` for the `_repr_attr` here from what i can tell based on the get&list response: https://docs.gitlab.com/ee/api/merge_request_approvals.html#get-merge-request-level-rules

Done 🙂

tests/unit/objects/test_project_merge_request_approvals.py
nejch: Thanks, was confused by the lack of assert I think, and I missed that as I only checked the PR description :grin:

nejch (approved):

PR #2067, opened 2022-06-12
Specific type hints

### LINKED ISSUE #2062: Type hint on ListMixin.list is too strict

vectro:
Update the type hints on Mixin objects to be more specific so that clients can know the specific RESTObject that would be returned.

Closes  #2062


nejch:
Hi @vectro. thanks for the earlier discussions and for looking into this. We use `tox` to manage our environments and versions so you can use that before pushing to make sure all the test & lint jobs succeed consistently with the versions we track.

https://github.com/python-gitlab/python-gitlab/blob/main/CONTRIBUTING.rst#running-unit-tests

One of these is a commitlint check as we parse commits to create changelogs so this might need a bit of a squash and `amend` once it's ready for review :bow:

vectro:
@nejch I ran tox and see some errors that I'm not sure what to do with. What do you think?

- At gitlab/v4/cli.py:352 related to SidekiqManager but I'm not really sure what that class should look like. Should it even have an _obj_cls attribute?
- At tests/meta/test_mro.py:122 because Generic is now the final class in the MRO (I suppose) and not RESTManager.
- cz is unhappy but presumably that can be resolved by squashing commits.

nejch:
@vectro sorry for the super late response! I haven't checked this in depth myself but a few pointers:

- the SidekiqManager failure probably has to do with the fact you removed the optional `_obj_cls` there. We do have a few quirky managers without a defined `_obj_cls`, but some parts of the code rely on it so we default to `None` in the base.
- `test_mro.py` is a custom check and is explained in more detail here https://github.com/python-gitlab/python-gitlab/blob/30117a3b6a8ee24362de798b2fa596a343b8774f/tests/meta/test_mro.py. If a new approach to typing makes that redundant we can potentially get rid of that but not sure if that's the case here already. @JohnVillalovos might be able to help a bit more here as he did the work on that initially.
- cz: yes, just squash/amend the commit message according to https://github.com/python-gitlab/python-gitlab/blob/main/CONTRIBUTING.rst#commit-message-guidelines and you should be fine :)

I'm not sure your code gets checked by mypy as expected, as the generics part is defined outside the `if TYPE_CHECKING` block - unless I'm misunderstanding something here:

https://github.com/python-gitlab/python-gitlab/blob/bd81ee01365d796928922dad7b3606cce600a8ad/gitlab/mixins.py#L67-L79

Also keep in mind @JohnVillalovos did some work on typing recently so that might affect your rebase here.

PR #2069, opened 2022-06-14
test(projects): add unit tests for projects

### LINKED ISSUE #1708: Tests need to be implemented for tests/unit/test_projects.py
Currently we have a lot of unimplemented tests in tests/unit/test_projects.py.

Adding these tests would be a good first issue for someone to work on.

Comment below if you would like to work on this issue.

antoineauger:
Closes #1708

:hammer_and_wrench: with :heart: by [Siemens](https://opensource.siemens.com/)

### DIFF
--- a/tests/unit/objects/test_projects.py
+++ b/tests/unit/objects/test_projects.py
@@ -2,9 +2,13 @@
 GitLab API: https://docs.gitlab.com/ce/api/projects.html
 """
 
+from unittest.mock import mock_open, patch
+
 import pytest
 import responses
 
+from gitlab import exceptions
+from gitlab.const import DEVELOPER_ACCESS, SEARCH_SCOPE_ISSUES
 from gitlab.v4.objects import (
     Project,
     ProjectFork,
@@ -15,6 +19,11 @@ from gitlab.v4.objects import (
 from gitlab.v4.objects.projects import ProjectStorage
 
 project_content = {"name": "name", "id": 1}
+project_with_owner_content = {
+    "name": "name",
+    "id": 1,
+    "owner": {"id": 1, "username": "owner_username", "name": "owner_name"},
+}
 languages_content = {
     "python": 80.00,
     "ruby": 99.99,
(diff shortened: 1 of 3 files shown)


tests/unit/objects/test_projects.py
```diff
@@ -142,30 +281,263 @@ def resp_list_projects():
...
+            status=200,
+        )
+        rsps.add(
+            method=responses.PUT,
+            url="http://localhost/api/v4/projects/1/push_rule",
+            json=push_rules_content,
+            content_type="application/json",
+            status=201,
+        )
+        yield rsps
```
antoineauger, line 485: I'm not full satisfied with this fixture because the first GET response is duplicated in [`resp_delete_push_rules_project()`](https://github.com/python-gitlab/python-gitlab/pull/2069/files#diff-e84ffbb57aae6c0765f257e13ae65e6522deac96c73b275f6450df2d9f2c5308R491)

tests/unit/objects/test_projects.py
```diff
@@ -655,11 +655,6 @@ def test_delete_project(project, resp_delete_project):
     project.delete()
 
 
-@pytest.mark.skip(reason="missing test")
-def test_restore_project(gl):
-    pass
```
antoineauger, line 296: The `restore()` function is currently not implement for `projects` as far as I can see 🤔

antoineauger:
@nejch Here is a first try to reach the 100% coverage for projects/push rules 😄 

🏓

tests/unit/objects/test_projects.py
nejch: Oh, oops, not sure how it got into the test stubs then :) Thanks, opened https://github.com/python-gitlab/python-gitlab/issues/2070 for this!

max-wittig: yeah we don't really prioritize premium features.

tests/unit/objects/test_projects.py
nejch: Thanks @antoineauger - I think what we've done in these situations is just activate responses with a decorator and explicitly add individual responses inside the test cases as needed, then assert that they were fired.

But I think it's ok to keep it like this for now - we have some other duplicates elsewhere so I think we can do a cleanup in a dedicated PR :)

nejch (commented):
Thanks a lot for all the work here @antoineauger :open_mouth: I left only some tiny comment but it's non-blocking. I think we can leave the duplicate for now as we have to do more cleanup in other parts.

I realize adding the response fixtures now is quite tedious, will probably need to think of some lightweight wrappers to make it more manageable.

tests/unit/objects/test_projects.py
```diff
@@ -30,10 +39,49 @@
...
+    "forked_from_project": {"id": 1, "name": "name", "forks_count": 1},
+}
+project_starrers_content = {
+    "starred_since": "2019-01-28T14:47:30.642Z",
+    "user": {
+        "id": 1,
+        "name": "name",
+    },
+}
+delete_project_content = {"message": "202 Accepted"}
```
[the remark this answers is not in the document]
antoineauger: Changed in a51f848db4204b2f37ae96fd235ae33cb7c2fe98 to use `accepted_content` and `no_content` fixtures when applicable.

tests/unit/objects/test_projects.py
```diff
@@ -50,6 +98,58 @@ def resp_get_project():
         yield rsps
 
 
+@pytest.fixture
+def resp_post_project():
```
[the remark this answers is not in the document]
antoineauger: Sure, I made a pass to rename everything according to this logic in a51f848db4204b2f37ae96fd235ae33cb7c2fe98

antoineauger:
> Thanks a lot for all the work here @antoineauger open_mouth I left only some tiny comment but it's non-blocking. I think we can leave the duplicate for now as we have to do more cleanup in other parts.
> 
> I realize adding the response fixtures now is quite tedious, will probably need to think of some lightweight wrappers to make it more manageable.

Thanks for the valuable feedback @nejch, I think I addressed everything. Back to you for final check 🏓

nejch (approved):
Looks great, thanks again @antoineauger! Really happy to see this done, it was open for a long time :)

PR #2077, opened 2022-06-20
chore: fix issue found with pylint==2.14.3

JohnVillalovos:
A new error was reported when running pylint==2.14.3:
  gitlab/client.py:488:0: W1404: Implicit string concatenation found in call (implicit-str-concat)

Fixed this issue.


nejch:
Thanks @JohnVillalovos! Looks like a leftover from black commits or so :)

PR #2083, opened 2022-06-20
chore: add link to Commitizen in Github workflow

JohnVillalovos:
Add a link to the Commitizen website in the Github workflow. Hopefully
this will help people when their job fails.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -27,7 +27,7 @@ jobs:
           fetch-depth: 0
       - uses: actions/setup-python@v3
       - run: pip install --upgrade tox
-      - name: Run commitizen
+      - name: Run commitizen (https://commitizen-tools.github.io/commitizen/)
         run: tox -e cz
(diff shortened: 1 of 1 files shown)


nejch:
My suspicion based on previous PRs is that unfortunately not :rofl: but this definitely won't hurt, thanks @JohnVillalovos!

JohnVillalovos:
> My suspicion based on previous PRs is that unfortunately not :rofl: but this definitely won't hurt, thanks @JohnVillalovos! 

I have an idea about that but need to think more

nejch:
> > My suspicion based on previous PRs is that unfortunately not rofl but this definitely won't hurt, thanks @JohnVillalovos!
> 
> I have an idea about that but need to think more

IMO it'd be easiest with a pre-written bot to comment on the PRs on failure. I'm sure there are already some out there, on GitLab they use https://danger.systems/python/ and https://www.rubydoc.info/gems/gitlab-triage, there's also https://github.com/reviewdog/reviewdog though.

Issue #1130, opened 2020-07-07
Protected Environments not supported


nastrofaction:
## Description of the problem, including code/CLI snippet
Per API: https://docs.gitlab.com/ee/api/protected_environments.html we should be able to create protected environment but the code fails as it isn't supported

## Expected Behavior
```
project.protectedenvironments.create({
                'name': env,
                'deploy_access_levels': [{"access_level": gitlab.MAINTAINER_ACCESS}]
            })
```
should create the protected environment

## Actual Behavior
Doesn't work
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/gitlab/base.py", line 59, in __getattr__
    return self.__dict__["_updated_attrs"][name]
KeyError: 'protectedenvironments'
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/gitlab/base.py", line 62, in __getattr__
    value = self.__dict__["_attrs"][name]
KeyError: 'protectedenvironments'
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/gitlab/base.py", line 82, in __getattr__
    return self.__dict__["_parent_attrs"][name]
KeyError: 'protectedenvironments'
```

## Specifications

  - python-gitlab version: python-gitlab-2.3.1
  - API version you are using (v3/v4): v4
  - Gitlab server version (or gitlab.com): gitlab.com

max-wittig:
I don't have access to Gitlab Premium. That's why I'm not really implementing these features, but I would be open to review any MRs regarding EE features.

nastrofaction:
Thanks for responding @max-wittig For now I am using cURL to complete my automation through API for protected environments. Once this is complete; I'll try to come back and submit MR for this.

filip-malczak-bny:
@nastrofaction is there any update on this?

calve:
I added support for Protected Environments in https://github.com/python-gitlab/python-gitlab/pull/2084 which is merged but not released yet

Do note that the API use the name `protected_environments` and not `protectedenvironments` as specified in this issue

nickbroon:
Protected environment support is available since [v3.6.0](https://github.com/python-gitlab/python-gitlab/releases/tag/v3.6.0). Should this issues now be closed?

nejch:
[...]

PR #2084, opened 2022-06-21
feat: Add support for Protected Environments

calve:
- https://docs.gitlab.com/ee/api/protected_environments.html
- https://github.com/python-gitlab/python-gitlab/issues/1130

no write operation are implemented yet as I have no use case right now
and am not sure how it should be done

### DIFF
--- a/docs/gl_objects/protected_environments.rst
+++ b/docs/gl_objects/protected_environments.rst
@@ -1,0 +1,45 @@
+######################
+Protected environments
+######################
+
+You can list and manage protected environments in a project.
+
+References
+----------
+
+* v4 API:
+
+  + :class:`gitlab.v4.objects.ProjectProtectedEnvironment`
+  + :class:`gitlab.v4.objects.ProjectProtectedEnvironmentManager`
+  + :attr:`gitlab.v4.objects.Project.protected_environment`
+
+* GitLab API: https://docs.gitlab.com/ee/api/protected_environments.html
+
+Examples
+--------
+
+Get the list of protected environments for a project::
+
+    p_environments = project.protected_environments.list()
+
+Get a single protected environment::
+
+    p_environments = project.protected_environments.get('production')
+
+Protect an existing environment::
+
+    p_environment = project.protected_environments.create(
+        {
+            'name': 'production',
+            'deploy_access_levels': [
+                {'access_level': 40}
+            ],
+        }
+    )
+
+
+Unprotect a protected environment::
+
+    p_environment = project.protected_environments.delete('production')
+    # or
+    p_environment.delete()
--- a/gitlab/v4/objects/environments.py
+++ b/gitlab/v4/objects/environments.py
@@ -18,6 +18,8 @@ from gitlab.types import RequiredOptional
 __all__ = [
     "ProjectEnvironment",
     "ProjectEnvironmentManager",
+    "ProjectProtectedEnvironment",
+    "ProjectProtectedEnvironmentManager",
 ]
 
 
@@ -55,3 +57,30 @@ class ProjectEnvironmentManager(
         self, id: Union[str, int], lazy: bool = False, **kwargs: Any
     ) -> ProjectEnvironment:
         return cast(ProjectEnvironment, super().get(id=id, lazy=lazy, **kwargs))
+
+
+class ProjectProtectedEnvironment(ObjectDeleteMixin, RESTObject):
+    _id_attr = "name"
+    _repr_attr = "name"
+
+
+class ProjectProtectedEnvironmentManager(
+    RetrieveMixin, CreateMixin, DeleteMixin, RESTManager
+):
+    _path = "/projects/{project_id}/protected_environments"
+    _obj_cls = ProjectProtectedEnvironment
+    _from_parent_attrs = {"project_id": "id"}
+    _create_attrs = RequiredOptional(
+        required=(
+            "name",
+            "deploy_access_levels",
+        ),
+        optional=("required_approval_count", "approval_rules"),
+    )
+
+    def get(
+        self, id: Union[str, int], lazy: bool = False, **kwargs: Any
+    ) -> ProjectProtectedEnvironment:
+        return cast(
+            ProjectProtectedEnvironment, super().get(id=id, lazy=lazy, **kwargs)
+        )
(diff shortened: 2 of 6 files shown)


calve:
more specically, I guess all crud operation are inherited from the CRUD mixin but we should ideally implement `stop()`, `protect()` and `unprotect()` methods ?

JohnVillalovos (changes requested):
Thanks for the PR!

Overall looks good. In addition to the one thing I mentioned, would also be good to create a unit test for this.

gitlab/v4/objects/projects.py
```diff
@@ -175,6 +178,7 @@ class Project(RefreshMixin, SaveMixin, ObjectDeleteMixin, RepositoryMixin, RESTO
     pagesdomains: ProjectPagesDomainManager
     pipelines: ProjectPipelineManager
     pipelineschedules: ProjectPipelineScheduleManager
+    protectedenvironments: ProjectProtectedEnvironmentManager
```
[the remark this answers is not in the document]
calve: Sure. 

At some point you might want to proxy unmatching attributes to keep a naming coherence ?

```python
protected_environments: ProjectProtectedEnvironmentManager
protectedenvironments: self.protected_environments  # or anything that actually works
```

to avoid surprise as in 

```python
print(project.protected_environments, project.protectedvariables)
```

we might still ensure identity 

```
assert project.protected_environments is project.protectedenvironments
```.

Anyway I'm taking the freedom to resolve this conversation :)

File: gitlab/v4/objects/environments.py [hunk 2]
```diff
@@ -55,3 +57,26 @@ def get(
...
         return cast(ProjectEnvironment, super().get(id=id, lazy=lazy, **kwargs))
+
+
+class ProjectProtectedEnvironment(ObjectDeleteMixin, RESTObject):
+    _id_attr = "name"
+    _repr_attr = "name"
+
+
+class ProjectProtectedEnvironmentManager(
+    RetrieveMixin, CreateMixin, UpdateMixin, DeleteMixin, RESTManager
```
nejch, line 68: I'm not sure about `UpdateMixin` - does the API allow `PUT` on existing protected environments? Even protected branches don't support editing, so I'd think that no.

nejch:
Thanks for the great PR and quick rework @calve. It'd be cool if you could add a tiny bit of docs for this (see protected branches) if you have the time & energy as well :)

calve:
Thanks both of you for the quick review.

> BTW: There is also a group-level protected environments endpoint: https://docs.gitlab.com/ee/api/group_protected_environments.html

I'm inclined to let the implementation to a future reader ?

> I'm not sure about UpdateMixin - does the API allow PUT on existing protected environments? Even protected branches don't support editing, so I'd think that no.

> I'll also add a note in the docs for our protected branches/tags that create/delete does protect/unprotect.

And globally on the create/delete/protect/unprotect behavior, I have to say I did not have those usages right now and so haven't actually tested anything related to other operation than list/get.

I just realised the documentation isn't autogenerated as I initially thought, I'm going to amend the commit to add a specific pages.

gitlab/v4/objects/environments.py
calve: it looks like it does not per https://docs.gitlab.com/ee/api/protected_environments.html ?

I understand I should remove the UpdateMixin from the inheritance tree, right ?

nejch:
> And globally on the create/delete/protect/unprotect behavior, I have to say I did not have those usages right now and so haven't actually tested anything related to other operation than list/get.

> I just realised the documentation isn't autogenerated as I initially thought, I'm going to amend the commit to add a specific pages.

Thanks for the quick reply @calve. We auto-generate the API reference but add some prose docs with examples usually, as some people find that easier to grok.

Sounds good, the rest can be a follow-up :)

gitlab/v4/objects/environments.py
nejch: I think that's best @calve yes, otherwise some people get confused and try to use the `update()` methods. Thanks!

calve:
I amended the commit to 

 - remove the UpdateMixin, and the associated `.update()` method
 - set `_create_attrs` and `_list_filters` to what seems to be the actual spec
 - document that `create()` and `delete()` actually protect/unprotected existing environments
 - manually (and successfully) test a full CRUD cycle on my gitlab-ee instance

please let me know for any other changes or information :)

docs/gl_objects/protected_environments.rst
```diff
@@ -0,0 +1,45 @@
+######################
+Protected environments
+######################
+
+You can define a list of protected environments on a repository.
```
nejch, line 5:
```suggestion
You can list and manage protected environments in a project.
```

tests/unit/objects/test_environments.py
```diff
@@ -22,9 +22,36 @@ def resp_get_environment():
...
     assert isinstance(environment, ProjectEnvironment)
     assert environment.id == 1
     assert environment.last_deployment == "sometime"
     assert environment.name == "environment_name"
+
+
+def test_project_protected_environments(project, resp_get_protected_environment):
+    protected_environment = project.protected_environments.get(2)
+    assert isinstance(protected_environment, ProjectProtectedEnvironment)
+    assert protected_environment.id == 2
```
nejch, line 55:
```suggestion
```

nejch:
Thanks! Just a few more typos/nitpicks from me, then we should be good to go. Our handling of IDs is a bit different from how GitLab documents them, because we create objects to manipulate, so I've added some comments.

nejch (changes requested):

gitlab/v4/objects/branches.py
```diff
@@ -48,6 +48,9 @@ class ProjectProtectedBranchManager(NoUpdateMixin, RESTManager):
             "code_owner_approval_required",
         ),
     )
+    _delete_attrs = RequiredOptional(
+        required=("name",),
+    )
```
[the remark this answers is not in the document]
calve: why on earth did I commited a changes in `branch.py` :facepalm: I'm reverting this change, thanks for the review

nejch:
Thanks for the super quick reactions @calve. LGTM now, this should make it into the next scheduled release on the 28th.

gitlab/v4/objects/projects.py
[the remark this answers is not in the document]
nejch: @calve missed this one - yes that definitely makes sense for some of our other managers too, we did something similar for the breaking changes in v3.0.0 but should probably check all of our endpoints again. Thanks @JohnVillalovos for pointing this out :)

nejch (approved):

calve:
Thanks for your time, the super review and velocity :handshake:

PR #2085, opened 2022-06-22
refactor: do not recommend plain gitlab.const constants

nejch:
Since we now have enums, we shouldn't tell people to use the plain constants :) also cleans up the use of old constants in tests.

### DIFF
--- a/docs/gl_objects/access_requests.rst
+++ b/docs/gl_objects/access_requests.rst
@@ -43,7 +43,7 @@ Create an access request::
 Approve an access request::
 
     ar.approve()  # defaults to DEVELOPER level
-    ar.approve(access_level=gitlab.const.AccessLevel.MAINTAINER.value)  # explicitly set access level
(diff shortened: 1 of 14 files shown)


nejch:
@JohnVillalovos not sure, but this one could potentially also be squeezed in for the release as the old warning still points to plain constants, so this would avoid deprecating things twice.

docs/gl_objects/access_requests.rst
```diff
@@ -43,7 +43,7 @@ Create an access request::
 Approve an access request::
 
     ar.approve()  # defaults to DEVELOPER level
-    ar.approve(access_level=gitlab.const.AccessLevel.MAINTAINER.value)  # explicitly set access level
+    ar.approve(access_level=gitlab.const.AccessLevel.MAINTAINER)  # explicitly set access level
```
JohnVillalovos, line 46: Nice that don't need the `.value` 👍

JohnVillalovos (approved):

PR #2086, opened 2022-06-22
chore: don't use `Enum` as we don't need it.

JohnVillalovos:
For our use case `Enum` is overkill. We can just use simple classes
instead.

Looking for some feedback on this.

@nejch  When I saw how we are going from `gitlab.OWNER` -> `gitlab.const.OWNER` -> `gitlab.const.AccessLevel.OWNER.value`, it just seemed verbose, especially the `.value`. So that is why I thought of this idea. So with this it would be `gitlab.const.AccessLevel.OWNER`

### DIFF
--- a/gitlab/const.py
+++ b/gitlab/const.py
@@ -1,7 +1,5 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright (C) 2016-2017 Gauvain Pocentek <gauvain@pocentek.net>
-#
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU Lesser General Public License as published by
 # the Free Software Foundation, either version 3 of the License, or
@@ -15,7 +13,7 @@
 # You should have received a copy of the GNU Lesser General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-from enum import Enum, IntEnum
+from typing import Any
 
 from gitlab._version import __title__, __version__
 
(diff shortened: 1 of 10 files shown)


gitlab/const.py
```diff
@@ -56,8 +54,18 @@
...
 
+class GitlabConstant:
+    def __new__(cls, value: Any) -> str:  # type: ignore
+        # This allows us to get a string representation of a value.
+        # For example: AccessLevel(10)
+        for attr, attr_value in cls.__dict__.items():
+            if value == attr_value:
+                return attr
+        raise ValueError(f"{value!r} is not a valid {cls.__qualname__}")
+
```
JohnVillalovos, line 65: I only put the `__new__()` method in so we could do `AccessLevel(10)`. But not sure if we need/desire that.

nejch:
> For our use case `Enum` is overkill. We can just use simple classes instead.
> 
> Looking for some feedback on this.
>
> @nejch When I saw how we are going from `gitlab.OWNER` -> `gitlab.const.OWNER` -> `gitlab.const.AccessLevel.OWNER.value`, it just seemed verbose, especially the `.value`. So that is why I thought of this idea. So with this it would be `gitlab.const.AccessLevel.OWNER`

For _you and me_ it's overkill :grin: even the extended constants are for me, tbh - everything in `gitlab.const*` is very verbose to me. But they were a community contribution initially and so are the Enums now, and I think people have use cases for this: `levels = list(AccessLevel)`, easily enumerating over it for input validation, filtering/comparing levels using human-readable names etc.

Some examples I linked before are in
https://github.com/encode/httpx/blob/master/httpx/_status_codes.py and
https://github.com/gitlabform/gitlabform/blob/main/gitlabform/gitlab/__init__.py#L23-L45.

I think we don't even need the `.value` actually, it seems. I did a quick push to https://github.com/python-gitlab/python-gitlab/pull/2085 and it works, but a `__new__()` approach might be simpler yes, as is done in httpx. But I'd keep the enums now as multiple people probable have some use for it judging from the feedback in the other PR.

Regarding reducing verbosity, putting these things into enums groups them quite well into namespaces so from my perspective we could re-export them explicitly now, to e.g. `gitlab.Visibility`, `gitlab.AccessLevel` etc. We wouldn't need star imports anymore - and wouldn't need to force usage of `gitlab.const.` - would be pretty short with `gitlab.Visibility.PRIVATE` for example.

Issue #2063, opened 2022-06-09
python-gitlab uses implicit reexports


vectro:
python-gitlab does not define `__all__` which causes problems in client applications that are inspected with mypy.

I think the solution here is to just define `__all__` in `__init__.py` and maybe some other libraries that re-export.

nejch:
Thanks @vectro I think we have discussed fixing this in the past already but I can't find the issue. An explicit `__all__` probably makes sense but for historical reasons we still have some glob imports here so I'm not sure how maintainable/error-prone that is until we get rid of some of those root-level re-exports (a breaking change).

Perhaps dynamically extending `__all__` with the submodule's own exports might make sense as suggested here. https://github.com/microsoft/pyright/blob/main/docs/typed-libraries.md#library-interface

nejch:
I found it now, we previously briefly mentioned this in https://github.com/python-gitlab/python-gitlab/issues/1588. Good idea :)

PR #2087, opened 2022-06-23
chore(gitlab): fix implicit re-exports for mypy

nejch:
Closes https://github.com/python-gitlab/python-gitlab/issues/2063


nejch:
The sphinx patch handles the fact that we're now explicitly exporting attributes in 2 places.

/cc @vectro @dAnjou FYI as you initially reported this typing issue. This is kind of the cleanest I think I get to fixing implicit re-exports without breaking changes to our messy star imports in `gitlab/__init__.py` from legacy code. But if you have any better ideas I'm happy to follow up on this.

JohnVillalovos (approved):
Nice job finding a fix for the sphinx issue!

Issue #2080, opened 2022-06-20
Formalize logging


nejch:
In addition to that change, I think it makes sense to log (in DEBUG level) that a rate limit 429 has occurred, but IDK how the maintainers would like to add logging here, if at all. If anyone agrees with me on this I'd love to have guidance on how to add this logging.

_Originally posted by @bachsh in https://github.com/python-gitlab/python-gitlab/issues/2074#issuecomment-1159817354_

nejch:
@bachsh I've also wanted to add some logging here for some time, and remove the custom `enable_debug()` that's been there for ages in favor of using standard logging methods, environment variable to set the log level, etc. Changing that would be the first step IMO.

---

PS: Personally I'd just go with something like https://github.com/Delgan/loguru because I don't like the tedious stdlib logging setup, but this being a library, people won't like more dependencies probably.

PR #2093, opened 2022-06-25
chore: add the version of python-gitlab to Exception

JohnVillalovos:
People will post tracebacks of AttributeError exceptions but not post
information on the version of python-gitlab used.

Add the version of python-gitlab to the AttributeError exception
message.


nejch:
Thanks @JohnVillalovos! This is a specific implementation for one exception. If we do this, a more generic approach might be useful I think, otherwise we might have to litter our code with custom exception messages.

But I'm also not sure if exception messages are the right place for this. Did you get this idea from existing libraries? I think we should look into https://github.com/python-gitlab/python-gitlab/issues/2080, maybe in combination with exception hooks if we want this :thinking:  not sure, it just seems more like a logging topic IMO?

JohnVillalovos:
Good idea. Let me explore having it in the base GitlabExeception class. But probably not today...

nejch:
> Good idea. Let me explore having it in the base GitlabExeception class. But probably not today...

I would also look at https://stackoverflow.com/questions/5191830/how-do-i-log-a-python-error-with-debug-information as a clean option. Of course, we'd need a nicer setup for https://github.com/python-gitlab/python-gitlab/issues/2080 first.

PR #2099, opened 2022-06-25
docs(api): add separate page for advanced usage

nejch:
The "getting started" section for the API is getting a little long now, so I thought it'd make sense to split the advanced usage part out and add a separate page for it. This is just copy/paste pretty much.


JohnVillalovos (approved):
Thanks. Good idea.

PR #2100, opened 2022-06-26
test(pylint): enable pylint "unused-argument" check

JohnVillalovos:
Enable the pylint "unused-argument" check and resolve issues it found.

  * Quite a few functions were accepting `**kwargs` but not then
    passing them on through to the next level. Now pass `**kwargs` to
    next level.
  * Other functions had no reason to accept `**kwargs`, so remove it
  * And a few other fixes.

### DIFF
--- a/gitlab/v4/cli.py
+++ b/gitlab/v4/cli.py
@@ -381,7 +381,7 @@ def get_dict(
 
 class JSONPrinter:
     @staticmethod
-    def display(d: Union[str, Dict[str, Any]], **kwargs: Any) -> None:
+    def display(d: Union[str, Dict[str, Any]], **_kwargs: Any) -> None:
         import json  # noqa
 
         print(json.dumps(d))
@@ -390,7 +390,7 @@ class JSONPrinter:
     def display_list(
         data: List[Union[str, gitlab.base.RESTObject]],
         fields: List[str],
-        **kwargs: Any,
+        **_kwargs: Any,
     ) -> None:
         import json  # noqa
 
@@ -399,7 +399,7 @@ class JSONPrinter:
 
 class YAMLPrinter:
     @staticmethod
-    def display(d: Union[str, Dict[str, Any]], **kwargs: Any) -> None:
+    def display(d: Union[str, Dict[str, Any]], **_kwargs: Any) -> None:
         try:
             import yaml  # noqa
 
@@ -415,7 +415,7 @@ class YAMLPrinter:
     def display_list(
         data: List[Union[str, gitlab.base.RESTObject]],
         fields: List[str],
-        **kwargs: Any,
+        **_kwargs: Any,
     ) -> None:
         try:
             import yaml  # noqa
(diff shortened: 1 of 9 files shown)


gitlab/v4/objects/merge_request_approvals.py
```diff
@@ -119,7 +119,6 @@ def set_approvers(
         approver_ids: Optional[List[int]] = None,
         approver_group_ids: Optional[List[int]] = None,
         approval_rule_name: str = "name",
-        **kwargs: Any,
```
nejch, line 122: The `create()` called in this method supports passing **kwargs so we should pass it on IMO, e.g. for `sudo` and friends, as we document it on the intro page.

https://github.com/python-gitlab/python-gitlab/blob/665b2c8ec6491990bc838139f6afc2797dd895c9/gitlab/v4/objects/merge_request_approvals.py#L217

gitlab/v4/objects/services.py
```diff
@@ -267,7 +267,7 @@ def get(
         return cast(ProjectService, super().get(id=id, lazy=lazy, **kwargs))
 
     @cli.register_custom_action("ProjectServiceManager")
-    def available(self, **kwargs: Any) -> List[str]:
+    def available(self) -> List[str]:
```
nejch, line 270: Same as above, we should pass it on for `sudo` etc IMO.

gitlab/v4/cli.py
```diff
@@ -382,6 +382,7 @@ def get_dict(
 class JSONPrinter:
     @staticmethod
     def display(d: Union[str, Dict[str, Any]], **kwargs: Any) -> None:
+        _ = kwargs  # Disables PyLint 'unused-argument' error
```
nejch, line 385: It seems like this might make pylint happy and avoid the superfluous dummy variables?
```suggestion
    def display(d: Union[str, Dict[str, Any]], **_kwargs: Any) -> None:
```

nejch (changes requested):
Thanks @JohnVillalovos, just a few tiny comments mainly for always passing on `kwargs` as we claim in our docs that we support it for `sudo` (and maybe in a few cases would also be useful for pagination args).

gitlab/v4/cli.py
JohnVillalovos: Done

gitlab/v4/objects/merge_request_approvals.py
JohnVillalovos: Good catch. Done.

gitlab/v4/objects/services.py
JohnVillalovos: Pass it to what?

It just calls:
```
return list(self._service_attrs.keys())
```

nejch: Oh I'm dumb, I just saw `list()` and assumed it was the same as above :grin: thanks.

PR #2107, opened 2022-06-27
fix: iterator possible breaking change

JohnVillalovos:
Commit b6447211754e126f64e12fc735ad74fe557b7fb4 inadvertently
introduced a possible breaking change as it added a new argument
`iterator` and added it in between existing positional arguments.

Move the `iterator` argument to the end of the argument list and also
require it to be a keyword-only argument.


nejch:
I just changed this to a refactor so it doesnt show up in the changelog as it does not affect a released version

PR #2115, opened 2022-06-30
fix: don't ignore 'as_list=False' for getting runners as iterator

koreno:
The `gitlab.runners.all(as_list=False)` was ignoring the flag, returning a list regardless

### DIFF
--- a/gitlab/v4/objects/runners.py
+++ b/gitlab/v4/objects/runners.py
@@ -98,7 +98,10 @@ class RunnerManager(CRUDMixin, RESTManager):
         if scope is not None:
             query_data["scope"] = scope
         obj = self.gitlab.http_list(path, query_data, **kwargs)
-        return [self._obj_cls(self, item) for item in obj]
+        if isinstance(obj, list):
+            return [self._obj_cls(self, item, created_from_list=True) for item in obj]
+        else:
+            return base.RESTObjectList(self, self._obj_cls, obj)    
 
     @cli.register_custom_action("RunnerManager", ("token",))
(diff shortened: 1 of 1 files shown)


nejch:
Thanks for your contribution @koreno!

I think we could get this in, but ideally we would get rid of the custom `all()` methods and convert the `runners/all` into a proper manager so we can use `runners_all.list(iterator=True)` directly using the `ListMixin`. 

We did exactly the same for the `members/all` endpoint and deprecated `all()`: https://github.com/python-gitlab/python-gitlab/pull/1376, so we already have some precedent on how to do it (just ignore the `AllMixin` changes there, not relevant here).
I've wanted to do this for a while. Would you be interested in reworking the PR for that? Otherwise we'll start with a minimal change and follow-up.

That would actually fix https://github.com/python-gitlab/python-gitlab/issues/593 as well.

koreno:
I'm gonna have to let you take over from here, I can't really afford to dig in much further...

nejch:
> I'm gonna have to let you take over from here, I can't really afford to dig in much further...

Thanks for your work on this @koreno. to avoid double work (e.g. getting your initial fix in and then tackling #593), I'll get to this ASAP.

nejch:
Thanks again @koreno, I've opened https://github.com/python-gitlab/python-gitlab/pull/2167 to reuse our existing `list()` infrastructure and fix a related issue, replacing the custom method used here. Note that the new `runners_all.list()` will behave slightly differently, you'll still need to `get()` individual runners to get more details or manipulate them.

Should be released on the 28th.

Issue #2116, opened 2022-07-03
Get asset links issue if tag_name have "/" char


thcao-ampere:
## Description of the problem, including code/CLI snippet
```
releases = project.releases.list()
for rel in releases:
    links = rel.links.list()
```
https://github.com/python-gitlab/python-gitlab/blob/main/gitlab/v4/objects/releases.py#L43
```
_path = "/projects/{project_id}/releases/{tag_name}/assets/links"
```
If tag_name = "abc/xyz" --> `_path = "/projects/{project_id}/releases/abc/xyz/assets/links"`
We should replace "/" to "%2F" in tag_name,  `_path = "/projects/{project_id}/releases/abc%2Fxyz/assets/links"`

## Expected Behavior
Get asset link list from release

## Actual Behavior
Get exception `GitlabListError: 404: 404 Not Found`

## Specifications

  - python-gitlab version: all version
  - API version you are using (v3/v4): v4
  - Gitlab server version (or gitlab.com): gitlab.com

PR #2117, opened 2022-07-03
fix: ensure path elements are escaped

JohnVillalovos:
Ensure the path elements that are passed to the server are escaped.
For example a "/" will be changed to "%2F"

Closes: #2116


nejch:
Thanks for the quick fix @JohnVillalovos!

PR #2118, opened 2022-07-03
test(functional): replace len() calls with list membership checks

nejch:
See https://github.com/python-gitlab/python-gitlab/pull/1784#issuecomment-1172730920.

I've felt a bit weird for a while about having these magic numbers in tests, and they don't really test whether the actual resource created is or isn't in the list. This should hopefully also help with some of the race conditions potentially introduced with async gitlab operations.

I've left a few where it would need some further refactoring (e.g. fixture factories for group/member creation) or where I thought `len()` was still useful. It was already quite tedious so it might be tedious to review, I can split it into multiple commits if needed :)

### DIFF
--- a/tests/functional/api/test_clusters.py
+++ b/tests/functional/api/test_clusters.py
@@ -1,5 +1,5 @@
 def test_project_clusters(project):
-    project.clusters.create(
+    cluster = project.clusters.create(
         {
             "name": "cluster1",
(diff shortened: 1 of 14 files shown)


JohnVillalovos (approved):
@nejch  Wow! A lot of work there. Thanks!

Great idea.

PR #2125, opened 2022-07-04
docs(users): add docs about listing a user's projects

JohnVillalovos:
Add docs about listing a user's projects.

Update docs on the membership API to update the URL to the upstream
docs and also add a note that it requires Administrator access to use.

### DIFF
--- a/docs/gl_objects/users.rst
+++ b/docs/gl_objects/users.rst
@@ -192,6 +192,31 @@ Revoke (delete) an impersonation token for a user::
     i_t.delete()
 
 
+User projects
+=========================
+
+References
+----------
+
+* v4 API:
+
+  + :class:`gitlab.v4.objects.UserProject`
+  + :class:`gitlab.v4.objects.UserProjectManager`
+  + :attr:`gitlab.v4.objects.User.projects`
+
+* GitLab API: https://docs.gitlab.com/ee/api/projects.html#list-user-projects
+
+List visible projects in the user's namespace::
+
+    projects = user.projects.list()
+
+.. note::
+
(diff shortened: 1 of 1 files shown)


docs/gl_objects/users.rst
```diff
@@ -192,6 +192,31 @@ Revoke (delete) an impersonation token for a user::
...
+
+* v4 API:
+
+  + :class:`gitlab.v4.objects.UserProject`
+  + :class:`gitlab.v4.objects.UserProjectManager`
+  + :attr:`gitlab.v4.objects.User.projects`
+
+* GitLab API: https://docs.gitlab.com/ee/api/projects.html#list-user-projects
+
+List visible projects owned by the user::
```
nejch, line 209: Feels to me like the upstream docs are a bit contrived there, could we directly state it's only projects in the namespace and not those owned by the user?
 
```suggestion
List visible projects in the user's namespace::
```

JohnVillalovos: Sure!  Done.

docs/gl_objects/users.rst
```diff
@@ -218,6 +243,10 @@ List only direct group memberships::
 
     memberships = user.memberships.list(type='Namespace')
 
+.. note::
+
+    Administrator access is required.
```
nejch, line 248: Sorry, just noticed - I like the sentence here, but just to be consistent with the other notes about admin access:

```suggestion
    This endpoint requires admin access.
```

JohnVillalovos: Done 😊🔥

PR #2126, opened 2022-07-05
docs: update return type of pushrules

JohnVillalovos:
Update the return type of pushrules to surround None with back-ticks
to make it code-formatted.

### DIFF
--- a/docs/gl_objects/groups.rst
+++ b/docs/gl_objects/groups.rst
@@ -398,7 +398,7 @@ Create group push rules (at least one rule is necessary)::
 
     group.pushrules.create({'deny_delete_tag': True})
 
-Get group push rules (returns None is there are no push rules)::
+Get group push rules (returns ``None`` if there are no push rules)::
 
     pr = group.pushrules.get()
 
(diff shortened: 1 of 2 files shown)


docs/gl_objects/groups.rst
```diff
@@ -398,7 +398,7 @@ Create group push rules (at least one rule is necessary)::
 
     group.pushrules.create({'deny_delete_tag': True})
 
-Get group push rules (returns None is there are no push rules)::
+Get group push rules (returns ``None`` is there are no push rules)::
```
nejch, line 401:
```suggestion
Get group push rules (returns ``None`` if there are no push rules)::
```

docs/gl_objects/projects.rst
```diff
@@ -748,7 +748,7 @@ Create project push rules (at least one rule is necessary)::
 
     project.pushrules.create({'deny_delete_tag': True})
 
-Get project push rules (returns None is there are no push rules)::
+Get project push rules (returns ``None`` is there are no push rules)::
```
nejch, line 751:
```suggestion
Get project push rules (returns ``None`` if there are no push rules)::
```

nejch (changes requested):
Thanks @JohnVillalovos just a tiny suggestion while we're making changes here :)

docs/gl_objects/groups.rst
JohnVillalovos: Good catch. Will do.

nejch (approved):

PR #2127, opened 2022-07-05
docs: describe ROPC flow in place of password authentication

### LINKED ISSUE #2122: Need documented procedure for achieving an authenticated session

Can't get authenticated access using the code snippet attached to issue #380.

Please add to the documentation a Python code snippet showing how to achieve an authenticated session.   Note, authentication from within CI/CD is not sufficient, I need to be able to write an extractor that can analyze issue boards.

## Expected Behavior

Expected to achive Gitlab session with authenticated access

## Actual Behavior

Instead of a sign-in page, I got the Cloudflare "Checking your browser" page with no authentication token.

## Specifications

  - python-gitlab version: 3.6.0

nejch:
Closes #2122. I was quite blunt about not using password flows here, I think we shouldn't insist on doing this in 2022, so I also removed any usage from the docs.

### DIFF
--- a/docs/api-usage.rst
+++ b/docs/api-usage.rst
@@ -9,6 +9,15 @@ python-gitlab only supports GitLab API v4.
 
 To connect to GitLab.com or another GitLab instance, create a ``gitlab.Gitlab`` object:
 
+.. hint::
+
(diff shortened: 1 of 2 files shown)


JohnVillalovos:
Thanks a lot @nejch !

Issue #2135, opened 2022-07-09
Incorrect list filters in `gitlab project-merge-request-approval-rule list`


nejch:
### Discussed in https://github.com/python-gitlab/python-gitlab/discussions/2132

<div type='discussions-op-text'>

<sup>Originally posted by **MaximeCheramy** July  7, 2022</sup>
The GitLab API doesn't seem to have a name option: https://docs.gitlab.com/ee/api/merge_request_approvals.html#get-merge-request-level-rules</div>

PR #2136, opened 2022-07-09
fix(cli): remove irrelevant MR approval rule list filters

nejch:
Closes https://github.com/python-gitlab/python-gitlab/issues/2135


JohnVillalovos (approved):
Thanks @nejch

PR #2146, opened 2022-07-15
chore: enable mypy check `strict_equality`

JohnVillalovos:
Enable the `mypy` `strict_equality` check.

### DIFF
--- a/gitlab/v4/objects/users.py
+++ b/gitlab/v4/objects/users.py
@@ -3,7 +3,7 @@ GitLab API:
 https://docs.gitlab.com/ee/api/users.html
 https://docs.gitlab.com/ee/api/projects.html#list-projects-starred-by-a-user
 """
-from typing import Any, cast, Dict, List, Union
+from typing import Any, cast, Dict, List, Optional, Union
 
 import requests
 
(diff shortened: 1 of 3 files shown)


JohnVillalovos:
@nejch I am thinking it might be better to not return anything from `block` and `unblock`. As they are supposed to raise Exceptions if something goes wrong. The return values are undocumented. So am leaning towards just returning `None`. What do you think?

nejch:
> @nejch I am thinking it might be better to not return anything from `block` and `unblock`. As they are supposed to raise Exceptions if something goes wrong. The return values are undocumented. So am leaning towards just returning `None`. What do you think?

I'm thinking it might be best to just return whatever the server returns. This way we don't back ourselves into a corner which seems to have happened with the old `lint()` method that no longer returns all the fields the server provides:

https://github.com/python-gitlab/python-gitlab/blob/ed110bd131fd330e20ec55915b9452dbf8002e0b/gitlab/client.py#L407-L411

This way also any breaking changes are only because of GitLab and not us. WDYT?

JohnVillalovos:
> I'm thinking it might be best to just return whatever the server returns. This way we don't back ourselves into a corner which seems to have happened with the old `lint()` method that no longer returns all the fields the server provides:
> 
> This way also any breaking changes are only because of GitLab and not us. WDYT?

Works for me. I have updated it.

Thanks for the review.

Issue #1978, opened 2022-04-20
Warn user when detecting misconfigured server `external_url`?


nejch:
As seen in https://github.com/python-gitlab/python-gitlab/issues/1977, https://github.com/python-gitlab/python-gitlab/issues/1934 and older issues, paginated `list()` calls will fail when the server's external URL is not configured properly because we trust the `links` headers returned by the GitLab server:

https://github.com/python-gitlab/python-gitlab/blob/d27cc6a1219143f78aad7e063672c7442e15672e/gitlab/client.py#L1060

Maybe we should warn (or fail explicitly) if we detect the base url differs from the URL advertised back by the server?

(Just to avoid new issues popping up every now and then)

JohnVillalovos:
Sounds like a good idea. Maybe create an FAQ item to point to.

nejch:
User error in URL config could also potentially be detected by this, which is what seems to have happened in https://github.com/python-gitlab/python-gitlab/issues/2130 where a gitlab instance is served in a subdirectory on the domain.

PR #2149, opened 2022-07-17
fix: optionally keep user-provided base URL for pagination

iomarmochtar:
# Background

To enhance HTTP security our Gitlab is fronted by [IAP](https://cloud.google.com/iap) for example `https://gitlab.tools.domain.io` but for internal communication from runner to Gitlab API we create another endpoint that protected by firewall rules so then it only can be accessed from certain IP only, for example `https://runner.gitlab.tools.domain.io/` it actually using the reverse proxy that in the upstream destination will set `Host` header to the original one (`gitlab.tools.domain.io`). 

But when we run a python script inside Gitlab pipeline to the endpoint (`runner.gitlab.tools.domain.io`) and pass argument `all=True` or `iterator=True` [for loop all the data to all page](https://python-gitlab.readthedocs.io/en/stable/api-usage.html#pagination) will returning the base url as `gitlab.tools.domain.io` whereas it will causing the error due the request will be blocked by IAP.

# Expected

The base url will be persist same as the first time set in the next url so the request will be expected goes as the first page when using pagination iterrator (`all=True`).

# Fix

This MR will make sure the base in next url is the same. It's been tested in my case and working as expected.

### DIFF
--- a/gitlab/client.py
+++ b/gitlab/client.py
@@ -66,7 +66,8 @@ class Gitlab:
         user_agent: A custom user agent to use for making HTTP requests.
         retry_transient_errors: Whether to retry after 500, 502, 503, 504
             or 52x responses. Defaults to False.
-        persist_base_url: reconstruct next url if found not same as user provided url
+        keep_base_url: keep user-provided base URL for pagination if it
+            differs from response headers
     """
 
     def __init__(
@@ -86,7 +87,7 @@ class Gitlab:
         order_by: Optional[str] = None,
         user_agent: str = gitlab.const.USER_AGENT,
         retry_transient_errors: bool = False,
-        persist_base_url: bool = False,
+        keep_base_url: bool = False,
     ) -> None:
 
         self._api_version = str(api_version)
@@ -97,7 +98,7 @@ class Gitlab:
         #: Timeout to use for requests to gitlab server
         self.timeout = timeout
         self.retry_transient_errors = retry_transient_errors
-        self.persist_base_url = persist_base_url
+        self.keep_base_url = keep_base_url
         #: Headers that will be used in request to GitLab
         self.headers = {"User-Agent": user_agent}
 
@@ -1143,19 +1144,20 @@ class GitlabList:
                 search_api_url = re.search(r"(^.*?/api)", next_url)
                 if search_api_url:
                     next_api_url = search_api_url.group(1)
-                    if self._gl.persist_base_url:
+                    if self._gl.keep_base_url:
                         next_url = next_url.replace(
                             next_api_url, f"{self._gl._base_url}/api"
                         )
                     else:
                         utils.warn(
                             message=(
-                                f"The base url of the returned next page got "
-                                f"different with the user provided "
-                                f"{self._gl.url}/api* ~> {next_api_url}*, "
-                                f"since this may can lead to unexpected behaviour. "
-                                f"set argument persist_base_url to True for "
-                                f"resonctruct it with the origin one."
+                                f"The base URL in the server response"
+                                f"differs from the user-provided base URL "
+                                f"({self._gl.url}/api/ -> {next_api_url}/). "
+                                f"This may lead to unexpected behavior and  "
+                                f"broken pagination. Use `keep_base_url=True` "
+                                f"when initializing the Gitlab instance "
+                                f"to follow the user-provided base URL."
                             ),
                             category=UserWarning,
                         )
--- a/gitlab/client.py
+++ b/gitlab/client.py
@@ -66,6 +66,7 @@ class Gitlab:
         user_agent: A custom user agent to use for making HTTP requests.
         retry_transient_errors: Whether to retry after 500, 502, 503, 504
             or 52x responses. Defaults to False.
+        persist_base_url: reconstruct next url if found not same as user provided url
     """
 
     def __init__(
@@ -85,6 +86,7 @@ class Gitlab:
         order_by: Optional[str] = None,
         user_agent: str = gitlab.const.USER_AGENT,
         retry_transient_errors: bool = False,
+        persist_base_url: bool = False,
     ) -> None:
 
         self._api_version = str(api_version)
@@ -95,6 +97,7 @@ class Gitlab:
         #: Timeout to use for requests to gitlab server
         self.timeout = timeout
         self.retry_transient_errors = retry_transient_errors
+        self.persist_base_url = persist_base_url
         #: Headers that will be used in request to GitLab
         self.headers = {"User-Agent": user_agent}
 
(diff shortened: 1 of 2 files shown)


JohnVillalovos:
I wonder if this would be better to be a config option? As probably the most likely case when this occurs is that something is wrong with their config.

gitlab/client.py
```diff
@@ -1110,6 +1111,8 @@ def _query(
                 next_url = requests.utils.parse_header_links(result.headers["links"])[
                     0
                 ]["url"]
+            if not next_url.startswith(self._gl._base_url):
```
JohnVillalovos, line 1114: Since I'm not as smart as most people, I would appreciate a comment which explains why this is here and what it is doing.

iomarmochtar:
> I wonder if this would be better to be a config option? As probably the most likely case when this occurs is that something is wrong with their config.
do you mean by add as argument in `class Gitlab`, eg: `keep_base_url=False` so whether they have the same case by just set as `True` ?

iomarmochtar:
> I wonder if this would be better to be a config option? As probably the most likely case when this occurs is that something is wrong with their config.

do you mean by set the config as parameter in `class Gitlab`, eg: `keep_base_url=True`, once they have the same case then just set it as True ?

nejch:
Thanks for the work here @iomarmochtar! Have a look at https://github.com/python-gitlab/python-gitlab/issues/1978 as well, which covers mostly the same topic.

I agree with John we should at least not completely blindly follow either the server or the client-provided URL when there is a mismatch, hence my proposal in the issue above: I think we should issue a warning instead, unless explicitly configured to follow the base URL, and only then reconstruct the URL. Could we add that here? :bow: Thanks again!

---

For context: I actually would consider this approach a misconfiguration of the `external_url`, at least until GitLab supports multiple urls. E.g. the way your admins have configured it does not really scale - if your devs use other tech you'll have to implement this in the [node gitbeaker library](https://github.com/jdalrymple/gitbeaker/blob/44cb1f379b660b61042e9b6400749d3384198cdd/packages/core/src/infrastructure/RequestHelper.ts#L121), in the [terraform GitLab provider](https://github.com/gitlabhq/terraform-provider-gitlab/blob/5bc59f0fe470c15f17de0dd950f3ef876cacf12e/internal/provider/data_source_gitlab_group_membership.go#L151), and other libraries/apps, since most of these follow the link headers.

Instead I think they should use `clone_url` in the runner config (which was designed for this purpose), and if needed have additional proxying done in front of specific runner endpoints. Because this also affects other aspects including many other types of links that GitLab returns in its API responses.

nejch:
@iomarmochtar would you still like to work on this as an optional parameter as discussed above?

iomarmochtar:
thanks you for remind me @nejch ,almost forgot to continue it due office task. let me try in this weekend.

iomarmochtar:
push as requested, please review it @nejch @JohnVillalovos

tests/unit/test_gitlab.py
```diff
@@ -353,3 +353,60 @@ def test_gitlab_plain_const_does_not_warn(recwarn):
...
+        ),
+        # persist the base url
+        (
+            {"persist_base_url": True},
+            "<http://orig_host/api/v4/tests?per_page=1&page=2>;" ' rel="next"',
+            "http://localhost/api/v4/tests?per_page=1&page=2",
+            False,
+        ),
+    ],
+)
```
nejch, line 384: Thanks for adding the tests @iomarmochtar! I probably would have added a separate explicit test for warnings but this works too. Instead of adding comments, we can codify each parameter into `ids` and they will also get more human-readable test names displayed at runtime during test execution/failure:

```suggestion
@pytest.mark.parametrize(
    "kwargs,link_header,expected_next_url,show_warning",
    [
        (
            {},
            "<http://localhost/api/v4/tests?per_page=1&page=2>;" ' rel="next"',
            "http://localhost/api/v4/tests?per_page=1&page=2",
            False,
        ),
        (
            {},
            "<http://orig_host/api/v4/tests?per_page=1&page=2>;" ' rel="next"',
            "http://orig_host/api/v4/tests?per_page=1&page=2",
            True,
        ),
        (
            {"persist_base_url": True},
            "<http://orig_host/api/v4/tests?per_page=1&page=2>;" ' rel="next"',
            "http://localhost/api/v4/tests?per_page=1&page=2",
            False,
        ),
    ],
    ids=["url-match-does-not-warn", "url-mismatch-warns", "url-mismatch-keeps-url"]
)
```

gitlab/client.py
```diff
@@ -65,6 +66,7 @@ class Gitlab:
         user_agent: A custom user agent to use for making HTTP requests.
         retry_transient_errors: Whether to retry after 500, 502, 503, 504
             or 52x responses. Defaults to False.
+        persist_base_url: reconstruct next url if found not same as user provided url
```
nejch, line 69: I like your original wording of `keep base url`. Should we use that? It's shorter :) 

Maybe we can also a bit more specific about where it's used (might not fit in one line, just adjust accordingly if needed):

```suggestion
        keep_base_url: keep user-provided base URL for pagination if it differs from response headers
```

gitlab/client.py
```diff
@@ -1130,6 +1134,29 @@ def _query(
...
+                            message=(
+                                f"The base url of the returned next page got "
+                                f"different with the user provided "
+                                f"{self._gl.url}/api* ~> {next_api_url}*, "
+                                f"since this may can lead to unexpected behaviour. "
+                                f"set argument persist_base_url to True for "
+                                f"resonctruct it with the origin one."
+                            ),
+                            category=UserWarning,
+                        )
```
nejch, line 1159: This gets us into quite a bit of nesting here and a big chunk inside the `try` block. We could probably avoid some with the use of `continue`, but if it's not feasible here just dismiss it and we can refactor later.

gitlab/client.py
```diff
@@ -1130,6 +1134,29 @@ def _query(
...
+                    else:
+                        utils.warn(
+                            message=(
+                                f"The base url of the returned next page got "
+                                f"different with the user provided "
+                                f"{self._gl.url}/api* ~> {next_api_url}*, "
+                                f"since this may can lead to unexpected behaviour. "
+                                f"set argument persist_base_url to True for "
+                                f"resonctruct it with the origin one."
```
nejch, line 1156: Maybe:

```suggestion
                                f"The base URL in the server response"
                                f"differs from the user-provided base URL "
                                f"({self._gl.url}/api/ -> {next_api_url}/). "
                                f"This may lead to unexpected behavior and  "
                                f"broken pagination. Use `keep_base_url=True` "
                                f"when initializing the Gitlab instance "
                                f"to follow the user-provided base URL."
```

nejch (changes requested):
Thanks again @iomarmochtar. I have a few small comments

gitlab/client.py
iomarmochtar: this is the result of some warning from linter, it was shorter before. but yeah we can refactor it later.

tests/unit/test_gitlab.py
iomarmochtar: done, set as suggested

gitlab/client.py
iomarmochtar: done, set as suggested

gitlab/client.py
iomarmochtar: done, set as suggested

nejch:
Thanks @iomarmochtar. I have some ideas to reuse this elsewhere that I think are outside the scope of this PR, so I'll merge this as is, and I will do a little refactor after.

PR #2157, opened 2022-07-21
chore: enable mypy check `warn_return_any`

JohnVillalovos:
Update code so that the `warn_return_any` check passes.

### DIFF
--- a/gitlab/base.py
+++ b/gitlab/base.py
@@ -21,7 +21,7 @@ import json
 import pprint
 import textwrap
 from types import ModuleType
-from typing import Any, Dict, Iterable, Optional, Type, Union
+from typing import Any, Dict, Iterable, Optional, Type, TYPE_CHECKING, Union
 
 import gitlab
 from gitlab import types as g_types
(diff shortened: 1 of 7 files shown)


gitlab/base.py
```diff
@@ -245,14 +245,22 @@ def get_id(self) -> Optional[Union[int, str]]:
         """Returns the id of the resource."""
         if self._id_attr is None or not hasattr(self, self._id_attr):
             return None
-        return getattr(self, self._id_attr)
+        id_val = getattr(self, self._id_attr)
+        if id_val is None:
+            return None
+        if TYPE_CHECKING:
+            assert isinstance(id_val, (int, str))
+        return id_val
```
nejch, line 253: I wonder if mypy forcing us to do these kinds of things is showing some underlying issues and we should refactor our code. WDYT? Just thinking about what strict mypy is doing to our runtime code. I haven't explored in more detail though. Or maybe there's some magic in the `typing` that'd make mypy understand a bit better. I guess what we're looking for is something like "typing overloads based on the type of instance attributes"

E.g. I didn't think I'd ever see this kind of code:
```python
        if id_val is None:
            return None
```

JohnVillalovos: I simplified it

PR #2161, opened 2022-07-21
feat: add support for filtering jobs by scope

nickbroon:
See: 'scope' here: https://docs.gitlab.com/ee/api/jobs.html#list-project-jobs


nejch:
Thanks a lot for catching this @nickbroon, sometimes the CLI gets left behind a bit :)

Issue #2172, opened 2022-07-23
Unit tests are failing locally


JohnVillalovos:
On my local system `tox -e py3` is now failing:

```
___________________________________________________________________________________________________________________ test_gitlab_from_config_without_files_raises ____________________________________________________________________________________________________________________

    def test_gitlab_from_config_without_files_raises():
        with pytest.raises(GitlabConfigMissingError, match="non-existing"):
>           gitlab.Gitlab.from_config("non-existing")

tests/unit/test_gitlab.py:305:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
gitlab/client.py:257: in from_config
    config = gitlab.config.GitlabConfigParser(
gitlab/config.py:129: in __init__
    self._parse_config()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <gitlab.config.GitlabConfigParser object at 0x7f44fc519060>

    def _parse_config(self) -> None:
        _config = configparser.ConfigParser()
        _config.read(self._files)

        if self.gitlab_id and not _config.has_section(self.gitlab_id):
>           raise GitlabDataError(
                f"A gitlab id was provided ({self.gitlab_id}) "
                "but no config section found"
            )
[...]

PR #2173, opened 2022-07-23
test: fix broken test if user had config files

JohnVillalovos:
Use `monkeypatch` to ensure that no config files are reported for the
test.

Closes: #2172

### DIFF
--- a/tests/unit/test_gitlab.py
+++ b/tests/unit/test_gitlab.py
@@ -20,6 +20,7 @@ import copy
 import logging
 import pickle
 from http.client import HTTPConnection
+from typing import List, Optional, Union
 
 import pytest
 import responses
@@ -300,7 +301,11 @@ def test_gitlab_from_config(default_config):
     gitlab.Gitlab.from_config("one", [config_path])
 
 
-def test_gitlab_from_config_without_files_raises():
+def test_gitlab_from_config_without_files_raises(monkeypatch):
+    def no_files(config_files: Optional[List[str]] = None) -> Union[str, List[str]]:
+        return []
+
+    monkeypatch.setattr(gitlab.config, "_get_config_files", no_files)
     with pytest.raises(GitlabConfigMissingError, match="non-existing"):
         gitlab.Gitlab.from_config("non-existing")
 


JohnVillalovos:
PR #2174 demonstrates the issue.

nejch (changes requested):
Thanks for the catch @JohnVillalovos! We already have some scaffolding for mocking/prepping a clean config environment but perhaps it was not broad enough in where we placed it. It seems we should have a clean environment for _all tests_ by adding that to the root `tests/conftest.py` e.g.:

1. Move this `mock_clean_env` fixture to the root conftest with `autouse=True` - no test _ever_ should break from a user-supplied `PYTHON_GITLAB_CFG`:
https://github.com/python-gitlab/python-gitlab/blob/f0152dc3cc9a42aa4dc3c0014b4c29381e9b39d6/tests/unit/test_config.py#L132-L134

2. Move more logic to that fixture that we already use to ensure an empty list of default files - same here, no test _ever_ should be affected by presence of local files:
https://github.com/python-gitlab/python-gitlab/blob/f0152dc3cc9a42aa4dc3c0014b4c29381e9b39d6/tests/unit/test_config.py#L144-L145

3. In any test that does specifically need that, override it again if needed (I think we mostly don't, as we explicitly define things) like we already do here:
https://github.com/python-gitlab/python-gitlab/blob/f0152dc3cc9a42aa4dc3c0014b4c29381e9b39d6/tests/unit/test_config.py#L138

More approaches are discussed here https://stackoverflow.com/questions/38748257/disable-autouse-fixtures-on-specific-pytest-marks in case the simple override is not enough.

WDYT? This would then also be more local/CI-agnostic and consistent across all runs and we can stop worrying about config even before we hit CI.

JohnVillalovos:
That sounds reasonable but I can't help but think this should be merged now and the improvements you mentioned be done afterwards. As at the moment tests are broken for anyone who has a config.

I was going to work on arrays tomorrow and then could try to take a stab at your ideas.

nejch:
Sounds good also! I'll merge this for now.

PR #2174, opened 2022-07-23
test(ci): create a dummy ~/.python-gitlab.cfg file

JohnVillalovos:
Create a working ~/.python-gitlab.cfg file when running the unit
tests in the CI. This is to ensure our unit tests still work if a config file
exists.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -52,6 +52,8 @@ jobs:
           python-version: ${{ matrix.python.version }}
       - name: Install dependencies
         run: pip3 install tox pytest-github-actions-annotate-failures
+      - name: Setup dummy config
+        run: tests/ci_tools/setup-dummy-config.sh
       - name: Run tests
         env:
           TOXENV: ${{ matrix.python.toxenv }}
--- a/tests/ci_tools/setup-dummy-config.sh
+++ b/tests/ci_tools/setup-dummy-config.sh
@@ -1,0 +1,29 @@
+#!/bin/bash
+
+set -u
+
+if [[ -z "${CI:-}" || -z "${GITHUB_ACTIONS:-}" ]]; then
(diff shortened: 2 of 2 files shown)


JohnVillalovos:
Once PR #2173 is merged, then this will pass the CI.

nejch:
We have since added https://github.com/python-gitlab/python-gitlab/pull/2187 where we ignore default files including /etc/python-gitlab.cfg now at root-level.

JohnVillalovos:
> We have since added #2187 where we ignore default files including /etc/python-gitlab.cfg now at root-level.

True. But what if in the future we break it accidentally...

I don't have that strong of feelings about this but just thought a little extra safety.

JohnVillalovos:
It's fine with me if you want to go ahead and just close it 👍

nejch:
I'm still thinking about what to do with it :D for me not having extra bash scripts would be nicer*, as it usually doesn't make sense to lint them and other contributors would then get comfortable and start adding more and more potentially, but I also get your point. Maybe also a `tests/conftest.py` fixture that is autoused and always first would be an option but probably also overkill.

*I'm biased because I removed a bunch of them https://github.com/python-gitlab/python-gitlab/tree/v2.2.0/tools :D

nejch:
I think this could also be implemented as a top-level fixture, to keep things in one language and less code.

`tests/conftest.py`:

```python
import os
from pathlib import Path

def dummy_config() -> None:
    """
    Creates a dummy config file that should never affect our tests.
    See https://github.com/python-gitlab/python-gitlab/issues/2172.
    """
    config = Path.home() / ".python-gitlab.cfg"
    if config.exists() or "CI" not in os.environ:
        return

    config.write_text("""[global]
default = gitlab

[gitlab]
url = https://gitlab.example.com
private_token = not-a-valid-token
""")

Issue #1972, opened 2022-04-10
Get existence of a namespace


AliMirlou:
## Description of the problem, including code/CLI snippet

This is a feature request.

There is an endpoint in the Gitlab API for checking the existence of a namespace but it's not implemented in the Python client.
I know that I can check the existence by using the `get` method of namespace manager but it would be more efficient to have a `exists` method as well.

## Specifications

  - python-gitlab version: 3.2.0
  - API version you are using (v3/v4): v4
  - Gitlab server version (or gitlab.com): 14.7.0

nejch:
@max-wittig it seems there is a GET endpoint for this now in the upstream API, I wasn't aware of it either but I think this is what they're referring to:
https://docs.gitlab.com/ee/api/namespaces.html#get-existence-of-a-namespace

max-wittig:
Ahh nice. Sorry, then it's fine.

PR #2175, opened 2022-07-23
feat(namespaces): add support for namespace existence API

nejch:
Closes https://github.com/python-gitlab/python-gitlab/issues/1972

### DIFF
--- a/gitlab/v4/objects/namespaces.py
+++ b/gitlab/v4/objects/namespaces.py
@@ -1,7 +1,10 @@
-from typing import Any, cast, Union
+from typing import Any, cast, TYPE_CHECKING, Union
 
+from gitlab import cli
+from gitlab import exceptions as exc
 from gitlab.base import RESTManager, RESTObject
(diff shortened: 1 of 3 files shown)


gitlab/v4/objects/namespaces.py
```diff
@@ -20,3 +23,25 @@ class NamespaceManager(RetrieveMixin, RESTManager):
 
     def get(self, id: Union[str, int], lazy: bool = False, **kwargs: Any) -> Namespace:
         return cast(Namespace, super().get(id=id, lazy=lazy, **kwargs))
+
+    @cli.register_custom_action("NamespaceManager", ("namespace", "parent_id"))
```
nejch, line 27: I only defined the optionals for CLI help, not in explicit args, so we don't have to manually maintain them. Might be a little different from older custom methods but I think this may be easier long-term.

JohnVillalovos (approved):

Issue #2177, opened 2022-07-24
Make tests so that will not be impacted by any existing config files


JohnVillalovos:
Capturing the comments that were left by @nejch in #2173

Thanks for the catch @JohnVillalovos! We already have some scaffolding for mocking/prepping a clean config environment but perhaps it was not broad enough in where we placed it. It seems we should have a clean environment for _all tests_ by adding that to the root `tests/conftest.py` e.g.:

1. Move this `mock_clean_env` fixture to the root conftest with `autouse=True` - no test _ever_ should break from a user-supplied `PYTHON_GITLAB_CFG`:
https://github.com/python-gitlab/python-gitlab/blob/f0152dc3cc9a42aa4dc3c0014b4c29381e9b39d6/tests/unit/test_config.py#L132-L134

2. Move more logic to that fixture that we already use to ensure an empty list of default files - same here, no test _ever_ should be affected by presence of local files:
https://github.com/python-gitlab/python-gitlab/blob/f0152dc3cc9a42aa4dc3c0014b4c29381e9b39d6/tests/unit/test_config.py#L144-L145

3. In any test that does specifically need that, override it again if needed (I think we mostly don't, as we explicitly define things) like we already do here:
https://github.com/python-gitlab/python-gitlab/blob/f0152dc3cc9a42aa4dc3c0014b4c29381e9b39d6/tests/unit/test_config.py#L138

More approaches are discussed here https://stackoverflow.com/questions/38748257/disable-autouse-fixtures-on-specific-pytest-marks in case the simple override is not enough.

WDYT? This would then also be more local/CI-agnostic and consistent across all runs and we can stop worrying about config even before we hit CI.

PR #2187, opened 2022-07-26
test: always ensure clean config environment

nejch:
Closes https://github.com/python-gitlab/python-gitlab/issues/2177

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,9 +1,27 @@
 import pytest
 
+import gitlab
+
 
 @pytest.fixture(scope="session")
 def test_dir(pytestconfig):
     return pytestconfig.rootdir / "tests"
+
+
+@pytest.fixture(autouse=True)
+def mock_clean_config(monkeypatch):
+    """Ensures user-defined environment variables do not interfere with tests."""
+    monkeypatch.delenv("PYTHON_GITLAB_CFG", raising=False)
+    monkeypatch.delenv("GITLAB_PRIVATE_TOKEN", raising=False)
+    monkeypatch.delenv("GITLAB_URL", raising=False)
+    monkeypatch.delenv("CI_JOB_TOKEN", raising=False)
+    monkeypatch.delenv("CI_SERVER_URL", raising=False)
+
+
+@pytest.fixture(autouse=True)
+def default_files(monkeypatch):
+    """Ensures user configuration files do not interfere with tests."""
+    monkeypatch.setattr(gitlab.config, "_DEFAULT_FILES", [])
 
 
 @pytest.fixture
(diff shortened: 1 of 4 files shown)


tests/conftest.py
```diff
@@ -1,11 +1,29 @@
...
 
 
+@pytest.fixture(autouse=True)
+def mock_clean_config(monkeypatch):
+    """Ensures user-defined environment variables do not interfere with tests."""
+    monkeypatch.delenv("PYTHON_GITLAB_CFG", raising=False)
+    monkeypatch.delenv("GITLAB_PRIVATE_TOKEN", raising=False)
+    monkeypatch.delenv("GITLAB_URL", raising=False)
+    monkeypatch.delenv("CI_JOB_TOKEN", raising=False)
+    monkeypatch.delenv("CI_SERVER_URL", raising=False)
```
nejch, line 18: Not really needed because of tox, but thought I'd add it anyway since we were doing it in other parts.

tests/unit/test_config.py
```diff
@@ -100,6 +100,12 @@
 """
 
 
+@pytest.fixture()
+def default_files(monkeypatch):
+    """Overrides mocked default files from conftest.py as we have our own mocks here."""
+    monkeypatch.setattr(gitlab.config, "_DEFAULT_FILES", config._DEFAULT_FILES)
+
```
JohnVillalovos, line 107: Why is this function in this file? I'm confused how it is used.

JohnVillalovos (commented):
Thanks for doing this!

Left one question for something I didn't understand.

tests/unit/test_config.py
nejch: The global conftest now removes all `_DEFAULT_FILES` so the default behavior in tests is that it won't find any of these files.

In `test_config.py` we have tests that expect `_DEFAULT_FILES` to match our production list, and have their own mocks that mock opening these files instead. So by reverting to this list in the module we let the existing test mocks take care of it. The override works kind of like python's LEGB for variables, the module level fixture overrides the root one.

Maybe a better way long-term would be to completely rework the tests in this module to mock things differently, but this resulted in the smallest diff.

JohnVillalovos: Okay. Seems to be because it is being overridden: https://docs.pytest.org/en/latest/how-to/fixtures.html#overriding-fixtures-on-various-levels

Would be nice if `autouse=True` is in the function definition, for understand-ability. I realize it isn't needed.

WDYT?

nejch: Oh yes, that was actually supposed to be there :facepalm: added now. And yes, that links explains the override much better than my mini-novel :grin:

JohnVillalovos (approved):

JohnVillalovos:
Thanks @nejch !

PR #2188, opened 2022-07-26
test: attempt to make functional test startup more reliable

JohnVillalovos:
The functional tests have been erratic. Current theory is that we are
starting the tests before the GitLab container is fully up and
running.

  * Add checking of the Health Check[1] endpoints.
  * Add a 20 second delay after we believe it is up and running.
  * Increase timeout from 300 to 400 seconds

[1] https://docs.gitlab.com/ee/user/admin_area/monitoring/health_check.html

### DIFF
--- a/tests/functional/conftest.py
+++ b/tests/functional/conftest.py
@@ -1,20 +1,46 @@
 import logging
+import pathlib
 import tempfile
 import time
 import uuid
-from pathlib import Path
 from subprocess import check_output
 
 import pytest
+import requests
 
 import gitlab
 import gitlab.base
 from tests.functional import helpers
 
+SLEEP_TIME = 10
+
 
 @pytest.fixture(scope="session")
-def fixture_dir(test_dir):
+def fixture_dir(test_dir) -> pathlib.Path:
     return test_dir / "functional" / "fixtures"
+
+
+@pytest.fixture(scope="session")
+def gitlab_service_name() -> str:
+    """The "service" name is the one defined in the `docker-compose.yml` file"""
+    return "gitlab"
+
+
+@pytest.fixture(scope="session")
+def gitlab_container_name() -> str:
+    """The "container" name is the one defined in the `docker-compose.yml` file
+    for the "gitlab" service"""
+    return "gitlab-test"
+
+
+@pytest.fixture(scope="session")
+def gitlab_docker_port(docker_services, gitlab_service_name: str) -> int:
+    return docker_services.port_for(service=gitlab_service_name, container_port=80)
+
+
+@pytest.fixture(scope="session")
+def gitlab_url(docker_ip: str, gitlab_docker_port: int) -> str:
+    return f"http://{docker_ip}:{gitlab_docker_port}"
 
 
 def reset_gitlab(gl: gitlab.Gitlab) -> None:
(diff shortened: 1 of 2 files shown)


tests/functional/conftest.py
```diff
@@ -129,15 +132,30 @@ def check_is_alive():
     Return a healthcheck function fixture for the GitLab container spinup.
     """
 
-    def _check(container: str, start_time: float) -> bool:
+    def _check(
+        *, container: str, start_time: float, docker_ip: str, docker_port: int
```
nejch, line 136: We don't necessarily need to pass `docker_ip` and `docker_port` in here, we could directly use the `docker_ip` and `docker_services` fixtures in this fixture directly.

Then we again don't need to have the keywords discussion :rofl:

nejch: e.g. you could have
```python
@pytest.fixture
def gitlab_url(docker_ip, docker_services):
    return f"http://{docker_ip}:{docker_services.port_for('gitlab-test')}"
```

JohnVillalovos: Thanks. Good idea. I did that and even more fixtures 😊

tests/functional/conftest.py
```diff
@@ -167,31 +215,41 @@ def _wait(timeout=30, step=0.5):
...
+    fixture_dir: pathlib.Path,
+):
     config_file = temp_dir / "python-gitlab.cfg"
-    port = docker_services.port_for("gitlab", 80)
 
     start_time = time.perf_counter()
     logging.info("Waiting for GitLab container to become ready.")
     docker_services.wait_until_responsive(
-        timeout=300,
+        timeout=400,
```
nejch, line 231: From what I could tell in the logs, the issue was happening way before the timeout actually (e.g. already at 2.5 minutes or so), when the logs state reconfgiured. Maybe we could keep it at 300 and see how that goes? 300 seconds is already a sad amount of time for a service to start (:grin:) and maybe these new checks can improve our handling of startup.

tests/functional/conftest.py
```diff
@@ -1,22 +1,48 @@
...
 from subprocess import check_output
 
 import pytest
+import requests
 
 import gitlab
 import gitlab.base
 from tests.functional import helpers
 
+SLEEP_TIME = 20
```
nejch, line 15: Could we try with 10 sec initially? I've been using this simple check myself for another project and it seems to be ok:

```python
def service_ready(gitlab_url: str) -> None:
    for _ in GITLAB_MAX_ATTEMPTS:
        try:
            response = requests.get(f"{gitlab_url}/-/health")
            assert "GitLab OK" in getattr(response, "text", "")
        except (requests.ConnectionError, AssertionError):
            sleep(10)
            continue

        return

    raise TimeoutError("GitLab health checks timed out")
```

tests/functional/conftest.py
JohnVillalovos: Sure. Done. But I guess I don't see the issue with making it even 1,000 seconds 😆 Since if it doesn't come up in the timeout period then the job fails.

tests/functional/conftest.py
JohnVillalovos: Okay. I guess you like to live on the edge of danger! I prefer a larger margin of error...

tests/functional/conftest.py
nejch: Hehe I think I (at least used to) run these tests locally rather than in CI more often than you - basically I don't want people to get stuck in the dark (with an unhealthy container that's never getting out of 502 for whatever reason) for too long when running tests. So at some point we have to say "ok, something is really wrong, this container will never properly start up"

tests/functional/conftest.py
nejch: I'm probably biased as I spent too much time optimizing this from a hardcoded 300 seconds wait a while back :sweat_smile:

PR #2189, opened 2022-07-26
chore: revert "test(functional): simplify token creation"

nejch:
This reverts commit 67ab24fe5ae10a9f8cc9122b1a08848e8927635d.


nejch:
Seems to be going well, let's revert and I'll play with this approach at some later point.

PR #2194, opened 2022-07-27
test(functional): bump GitLab docker image to 15.2.0-ee.0

JohnVillalovos:
Use the GitLab docker image 15.2.0-ee.0 in the functional testing.


tests/functional/api/test_clusters.py
```diff
@@ -1,4 +1,10 @@
-def test_project_clusters(project):
+import pytest
+
+
+def test_project_clusters(project, gitlab_version):
+    if gitlab_version.major >= 15:
+        # https://gitlab.com/groups/gitlab-org/configure/-/epics/8
+        pytest.skip("Project clusters are deprecated")
```
nejch, line 7: Should we maybe remove these cluster (group + project) tests entirely? The cert-based integration was already deprecated in 14.5 so we probably don't need to "officially" test this anymore, since we don't run tests against older versions.

JohnVillalovos: Done

tests/functional/api/test_topics.py
```diff
@@ -4,11 +4,18 @@
...
-    topic = gl.topics.create({"name": "my-topic", "description": "My Topic"})
+    create_dict = {"name": "my-topic", "description": "My Topic"}
+    if gitlab_version.major >= 15:
+        create_dict["title"] = "my topic title"
+    topic = gl.topics.create(
+        {"name": "my-topic", "title": "my topic title", "description": "My Topic"}
+    )
     assert topic.name == "my-topic"
+    if gitlab_version.major >= 15:
+        assert topic.title == "my topic title"
```
nejch, line 18: FWIW I don't think we need to test multiple major versions as tests get complicated quickly, but I agree it'll be helpful having version checks when developing new features.

PR #2199, opened 2022-07-28
Support downloading archive subpaths

orf:
https://docs.gitlab.com/ee/api/repositories.html#get-file-archive

The archive endpoint supports a `path` argument. This change adds support for it.

### DIFF
--- a/gitlab/v4/objects/repositories.py
+++ b/gitlab/v4/objects/repositories.py
@@ -201,6 +201,7 @@ class RepositoryMixin(_RestObjectBase):
         action: Optional[Callable[..., Any]] = None,
         chunk_size: int = 1024,
         format: Optional[str] = None,
+        path: Optional[str] = None,
         *,
         iterator: bool = False,
         **kwargs: Any,
@@ -218,6 +219,7 @@ class RepositoryMixin(_RestObjectBase):
                 data
             chunk_size: Size of each chunk
             format: file format (tar.gz by default)
+            path: The subpath of the repository to download (all files by default)
             **kwargs: Extra options to send to the server (e.g. sudo)
 
         Raises:
@@ -227,14 +229,16 @@ class RepositoryMixin(_RestObjectBase):
         Returns:
             The binary data of the archive
         """
-        path = f"/projects/{self.encoded_id}/repository/archive"
+        url_path = f"/projects/{self.encoded_id}/repository/archive"
         if format:
-            path += "." + format
+            url_path += "." + format
         query_data = {}
         if sha:
             query_data["sha"] = sha
+        if path is not None:
+            query_data["path"] = path
         result = self.manager.gitlab.http_get(
-            path, query_data=query_data, raw=True, streamed=streamed, **kwargs
+            url_path, query_data=query_data, raw=True, streamed=streamed, **kwargs
         )
         if TYPE_CHECKING:
             assert isinstance(result, requests.Response)


gitlab/v4/objects/repositories.py
```diff
@@ -201,6 +201,7 @@ def repository_archive(
         action: Optional[Callable[..., Any]] = None,
         chunk_size: int = 1024,
         format: Optional[str] = None,
+        path: str = None,
```
JohnVillalovos, line 204:
```suggestion
        path: Optional[str] = None,
```

gitlab/v4/objects/repositories.py
```diff
@@ -233,6 +235,8 @@ def repository_archive(
         query_data = {}
         if sha:
             query_data["sha"] = sha
+        if path:
```
JohnVillalovos, line 238:
```suggestion
        if path is not None:
```

JohnVillalovos (changes requested):

JohnVillalovos (approved):

JohnVillalovos:
@orf Thanks for the contribution!

gitlab/v4/objects/repositories.py
```diff
@@ -233,6 +235,8 @@ def repository_archive(
         query_data = {}
         if sha:
             query_data["sha"] = sha
+        if path is not None:
```
JohnVillalovos, line 238: Will need to change the name of the other `path` variable on line 232 and on line 241.

JohnVillalovos: Maybe to `url_path`

JohnVillalovos (changes requested):
Looks like conflict with the pre-existing `path` variable.

nejch:
@JohnVillalovos for later: we could potentially also prefix these internal python-gitlab variables in the mixins and custom methods to always avoid clashes with GitLab API attributes. So this would be `_path` here for example. Not sure how ugly that would get.

@orf sometimes our variables happen to overlap with upstream API attributes hence the suggestion.

```diff
diff --git a/gitlab/v4/objects/repositories.py b/gitlab/v4/objects/repositories.py
index 8f99e5d..b8dbdd8 100644
--- a/gitlab/v4/objects/repositories.py
+++ b/gitlab/v4/objects/repositories.py
@@ -229,16 +229,16 @@ class RepositoryMixin(_RestObjectBase):
         Returns:
             The binary data of the archive
         """
-        path = f"/projects/{self.encoded_id}/repository/archive"
+        url_path = f"/projects/{self.encoded_id}/repository/archive"
         if format:
-            path += "." + format
+            url_path += "." + format
         query_data = {}
         if sha:
             query_data["sha"] = sha
         if path is not None:
             query_data["path"] = path
         result = self.manager.gitlab.http_get(
-            path, query_data=query_data, raw=True, streamed=streamed, **kwargs
+            url_path, query_data=query_data, raw=True, streamed=streamed, **kwargs
         )
         if TYPE_CHECKING:
             assert isinstance(result, requests.Response)
```

orf:
Thank you both! I've applied your suggestions 🙏

nejch (approved):

nejch:
I guess we can get this one in, I hope you don't mind @JohnVillalovos ;) 

@orf thanks again! I think I recognized your username from the loguru saga, fun popcorn material. :)

JohnVillalovos:
> I guess we can get this one in, I hope you don't mind @JohnVillalovos ;)

Thanks for reviewing it @nejch !

PR #2201, opened 2022-07-29
chore: make code PEP597 compliant

JohnVillalovos:
Use `encoding="utf-8"` in `open()` and open-like functions.

https://peps.python.org/pep-0597/

### DIFF
--- a/gitlab/config.py
+++ b/gitlab/config.py
@@ -135,7 +135,7 @@ class GitlabConfigParser:
 
     def _parse_config(self) -> None:
         _config = configparser.ConfigParser()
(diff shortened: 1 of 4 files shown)


nejch:
Thanks @JohnVillalovos!

Slightly related to this, I've seen some linters suggest to always use the standard `with open() ... as f:` pattern as suggested in https://docs.python.org/3/tutorial/inputoutput.html#reading-and-writing-files, to avoid arbitrary file descriptor names and shadowing `file`. Also discourages creating nested managers I assume. But I can't find an example now.

PR #2212, opened 2022-08-02
chore: remove broad Exception catching from `config.py`

JohnVillalovos:
Change "except Exception:" catching to more granular exceptions.

A step in enabling the "broad-except" check in pylint.

### DIFF
--- a/gitlab/config.py
+++ b/gitlab/config.py
@@ -17,6 +17,8 @@ HELPER_PREFIX = "helper:"
 
 HELPER_ATTRIBUTES = ["job_token", "http_password", "private_token", "oauth_token"]
 
+_CONFIG_PARSER_ERRORS = (configparser.NoOptionError, configparser.NoSectionError)
+
 
 def _resolve_file(filepath: Union[Path, str]) -> str:
     resolved = Path(filepath).resolve(strict=True)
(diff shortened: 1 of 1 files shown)


gitlab/config.py
```diff
@@ -165,108 +165,102 @@ def _parse_config(self) -> None:
...
-                pass
-        except Exception:
+            self.ssl_verify = _config.get(self.gitlab_id, "ssl_verify")
+        except (configparser.NoOptionError, configparser.NoSectionError):
             pass
 
         try:
             self.timeout = _config.getint("global", "timeout")
-        except Exception:
+        except (configparser.NoOptionError, configparser.NoSectionError):
```
nejch, line 183: We were probably silently catching a lot of `ValueError` and `TypeError` in these excepts so far, especially for `getboolean` and `getint`. Not sure if we should also handle that in here? Haven't tested with offending values yet :)

nejch:
Thanks @JohnVillalovos! Just 2 questions.

FWIW I recently realized we are rolling our own homebrew version of configparser (with all this try/except spaghetti) when python3's (and previously `SafeConfigParser`) parser actually handles section- vs. `default` defined options natively.

But probably not worth migrating that breaking change now that tomllib is coming to the stdlib and might be more attractivein the future to bring the config format in line with trends in the python world.

nejch:
Cool, I think we can get this in and double check if any new errors are thrown before the 28th :)

nejch (approved):

PR #2219, opened 2022-08-04
fix(client): ensure encoded query params are never duplicated 

### LINKED ISSUE #2218: URL-encoded kwargs are infinitely duplicated in query params (v3.7.0 regression for arrays)

With #1699 we added array params for query params in GET requests. But GitLab also returns that param, URL-encoded, in the `next` URL which we then follow in listing endpoints. Because one is URL-encoded and the other isn't, the param gets added at the end of the query again. With `list()` calls this happens over and over again and causes 502 due to super long query strings eventually.

minimal reproduction:

```python
import gitlab

gl = gitlab.Gitlab()
gl.enable_debug()

projects = gl.projects.list(topics=["python"], per_page=1, iterator=True)

for project in projects:
    # watch the `send` URL grow continuously until triggering 502
    pass

```

## Expected Behavior

If a kwarg/param (URL-encoded or not) that we provide is already in the provided URL, it should never be duplicated when we construct the final request URL.

## Actual Behavior

`http_request()` and requests can't tell if our kwargs match those already in the URL params and they get duplicated to oblivion, causing 502 once the URL query is too long for the server.

## Specifications

  - python-gitlab version: v3.7.0
  - Gitlab server version (or gitlab.com): all

nejch:
Closes #2218 

This provides a small generic fix at the http_request level instead of messing even more with our own flows.

(I've always thought it was a bit weird that GitLab returns a lot of boilerplate query params in ther `next` header links, so I think this will also make other things more robust in the future).

### DIFF
--- a/gitlab/client.py
+++ b/gitlab/client.py
@@ -20,6 +20,7 @@ import os
 import re
 import time
 from typing import Any, cast, Dict, List, Optional, Tuple, TYPE_CHECKING, Union
+from urllib import parse
 
 import requests
 import requests.utils
@@ -677,10 +678,14 @@ class Gitlab:
             GitlabHttpError: When the return code is not 2xx
         """
         query_data = query_data or {}
-        url = self._build_url(path)
+        raw_url = self._build_url(path)
 
-        params: Dict[str, Any] = {}
+        # parse user-provided URL params to ensure we don't add our own duplicates
+        parsed = parse.urlparse(raw_url)
+        params = parse.parse_qs(parsed.query)
         utils.copy_dict(src=query_data, dest=params)
+
+        url = raw_url.replace(parsed.query, "").strip("?")
 
         # Deal with kwargs: by default a user uses kwargs to send data to the
         # gitlab server, but this generates problems (python keyword conflicts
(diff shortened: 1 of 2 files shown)


gitlab/client.py
```diff
@@ -677,11 +678,15 @@ def http_request(
...
-        url = self._build_url(path)
+        raw_url = self._build_url(path)
 
-        params: Dict[str, Any] = {}
+        # parse user-provided URL params to ensure we don't add our own duplicates
+        parsed = parse.urlparse(raw_url)
+        params = parse.parse_qs(parsed.query)
         utils.copy_dict(src=query_data, dest=params)
 
+        url = raw_url.replace(parsed.query, "").strip("?")
```
JohnVillalovos, line 688: Why not use `parse.urlunparse`? 

```
url = parse.urlunparse(parsed._replace(query=""))
```

nejch: I saw that on stackoverflow as well but it uses a private method (`_replace`) so didn't want to touch that :grin:

JohnVillalovos: Oh it is documented in the official docs:

https://docs.python.org/3/library/urllib.parse.html

```
As is the case with all named tuples, the subclass has a few additional methods and attributes that are particularly useful. One such method is _replace(). The _replace() method will return a new ParseResult object replacing specified fields with new values.
```

PR #2221, opened 2022-08-04
chore: use `urlunparse` instead of string replace

JohnVillalovos:
Use the `urlunparse()` function to reconstruct the URL without the
query parameters.

### DIFF
--- a/gitlab/client.py
+++ b/gitlab/client.py
@@ -685,7 +685,7 @@ class Gitlab:
         params = parse.parse_qs(parsed.query)
         utils.copy_dict(src=query_data, dest=params)
 
-        url = raw_url.replace(parsed.query, "").strip("?")
+        url = parse.urlunparse(parsed._replace(query=""))
 
         # Deal with kwargs: by default a user uses kwargs to send data to the
         # gitlab server, but this generates problems (python keyword conflicts


nejch:
@JohnVillalovos this uses a private interface, see my comment at https://github.com/python-gitlab/python-gitlab/pull/2219#discussion_r937927383.

JohnVillalovos:
> @JohnVillalovos this uses a private interface, see my comment at [#2219 (comment)](https://github.com/python-gitlab/python-gitlab/pull/2219#discussion_r937927383).

It is documented in the official docs, which to me means it is supposed to be used.

https://docs.python.org/3/library/urllib.parse.html

> As is the case with all named tuples, the subclass has a few additional methods and attributes that are particularly useful. One such method is _replace(). The _replace() method will return a new ParseResult object replacing specified fields with new values.

JohnVillalovos:
My theory is that because the `urllib.parse.ParseResult` is a `namedtuple` they preface functions with underscores.

The `namedtuple` documentation sort-of mentions that: https://docs.python.org/3/library/collections.html#collections.namedtuple

> In addition to the methods inherited from tuples, named tuples support three additional methods and two attributes. To prevent conflicts with field names, the method and attribute names start with an underscore.

nejch:
Ok, I will begrudgingly merge this :cat:

PR #2222, opened 2022-08-04
feat(client): warn user on misconfigured URL in `auth()`

nejch:
Closes https://github.com/python-gitlab/python-gitlab/issues/1978.

3 steps here:
* https://github.com/python-gitlab/python-gitlab/commit/7b61ae6a561ca39399e3feedbe9aa2a701ca8a1f removes a hack to handle gitlab's non-standard headers. That was only valid for 13.0 and fixed in 13.1 - super old now and I wouldn't say we're expected to support that.
* https://github.com/python-gitlab/python-gitlab/commit/31c9c2f00f49500de9c74c2eb3f7f06036521755 refactors the nested code out into a slightly more generic helper (that's where it should live anyway, as most of the attributes are from the Gitlab instance)
* https://github.com/python-gitlab/python-gitlab/commit/d71c1dbe349dca044bf9e236c937b2d4d6923d06 adds it to `auth()` so the user gets a warning ASAP.

Other than that I haven't changed any of the `str`/`None` handling to keep the `next_url` code the same in `_query`.

### DIFF
--- a/gitlab/client.py
+++ b/gitlab/client.py
@@ -1130,13 +1130,8 @@ class GitlabList:
         query_data = query_data or {}
         result = self._gl.http_request("get", url, query_data=query_data, **kwargs)
         try:
-            links = result.links
-            if links:
-                next_url = links["next"]["url"]
-            else:
-                next_url = requests.utils.parse_header_links(result.headers["links"])[
-                    0
-                ]["url"]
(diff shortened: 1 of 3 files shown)


JohnVillalovos (approved):
Looks good. Thanks @nejch

Issue #658, opened 2018-12-09
Review ownership/licensing of the project


gpocentek:
My name is everywhere in the code, we need to update information to make the project belong to the python-gitlab team.

Things to do:
* [ ] update `gitlab/__init__.py` (`__author__`, `__email__`, `__copyright__`)
* [ ] review the files headers for copyright update

max-wittig:
@bufferoverflow What do you think? I would just close it. From my point of view, it's totally fine if `@gpocentek` is in the copyrights. He is the original creator.

bufferoverflow:
I'm unsure how author is handled and if there is e.g. a maintainer field within setup.py. 

https://pypi.org/project/python-gitlab/ shows the maintainers based on pypi access, maybe we can sync that with the content of setup.py

PR #2224, opened 2022-08-05
refactor(build): build project using PEP 621

nejch:
PEP 621 now has an authors & maintainers concept. So after #2223, this closes https://github.com/python-gitlab/python-gitlab/issues/658:

- [ ] Add John to GitHub.com `python-gitlab/dev` team
- [ ] Add John to GitLab.com `python-gitlab` group
- [ ] Add John and Nejc to pypi.org maintainers
- [ ] We have https://pypi.org/project/gitlab/ now, check if it's any use to us (https://pypi.org/project/pypi-alias/ etc)
- [ ] Check if it's ok to use `pypa/build` or do 4.0 so downstream package maintainers aren't surprised
- [ ] Wait for setuptools with PEP 660 and drop setup.cfg?


tox.ini
```diff
@@ -15,7 +15,7 @@ isolated_build = True
 deps = -r{toxinidir}/requirements.txt
        -r{toxinidir}/requirements-test.txt
 commands =
-  pytest tests/unit tests/meta {posargs}
+  pytest tests/unit {posargs}
```
JohnVillalovos, line 18: I really don't like this change. I want the meta tests to run when the standard `py3` tests run. Now I have one more tox target to remember to run.

Also I can easily see contributors failing to run the `meta` test manually and have things fail when it hits the CI.

tests/meta/test_dists.py
```diff
@@ -1,14 +1,13 @@
 import tarfile
 import zipfile
 from pathlib import Path
-from sys import version_info
+from subprocess import CompletedProcess, run
+from sys import executable, version_info
```
JohnVillalovos, line 5:
```
import subprocess
import sys
```

I prefer just importing the module. And doesn't seem like much more work to do `sys.executable` for example.

Google's style guide also recommends this, with a few exceptions:
https://google.github.io/styleguide/pyguide.html#22-imports

pyproject.toml
```diff
@@ -6,24 +66,7 @@ order_by_type = false
...
-no_implicit_optional = true
-no_implicit_reexport = true
-strict_equality = true
-warn_redundant_casts = true
-warn_return_any = true
-warn_unused_configs = true
-warn_unused_ignores = true
-
-# The following need to have changes made to be able to enable them:
-# disallow_untyped_calls = true
```
JohnVillalovos, line 26: Oh cool. The `types-requests` library is now updated.

JohnVillalovos (commented):
Overall I like this. But my biggest concern is removing the `meta` checks for the `py3` checks.

tests/meta/test_dists.py
nejch: I actually had that before for subprocess, but black mangles the formatting. I guess at least for typing purposes I'd like to have concise imports to not explode functions. But just skimming through that chapter, I'm not sure you'd like some of the rest of google's style recommendations on imports :grin:

tox.ini
nejch: @JohnVillalovos this PR isn't really fully ready for review. I will reopen it when it's done. I need to deduplicate the slow meta builds because they're now done in a virtualenv, then `meta` can be run in the standard `tox` suite. But I don't really see them as part of the unit tests, more like a custom mypy/pylint/isort job, hence the removal from the unit test target.

> Also I can easily see contributors failing to run the `meta` test manually and have things fail when it hits the CI.

I just forgot to add the meta target to the env list, I will add it when this is done.

PR #2230, opened 2022-08-09
docs(api): describe use of lower-level methods

nejch:
Continuation of #1846 and goes towards #1842.

### DIFF
--- a/docs/api-levels.rst
+++ b/docs/api-levels.rst
@@ -86,7 +86,7 @@ For example, if for whatever reason you want to fetch allowed methods for an end
 
     >>> gl = gitlab.Gitlab(private_token=private_token)
     >>>
-    >>> response = gl.http_request("OPTIONS", "/projects")
+    >>> response = gl.http_request(verb="OPTIONS", path="/projects")
     >>> response.headers["Allow"]
     'OPTIONS, GET, POST, HEAD'
 
@@ -94,6 +94,11 @@ Or get the total number of a user's events with a customized HEAD request:
 
 .. code-block:: python
 
-    >>> response = gl.http_request("HEAD", "/events", query_params={"sudo": "some-user"}, timeout=10)
+    >>> response = gl.http_request(
+            verb="HEAD",
+            path="/events",
+            query_params={"sudo": "some-user"},
+            timeout=10
+        )
     >>> response.headers["X-Total"]
     '123'
--- a/docs/api-levels.rst
+++ b/docs/api-levels.rst
@@ -1,0 +1,96 @@
+################
+Lower-level APIs
+################
+
+``python-gitlab``'s API levels provide different degrees of convenience, control and stability.
+
+Main interface - ``Gitlab``, managers and objects
+=================================================
+
+As shown in previous sections and examples, the high-level API interface wraps GitLab's API
+endpoints and makes them available from the ``Gitlab`` instance via managers that create
+objects you can manipulate.
+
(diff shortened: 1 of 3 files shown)


docs/api-levels.rst
```diff
@@ -0,0 +1,87 @@
...
+The available methods are:
+
+* ``http_get()``
+* ``http_post()``
+* ``http_put()``
+* ``http_delete()``
+* ``http_list()`` (a wrapper around ``http_get`` handling pagination, including with lazy generators)
+* ``http_head()`` (only returns the header dictionary)
+
+Lower-lower-level API - HTTP requests
```
nejch, line 64: I took this funny terminology from requests - https://requests.readthedocs.io/en/latest/api/#lower-lower-level-classes.
I thought people interested in this might already be familiar with requests docs.

docs/api-levels.rst
```diff
@@ -0,0 +1,87 @@
...
+=====================================
+
+.. important::
+
+    This is mostly intended for internal use in python-gitlab and may have a less stable interface than
+    higher-level APIs.
```
JohnVillalovos, line 70: It is highly recommended to use keyword arguments when calling the interfaces to lessen the chances of a change to the interface impacting your code.

Or something to that effect.

docs/api-levels.rst
```diff
@@ -0,0 +1,87 @@
...
+HTTP methods not currently implemented in the library - while still making use of all of the client's
+options and authentication methods.
+
+For example, if for whatever reason you want to fetch allowed methods for an endpoint at runtime:
+
+.. code-block:: python
+
+    >>> gl = gitlab.Gitlab(private_token=private_token)
+    >>>
+    >>> response = gl.http_request("OPTIONS", "/projects")
```
JohnVillalovos, line 85: Can keyword arguments be used here? As I'm not 100% sure what the arguments are at first glance.

docs/api-levels.rst
```diff
@@ -0,0 +1,87 @@
...
+allows you to write idiomatic Python code when interacting with the API.
+
+Lower-level API - HTTP methods
+==============================
+
+.. danger::
+
+   At this point, python-gitlab will no longer take care of URL-encoding and other transformations
+   needed to correctly pass API parameter types. You have to construct these correctly yourself.
+
```
JohnVillalovos, line 24: Maybe there could be an explanation of what it does do? For example it does handle the authorization. What other benefits does it have over just using the requests library?

nejch: Ah good idea. Maybe we can add some text in the intro on what the library does and link it here.

nejch: Done, thanks - I've added a reference to some basic features.

docs/api-levels.rst
nejch: Done :)

docs/api-levels.rst
nejch: I've looked at all the well-known http libraries and they all have the `request()` interface done like this (stdlib's http.client, urllib3, requests, httpx, httpcore, aiohttp), with those that enforce keyword-only args doing it after the method+url pair. So I thought this would make sense for people looking at lower-level APIs. I've added another example where we add some kwargs. If you feel really strongly about this I can still change it? :)

nejch: I ended up adding this in https://github.com/python-gitlab/python-gitlab/pull/2230/commits/af3d98bd2d13e77fe26fd58f626759e8515a3f3e.

JohnVillalovos (approved):

Issue #1495, opened 2021-06-03
No merge_base API support while GitLab v4 API has.


ZSmallX:
## Description of the problem, including code/CLI snippet
No merge_base API support, while GitLab v4 has this API.

See https://docs.gitlab.com/ee/api/repositories.html#merge-base

## Expected Behavior
Can get merge_base via python-gitlab

## Actual Behavior
No API found.

## Specifications

  - python-gitlab version: 1.15.0
  - API version you are using (v3/v4): v4
  - Gitlab server version (or gitlab.com): 11.4.0

PR #1577, opened 2021-08-26
Added merge_base API endpoint support

GerevSec:
Added simple merge_base API endpoint support, based on the repository_compare code.

Closes https://github.com/python-gitlab/python-gitlab/issues/1495.

PR #2236, opened 2022-08-14
feat: add support for merge_base API

nejch:
Closes https://github.com/python-gitlab/python-gitlab/issues/1495
Replaces https://github.com/python-gitlab/python-gitlab/pull/1577
Depended on https://github.com/python-gitlab/python-gitlab/pull/1699

### DIFF
--- a/gitlab/v4/objects/repositories.py
+++ b/gitlab/v4/objects/repositories.py
@@ -10,7 +10,7 @@ import requests
 import gitlab
 from gitlab import cli
 from gitlab import exceptions as exc
-from gitlab import utils
+from gitlab import types, utils
 
 if TYPE_CHECKING:
     # When running mypy we use these as the base classes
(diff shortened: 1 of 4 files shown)


gitlab/v4/objects/repositories.py
```diff
@@ -246,6 +246,32 @@ def repository_archive(
             result, streamed, action, chunk_size, iterator=iterator
         )
 
+    @cli.register_custom_action("Project", ("refs",))
+    @exc.on_http_error(exc.GitlabGetError)
+    def repository_merge_base(
+        self, refs: List[str], **kwargs: Any
+    ) -> Union[Dict[str, Any], requests.Response]:
```
JohnVillalovos, line 253: This made me think if we should start forcing these to just be a `Dict`?

JohnVillalovos (approved):
@nejch Looks good to me. Left a comment, but for sure could be something we don't do or do later.

This is approved, feel free to merge with my blessing!

gitlab/v4/objects/repositories.py
nejch: Thanks @JohnVillalovos! Makes sense to me! I was looking at doing this via overloading `http_get` to narrow the types there already, depending on the streamed/raw arguments but it got a bit out of hand. Then I think we wouldn't need to do it in "downstream" methods, IIUC. I will look into it again :)

But it requires quite a few overloads for all the possible signatures. This is what I was looking at https://medium.com/analytics-vidhya/making-sense-of-typing-overload-437e6deecade.

PR #2247, opened 2022-08-22
chore(ci): make pytest annotations work

nejch:
As discussed on Signal :)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -51,7 +51,7 @@ jobs:
         with:
           python-version: ${{ matrix.python.version }}
       - name: Install dependencies
-        run: pip3 install tox pytest-github-actions-annotate-failures
+        run: pip install tox
       - name: Run tests
         env:
           TOXENV: ${{ matrix.python.toxenv }}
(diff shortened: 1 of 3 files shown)


tox.ini
```diff
@@ -5,7 +5,7 @@ skip_missing_interpreters = True
 envlist = py310,py39,py38,py37,flake8,black,twine-check,mypy,isort,cz,pylint
 
 [testenv]
-passenv = GITLAB_IMAGE GITLAB_TAG PY_COLORS NO_COLOR FORCE_COLOR DOCKER_HOST PWD
+passenv = GITHUB_ACTIONS GITLAB_IMAGE GITLAB_TAG PY_COLORS NO_COLOR FORCE_COLOR DOCKER_HOST PWD
```
nejch, line 8: Seems like we need it.

https://github.com/utgwkk/pytest-github-actions-annotate-failures/blob/55bba59b58955c26cb657dee39b8551909277b88/pytest_github_actions_annotate_failures/plugin.py#L25-L28

.github/workflows/test.yml
```diff
@@ -51,7 +51,7 @@ jobs:
         with:
           python-version: ${{ matrix.python.version }}
       - name: Install dependencies
-        run: pip3 install tox pytest-github-actions-annotate-failures
+        run: pip3 install tox
```
JohnVillalovos, line 54: Maybe just `pip` (instead of `pip3`) like the rest of them?

JohnVillalovos (approved):

.github/workflows/test.yml
nejch: Done :)

PR #2248, opened 2022-08-22
wip: testing

JohnVillalovos:


nejch:
Thanks @JohnVillalovos - should we turn this into a real PR to only get annotations from the coverage jobs (and functional)?

JohnVillalovos:
> Thanks @JohnVillalovos - should we turn this into a real PR to only get annotations from the coverage jobs (and functional)?

If you want to take this over and use it to become a real PR, works for me. I guess could iterate on options with this failing test and then when finally ready remove the bad assert.

nejch:
Sounds great! We can amend at the end.

nejch:
@JohnVillalovos maybe I misunderstood here, should we close this PR or continue here?

JohnVillalovos:
Let's just close this one

Issue #2253, opened 2022-08-24
RFE: Support deployment approve/reject


JohnVillalovos:
### Discussed in https://github.com/python-gitlab/python-gitlab/discussions/2252

<div type='discussions-op-text'>

<sup>Originally posted by **jbotello7381** August 23, 2022</sup>
I was playing with deployment and gitlab python library, and notice there's no method to approve or reject a deployment? i was able to do it via REST API from my python code, but not via the python wrapper.  I'm using 3.8.1 of the gitlab python library.</div>

So it looks like it is this API:
https://docs.gitlab.com/ee/api/deployments.html#approve-or-reject-a-blocked-deployment

And it doesn't look like there is any support for that yet in:
https://github.com/python-gitlab/python-gitlab/blob/main/gitlab/v4/objects/deployments.py

Implementing it will look similar to how the "approve" method is done for a merge request.

https://github.com/python-gitlab/python-gitlab/blob/f67514e5ffdbe0141b91c88366ff5233e0293ca2/gitlab/v4/objects/merge_requests.py#L263-L290

PR #2254, opened 2022-08-24
feat: add support for deployment approval endpoint

JohnVillalovos:
Add support for the deployment approval endpoint[1]

[1] https://docs.gitlab.com/ee/api/deployments.html#approve-or-reject-a-blocked-deployment
Closes: #2253

### DIFF
--- a/gitlab/v4/objects/deployments.py
+++ b/gitlab/v4/objects/deployments.py
@@ -1,5 +1,11 @@
-from typing import Any, cast, Union
+"""
+GitLab API:
+https://docs.gitlab.com/ee/api/deployments.html
+"""
+from typing import Any, cast, Dict, Optional, TYPE_CHECKING, Union
 
+from gitlab import cli
+from gitlab import exceptions as exc
 from gitlab.base import RESTManager, RESTObject
 from gitlab.mixins import CreateMixin, RetrieveMixin, SaveMixin, UpdateMixin
 from gitlab.types import RequiredOptional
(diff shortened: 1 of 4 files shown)


gitlab/v4/objects/deployments.py
```diff
@@ -15,6 +21,50 @@
 class ProjectDeployment(SaveMixin, RESTObject):
     mergerequests: ProjectDeploymentMergeRequestManager
 
+    @cli.register_custom_action(
+        "ProjectDeployment",
+        mandatory=("status",),
+        optional=("comment", "represented_as"),
+    )
+    @exc.on_http_error(exc.GitlabDeploymentApprovalError)
+    def approval(
```
nejch, line 30: Thanks @JohnVillalovos that was quick!

Do you think it makes sense to implement this as an additional manager with a CreateMixin perhaps? I am looking at the upstream planning and I see there may also be _skip_ and _revoke_ actions soon:
* https://gitlab.com/gitlab-org/gitlab/-/issues/353963
* https://gitlab.com/gitlab-org/gitlab/-/issues/352914

I'm also happy to merge this as is. Just again the endpoint on the GitLab side seems a bit inconsistent so I was wondering if this makes sense to future-proof extending it. It'd be a bit more verbose though.

nejch (commented):
Thanks LGTM! Just a quick question mostly about upstream API weirdness.

gitlab/v4/objects/deployments.py
JohnVillalovos: I'll admit I'm confused. How would that work? I'm not quite sure what you mean.

I notice that the current `ProjectDeploymentManager` does use `CreateMixin`. But not sure what is being suggested here.

nejch: Like we have `ProjectMergeRequestApprovalManager` as `mr.approvals`, this could be `ProjectDeploymentApprovalManager` as `deployment.approvals`, and use `_create_attrs` for validation instead of custom checks.

You'd then use 
`deployment.approvals.create({"status": "rejected", "comment": "Fails CI", "represented_as": "security")`
instead of
`deployment.approval(status="rejected", comment="Fails CI", represented_as="security")`

it would also open things up for `approval.skip()` and `approval.revoke()` for the planned actions above.

JohnVillalovos: Thanks for the clarification!

TL;DR: I am leaning towards merging as it is now for the sake of simplicity. But for sure willing to have another manager.

1. That `create` syntax sure looks ugly to me.  I would rather have a custom method that looks like `approve` does now. But that could be done. just not using the `CreateMixin`
2. Looking at the GitLab issues, I don't see any proposed MRs yet. So unsure how the API endpoints look at this time. Or if they will even get merged.
3. It seems like we already have dot attribute access for `deployment.status`

So that being said I guess I like the current version. But if do decide to create another manager I would prefer to not use `create` but some other way.  Maybe `deployment.approvals.approval(status="approved")`

What do you think @nejch ?

nejch: Deployment status is not the same as approval status :) I agree it's a bit funky but `approval()` to me sounds like it would get the details of an existing one rather than create a new one, and we've gone this way with other weird endpoints like `lint`. It would allow code to be more declarative IMO.

But anyway I just saw the GitLab functionality is actually still beta. so maybe it'll change on their end to make more sense for us as well.

PR #2255, opened 2022-08-24
chore: fix issue if only run test_gitlab.py func test

JohnVillalovos:
Make it so can run just the test_gitlab.py functional test.

For example:
$ tox -e api_func_v4 -- -k test_gitlab.py


nejch:
I've been checking the ResourceWarning topic in requests - and I think we should eventually make sure to get rid of it completely so I'd merge this for now to get the other PR in, and then later ensure a clean run without warnings in the test run and then potentially clean up if needed :)

PR #2279, opened 2022-09-22
feat: Add reset_approvals api

zampierilucas:
Added the newly added reset_approvals merge request api.
https://docs.gitlab.com/ee/api/merge_requests.html#reset-approvals-of-a-merge-request

Signed-off-by: Lucas Zampieri <lzampier@redhat.com>

### DIFF
--- a/docs/gl_objects/merge_requests.rst
+++ b/docs/gl_objects/merge_requests.rst
@@ -189,6 +189,10 @@ Attempt to rebase an MR::
 
     mr.rebase()
 
+Clear all approvals of a merge request (possible with project or group access tokens only)::
+
+    mr.reset_approvals()
+
 Get status of a rebase for an MR::
 
     mr = project.mergerequests.get(mr_id, include_rebase_in_progress=True)
--- a/gitlab/v4/objects/merge_requests.py
+++ b/gitlab/v4/objects/merge_requests.py
@@ -328,6 +328,24 @@ class ProjectMergeRequest(
         return self.manager.gitlab.http_put(path, post_data=data, **kwargs)
 
     @cli.register_custom_action("ProjectMergeRequest")
+    @exc.on_http_error(exc.GitlabMRResetApprovalError)
+    def reset_approvals(
+        self, **kwargs: Any
+    ) -> Union[Dict[str, Any], requests.Response]:
+        """Clear all approvals of the merge request.
+
+        Args:
+            **kwargs: Extra options to send to the server (e.g. sudo)
+
+        Raises:
+            GitlabAuthenticationError: If authentication is not correct
+            GitlabMRResetApprovalError: If reset approval failed
+        """
+        path = f"{self.manager.path}/{self.encoded_id}/reset_approvals"
+        data: Dict[str, Any] = {}
+        return self.manager.gitlab.http_put(path, post_data=data, **kwargs)
+
+    @cli.register_custom_action("ProjectMergeRequest")
     @exc.on_http_error(exc.GitlabGetError)
     def merge_ref(self, **kwargs: Any) -> Union[Dict[str, Any], requests.Response]:
         """Attempt to merge changes between source and target branches into
--- a/tests/functional/api/test_merge_requests.py
+++ b/tests/functional/api/test_merge_requests.py
@@ -117,6 +117,15 @@ def test_merge_request_rebase(project):
     assert mr.rebase()
 
 
+def test_merge_request_reset_approvals(gitlab_url, project):
+    bot = project.access_tokens.create({"name": "bot", "scopes": ["api"]})
+    bot_gitlab = gitlab.Gitlab(gitlab_url, private_token=bot.token)
+    bot_project = bot_gitlab.projects.get(project.id, lazy=True)
+
+    mr = bot_project.mergerequests.list()[0]
+    assert mr.reset_approvals()
+
+
 @pytest.mark.skip(reason="flaky test")
 def test_merge_request_merge(project):
     mr = project.mergerequests.list()[0]
--- a/gitlab/exceptions.py
+++ b/gitlab/exceptions.py
@@ -205,6 +205,10 @@ class GitlabMRRebaseError(GitlabOperationError):
     pass
 
 
+class GitlabMRResetApprovalError(GitlabOperationError):
+    pass
+
+
 class GitlabMRClosedError(GitlabOperationError):
     pass
 


gitlab/v4/objects/merge_requests.py
```diff
@@ -327,6 +327,24 @@ def rebase(self, **kwargs: Any) -> Union[Dict[str, Any], requests.Response]:
         data: Dict[str, Any] = {}
         return self.manager.gitlab.http_put(path, post_data=data, **kwargs)
 
+    @cli.register_custom_action("ProjectMergeRequest")
+    @exc.on_http_error(exc.GitlabMRResetApprovalError)
+    def reset_approvals(
+        self, **kwargs: Any
+    ) -> Union[Dict[str, Any], requests.Response]:
+        """Attempt to reset approvals onto the target Merge Request
```
nejch, line 335: Maybe we can use the upstream wording, `attempt` was more for rebase I'd say:
```suggestion
        """Clear all approvals of the merge request.
```

docs/gl_objects/merge_requests.rst
```diff
@@ -189,6 +189,12 @@ Attempt to rebase an MR::
 
     mr.rebase()
 
+
+Attempt to reset all MR approval::
```
nejch, line 193: Same as below, and maybe we could mention this only works for bots to avoid questions from users in the future, WDYT?

```suggestion
Clear all approvals of a merge request (possible with project or group access tokens only)::
```

docs/gl_objects/merge_requests.rst
```diff
@@ -189,6 +189,12 @@ Attempt to rebase an MR::
 
     mr.rebase()
 
+
```
nejch, line 192:
```suggestion
```

docs/gl_objects/merge_requests.rst
```diff
@@ -189,6 +189,12 @@ Attempt to rebase an MR::
...
+Attempt to reset all MR approval::
+
+    mr.reset_approvals()
+
+
```
nejch, line 197:
```suggestion
```

gitlab/v4/objects/merge_requests.py
```diff
@@ -327,6 +327,24 @@ def rebase(self, **kwargs: Any) -> Union[Dict[str, Any], requests.Response]:
...
+
+        Args:
+            **kwargs: Extra options to send to the server (e.g. sudo)
+
+        Raises:
+            GitlabAuthenticationError: If authentication is not correct
+            GitlabMRResetApprovalError: If reset approval failed
+        """
+        path = f"{self.manager.path}/{self.encoded_id}/reset_approvals"
+        data: Dict[str, Any] = {}
```
nejch, line 345: I don't think we necessarily to supply the empty data here, but I see we do the same for rebase above, so let's leave it.

nejch (changes requested):
Thanks a lot @zampierilucas, looks great, I mainly just had a few style comments for docs.

I see you're getting a 404 on the functional test, did this work locally?

I've retried just in case, but I'm not sure if you need to first ensure that an approval exists for this to work, or if we maybe have an issue because this is done with an admin token and not a resource access token that this endpoint requires.

nejch:
Oh I'm sorry @zampierilucas the only reason this is probably failing is that we're on an old GitLab image. Could you bump the docker image version here?
https://github.com/python-gitlab/python-gitlab/blob/main/tests/functional/fixtures/.env#L2

Something's wrong with our renovate setup, this was supposed to be auto-updated. After that's done, let's see if it works with an admin token or if we get a 401 as well in tests. Thanks again :)

JohnVillalovos:
> Oh I'm sorry @zampierilucas the only reason this is probably failing is that we're on an old GitLab image. Could you bump the docker image version here? https://github.com/python-gitlab/python-gitlab/blob/main/tests/functional/fixtures/.env#L2

I have created a PR to update the Docker image: https://github.com/python-gitlab/python-gitlab/pull/2280

docs/gl_objects/merge_requests.rst
zampierilucas: Added.

gitlab/v4/objects/merge_requests.py
zampierilucas: I'm thinking that maybe that is there, so we can `skip_ci` to rebase as data? https://docs.gitlab.com/ee/api/merge_requests.html#rebase-a-merge-request

If that makes sense, we could remove it from reset_approvalls, OTOH it doesn't really affect the functionality of reset_approvals, and it can be ready if reset_approvalls is expanded to, lets say, reset a single approval, wdyt?

zampierilucas:
Thank you both @nejch and @JohnVillalovos for your feedback, I've applied @nejch suggestion to my last push, and rebase this branch on top of `jlvillal/docker_image`, so let's see if the pipeline passes now :D

gitlab/v4/objects/merge_requests.py
nejch: Oh, that wouldn't actually work as we just define an empty dict in the method, but nice find! Unrelated, but we should also support that in rebase :) That's also a recent feature IIRC.

nejch:
Thanks @zampierilucas and @JohnVillalovos! I merged the other PR and we can start fresh here.

So now we are hitting 401 as I suspected above because we run it as the `root` user. @zampierilucas two options:
* You might need to create a project access token in that specific test case and override `private_token`, e.g. by creating a new Gitlab instance for that test
* Alternatively you can write a unit test instead and just write some mocked fixtures.

I think adding the project access token will be quicker and would also test behavior against future gitlab versions so I think that might be nicer :) let us know if you can try and add that. See docs here: https://python-gitlab.readthedocs.io/en/stable/gl_objects/project_access_tokens.html#examples

tests/functional/api/test_merge_requests.py
```diff
@@ -117,6 +117,11 @@ def test_merge_request_rebase(project):
     assert mr.rebase()
 
 
+def test_merge_request_reset_approvals(project):
+    mr = project.mergerequests.list()[0]
+    assert mr.reset_approvals()
```
nejch, line 122: I think even just something small like this should get rid of the 401, @zampierilucas:

```suggestion
def test_merge_request_reset_approvals(gitlab_url, project):
    bot = project.access_tokens.create({"name": "bot", "scopes": ["api"]})
    bot_gitlab = gitlab.Gitlab(gitlab_url, private_token=bot.token)
    bot_project = bot_gitlab.projects.get(project.id, lazy=True)

    mr = bot_project.mergerequests.list()[0]
    assert mr.reset_approvals()
```

zampierilucas: Yup, that did it, I have been trying it this afternoon, but couldn't quite figure out the gitlab.Gitlab(), turns out I was missing the gitlab_url, thanks @nejch, I just pushed an update with your suggestion :D

JohnVillalovos (approved):

nejch (approved):
Thanks for the quick rework @zampierilucas! :) This should be available with the next release on the 28th.

Issue #2287, opened 2022-09-30
project-commit merge-requests call doesn't work


ramarro123:
calling

MRID=$(gitlab -o json project-commit merge-requests --project-id ${id} --id ${COMMIT_ID})

with 
COMMIT_ID 2fdd3e323446d24ca80e36e41aa710a853662fac
and id = 63

(valid and working)

give  as output

Traceback (most recent call last):
  File "/usr/local/bin/gitlab", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.9/site-packages/gitlab/cli.py", line 377, in main
    gitlab.v4.cli.run(
  File "/usr/local/lib/python3.9/site-packages/gitlab/v4/cli.py", line 549, in run
    printer.display_list(data, fields, verbose=verbose)
  File "/usr/local/lib/python3.9/site-packages/gitlab/v4/cli.py", line 415, in display_list
    print(json.dumps([get_dict(obj, fields) for obj in data]))
  File "/usr/local/lib/python3.9/site-packages/gitlab/v4/cli.py", line 415, in <listcomp>
    print(json.dumps([get_dict(obj, fields) for obj in data]))
  File "/usr/local/lib/python3.9/site-packages/gitlab/v4/cli.py", line 397, in get_dict
    return obj.attributes
AttributeError: 'dict' object has no attribute 'attributes'

PR #2290, opened 2022-09-30
fix(cli): handle list response for json/yaml output

JohnVillalovos:
Handle the case with the CLI where a list response is returned from GitLab and json/yaml output is requested.

Closes: #2287


tests/functional/cli/test_cli_repository.py
```diff
@@ -45,6 +49,49 @@ def test_list_all_commits(gitlab_cli, project):
     assert len(ret_all.stdout) > len(ret.stdout)
 
 
+def test_commit_merge_requests(gitlab_cli, project, merge_request, wait_for_sidekiq):
+    """This tests the `project-commit merge-requests` command and also tests
+    that we can print the result using the `json` formatter"""
+    # create and then merge a merge-request
+    source_branch = "test_commit_merge_requests"
+    mr = merge_request(source_branch=source_branch)
```
nejch, line 57: I didn't see that variable used anywhere- maybe we don't need it?

```suggestion
    mr = merge_request(source_branch="test_commit_merge_requests")
```

tests/functional/cli/test_cli_repository.py
```diff
@@ -45,6 +49,49 @@ def test_list_all_commits(gitlab_cli, project):
...
+    merge_result = mr.merge(should_remove_source_branch=True)
+    result = wait_for_sidekiq(timeout=60)
+    assert result is True, "sidekiq process should have terminated but did not"
```
nejch, line 60: I see we are adding this assert in all the tests - maybe we should just put this in the fixture itself to avoid repetition? Or is that too strict.

tests/functional/cli/test_cli_repository.py
```diff
@@ -45,6 +49,49 @@ def test_list_all_commits(gitlab_cli, project):
     assert len(ret_all.stdout) > len(ret.stdout)
 
 
+def test_commit_merge_requests(gitlab_cli, project, merge_request, wait_for_sidekiq):
+    """This tests the `project-commit merge-requests` command and also tests
+    that we can print the result using the `json` formatter"""
```
nejch, line 54: Just an idea, if we're testing mostly the list output, we could simplify this test a lot by not relying on MRs and just getting the diff of the HEAD: `gitlab -o json project-commit diff --project-id project.id --id project.default_branch`. This command has the same issue.

nejch (commented):
Nice fix @JohnVillalovos! Just a quick question on tests as I'm seeing some more repetition now the sidekiq handling.

tests/functional/cli/test_cli_repository.py
JohnVillalovos: Well consumers of the fixture don't necessarily have to merge. And it is doing additional things before we would hand things back to the cleanup part of the fixture.

So in this case would still want to have sidekiq finish (to be safe). So I'm not sure how to move it to the fixture...

JohnVillalovos: One idea could be to add a new argument to `wait_for_sidekiq`. Maybe `wait_for_sidekiq(timeout=60, fail_if_timeout=True)` ?

tests/functional/cli/test_cli_repository.py
JohnVillalovos: Sure. But it does allows us to test both `project-commit merge-requests` and the list issue. I guess since I already wrote this one I don't want to throw away the testing of  `merge-requests`.

Agree it isn't the simplest test for sure! Took me awhile to get it all going.

WDYT?

tests/functional/cli/test_cli_repository.py
JohnVillalovos: Done.

tests/functional/cli/test_cli_repository.py
nejch: Yes, I think the way it was done in #2297 looks nice :)

tests/functional/cli/test_cli_repository.py
nejch: True, makes sense too. was just thinking as this had to re-create some of the MR creation logic. Let's leave it as is then :)

tests/functional/cli/test_cli_repository.py
JohnVillalovos: > Yes, I think the way it was done in #2297 looks nice :)

Done

nejch (approved):

PR #2297, opened 2022-10-05
chore: simplify `wait_for_sidekiq` usage

JohnVillalovos:
Simplify usage of `wait_for_sidekiq` by putting the assert if it timed out inside the function rather than after calling it.


nejch (approved):
Thanks @JohnVillalovos that looks clean and also separates the test asserts from test framework errors :)

PR #2305, opened 2022-10-08
docs(api): describe the list() and all() runners' functions

lmilbaum:
A follow up from #2304
Also, added a rst lint precommit checks

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -38,3 +38,9 @@ repos:
           - types-PyYAML==6.0.12
           - types-requests==2.28.11
           - types-setuptools==64.0.1
+  - repo: https://github.com/pre-commit/pygrep-hooks
+    rev: v1.9.0
+    hooks:
+      - id: rst-backticks
+      - id: rst-directive-colons
+      - id: rst-inline-touching-normal
--- a/docs/gl_objects/runners.rst
+++ b/docs/gl_objects/runners.rst
@@ -28,7 +28,11 @@ Reference
 Examples
 --------
 
-Use the ``list()`` and ``all()`` methods to list runners.
+Use the ``runners.list()`` and ``runners_all.list()`` methods to list runners.
+``runners.list()`` - Get a list of specific runners available to the user
+``runners_all.list()``  - Get a list of all runners in the GitLab instance 
+(specific and shared). Access is restricted to users with administrator access.
+
 
 Both methods accept a ``scope`` parameter to filter the list. Allowed values
 for this parameter are:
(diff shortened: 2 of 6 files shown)


docs/gl_objects/runners.rst
```diff
@@ -29,6 +29,10 @@ Examples
 --------
 
 Use the ``list()`` and ``all()`` methods to list runners.
+``list()`` - Get a list of specific runners available to the user
+``all()``  - Get a list of all runners in the GitLab instance (specific and 
+shared). Access is restricted to users with administrator access.
```
nejch, line 34: Thanks for the PR @lmilbaum! I realized we have some outdated docs here. Would you mind updating this as well?

We no longer recommend the `all()` method, but instead have a separate manager which brings more consistent behavior between the two endpoints. Like this:

```suggestion
Use the ``runners.list()`` and ``runners_all.list()`` methods to list runners.
``runners.list()`` - Get a list of specific runners available to the user
``runners_all.list()``  - Get a list of all runners in the GitLab instance (specific and 
shared). Access is restricted to users with administrator access.
```

We would also need to change the text below, but I can't suggest it here in the PR - like so:
```diff
-* ``specific`` (``all()`` only)
-* ``shared`` (``all()`` only)
+* ``specific`` (``runners_all.list()`` only)
+* ``shared`` (``runners_all.list()`` only)
```

Thanks again for the PR :bow:

lmilbaum: NP. Pushed a fix.

.pre-commit-config.yaml
```diff
@@ -38,3 +38,9 @@ repos:
           - types-PyYAML==6.0.12
           - types-requests==2.28.11
           - types-setuptools==64.0.1
+  - repo: https://github.com/pre-commit/pygrep-hooks
+    rev: v1.9.0
+    hooks:
+      - id: rst-backticks
+      - id: rst-directive-colons
+      - id: rst-inline-touching-normal
```
nejch, line 46: Nice, I guess these are quite fast and don't slow down the pre-commit hooks much right? It might make sense to fix the other issues in a separate commit if you want to already add this in the PR here and get green CI.

lmilbaum: I don't think those additional checks slow down the pre-commit. LMK if you notice a degradation in CI performance.
Fixed the issues discovered by the new checks.

nejch: Just gave it a quick try locally and the new hooks are much faster than all our existing linters even when running all files, so looks good :)

nejch (approved):
Thanks again for this @lmilbaum!

PR #2309, opened 2022-10-11
chore: renovate and precommit cleanup

lmilbaum:
1. Cleanup pre-commit configuration file 
2. Remove the installation of pip packages in the precommit CI stage. precommit is responsible for installing the dependencies according to the repo revision.
3. renovate configuration file validation by precommit


.renovaterc.json
```diff
@@ -4,39 +4,46 @@
...
-        "ignoreUnstable": false,
-        "groupName": "black"
-     }
+      "matchPackagePatterns": [
+        "(^|/)black$"
+      ],
+      "versioning": "pep440",
+      "ignoreUnstable": false,
+      "groupName": "black"
+    }
```
nejch, line 47: Since we're doing some cleanup here, I think we can remove the custom black handling entirely. We had this because black was beta for a really long time and updates were ignored.

With https://black.readthedocs.io/en/latest/the_black_code_style/index.html#stability-policy I think we can get rid of this (sorry, you'll have to change this locally, github won't let me suggest changes to lines above):
```suggestion
```

.renovaterc.json
```diff
@@ -4,39 +4,46 @@
...
-  "pip_requirements": {
-    "fileMatch": ["^requirements(-[\\w]*)?\\.txt$"]
-  },
+  "enabledManagers": [
+    "pip_requirements",
+    "pre-commit"
+  ],
+  "ignoreDeps": [
+    "relekang/python-semantic-release"
+  ],
```
nejch, line 13: I think we can also delete this, we had it because semantic-release was buggy for a while affecting our releases, but it's been running stable for a few months now after manually bumping it:

```suggestion
```

.renovaterc.json
```diff
@@ -4,39 +4,46 @@
...
     "schedule:weekly"
   ],
-  "ignoreDeps": ["relekang/python-semantic-release"],
-  "pip_requirements": {
-    "fileMatch": ["^requirements(-[\\w]*)?\\.txt$"]
-  },
+  "enabledManagers": [
+    "pip_requirements",
+    "pre-commit"
+  ],
```
nejch, line 10: Just a quick question, was this suggested by `renovate-config-validator`? IIRC `enabledManagers` is an allowlist and others will be ignored, so we might need to maintain this list manually (e.g. in https://github.com/python-gitlab/python-gitlab/blob/main/Dockerfile.

nejch (commented):
Thanks a lot for the cleanup here @lmilbaum. It made me take another look at the setup and I think we can get rid of some leftover config so I just had a few quick questions :)

.renovaterc.json
lmilbaum: Sure. Removing those lines.

.renovaterc.json
lmilbaum: Sure. Removing.

.renovaterc.json
lmilbaum: It was not suggested by `renovate-config-validator`. Not sure I am following the rest of your comment.

lmilbaum: Hmmm. I am giving it another thought. It is better to remove this section and let precommit take into consideration all the managers.

.github/workflows/pre_commit.yml
```diff
@@ -33,7 +33,7 @@ jobs:
       - uses: actions/setup-python@v4
         with:
           python-version: "3.10"
-      - run: pip install --upgrade -r requirements.txt -r requirements-lint.txt pre-commit
+      - run: pip install pre-commit
```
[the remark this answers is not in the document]
lmilbaum: Thanks. BTW, it might be more elegant to move the pre-commit package to a requirements file with a pinned version. WDYT?

.renovaterc.json
nejch: That's true, but I agree it's best to let renovate pick up any potential updates yes :) thanks.

.github/workflows/pre_commit.yml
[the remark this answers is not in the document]
nejch: Ah, that might make sense, although users will probably still have it installed globally with a version we don't really control (https://pre-commit.com/#installation). I'm ok either way, making that change here or in a follow-up PR as well :)

nejch (approved):
LGTM, thanks again @lmilbaum!

PR #2310, opened 2022-10-14
feat(api): add support for topics merge API

nejch:
https://docs.gitlab.com/ee/api/topics.html#merge-topics

### DIFF
--- a/docs/gl_objects/topics.rst
+++ b/docs/gl_objects/topics.rst
@@ -46,3 +46,7 @@ Delete a topic::
 
     # or
     gl.topics.delete(topic_id)
+
+Merge a source topic into a target topic::
+
+    gl.topics.merge(topic_id, target_topic_id)
--- a/gitlab/v4/objects/topics.py
+++ b/gitlab/v4/objects/topics.py
@@ -1,5 +1,7 @@
-from typing import Any, cast, Union
+from typing import Any, cast, Dict, TYPE_CHECKING, Union
 
+from gitlab import cli
+from gitlab import exceptions as exc
 from gitlab import types
 from gitlab.base import RESTManager, RESTObject
 from gitlab.mixins import CRUDMixin, ObjectDeleteMixin, SaveMixin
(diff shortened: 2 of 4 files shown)


nejch:
@max-wittig could you take a quick look at this one?

bufferoverflow (approved):

docs/gl_objects/topics.rst
```diff
@@ -46,3 +46,7 @@ Delete a topic::
 
     # or
     gl.topics.delete(topic_id)
+
+Merge a source topic into a target topic::
```
JohnVillalovos, line 50: Should it mention that it requires admin permission?

gitlab/v4/objects/topics.py
```diff
@@ -29,3 +31,39 @@ class TopicManager(CRUDMixin, RESTManager):
...
         return cast(Topic, super().get(id=id, lazy=lazy, **kwargs))
+
+    @cli.register_custom_action(
+        "TopicManager",
+        mandatory=("source_topic_id", "target_topic_id"),
+    )
+    @exc.on_http_error(exc.GitlabMRClosedError)
+    def merge(
+        self,
+        source_topic_id: Union[int, str],
```
JohnVillalovos, line 42: The docs say it only allows an integer.

max-wittig (approved):

gitlab/v4/objects/topics.py
nejch: I think it just needs to be IDs, as long as the strings contain the IDs this will work: `gl.topics.merge("1", "2")`. I can change it, in the end it's just hints but technically it will be a bit too strict :)

docs/gl_objects/topics.rst
nejch: Most of these endpoints here need admin access so we have this above, did you mean in addition to that?

> This endpoint requires admin access for creating, updating and deleting objects.

JohnVillalovos: Ah I didn't scroll back to see line 18.

gitlab/v4/objects/topics.py
JohnVillalovos: Makes sense. Thanks.

JohnVillalovos (approved):

PR #2311, opened 2022-10-15
chore: topic functional tests

lmilbaum:
A followup to @nejch's contribution #2310

### DIFF
--- a/tests/functional/api/test_topics.py
+++ b/tests/functional/api/test_topics.py
@@ -10,19 +10,25 @@ def test_topics(gl, gitlab_version):
     create_dict = {"name": "my-topic", "description": "My Topic"}
     if gitlab_version.major >= 15:
         create_dict["title"] = "my topic title"
-    topic = gl.topics.create(
-        {"name": "my-topic", "title": "my topic title", "description": "My Topic"}
-    )
+    topic = gl.topics.create(create_dict)
     assert topic.name == "my-topic"
+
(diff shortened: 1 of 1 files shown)


tests/functional/api/test_topics.py
```diff
@@ -10,19 +10,25 @@ def test_topics(gl, gitlab_version):
...
 
-    topic.delete()
+    create_dict = {"name": "my-second-topic", "description": "My Second Topic"}
+    if gitlab_version.major >= 15:
+        create_dict["title"] = "my second topic title"
+    topic2 = gl.topics.create(create_dict)
+    merged_topic = gl.topics.merge(topic.id, topic2.id)
+    assert merged_topic.id == topic2.id
+
+    topic2.delete()
```
nejch, line 33: This looks great already @lmilbaum just note that we don't return a real object on `merge()`, just the server data, so we can't manipulate it like that.

But I think the server returns enough information that we could even create a proper Topic object that we can then manipulate - might be a nicer interface if we refactor that! :)

I know this is just a draft, just thought I'd chime in.

lmilbaum: Thanks @nejch for you feedback. The latest change I've pushed just passed the tests. Does that looks good to you?

nejch: Looks great @lmilbaum, thanks a lot for the tests! :tada:

nejch (approved):

PR #2315, opened 2022-10-16
refactor: pytest-docker fixtures

lmilbaum:
Splitting out the pytest-docker fixtures from conftest.py file. 
At first I was thinking to gather those functions in conftest.py file and adding a starting and closing comment. 
On a second thought, moving those functions to a second fixtures file might make more sense.

### DIFF
--- a/tests/functional/conftest.py
+++ b/tests/functional/conftest.py
@@ -12,6 +12,7 @@ import requests
 import gitlab
 import gitlab.base
 from tests.functional import helpers
+from tests.functional.fixtures.docker import *  # noqa
 
 SLEEP_TIME = 10
 
@@ -32,11 +33,6 @@ def gitlab_version(gl) -> GitlabVersion:
     version, revision = gl.version()
     major, minor, patch = version.split(".")
     return GitlabVersion(major=major, minor=minor, patch=patch, revision=revision)
-
-
-@pytest.fixture(scope="session")
-def docker_compose_command():
-    return "docker compose"
 
 
 @pytest.fixture(scope="session")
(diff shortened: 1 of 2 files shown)


tests/functional/pytest_docker_fixtures.py
```diff
@@ -0,0 +1,29 @@
+import pytest
+
+# pytest-docker fixtures
+# https://github.com/avast/pytest-docker#available-fixtures
```
nejch, line 4: Should we make it a module docstring? I think then people can see it in IDEs when hovering over imports.
```suggestion
"""
pytest-docker fixture overrides.
See https://github.com/avast/pytest-docker#available-fixtures.
"""
import pytest
```

tests/functional/conftest.py
```diff
@@ -12,6 +12,7 @@
 import gitlab
 import gitlab.base
 from tests.functional import helpers
+from tests.functional.pytest_docker_fixtures import *  # noqa
```
nejch, line 15: One interesting thing is that our `tests/functional/fixtures/` directory contains all the fixture files related to this docker setup. So maybe we could move this to `tests/functional/fixtures/docker.py` if that makes sense? It would probably need an `__init__.py` in that folder. WDYT?

```suggestion
from tests.functional.fixtures.docker import *  # noqa
```

nejch (commented):
Thanks @lmilbaum, I guess it does make sense to factor this out as our conftest grows. Just 2 little questions about file placement :)

tests/functional/pytest_docker_fixtures.py
lmilbaum: Good idea.

tests/functional/conftest.py
lmilbaum: Totally makes sense

nejch (approved):
Thanks for the cleanup @lmilbaum!

PR #2320, opened 2022-10-17
refactor: pre-commit triggered from tox

lmilbaum:
The project is using two liniting frameworks: tox and pre-commit.  Each has its own configuration files and a CI workflow. 
The purpose of this PR is to consolidate linting  into one framework: tox
Discussion -  #2321

This PR is a baby step towards a full transition

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -2,7 +2,7 @@
 minversion = 1.6
 skipsdist = True
 skip_missing_interpreters = True
-envlist = py310,py39,py38,py37,flake8,black,twine-check,mypy,isort,cz,pylint
+envlist = py310,py39,py38,py37,flake8,black,twine-check,mypy,isort,cz,pylint,pre-commit
 
 [testenv]
 passenv =
@@ -123,3 +123,8 @@ commands =
 [testenv:smoke]
 deps = -r{toxinidir}/requirements-test.txt
 commands = pytest tests/smoke {posargs}
+
+[testenv:pre-commit]
+skip_install = true
+deps = -r requirements-precommit.txt
+commands = pre-commit run --all-files --show-diff-on-failure
--- a/.github/workflows/pre_commit.yml
+++ b/.github/workflows/pre_commit.yml
@@ -33,9 +33,7 @@ jobs:
       - uses: actions/setup-python@v4.3.0
         with:
           python-version: "3.10"
-      - name: Install pre-commit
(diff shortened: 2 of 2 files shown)


JohnVillalovos:
I personally don't like this as I do NOT use pre-commit. I like to do everything in tox. And I do not want to use pre-commit.

lmilbaum:
> I personally don't like this as I do NOT use pre-commit. I like to do everything in tox. And I do not want to use pre-commit.

Good point. I should have raised this topic  in a discussion.

requirements-lint.txt
```diff
@@ -1,4 +1,5 @@
 argcomplete==2.0.0
+astroid==2.12.12
```
nejch, line 2: I guess we'll want to address the real pylint issue at some point, but right now this should fix it:

```suggestion
astroid==2.12.11
```

lmilbaum: That worked. Thanks.

tox.ini
```diff
@@ -123,3 +123,8 @@ commands =
 [testenv:smoke]
 deps = -r{toxinidir}/requirements-test.txt
 commands = pytest tests/smoke {posargs}
+
+[testenv:pre-commit]
+skip_install = true
+deps = -r requirements-precommit.txt
+commands = pre-commit run --all-files --show-diff-on-failure
```
nejch, line 130: If I understand correctly - this, together with `pre-commit` in the `envlist`, will actually still run duplicate checks when you run a default `tox`. I'd either take them out from the envlist, or specifically define which pre-commit hook ids we want to run. Otherwise, LGTM! :)

lmilbaum: Correct. It will run duplicate checks. I would prefer to leave it like that for the time being until @JohnVillalovos provides his feedback on his user expriance. Does that make sense?

nejch: If it doesn't add too much runtime I don't mind and sounds ok to me if we deduplicate them in the end, will let @JohnVillalovos chime in as I joined in a bit late in this thread :)

JohnVillalovos (approved):
This works for me. As I usually run a subset of the tox checks in my development process.

I like that I don't have to run `pre-commit` and can just use `tox` 🙂

So I approve but let @nejch do the merging if he approves.

nejch:
`time tox` gives me 50s versus 20s before so running the default env list takes a bit long now,  but let's get this in and work on that later :)

Issue #2325, opened 2022-10-18
[feature request] auto-retry when HTTP 409 Conflict: Resource lock is received


pspacek:
## Description of the problem, including code/CLI snippet
Gitlab API requests can fail with error `Conflict: Resource lock`, and it is tedious to write retry loop around all call sites.

I propose to add an (optional?) feature to auto-retry API calls which fail with this particular error.

In my opinion this is the same case as https://peps.python.org/pep-0475/ .

I'm happy to discuss details like retry limit etc. Thank you for considering this.

## Expected Behavior
API calls are automatically retried when server replies with "HTTP 409 Conflict: Resource lock".

## Actual Behavior
```
Traceback (most recent call last):
  File "/usr/lib/python3.10/site-packages/gitlab/exceptions.py", line 333, in wrapped_f
    return f(*args, **kwargs)
  File "/usr/lib/python3.10/site-packages/gitlab/v4/objects/jobs.py", line 32, in cancel
    result = self.manager.gitlab.http_post(path, **kwargs)
  File "/usr/lib/python3.10/site-packages/gitlab/client.py", line 1015, in http_post
    result = self.http_request(
  File "/usr/lib/python3.10/site-packages/gitlab/client.py", line 798, in http_request
    raise gitlab.exceptions.GitlabHttpError(
gitlab.exceptions.GitlabHttpError: 409: 409 Conflict: Resource lock

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/pspacek/w/isc/gitlab/docs_test.py", line 99, in <module>
    main()
  File "/home/pspacek/w/isc/gitlab/docs_test.py", line 53, in main
    proj.jobs.get(job.id, lazy=True).cancel()
  File "/usr/lib/python3.10/site-packages/gitlab/cli.py", line 71, in wrapped_f
    return f(*args, **kwargs)
  File "/usr/lib/python3.10/site-packages/gitlab/exceptions.py", line 335, in wrapped_f
    raise error(e.error_message, e.response_code, e.response_body) from e
gitlab.exceptions.GitlabJobCancelError: 409: 409 Conflict: Resource lock
```

## Specifications

  - python-gitlab version: 3.10.0
  - API version you are using (v3/v4): v4
  - Gitlab server version (or gitlab.com): gitlab.isc.org

pspacek:
[...]

PR #2326, opened 2022-10-18
feat(client): automatically retry on HTTP 409 Resource lock

pspacek:
Fixes: #2325

### DIFF
--- a/docs/api-usage-advanced.rst
+++ b/docs/api-usage-advanced.rst
@@ -123,9 +123,14 @@ GitLab server can sometimes return a transient HTTP error.
 python-gitlab can automatically retry in such case, when
 ``retry_transient_errors`` argument is set to ``True``.  When enabled,
 HTTP error codes 500 (Internal Server Error), 502 (502 Bad Gateway),
-503 (Service Unavailable), and 504 (Gateway Timeout) are retried.
-Additionally the HTTP error code 409 (Conflict) is retried if the text message
-mentions "Resource lock". It will retry until reaching the ``max_retries``
+503 (Service Unavailable), 504 (Gateway Timeout), and Cloudflare
+errors (520-530) are retried.
+
+Additionally, HTTP error code 409 (Conflict) is retried if the reason
+is a
+`Resource lock <https://gitlab.com/gitlab-org/gitlab/-/blob/443c12cf3b238385db728f03b2cdbb4f17c70292/lib/api/api.rb#L111>`__.
+
+It will retry until reaching the ``max_retries``
 value. By default, ``retry_transient_errors`` is set to ``False`` and an
 exception is raised for these errors.
 
--- a/docs/api-usage-advanced.rst
+++ b/docs/api-usage-advanced.rst
@@ -123,9 +123,11 @@ GitLab server can sometimes return a transient HTTP error.
 python-gitlab can automatically retry in such case, when
 ``retry_transient_errors`` argument is set to ``True``.  When enabled,
 HTTP error codes 500 (Internal Server Error), 502 (502 Bad Gateway),
-503 (Service Unavailable), and 504 (Gateway Timeout) are retried. It will retry until reaching
-the ``max_retries`` value. By default, ``retry_transient_errors`` is set to ``False`` and an exception
-is raised for these errors.
+503 (Service Unavailable), and 504 (Gateway Timeout) are retried.
+Additionally the HTTP error code 409 (Conflict) is retried if the text message
+mentions "Resource lock". It will retry until reaching the ``max_retries``
+value. By default, ``retry_transient_errors`` is set to ``False`` and an
+exception is raised for these errors.
 
 .. code-block:: python
 
--- a/gitlab/client.py
+++ b/gitlab/client.py
@@ -780,7 +780,10 @@ class Gitlab:
                 return result
 
             if (429 == result.status_code and obey_rate_limit) or (
-                result.status_code in gitlab.const.RETRYABLE_TRANSIENT_ERROR_CODES
+                (
+                    result.status_code in gitlab.const.RETRYABLE_TRANSIENT_ERROR_CODES
+                    or (result.status_code == 409 and "Resource lock" in result.reason)
+                )
                 and retry_transient_errors
             ):
                 # Response headers documentation:
--- a/gitlab/client.py
+++ b/gitlab/client.py
@@ -779,13 +779,20 @@ class Gitlab:
             if 200 <= result.status_code < 300:
                 return result
 
-            if (429 == result.status_code and obey_rate_limit) or (
-                (
-                    result.status_code in gitlab.const.RETRYABLE_TRANSIENT_ERROR_CODES
-                    or (result.status_code == 409 and "Resource lock" in result.reason)
-                )
-                and retry_transient_errors
-            ):
+            def should_retry() -> bool:
+                if result.status_code == 429 and obey_rate_limit:
+                    return True
+
+                if not retry_transient_errors:
+                    return False
+                if result.status_code in gitlab.const.RETRYABLE_TRANSIENT_ERROR_CODES:
+                    return True
+                if result.status_code == 409 and "Resource lock" in result.reason:
+                    return True
+
+                return False
+
+            if should_retry():
                 # Response headers documentation:
                 # https://docs.gitlab.com/ee/user/admin_area/settings/user_and_ip_rate_limits.html#response-headers
                 if max_retries == -1 or cur_retries < max_retries:
--- a/tests/unit/test_gitlab_http_methods.py
+++ b/tests/unit/test_gitlab_http_methods.py
@@ -372,6 +372,63 @@ def test_http_request_302_put_raises_redirect_error(gl):
     assert "http://example.com/api/v4/user/status" in error_message
 
 
+def test_http_request_on_409_resource_lock_retries(gl_retry):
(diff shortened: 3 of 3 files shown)


lmilbaum:
Unit tests might increase the confidence that the code is working as expected. Would you mind adding unit tests?

docs/api-usage-advanced.rst
```diff
@@ -123,9 +123,11 @@ GitLab server can sometimes return a transient HTTP error.
...
 ``retry_transient_errors`` argument is set to ``True``.  When enabled,
 HTTP error codes 500 (Internal Server Error), 502 (502 Bad Gateway),
-503 (Service Unavailable), and 504 (Gateway Timeout) are retried. It will retry until reaching
-the ``max_retries`` value. By default, ``retry_transient_errors`` is set to ``False`` and an exception
-is raised for these errors.
+503 (Service Unavailable), and 504 (Gateway Timeout) are retried.
+Additionally the HTTP error code 409 (Conflict) is retried if the text message
+mentions "Resource lock". It will retry until reaching the ``max_retries``
+value. By default, ``retry_transient_errors`` is set to ``False`` and an
+exception is raised for these errors.
```
nejch, line 130: Maybe we could link to the Gitlab code for this, it'll be good for future us as well.

```suggestion
503 (Service Unavailable), and 504 (Gateway Timeout) are retried.

Additionally, HTTP error code 409 (Conflict) is retried if the reason
is a 
`Resource lock <https://gitlab.com/gitlab-org/gitlab/-/blob/443c12cf3b238385db728f03b2cdbb4f17c70292/lib/api/api.rb#L111>`__.

It will retry until reaching the ``max_retries``
value. By default, ``retry_transient_errors`` is set to ``False`` and an
exception is raised for these errors.
```

gitlab/client.py
```diff
@@ -764,7 +764,10 @@ def http_request(
                 return result
 
             if (429 == result.status_code and obey_rate_limit) or (
-                result.status_code in RETRYABLE_TRANSIENT_ERROR_CODES
+                (
+                    result.status_code in RETRYABLE_TRANSIENT_ERROR_CODES
+                    or (result.status_code == 409 and "Resource lock" in result.reason)
+                )
```
nejch, line 770: Hehe, it seems like pylint doesn't like the chunky if condition - let's try what I suggested in https://github.com/python-gitlab/python-gitlab/issues/2325#issuecomment-1282310403.

nejch (changes requested):
@pspacek thanks for the quick work here! Just a few nits.

Regarding unit tests and your question in https://github.com/python-gitlab/python-gitlab/issues/2325#issuecomment-1282343954.

We have some vendored code from httmock for a similar reason (redirects). I think you might be able to reuse it here?
https://github.com/python-gitlab/python-gitlab/blob/main/tests/unit/helpers.py

See how it's used in the redirect tests:
https://github.com/python-gitlab/python-gitlab/blob/main/tests/unit/test_gitlab_http_methods.py#L310-L317

Let us know if you need help with that :) That's a shame about responses though.

nejch:
And @pspacek we're also ok to go ahead without the unit tests if it's too complicated and we open a follow-up issue to resolve on our end :)

nejch:
Hi @pspacek! Let us know if you'd like to finish this PR, or if we should push to your branch to get this merged (either way you'll be credited as the commit author). Thanks again for the contribution!

pspacek:
@nejch I'm sorry, I have no time to work on this. Feel free to do whatever you like.

gitlab/client.py
nejch: Done in https://github.com/python-gitlab/python-gitlab/pull/2326/commits/25446cf3297453ef58abc66e701813475423fe59.

docs/api-usage-advanced.rst
nejch: Done in https://github.com/python-gitlab/python-gitlab/pull/2326/commits/25446cf3297453ef58abc66e701813475423fe59.

tests/unit/test_gitlab_http_methods.py
```diff
@@ -372,6 +372,63 @@ def response_callback(
...
+        response: requests.models.Response,
+    ) -> requests.models.Response:
+        """We need a callback that adds a resource lock reason only on first call"""
+        nonlocal retried
+
+        if not retried:
+            response.reason = "Resource lock"
+
+        retried = True
+        return response
```
nejch, line 389: We're mostly doing this because `responses` doesn't support mocking `Response.reason` as @pspacek already found out in https://github.com/python-gitlab/python-gitlab/issues/2325#issuecomment-1282343954.

nejch (approved):
This LGTM but I added a commit in 25446cf3297453ef58abc66e701813475423fe59 to finish the PR so will need someone else to do the merge.

Thanks again @pspacek!

nejch:
> Unit tests might increase the confidence that the code is working as expected. Would you mind adding unit tests?

@lmilbaum I've added a commit on top of the author's to rework and add tests (https://github.com/python-gitlab/python-gitlab/pull/2326/commits/1a37bab35d92d442875829dab9a45a701277400a). Maybe you can take a look and we can merge before the next release.

tests/unit/test_gitlab_http_methods.py
```diff
@@ -372,6 +372,63 @@ def response_callback(
...
+
+    with responses.RequestsMock(response_callback=response_callback) as rsps:
+        rsps.add(
+            method=responses.GET,
+            url=url,
+            status=409,
+            match=helpers.MATCH_EMPTY_QUERY_PARAMS,
+        )
+        rsps.add(
```
lmilbaum, line 398: Not sure I understand what is going on here. Is this command overrides the one in line 392?

gitlab/client.py
```diff
@@ -779,10 +779,20 @@ def http_request(
             if 200 <= result.status_code < 300:
                 return result
 
-            if (429 == result.status_code and obey_rate_limit) or (
-                result.status_code in gitlab.const.RETRYABLE_TRANSIENT_ERROR_CODES
-                and retry_transient_errors
-            ):
+            def should_retry() -> bool:
+                if result.status_code == 429 and obey_rate_limit:
```
lmilbaum, line 783: Probably a nit pick but I would suggest using the `Any` function. Not a show stopper for this PR though.

tests/unit/test_gitlab_http_methods.py
```diff
@@ -372,6 +372,63 @@ def response_callback(
     assert "http://example.com/api/v4/user/status" in error_message
 
 
+def test_http_request_on_409_resource_lock_retries(gl_retry):
+    url = "http://localhost/api/v4/user"
+    retried = False
+
+    def response_callback(
+        response: requests.models.Response,
```
lmilbaum, line 380: Is there a difference between `requests.modles.Response` and `requests.Response`?

lmilbaum (changes requested):

tests/unit/test_gitlab_http_methods.py
nejch: This adds a second response to the `rsps` object which tracks our responses for this context. It ensures the first mocked response will be a 409 and the second a 200, mimicking a successful response after retry.

See https://github.com/getsentry/responses#main-interface, and as we are using a context manager, also https://github.com/getsentry/responses#responses-as-a-context-manager.

tests/unit/test_gitlab_http_methods.py
nejch: Not from my perspective, I'm ok either way! But this is how it is already used in `tests/unit/test_gitlab_http_methods.py` so this was just consistent with the other occurrences. I didn't add the initial code in the other files, I'm guessing IDEs auto-complete it where it is defined when they try to auto-detect the type.

gitlab/client.py
nejch: Not sure I understand here - did you mean `all()`?

The previous implementation with the 409 condition added was too complex for pylint so I split into a more explicit helper with separate if branches.

lmilbaum: Well, it could be a combination of `all` and `any`.  I get that the result my be too complex for pylint and for maintenance in general.

tests/unit/test_gitlab_http_methods.py
lmilbaum: Thank you for the clarification

tests/unit/test_gitlab_http_methods.py
lmilbaum: Thanks. The `Responses` package is a new one for me.

lmilbaum (approved):
LGTM

gitlab/client.py
nejch: Yes, I think for now this might be the most readable option. Maybe the match statement once we're on python 3.10 will make life a bit easier :) Thanks!

gitlab/client.py
```diff
@@ -779,10 +779,20 @@ def http_request(
...
+                    return True
+
+                if not retry_transient_errors:
+                    return False
+                if result.status_code in gitlab.const.RETRYABLE_TRANSIENT_ERROR_CODES:
+                    return True
+                if result.status_code == 409 and "Resource lock" in result.reason:
```
rdesgroppes, line 790: Doesn't seem to work for us - could it be because `Resource lock` is in the response body rather than the HTTP reason (`Conflict`)?

Issue #1192, opened 2020-09-16
Project Push Rules get method returning NoneType


L30Bola:
## Description of the problem, including code/CLI snippet

I have this function

```python
def set_project_push_rules_configuration(gitlab, project_id, **kwargs):
    """
    On kwargs you pass arguments that are mutable using the Gitlab Project's Push Rules API
    Documentation on that: https://docs.gitlab.com/ee/api/projects.html#push-rules
    """

    defaults = {
        "prevent_secrets": True,
    }
    defaults.update(kwargs)
    project = gitlab.projects.get(project_id)
    project_push_rules = project.pushrules.get()
    project_push_rules = set_attrs(project_push_rules, defaults)
    try:
        project_push_rules.save()
    except GitlabUpdateError:
        exception(
            "Couldn't update "
            + project.name
            + "'s (ID: "
            + str(project_id)
            + ") Push Rules configurations!"
        )
```

The above function is available on a Flask App that receives all Gitlab System Hooks and is triggered when a repository is created (the `project_create` event), alongside some other functions:

```python
 if request_json["event_name"] == "project_create" and request_json[
        "path_with_namespace"
    ].startswith("infrastructure/"):
        set_project_configuration(gl, request_json["project_id"])
        remove_project_approval_rules(gl, request_json["project_id"])
        set_project_merge_request_approvals_configuration(
            gl, request_json["project_id"]
        )
        create_basic_directory_structure(gl, request_json["project_id"])
        set_project_master_branch_protected(gl, request_json["project_id"])
        set_project_push_rules_configuration(gl, request_json["project_id"])
```

## Expected Behavior

This `project_push_rules = project.pushrules.get()`, should return a `<ProjectPushRules>` object.

## Actual Behavior

When I try on an already created repository, the above code works. But from the system hook `project_create` event, the line that gets the repository Push Rules (`project_push_rules = project.pushrules.get()`) return None, which then leads to `AttributeError: 'NoneType' object has no attribute 'prevent_secrets'`.

## Specifications

  - python-gitlab version: 2.5.0
  - API version you are using (v3/v4): v4
  - Gitlab server version (or gitlab.com): 13.3.5-ee

The project Pipfile:
```
[[source]]
url = "https://pypi.python.org/simple"
verify_ssl = true
name = "pypi"

[requires]
python_version = "3.8"

[packages]
python-gitlab = "==2.5.0"
requests = "*"
Flask = "==1.1.2"
[...]

PR #2332, opened 2022-10-19
test: fix `test_project_push_rules` test

JohnVillalovos:
Make the `test_project_push_rules` test work.

### DIFF
--- a/tests/functional/api/test_push_rules.py
+++ b/tests/functional/api/test_push_rules.py
@@ -1,11 +1,14 @@
 import pytest
 
+import gitlab
+
 
 @pytest.mark.gitlab_premium
-@pytest.mark.xfail(reason="need to relax RESTObject init for non-dict responses")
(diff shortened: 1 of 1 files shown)


nejch (commented):
I was wondering if raising `GitlabParsingError` is correct for a "valid" response, IIRC we added it for malformed responses and not for real GitLab responses (although here I would expect an empty dict or 404 TBH). Maybe a more specific error or an upstream fix to not return None would be clearer to users.

Here is an old issue that I forgot about completely: https://github.com/python-gitlab/python-gitlab/issues/1192

And the upstream issue: https://gitlab.com/gitlab-org/gitlab/-/issues/8829

nejch:
Makes sense, I just wanted to track that somewhere hence the xfail. I've opened an issue for that :)

nejch (approved):

PR #2336, opened 2022-10-20
feat: Initial support for httpx

lmilbaum:
A followup of #1036
Two clients defined: `requests` and `httpx`. As long as `httpx` is not fully implemented and adopted by the community, keeping the `requests` client as the default.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -26,6 +26,7 @@ repos:
         additional_dependencies:
           - argcomplete==2.0.0
           - pytest==7.1.3
+          - httpx==0.23.0
           - requests==2.28.1
           - requests-toolbelt==0.9.1
         files: 'gitlab/'
@@ -37,6 +38,7 @@ repos:
         additional_dependencies:
           - pytest==7.1.3
           - responses==0.21.0
+          - httpx==0.23.0
           - types-PyYAML==6.0.12
           - types-requests==2.28.11.2
           - types-setuptools==64.0.1
--- a/gitlab/client.py
+++ b/gitlab/client.py
@@ -1,14 +1,11 @@
 """Wrapper for the GitLab API."""
-
 import os
 import re
-import time
 from typing import Any, cast, Dict, List, Optional, Tuple, Type, TYPE_CHECKING, Union
-from urllib import parse
+import sys
 
 import requests
 import requests.utils
-from requests_toolbelt.multipart.encoder import MultipartEncoder  # type: ignore
 
 import gitlab
 import gitlab.config
@@ -16,13 +13,13 @@ import gitlab.const
 import gitlab.exceptions
 from gitlab import http_backends, utils
 
-REDIRECT_MSG = (
-    "python-gitlab detected a {status_code} ({reason!r}) redirection. You must update "
-    "your GitLab URL to the correct URL to avoid issues. The redirection was from: "
-    "{source!r} to {target!r}"
-)
+from .http_clients._request_sclient import _Requests_Client
 
-RETRYABLE_TRANSIENT_ERROR_CODES = [500, 502, 503, 504] + list(range(520, 531))
+if sys.version_info >= (3, 8):
+    from typing import Literal
+else:
+    from typing_extensions import Literal
+
 
 # https://docs.gitlab.com/ee/api/#offset-based-pagination
 _PAGINATION_URL = (
@@ -60,6 +57,8 @@ class Gitlab:
         RequestsBackend http_backend: Backend that will be used to make http requests
     """
 
+    http_client: Any  # Union[_RequestsClient, _HttpxClient]
+
     def __init__(
         self,
         url: Optional[str] = None,
@@ -71,6 +70,12 @@ class Gitlab:
         http_password: Optional[str] = None,
         timeout: Optional[float] = None,
         api_version: str = "4",
+<<<<<<< HEAD
+=======
+        http_library: Optional[Literal["requests", "httpx"]] = "requests",
+        session: Optional[requests.Session] = None,
+        http_client: Any = None,  # Optional[Union[requests.Session,httpx.Client]]
+>>>>>>> 884daf3 (feat: Initial support for httpx)
         per_page: Optional[int] = None,
         pagination: Optional[str] = None,
         order_by: Optional[str] = None,
@@ -79,7 +84,42 @@ class Gitlab:
         keep_base_url: bool = False,
         **kwargs: Any,
     ) -> None:
+<<<<<<< HEAD
+=======
+
+        # We only support v4 API at this time
+        if api_version not in ("4",):
+            raise ModuleNotFoundError(f"gitlab.v{api_version}.objects")
+        # NOTE: We must delay import of gitlab.v4.objects until now or
+        # otherwise it will cause circular import errors
+        from gitlab.v4 import objects
+
+>>>>>>> 884daf3 (feat: Initial support for httpx)
         self._api_version = str(api_version)
+        self.job_token = job_token
+
+        if http_library == "requests":
+            self.http_client = _Requests_Client(
+                url=url,
+                session=session,
+                retry_transient_errors=retry_transient_errors,
+                ssl_verify=ssl_verify,
+                timeout=timeout,
+                http_username=http_username,
+                http_password=http_password,
+                job_token=job_token,
+                oauth_token=oauth_token,
+                private_token=private_token,
+                user_agent=user_agent,
+            )
+            self.session = self.http_client.get_session
+            self._http_auth = self.http_client.get_auth_info()
+            self.headers = self.http_client.get_headers
+        elif http_library == "httpx":
+            from .http_clients._httpx_client import _Httpx_Client
+
+            self.http_client = _Httpx_Client(http_client=http_client)
+
         self._server_version: Optional[str] = None
         self._server_revision: Optional[str] = None
         self._base_url = self._get_base_url(url)
@@ -88,16 +128,13 @@ class Gitlab:
         self.timeout = timeout
         self.retry_transient_errors = retry_transient_errors
         self.keep_base_url = keep_base_url
-        #: Headers that will be used in request to GitLab
-        self.headers = {"User-Agent": user_agent}
 
         #: Whether SSL certificates should be validated
         self.ssl_verify = ssl_verify
 
         self.private_token = private_token
-        self.http_username = http_username
-        self.http_password = http_password
         self.oauth_token = oauth_token
+<<<<<<< HEAD
         self.job_token = job_token
         self._set_auth_info()
 
@@ -107,17 +144,12 @@ class Gitlab:
         )
         self.http_backend = http_backend(**kwargs)
         self.session = self.http_backend.client
+=======
+>>>>>>> 884daf3 (feat: Initial support for httpx)
 
         self.per_page = per_page
         self.pagination = pagination
         self.order_by = order_by
-
-        # We only support v4 API at this time
-        if self._api_version not in ("4",):
-            raise ModuleNotFoundError(f"gitlab.v{self._api_version}.objects")
-        # NOTE: We must delay import of gitlab.v4.objects until now or
-        # otherwise it will cause circular import errors
-        from gitlab.v4 import objects
 
         self._objects = objects
         self.user: Optional[objects.CurrentUser] = None
@@ -203,7 +235,7 @@ class Gitlab:
         return self
 
     def __exit__(self, *args: Any) -> None:
-        self.session.close()
+        self.http_client.__exit__(self.http_client, *args)
 
     def __getstate__(self) -> Dict[str, Any]:
         state = self.__dict__.copy()
@@ -243,7 +275,10 @@ class Gitlab:
         cls,
         gitlab_id: Optional[str] = None,
         config_files: Optional[List[str]] = None,
+<<<<<<< HEAD
         **kwargs: Any,
+=======
+>>>>>>> 884daf3 (feat: Initial support for httpx)
     ) -> "Gitlab":
         """Create a Gitlab connection from configuration files.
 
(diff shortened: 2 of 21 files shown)


requirements.txt
```diff
@@ -1,2 +1,3 @@
-requests==2.28.1
+httpx==0.23.0
```
nejch, line 1: Will also need an `extra_requires` entry in setup.py, so people ensure they have it with `pip install python-gitlab[httpx]`.

gitlab/clients/httpx_client.py
```diff
@@ -0,0 +1,11 @@
+from typing import Optional
```
nejch, line 1: Not sure if we need separate modules at this point as they're quite small.

gitlab/clients/httpx_client.py
```diff
@@ -0,0 +1,11 @@
...
+
+import httpx
+
+
+class Httpx_Client:
```
nejch, line 6: I would follow camelcase and make them private here, so people know it's an internal implementation only :)

```suggestion
class _HttpxClient:
```

gitlab/client.py
```diff
@@ -2,10 +2,12 @@
 
 import os
 import re
+import sys
 import time
 from typing import Any, cast, Dict, List, Optional, Tuple, TYPE_CHECKING, Union
 from urllib import parse
 
+import httpx
```
nejch, line 10: This should probably be in a try/except clause and we should only throw an error if people explicitly want to use httpx. We don't want to pull in a new dependency by default if not used, IMO.

I did something similar for graphql here:

https://github.com/python-gitlab/python-gitlab/blob/b8ca5ffb69dce6a7f2f1eb07aeec2216d6817bc5/gitlab/client.py#L1248-L1258

gitlab/client.py
```diff
@@ -97,8 +109,11 @@ def __init__(
         self.job_token = job_token
         self._set_auth_info()
 
-        #: Create a session object for requests
-        self.session = session or requests.Session()
+        if http == "requests":
+            self._requests_client = Requests_Client(session=session)
+            self.session = self._requests_client.get_session
+        elif http == "httpx":
+            self._httpx_client = Httpx_Client(client=client)
```
nejch, line 116: Maybe just `self._client` is enough :) as they are mostly API-compatible for making requests.

But I think we will need some more overriding based on multipart-encoding in some of our helper methods. I would try to make a matrix in functional test where we run them with httpx and then you should start seeing some issues with that.

nejch (commented):
@lmilbaum I made a quick initial pass. Would be interested to see how this runs on functional tests. Thanks a lot for working on this, it's something I've been thinking of myself for a while!

gitlab/client.py
```diff
@@ -16,6 +18,14 @@
 import gitlab.exceptions
 from gitlab import utils
 
+from .clients.httpx_client import Httpx_Client
+from .clients.requests_client import Requests_Client
+
+if sys.version_info >= (3, 8):
+    from typing import Literal
+else:
+    from typing_extensions import Literal
```
nejch, line 27: If we do this, it also needs to be a conditional dependency in setup.py when python =< 3.7.

requirements.txt
lmilbaum: Added to the "install_requires" key where the reqests package is listed

gitlab/clients/httpx_client.py
lmilbaum: The intention is moving the functionality to the clients classes.  This PR will lay the ground for the coming changes.

gitlab/clients/httpx_client.py
lmilbaum: Done

gitlab/client.py
lmilbaum: Makes sense. Just on quick question. How can we know upfront (before the package is installed) if a user wants to use one of the packages and not the other?

gitlab/client.py
lmilbaum: I am not sure I know what `multipart-encoding` is

gitlab/client.py
lmilbaum: OK. Worth adding a functional test which checks that. I've implemented something similar in a different project where I used a container to install the package and use it as the environment under test.

gitlab/client.py
nejch: I think we can import at the top initially with a try/except and then later when it needs to be used (e.g. when the user provides the httpx client), if it is `None` we can raise an ImportError.
```python
try:
    import httpx
except ImportError:
    httpx = None

# ...

if not httpx:
     raise ImportError("... please install with python-gitlab[httpx]")
```

We do something similar in our cli modules:
https://github.com/python-gitlab/python-gitlab/blob/9b88132078ed37417c2a45369b4976c9c67f7882/gitlab/v4/cli.py#L421-L430

httpx itself is actually a good example of this kind of pattern with all its optional transports:

https://github.com/encode/httpx/blob/14a1704be507cf435fa84da75c249c2ecd1c3981/httpx/_transports/default.py#L158-L164

gitlab/client.py
nejch: Let's see when we start running functional tests with httpx. I think we will run into issues then and we can debug.

requirements.txt
nejch: Ah, see my other comment about optional dependencies - IMO extras are a good place for this to avoid adding it by default :)

gitlab/clients/httpx_client.py
nejch: Ah ok :) I'd personally go for `httpx.py` and `requests.py` as we already have the namespace.

gitlab/client.py
lmilbaum: IMO, it is in the scope of this PR.  This is how a maintainer can get confidence that the contributed change is not breaking something that worked before. Will work on it.

requirements.txt
lmilbaum: Yeah. Pushed a fix.

gitlab/clients/httpx_client.py
lmilbaum: I like the file name to be aligned with the class name

gitlab/client.py
lmilbaum: @nejch I added a new workflow which installs the package and test whether `httpx` package can be imported or not. Not sure if this covers what you mentioned in your first comment. WDYT?

lmilbaum:
@nejch The PR is waiting for review quite some time. I can understand that it is not a trivial PR. How about me setting a call for a merge party?

nejch:
> @nejch The PR is waiting for review quite some time. I can understand that it is not a trivial PR. How about me setting a call for a merge party?

Sorry for the delay @lmilbaum, I was away a few days and there was a backlog of smaller PRs/issues :sweat_smile: 

I have a few pending comments and I wanted to have a closer look before finishing review as it's a larger change. We can also set up a call for later this week, sure, although I'll try to do another pass in the meantime. I'll send my pending comments for now.

gitlab/client.py
```diff
@@ -75,7 +81,34 @@ def __init__(
...
+                http_password=http_password,
+                job_token=job_token,
+                oauth_token=oauth_token,
+                private_token=private_token,
+            )
+            self.session = self.client.get_session
+        elif http == "httpx":
+            from .clients._httpxclient import _HttpxClient
+
+            self.client = _HttpxClient(client=client)
```
nejch, line 110: Is it necessary that we offload all of this to the backend clients? It mostly just sets up auth headers, so ideally we could keep most of the logic in the existing client.

I have a suspicion some of this is needed because of `HTTPBasicAuth` in `_set_auth_info()`. In fact, this is sort of dead code as GitLab has long disabled basic auth via username/password. So instead we could perhaps remove that part of the code. WDYT?

setup.py
```diff
@@ -52,5 +52,6 @@ def get_version() -> str:
     extras_require={
         "autocompletion": ["argcomplete>=1.10.0,<3"],
         "yaml": ["PyYaml>=5.2"],
+        "httpx": ["httpx==0.23.0"],
```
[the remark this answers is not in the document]
lmilbaum: Makes sense.

tests/install/test_install.py
```diff
@@ -0,0 +1,7 @@
+def test_install() -> None:
+    try:
+        import httpx  # noqa
+
+        assert False
+    except ImportError:
+        assert True
```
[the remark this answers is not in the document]
lmilbaum: Right. I haven't used it for a while. Thanks.

gitlab/client.py
lmilbaum: I would suggest to break it into two steps. Right now to do the minimal change of moving the same code from the general client to the requests client and do some cleanup in a follow up PR.

gitlab/client.py
```diff
@@ -384,60 +407,6 @@ def version(self) -> Tuple[str, str]:
 
         return cast(str, self._server_version), cast(str, self._server_revision)
 
-    @gitlab.exceptions.on_http_error(gitlab.exceptions.GitlabVerifyError)
-    def lint(self, content: str, **kwargs: Any) -> Tuple[bool, List[str]]:
```
JohnVillalovos, line 388: Why is this method moved? It makes  it harder to review the changes, IMHO.

gitlab/client.py
```diff
@@ -384,60 +407,6 @@ def version(self) -> Tuple[str, str]:
...
-            category=DeprecationWarning,
-        )
-        post_data = {"content": content}
-        data = self.http_post("/ci/lint", post_data=post_data, **kwargs)
-        if TYPE_CHECKING:
-            assert not isinstance(data, requests.Response)
-        return (data["status"] == "valid", data["errors"])
-
-    @gitlab.exceptions.on_http_error(gitlab.exceptions.GitlabMarkdownError)
-    def markdown(
```
JohnVillalovos, line 414: Why is this method moved? It makes  it harder to review the changes, IMHO.

gitlab/clients/_httpxclient.py
```diff
@@ -0,0 +1,103 @@
...
+        self,
+        url: Optional[str] = None,
+        client: Optional[httpx.Client] = None,
+    ) -> None:
+        self._client = client or httpx.Client()
+        self._base_url = self._get_base_url(url)
+
+    def __enter__(self: _HttpxClient) -> _HttpxClient:
+        return self
+
```
JohnVillalovos, line 29: Should there be an `__exit__` method?

gitlab/client.py
```diff
@@ -1,34 +1,36 @@
...
+    from typing_extensions import Literal
 
-RETRYABLE_TRANSIENT_ERROR_CODES = [500, 502, 503, 504] + list(range(520, 531))
 
 # https://docs.gitlab.com/ee/api/#offset-based-pagination
 _PAGINATION_URL = (
     f"https://python-gitlab.readthedocs.io/en/v{gitlab.__version__}/"
     f"api-usage.html#pagination"
 )
+REDIRECT_MSG = (
```
JohnVillalovos, line 29: Maybe this could use the one in the requestsclient.py file or vice-versa.

Or move it to the constants file. Basically only have one copy of it, if possible.

gitlab/mixins.py
```diff
@@ -56,7 +56,7 @@ class HeadMixin(_RestManagerBase):
     @exc.on_http_error(exc.GitlabHeadError)
     def head(
         self, id: Optional[Union[str, int]] = None, **kwargs: Any
-    ) -> "requests.structures.CaseInsensitiveDict[Any]":
+    ) -> Union["requests.structures.CaseInsensitiveDict[Any]", Any]:
```
JohnVillalovos, line 59: Why is the `Any` needed? And for all the other ones.

gitlab/clients/_httpxclient.py
```diff
@@ -0,0 +1,103 @@
+from __future__ import annotations
+
```
JohnVillalovos, line 2: Not sure what others think. But I would prefer the name to be either `_httpx_client` or `_httpx`. And the same idea on the current `_requestsclient`. But others may hate that idea!

JohnVillalovos (changes requested):
@lmilbaum Thanks for all the work on this!!

I just did a quick review. If you disagree with my comments don't be afraid to say so 🙂 I am often wrong!

One concern I have is all the adding of `Any` to a lot of the type-hints.

lmilbaum:
> @lmilbaum Thanks for all the work on this!!
> 
> I just did a quick review. If you disagree with my comments don't be afraid to say so 🙂 I am often wrong!
> 
> One concern I have is all the adding of `Any` to a lot of the type-hints.

Yeah. I can understand why. I didn't find any other solution. The understanding that this is an interim step, helps me go for this option.

.github/workflows/functional.yml
```diff
@@ -0,0 +1,50 @@
+name: "Functional"
```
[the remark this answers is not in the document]
lmilbaum: Right :-). Changed

gitlab/client.py
```diff
@@ -66,7 +70,9 @@ def __init__(
         http_password: Optional[str] = None,
         timeout: Optional[float] = None,
         api_version: str = "4",
+        http: Optional[Literal["requests", "httpx"]] = "requests",
```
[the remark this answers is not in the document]
lmilbaum: Yeah. Yet another agreement :-)

.pre-commit-config.yaml
```diff
@@ -23,9 +23,10 @@ repos:
     rev: v2.15.5
     hooks:
       - id: pylint
-        additional_dependencies:
+        additional_dependencies:
```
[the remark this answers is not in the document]
lmilbaum: By mistake. Removed.

gitlab/client.py
lmilbaum: I agree. It is probably because I moved few function from this file to another.

gitlab/mixins.py
lmilbaum: Because the httpx client function will return different type which I didn't want to set at this stage. Right now it is a placeholder for the next implementation phase.

gitlab/clients/_httpxclient.py
lmilbaum: I don't have any strong feeling to either option.

nejch:
>> @lmilbaum Thanks for all the work on this!!
>> I just did a quick review. If you disagree with my comments don't be afraid to say so slightly_smiling_face I am often wrong!
>> One concern I have is all the adding of Any to a lot of the type-hints.
>
> Yeah. I can understand why. I didn't find any other solution. The understanding that this is an interim step, helps me go for this option.

Thanks both for driving this, sorry again this is taking a while for me to get back into, I thought this would be easier and only the most low-level stuff factored out, similar to how it's done in https://python-redmine.com/advanced/request_engines.html /  https://github.com/maxtepkeev/python-redmine/tree/master/redminelib/engines. So just wanted to make sure we take a good look before we merge if there's a simple way :)

lmilbaum:
> > > @lmilbaum Thanks for all the work on this!!
> > > I just did a quick review. If you disagree with my comments don't be afraid to say so slightly_smiling_face I am often wrong!
> > > One concern I have is all the adding of Any to a lot of the type-hints.
> > 
> > 
> > Yeah. I can understand why. I didn't find any other solution. The understanding that this is an interim step, helps me go for this option.
> 
> Thanks both for driving this, sorry again this is taking a while for me to get back into, I thought this would be easier and only the most low-level stuff factored out, similar to how it's done in https://python-redmine.com/advanced/request_engines.html / https://github.com/maxtepkeev/python-redmine/tree/master/redminelib/engines. So just wanted to make sure we take a good look before we merge if there's a simple way :)

I liked the approach taken in `python-redmine`.  The question I have with this approach is wether it goes along with our approach of eventually dropping the `request` http client/engine and stay with only `httpx`.

lmilbaum:
Would breaking this PR into smaller pieces would make sense?

nejch:
> Would breaking this PR into smaller pieces would make sense?

I think this might help yes @lmilbaum!
It would probably be a smaller PR if we don't take out all the methods as @JohnVillalovos suggested:

https://github.com/python-gitlab/python-gitlab/pull/2336/files#r1010073025
https://github.com/python-gitlab/python-gitlab/pull/2336/files#r1010073368

Also, it would potentially be a smaller PR if we ensure we keep public `http_*` methods in the main client so that we don't need to change the tests, because this is a public API on the Gitlab instance. We may just have private methods in backend clients that these wrap, but we somehow need to support this for now IMO.

https://python-gitlab.readthedocs.io/en/stable/api-levels.html

JohnVillalovos:
So biggest issue with this for me is the type-hints. Everything having `Any` added. I worked a LONG time adding type-hints to the code and this kind of throws it all away :(

So I think we need to figure out a solution that doesn't use `Any`.

lmilbaum:
Thank you for the feedback @nejch and @JohnVillalovos. I've started breaking this up into smaller PRs. Take a look at #2382

Issue #2338, opened 2022-10-21
`error: Value of type "Union[GitlabList, List[Dict[str, Any]]]" is not indexable  [index]`


stdedos:
## Description of the problem, including code/CLI snippet

```py
	....

    ci_commit_sha = os.environ["CI_COMMIT_SHA"]
    mrs_related = project.commits.get(ci_commit_sha).merge_requests()

    if not mrs_related:
        raise ValueError(f"ci_commit_sha '{ci_commit_sha}' has no associated MRs!")

    if len(mrs_related) > 1:
        msg = f"ci_commit_sha '{ci_commit_sha}' has >1 associated MRs: [{csv(mrs_related)}]"
        warnings.warn(msg)

    return mrs_related[0]["iid"]
```

## Expected Behavior

_No mypy errors_

## Actual Behavior

_mypy errors_

## Specifications

  - python-gitlab version: 3.10.0
  - API version you are using (v3/v4): v4
  - Gitlab server version (or gitlab.com): gitlab.com

JohnVillalovos:
That sounds correct. As `GitlabList` is an iterator and not indexable.

You may want to put an `assert` in your code:
```
mrs_related = project.commits.get(ci_commit_sha).merge_requests()

assert isinstance(mrs_related, list)

# or
if TYPE_CHECKING:
    assert isinstance(mrs_related, list)
```

nejch:
Hehe @JohnVillalovos we discussed this in https://gitter.im/python-gitlab/Lobby?at=63528e2388e38704c020c5ff and I suggested the same but said to open an issue so we can track it.

It is a list in this case. We would need overloading to infer depending on whether `iterator=True` was passed:

```python
>>> mrs_related = commit.merge_requests()
>>> type(mrs_related)
<class 'list'>
>>> mrs_related_iterator = commit.merge_requests(iterator=True)
>>> type(mrs_related_iterator)
<class 'gitlab.client.GitlabList'>
```

stdedos:
Can I suggest using a different "external" interface, but the same internal? 😕 (`merge_requests` and `merge_requests_iterator`, with the original `merge_requests` as `_merge_requests`).

_or idk how else would it be possible to statically write such typing (and therefore I'm interested to see the PR fixing only this, if possible)_

[...]

PR #2339, opened 2022-10-22
chore: add basic typing to functional tests

nejch:
Related to https://github.com/python-gitlab/python-gitlab/issues/2338.

This was going to be the start of adding a bit of pragmatic/non-pedantic typing to tests, not so much to type the test code itself, but to catch typing issues in our own package before it gets to users, and then gradually removing suppressions with fixes.

In the end I think it might do more damage than good to readability and productivity because of the amount of additional boilerplate for `test_*` functions needed for no gain (as some maintainers suggested in https://github.com/pytest-dev/pytest/issues/5981). Maybe with some kind of plugin this can be avoided in an elegant way. It might make sense to at least type fixtures when explicitly used in tests eventually.

But since I started something might as well open a PR (folders done in separate commits). I added some docs instead on narrowing in user code.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -35,6 +35,7 @@ repos:
       - id: mypy
         args: []
         additional_dependencies:
+          - pytest==7.1.3
           - types-PyYAML==6.0.12
           - types-requests==2.28.11.2
           - types-setuptools==64.0.1
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -36,6 +36,7 @@ repos:
         args: []
         additional_dependencies:
           - pytest==7.1.3
+          - responses==0.21.0
           - types-PyYAML==6.0.12
           - types-requests==2.28.11.2
           - types-setuptools==64.0.1
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -30,7 +30,7 @@ repos:
           - requests-toolbelt==0.9.1
         files: 'gitlab/'
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v0.961
+    rev: v0.981
     hooks:
       - id: mypy
         args: []
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,13 +14,15 @@ module = [
     "docs.ext.*",
     "tests.functional.*",
     "tests.functional.api.*",
-    "tests.meta.*",
     "tests.unit.*",
 ]
 ignore_errors = true
 
 [[tool.mypy.overrides]]
-module = "tests.smoke.*"
+module = [
+    "tests.meta.*",
+    "tests.smoke.*",
+]
 disable_error_code = ["no-untyped-def"]
 
 [tool.semantic_release]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,11 +8,11 @@ files = "."
 exclude = "build/.*"
 strict = true
 
-[[tool.mypy.overrides]] # Overrides for currently untyped modules
+# Overrides for currently untyped modules
+[[tool.mypy.overrides]]
 module = [
     "docs.*",
     "docs.ext.*",
-    "tests.functional.api.*",
     "tests.unit.*",
 ]
 ignore_errors = true
@@ -20,6 +20,7 @@ ignore_errors = true
 [[tool.mypy.overrides]]
 module = [
     "tests.functional.*",
+    "tests.functional.api.*",
     "tests.meta.*",
     "tests.smoke.*",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,7 +12,6 @@ strict = true
 module = [
     "docs.*",
     "docs.ext.*",
-    "tests.functional.*",
     "tests.functional.api.*",
     "tests.unit.*",
 ]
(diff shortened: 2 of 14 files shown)


tests/smoke/test_dists.py
```diff
@@ -18,9 +18,9 @@
 
 
 @pytest.fixture(scope="function")
-def build():
-    sandbox.run_setup("setup.py", ["--quiet", "clean", "--all"])
-    return sandbox.run_setup("setup.py", ["--quiet", "sdist", "bdist_wheel"])
+def build() -> None:
+    sandbox.run_setup("setup.py", ["--quiet", "clean", "--all"])  # type: ignore[no-untyped-call]
+    sandbox.run_setup("setup.py", ["--quiet", "sdist", "bdist_wheel"])  # type: ignore[no-untyped-call]
```
nejch, line 23: Not available in typeshed, but not worth it as setup.py-based builds already raise deprecation warnings.

lmilbaum:
@nejch One quick observation and feedback if I may.  In this project it is common to submit PRs with multiple commits. There are arguments for supporting this practice vs the other which I am practicing - one commit per PR. The argument which I find useful is that CI is invoked per PR, not per commit. Therefore, there might be commits in a PR which might break CI.  This comment can be ignored in case a PR is merged to main while squashing all the commits into one commit.

JohnVillalovos:
> @nejch One quick observation and feedback if I may. In this project it is common to submit PRs with multiple commits. There are arguments for supporting this practice vs the other which I am practicing - one commit per PR. The argument which I find useful is that CI is invoked per PR, not per commit. Therefore, there might be commits in a PR which might break CI. This comment can be ignored in case a PR is merged to main while squashing all the commits into one commit.

I am fine with multiple commits per PR. And also don't usually squash commits unless things are a mess.

Yes CI could break in-between commits, but not sure that is a problem. Especially for a volunteer project of this type.  If we were writing code for medical devices or launching rockets into space, then my opinion would be different. I would also expect to be getting paid for this 🙂

nejch:
> @nejch One quick observation and feedback if I may. In this project it is common to submit PRs with multiple commits. There are arguments for supporting this practice vs the other which I am practicing - one commit per PR. The argument which I find useful is that CI is invoked per PR, not per commit. Therefore, there might be commits in a PR which might break CI. This comment can be ignored in case a PR is merged to main while squashing all the commits into one commit.

@lmilbaum no hard requirements from me here, I generally also don't like extra commits for fixing up code (which I violated right here https://github.com/python-gitlab/python-gitlab/pull/2339/commits/57ec33cae1b520a26ce597fcdb6cc4058f104df3), but the multiple commits here were otherwise pretty intentional and each was _meant_ to be a working baseline :grin: 

We generate changelogs from commit messages so I want them to be meaningful (the docs addition here would not end up in release notes if it was just in a test/chore commit). And if someone is doing git blame 4 years from now on, https://github.com/python-gitlab/python-gitlab/pull/2339/commits/2dbb473e5f4c51ca7e75ae25fe25d2396f932ee8 seems cleaner to me than including it in a big chunky commit with unrelated test changes. 

This PR was a bit extreme and probably not the best example, I usually would also do around 1 commit per PR. Just when it's doing these kinds of tedious chores over a bunch of files, I do multiple commits each relevant to its topic or scope rather than 1 commit across the entire codebase when changing things. Just a personal thing, not trying to say this project should be doing it this way :)

>  If we were writing code for medical devices or launching rockets into space, then my opinion would be different.

Haha good one but I can confidently tell you this project has much better commit hygiene than a lot of code that ends up running critical infrastructure :sweat_smile:

There are a lot of opinions here I guess, some people don't like single commits with rebases as reviewing can be harder and authorship from review suggestions is lost etc.

JohnVillalovos:
One thing I forgot to mention. I do not like multiple commits when they are fixes to review comments.  Those should be rebased.

I usually do prefer it when it is one commit per PR. But I can see times when that might not be the best approach. So I guess it is "it depends" 😊

lmilbaum:
@JohnVillalovos @nejch Thank you for taking the time to provide a thoughtful feedback. Maybe, it worth taking into consideration this topic when PRs are reviewed.

JohnVillalovos (approved):
Thanks! Sorry I didn't notice before it had left "Draft" status.

.pre-commit-config.yaml
```diff
@@ -30,11 +30,13 @@ repos:
           - requests-toolbelt==0.9.1
         files: 'gitlab/'
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v0.961
```
lmilbaum, line 33: It is interesting to check if renovate caches those cases.

pyproject.toml
```diff
@@ -6,37 +6,26 @@ order_by_type = false
...
     "docs.ext.*",
-    "tests.*",
-    "tests.functional.*",
-    "tests.functional.api.*",
     "tests.unit.*",
-    "tests.smoke.*"
 ]
 ignore_errors = true
 
+[[tool.mypy.overrides]]
```
lmilbaum, line 20: What does it mean to have this section the second time?

tests/functional/conftest.py
```diff
@@ -56,7 +56,8 @@ def gitlab_container_name() -> str:
 
 @pytest.fixture(scope="session")
 def gitlab_docker_port(docker_services, gitlab_service_name: str) -> int:
-    return docker_services.port_for(service=gitlab_service_name, container_port=80)
+    port: int = docker_services.port_for(gitlab_service_name, container_port=80)
```
lmilbaum, line 59: Out of curiosity, what is the need for a local variable?

PR #2347, opened 2022-10-28
feat(api): add application statistics

### LINKED ISSUE #2264: Add application statistics

The Gitlab provides some general statistics via `/application/statistics`.
See https://docs.gitlab.com/ee/api/statistics.html

## Expected Behavior

These data are available via `gitlab.statistics()`.

## Actual Behavior

Nothing.

## Specifications

  - python-gitlab version:

Shreya-7:
resolves #2264

### DIFF
--- a/tests/functional/api/test_statistics.py
+++ b/tests/functional/api/test_statistics.py
@@ -1,0 +1,12 @@
+"""
+GitLab API: https://docs.gitlab.com/ee/api/statistics.html
+"""
+
+
+def test_get_statistics(gl):
+    statistics = gl.statistics.get()
+
+    assert statistics.snippets.isdigit()
+    assert statistics.users.isdigit()
+    assert statistics.groups.isdigit()
+    assert statistics.projects.isdigit()
--- a/tests/functional/conftest.py
+++ b/tests/functional/conftest.py
@@ -94,7 +94,18 @@ def reset_gitlab(gl: gitlab.Gitlab) -> None:
             helpers.safe_delete(deploy_token)
         logging.info(f"Deleting project: {project.path_with_namespace!r}")
         helpers.safe_delete(project)
+
     for group in gl.groups.list():
+
+        # skip deletion of a descendant group to prevent scenarios where parent group
+        # gets deleted leaving a dangling descendant whose deletion will throw 404s.
+        if group.parent_id:
+            logging.info(
+                f"Skipping deletion of {group.full_path} as it is a descendant "
+                f"group and will be removed when the parent group is deleted"
+            )
+            continue
+
         for deploy_token in group.deploytokens.list():
             logging.info(
                 f"Deleting deploy token: {deploy_token.username!r} in "
@@ -110,7 +121,7 @@ def reset_gitlab(gl: gitlab.Gitlab) -> None:
         logging.info(f"Deleting variable: {variable.key!r}")
         helpers.safe_delete(variable)
     for user in gl.users.list():
-        if user.username != "root":
+        if user.username not in ["root", "ghost"]:
             logging.info(f"Deleting user: {user.username!r}")
             helpers.safe_delete(user, hard_delete=True)
 
--- a/tests/unit/objects/test_statistics.py
+++ b/tests/unit/objects/test_statistics.py
@@ -1,0 +1,49 @@
+"""
+GitLab API: https://docs.gitlab.com/ee/api/statistics.html
+"""
+
+import pytest
+import responses
+
+content = {
+    "forks": "10",
+    "issues": "76",
+    "merge_requests": "27",
+    "notes": "954",
+    "snippets": "50",
+    "ssh_keys": "10",
+    "milestones": "40",
+    "users": "50",
+    "groups": "10",
+    "projects": "20",
+    "active_users": "50",
+}
+
+
+@pytest.fixture
+def resp_application_statistics():
+
+    with responses.RequestsMock() as rsps:
+        rsps.add(
+            method=responses.GET,
+            url="http://localhost/api/v4/application/statistics",
+            json=content,
+            content_type="application/json",
+            status=200,
+        )
+
+        yield rsps
+
+
+def test_get_statistics(gl, resp_application_statistics):
+    statistics = gl.statistics.get()
+    assert statistics.forks == content["forks"]
+    assert statistics.merge_requests == content["merge_requests"]
+    assert statistics.notes == content["notes"]
+    assert statistics.snippets == content["snippets"]
+    assert statistics.ssh_keys == content["ssh_keys"]
+    assert statistics.milestones == content["milestones"]
+    assert statistics.users == content["users"]
+    assert statistics.groups == content["groups"]
+    assert statistics.projects == content["projects"]
+    assert statistics.active_users == content["active_users"]
--- a/docs/api-objects.rst
+++ b/docs/api-objects.rst
@@ -54,6 +54,7 @@ API examples
    gl_objects/search
    gl_objects/settings
    gl_objects/snippets
+   gl_objects/statistics
    gl_objects/system_hooks
    gl_objects/templates
    gl_objects/todos
--- a/docs/gl_objects/statistics.rst
+++ b/docs/gl_objects/statistics.rst
@@ -1,0 +1,21 @@
+##########
+Statistics
+##########
+
+Reference
+---------
+
+* v4 API:
+
+  + :class:`gitlab.v4.objects.ApplicationStatistics`
+  + :class:`gitlab.v4.objects.ApplicationStatisticsManager`
+  + :attr:`gitlab.Gitlab.statistics`
+
+* GitLab API: https://docs.gitlab.com/ee/api/statistics.html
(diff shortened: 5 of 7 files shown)


Shreya-7:
@nejch had a quick question. I've written a functional test for this functionality where I'm creating some entities (projects, groups, snippets, users, etc.) and then checking whether the statistics returned match or not (number of projects created = number from statistics).
However, this test keeps failing because there are a number of projects created in other functional tests that were not cleaned up, and hence those numbers are interfering with the test. Similarly for other entities.

I have two options here: either I go around modifying those tests to clean up any entities they create, or I do a count of created  (and undeleted) entities and just modify my test accordingly.
I would definitely prefer the first one, since it makes my test independent of others -- what do you think?

lmilbaum:
> @nejch had a quick question. I've written a functional test for this functionality where I'm creating some entities (projects, groups, snippets, users, etc.) and then checking whether the statistics returned match or not (number of projects created = number from statistics). However, this test keeps failing because there are a number of projects created in other functional tests that were not cleaned up, and hence those numbers are interfering with the test. Similarly for other entities.
> 
> I have two options here: either I go around modifying those tests to clean up any entities they create, or I do a count of created (and undeleted) entities and just modify my test accordingly. I would definitely prefer the first one, since it makes my test independent of others -- what do you think?

Jumping in if I may. The fixture `reset_gitlab` can do the trick.

tests/unit/objects/test_statistics.py
```diff
@@ -0,0 +1,49 @@
...
+    "ssh_keys": "10",
+    "milestones": "40",
+    "users": "50",
+    "groups": "10",
+    "projects": "20",
+    "active_users": "50",
+}
+
+
+@pytest.fixture
```
lmilbaum, line 23: Would it make sense to move the fixture code to conftest.py file?

tests/functional/api/test_statistics.py
```diff
@@ -0,0 +1,37 @@
+"""
+GitLab API: https://docs.gitlab.com/ee/api/statistics.html
+"""
+
+
+from tests.functional.conftest import reset_gitlab
```
lmilbaum, line 6: fixtures doesn't have to be imported. Just add the fixture to your function arguments and it will be executed.

tests/unit/objects/test_statistics.py
Shreya-7: I'm not sure if any other unit tests will need this particular fixture, I'm following the same pattern as other unit tests

Shreya-7:
> Jumping in if I may. The fixture `reset_gitlab` can do the trick.

Just what I was looking for, thanks!

tests/functional/api/test_statistics.py
Shreya-7: `reset_gitlab` isn't a fixture as of now - it's just a helper function as far as I can see. Not sure if we'd want to convert it into one as a function call is made to it within a different fixture.

tests/unit/objects/test_statistics.py
nejch: I also think in this case it makes sense to keep the fixtures here, close to the scope where they are used as we do with other endpoint-specific fixtures:

nejch:
Sorry @Shreya-7 I was away on a long weekend and catching up a bit!

> I have two options here: either I go around modifying those tests to clean up any entities they create, or I do a count of created (and undeleted) entities and just modify my test accordingly.
I would definitely prefer the first one, since it makes my test independent of others -- what do you think?

I would actually keep it even simpler than that, and only check that the attributes in the response can be successfully cast to an int but not actually assert on their values, that way we know the server response makes sense.

GitLab does a lot of creation and cleanup tasks asynchronously and I would not rely on absolute values here. We've actually tried to move away from that in some other tests - see https://github.com/python-gitlab/python-gitlab/pull/2118 for context.

That said, I see you've done a lot of work on getting `reset_gitlab` to behave. So we can keep that as it likely helps the testing in any case (in a separate commit maybe). WDYT?

tests/functional/conftest.py
```diff
@@ -94,7 +94,17 @@ def reset_gitlab(gl: gitlab.Gitlab) -> None:
             helpers.safe_delete(deploy_token)
         logging.info(f"Deleting project: {project.path_with_namespace!r}")
         helpers.safe_delete(project)
+
     for group in gl.groups.list():
+
+        # skip deletion of a descendant group to prevent scenarios where parent group gets deleted leaving a dangling descendant whose deletion will throw 404s.
+        # descendant groups cannot contain the `/` special character: https://docs.gitlab.com/ee/user/reserved_names.html#limitations-on-project-and-group-names
+        if "/" in group.full_path:
```
nejch, line 102: This might be simpler, check if it has a parent:
```suggestion
        if group.parent_id:
```

Shreya-7:
>I would actually keep it even simpler than that, and only check that the attributes in the response can be successfully cast to an int but not actually assert on their values, that way we know the server response makes sense.

This makes sense, let me try that!

> That said, I see you've done a lot of work on getting `reset_gitlab` to behave. So we can keep that as it likely helps the testing in any case (in a separate commit maybe). WDYT?

So should I open a new PR for this fix? I'm fine with it going in with this one, since my test does use this function

tests/functional/api/test_statistics.py
nejch: Agreed that this is a helper now, we could potentially move it to `tests.functional.helpers` later to avoid this confusion. Not needed for this PR though

tests/functional/api/test_statistics.py
```diff
@@ -0,0 +1,35 @@
...
+    )
+    gl.groups.create({"name": "gryffindor", "path": "gryffindor"})
+    gl.groups.create({"name": "slytherin", "path": "slytherin"})
+    gl.snippets.create(
+        {
+            "title": "uniquesnippet",
+            "file_name": "snippet394.py",
+            "content": "import gitlab",
+        }
+    )
```
nejch, line 29: Hehe just noticed the `hopefullythisdoesnthavetobeunique` ;) For what it's worth, with the current approach this entire block is not actually necessary. IMO you can remove it if you like, as GitLab should return these attributes even for a brand new instance without resources.

tests/functional/api/test_statistics.py
```diff
@@ -0,0 +1,35 @@
...
+
+    statistics = gl.statistics.get()
+    assert statistics.snippets.isdigit() == True
+    assert statistics.users.isdigit() == True
+    assert statistics.groups.isdigit() == True
+    assert statistics.projects.isdigit() == True
```
nejch, line 35: To make flake8 happy:

```suggestion
    assert statistics.snippets.isdigit()
    assert statistics.users.isdigit()
    assert statistics.groups.isdigit()
    assert statistics.projects.isdigit()
```

nejch (changes requested):
@Shreya-7 up to you if you want this as a separate PR, I'm fine with it in here too! Just a few more nits from me here.

tests/functional/conftest.py
```diff
@@ -94,7 +94,16 @@ def reset_gitlab(gl: gitlab.Gitlab) -> None:
...
         helpers.safe_delete(project)
+
     for group in gl.groups.list():
+
+        # skip deletion of a descendant group to prevent scenarios where parent group gets deleted leaving a dangling descendant whose deletion will throw 404s.
+        if group.parent_id:
+            logging.info(
+                f"Skipping deletion of {group.full_path} as it is a descendant group and will be removed when the parent group is deleted"
+            )
+            continue
```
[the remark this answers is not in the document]
Shreya-7: yep, it's the problem of whoever wants to use this function next! After all the changes, just realised I spent more time on a function I ended up not using in my tests than on the actual feature lol :')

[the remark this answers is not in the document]
Shreya-7: actually, I had to revert this change as it started to inexplicably fail CLI functional tests.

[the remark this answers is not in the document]
nejch: Hehe @Shreya-7 thanks for your patience and sorry for the rollercoaster ride here, still good to see this can affect other tests in the future (whichever way it's done with subgroups) and we'll keep this in mind. Could you maybe just squash your PR to a single commit here for brevity in the release notes? I can do it too if you've had enough :sweat_smile:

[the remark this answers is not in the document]
Shreya-7: no worries! Do we want the commit message to include the reset_gitlab fix as well or just the feature?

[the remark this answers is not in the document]
nejch: There was a bit more work as there were some merge commits in between so I just force-pushed a squashed commit t o your branch @Shreya-7 ;) I'll just wait for CI to run now and otherwise LGTM!

nejch (approved):
Thanks for all the work here @Shreya-7! Hopefully next time it's less of a bumpy ride ;)

tests/functional/conftest.py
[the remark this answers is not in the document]
Shreya-7: Thanks so much!!!

PR #2348, opened 2022-10-29
feat: add import project from remote support

### LINKED ISSUE #2260: Add support for remote-import API Endpoint

Hello Together,

GitLab supports the import from a remote storage, additionally to the export to a remote storage. This is really helpful, if you migrate a lot of big repositories to another GitLab Instance, it's increase the speed and also reduce the API calls and scripts complexity.

https://docs.gitlab.com/ee/api/project_import_export.html#import-a-file-from-a-remote-object-storage

## Expected Behavior

It would be great if gitlab-python would support this endpoint, so additionally to the import_project function, we could also use the import_project_from_remote function.

## Actual Behavior

At the moment, it is not possible to import a repository from a remote storage via import_project function. 

## Specifications

  - python-gitlab version: 3.8.1
  - Gitlab server version (or gitlab.com): 15.2.3

abhiandthetruth:
resolves #2260

### DIFF
--- a/gitlab/v4/objects/projects.py
+++ b/gitlab/v4/objects/projects.py
@@ -822,6 +822,8 @@ class ProjectManager(CRUDMixin, RESTManager):
         Args:
             file: Data or file object containing the project
             path: Name and path for the new project
+            name: The name of the project to import. If not provided,
+                defaults to the path of the project.
             namespace: The ID or path of the namespace that the project
                 will be imported to
             overwrite: If True overwrite an existing project with the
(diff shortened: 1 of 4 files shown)


abhiandthetruth:
@nejch can you please have a look on this PR? 
* On implementation of the rmote_import function, I have basically copied the logic from import_project, since the common code was not large enough to be refactored, I think into Mixin or a private function. Feel free to suggest any refactoring techinique I can use.
* On testing, have added kind of a negative functional test which asserts the expected error. I have verified the error from the gitlab spec, [here](https://gitlab.com/gitlab-org/gitlab/-/blob/master/spec/services/import/gitlab_projects/file_acquisition_strategies/remote_file_spec.rb#L27). I am using a ftp url which violates the validate_url predicate there and hence the error.
* I am not including s3 support here, will add that as separate PR, linked to separate issue(maybe I can raise it), once this goes through. This is my first one here so want this to be a little atomic :)
Thanks. Also if anyone else is also free to review and suggest changes

nejch:
@abhiandthetruth that makes sense! Will check again in a bit, and sure you can go ahead with a separate issue. Yes, currently when we add custom methods it still means copying a bit of code, that needs some rework later.

gitlab/v4/objects/projects.py
```diff
@@ -849,6 +851,49 @@ def import_project(
...
+            namespace: The ID or path of the namespace that the project
+                will be imported to
+            overwrite: If True overwrite an existing project with the
+                same path
+            override_params: Set the specific settings for the project
+            **kwargs: Extra options to send to the server (e.g. sudo)
+
+        Raises:
+            GitlabAuthenticationError: If authentication is not correct
+            GitlabListError: If the server failed to perform the request
```
nejch, line 880: This isn't exactly right, but I see we have the same in the other one. We'll fix that in all of htem later.

tests/functional/api/test_import_export.py
```diff
@@ -64,3 +66,15 @@ def test_project_import_export(gl, project, temp_dir):
...
+def test_project_remote_import(gl):
+    with pytest.raises(gitlab.exceptions.GitlabHttpError) as err_info:
+        gl.projects.remote_import(
+            "ftp://whatever.com/url", "remote-project", "remote-project", "root"
+        )
+    assert err_info.value.response_code == 400
+    assert (
+        "File url is blocked: Only allowed schemes are https"
+        in err_info.value.error_message
+    )
```
nejch, line 80: We could also use a `match=` argument here, but we can do this in a follow-up cleanup along with the list issue.

nejch (approved):
Thanks again @abhiandthetruth, I had some minor nits but realized we can fix them later with other bits of code.

abhiandthetruth:
Yup will clean them up in the other PR. Thanks! Waiting for my first merge in the repo!

tests/functional/api/test_import_export.py
abhiandthetruth: I considered this, but I wasn't sure of checking the status code through a regex, bcoz apparently match takes a string literal or regex. So went with this. We can discuss this on the PR for S3 support.

Issue #2349, opened 2022-10-30
`project-merge-request cancel-merge-when-pipeline-succeeds` fails with `gitlab.exceptions.GitlabMROnBuildSuccessError: 404: 404 Not Found`


TheDJVG:
## Description of the problem, including code/CLI snippet
When a pipeline is marked a merge on successful pipeline it seems this cannot be cancelled as the request is failing with a 404.

Running python-gitlab with `-d` shows that's it's doing a `PUT`:
```
DEBUG:urllib3.connectionpool:https://gitlab.<redacted>:443 "PUT /api/v4/projects/10/merge_requests/26/cancel_merge_when_pipeline_succeeds HTTP/1.1" 404 25
                                                           ^^^^^
```
The documentation however specfies a `POST` should be used: https://docs.gitlab.com/ee/api/merge_requests.html#cancel-merge-when-pipeline-succeeds

## Expected Behavior
The request to merge is cancelled when pipeline succeeded.

## Actual Behavior
```
# gitlab -v project-merge-request cancel-merge-when-pipeline-succeeds --project-id 10 --iid 26                    INT ✘ 
Traceback (most recent call last):
  File "/usr/lib/python3.10/site-packages/gitlab/exceptions.py", line 333, in wrapped_f
    return f(*args, **kwargs)
  File "/usr/lib/python3.10/site-packages/gitlab/v4/objects/merge_requests.py", line 188, in cancel_merge_when_pipeline_succeeds
    server_data = self.manager.gitlab.http_put(path, **kwargs)
  File "/usr/lib/python3.10/site-packages/gitlab/client.py", line 1067, in http_put
    result = self.http_request(
  File "/usr/lib/python3.10/site-packages/gitlab/client.py", line 798, in http_request
    raise gitlab.exceptions.GitlabHttpError(
gitlab.exceptions.GitlabHttpError: 404: 404 Not Found

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/bin/gitlab", line 33, in <module>
    sys.exit(load_entry_point('python-gitlab==3.10.0', 'console_scripts', 'gitlab')())
  File "/usr/lib/python3.10/site-packages/gitlab/cli.py", line 377, in main
    gitlab.v4.cli.run(
  File "/usr/lib/python3.10/site-packages/gitlab/v4/cli.py", line 542, in run
    data = g_cli.run()
[...]

PR #2350, opened 2022-10-30
fix: `mr.cancel_merge_when_pipeline_succeeds()`

JohnVillalovos:
* Call was incorrectly using a `PUT` method when should have used a `POST` method.
  * Changed return type to a `dict` as GitLab only returns {'status': 'success'} on success. Since the function didn't work previously, this should not impact anyone.
  * Updated the test fixture `merge_request` to add ability to create a pipeline.
  * Added functional test for `mr.cancel_merge_when_pipeline_succeeds()`

Fixes: #2349

### DIFF
--- a/gitlab/v4/objects/merge_requests.py
+++ b/gitlab/v4/objects/merge_requests.py
@@ -165,9 +165,7 @@ class ProjectMergeRequest(
 
     @cli.register_custom_action("ProjectMergeRequest")
     @exc.on_http_error(exc.GitlabMROnBuildSuccessError)
-    def cancel_merge_when_pipeline_succeeds(
-        self, **kwargs: Any
-    ) -> "ProjectMergeRequest":
+    def cancel_merge_when_pipeline_succeeds(self, **kwargs: Any) -> Dict[str, str]:
         """Cancel merge when the pipeline succeeds.
 
         Args:
@@ -179,17 +177,20 @@ class ProjectMergeRequest(
                 request
 
         Returns:
-            ProjectMergeRequest
+            dict of the parsed json returned by the server
         """
 
         path = (
             f"{self.manager.path}/{self.encoded_id}/cancel_merge_when_pipeline_succeeds"
         )
-        server_data = self.manager.gitlab.http_put(path, **kwargs)
+        server_data = self.manager.gitlab.http_post(path, **kwargs)
+        # 2022-10-30: The docs at
+        # https://docs.gitlab.com/ee/api/merge_requests.html#cancel-merge-when-pipeline-succeeds
+        # are incorrect in that the return value is actually just:
+        #   {'status': 'success'}  for a successful cancel.
         if TYPE_CHECKING:
             assert isinstance(server_data, dict)
-        self._update_attrs(server_data)
-        return ProjectMergeRequest(self.manager, server_data)
+        return server_data
 
     @cli.register_custom_action("ProjectMergeRequest")
     @exc.on_http_error(exc.GitlabListError)
--- a/tests/functional/api/test_merge_requests.py
+++ b/tests/functional/api/test_merge_requests.py
@@ -181,11 +181,27 @@ def test_merge_request_reset_approvals(gitlab_url, project, wait_for_sidekiq):
     assert mr.reset_approvals()
 
 
+def test_cancel_merge_when_pipeline_succeeds(project, merge_request, wait_for_sidekiq):
+    mr = merge_request(source_branch="test_merge_request_merge", create_pipeline=True)
+    # Set to merge when the pipeline succeeds, which should never happen
+    mr.merge(merge_when_pipeline_succeeds=True)
+    wait_for_sidekiq(timeout=60)
+
+    mr = project.mergerequests.get(mr.iid)
+    assert mr.merged_at is None
+    assert mr.merge_when_pipeline_succeeds is True
+    cancel = mr.cancel_merge_when_pipeline_succeeds()
+    assert cancel == {"status": "success"}
+
+
 def test_merge_request_merge(project, merge_request, wait_for_sidekiq):
     mr = merge_request(source_branch="test_merge_request_merge")
     mr.merge()
     wait_for_sidekiq(timeout=60)
 
+    mr = project.mergerequests.get(mr.iid)
+    assert mr.merged_at is not None
+    assert mr.merge_when_pipeline_succeeds is False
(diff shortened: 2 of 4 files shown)


gitlab/v4/objects/merge_requests.py
```diff
@@ -165,9 +165,7 @@ class ProjectMergeRequest(
 
     @cli.register_custom_action("ProjectMergeRequest")
     @exc.on_http_error(exc.GitlabMROnBuildSuccessError)
-    def cancel_merge_when_pipeline_succeeds(
-        self, **kwargs: Any
-    ) -> "ProjectMergeRequest":
+    def cancel_merge_when_pipeline_succeeds(self, **kwargs: Any) -> Dict[str, str]:
```
nejch, line 168: We just need a corresponding docstring change here (we have `Returns: ProjectMergeRequest` atm) (sorry can't comment on unchanged lines)

tests/functional/conftest.py
```diff
@@ -401,6 +401,21 @@ def _merge_request(*, source_branch: str):
...
+                    "content": """
+test:
+  rules:
+    - if: '$CI_PIPELINE_SOURCE == "merge_request_event"'
+  script:
+    - sleep 24h  # We don't expect this to finish
+""",
+                    "commit_message": "Add a simple pipeline",
+                }
+            )
```
nejch, line 418: One thought I had here, to keep the MR fixture generic/short we could also create a specific `merge_request_pipeline` or `merge_request_with_pipeline` that uses `merge_request`, the same way we do with `project_file` when we need a project that also has a file (https://github.com/python-gitlab/python-gitlab/blob/main/tests/functional/conftest.py#L432-L444). Then in tests we can just choose which fixture to use.

Although maybe the pattern is a bit different here :thinking:

tests/functional/api/test_merge_requests.py
```diff
@@ -181,11 +181,27 @@ def test_merge_request_reset_approvals(gitlab_url, project, wait_for_sidekiq):
...
 
+def test_cancel_merge_when_pipeline_succeeds(project, merge_request, wait_for_sidekiq):
+    mr = merge_request(source_branch="test_merge_request_merge", create_pipeline=True)
+    # Set to merge when the pipeline succeeds, which should never happen
+    mr.merge(merge_when_pipeline_succeeds=True)
+    wait_for_sidekiq(timeout=60)
+
+    mr = project.mergerequests.get(mr.iid)
+    assert mr.merged_at is None
+    assert mr.merge_when_pipeline_succeeds is True
```
nejch, line 192: just a nitpicky question here if we should follow pep8 as we are dealing with a boolean :)

nejch (changes requested):
Thanks @JohnVillalovos nice catch with the response gotcha! I think we just need to update the docstring as well.

Interesting, `GitlabMROnBuildSuccessError` also sounds a bit weird to me now here, could be there are some historical reasons for discrepancies here.

gitlab/v4/objects/merge_requests.py
JohnVillalovos: Done

tests/functional/api/test_merge_requests.py
JohnVillalovos: Not sure I understand.

Do you mean:
`assert mr.merge_when_pipeline_succeeds` ?

But that would pass if it was a 1 or a non-empty string or pretty much any value that is not False, 0, or an empty string.

tests/functional/conftest.py
JohnVillalovos: Sure, I can do that.

JohnVillalovos: Okay. I investigated doing that and it got complicated. More complicated, IMHO, then having the argument to the function.

As the `merge_request_with_pipeline` would need to replicate a large portion of `merge_request`

lmilbaum: A python question, is it possible to use one fixture in a second fixture? Could that work?

tests/functional/api/test_merge_requests.py
```diff
@@ -181,11 +181,27 @@ def test_merge_request_reset_approvals(gitlab_url, project, wait_for_sidekiq):
     assert mr.reset_approvals()
 
 
+def test_cancel_merge_when_pipeline_succeeds(project, merge_request, wait_for_sidekiq):
+    mr = merge_request(source_branch="test_merge_request_merge", create_pipeline=True)
+    # Set to merge when the pipeline succeeds, which should never happen
+    mr.merge(merge_when_pipeline_succeeds=True)
+    wait_for_sidekiq(timeout=60)
+
+    mr = project.mergerequests.get(mr.iid)
```
lmilbaum, line 190: A probably newbie question, do we need to get a handle of the mr even though it was already done in line 185?

tests/functional/conftest.py
lmilbaum, line 418: Does it make sense to move this block into a fixture and re-use it in other functions?

nejch: Ok thanks, let's not complicate :) @lmilbaum yes, generally that was my idea. But this fixture here is more like a [fixture factory](https://docs.pytest.org/en/6.2.x/fixture.html#factories-as-fixtures) so a slightly different approach as you call it in the test case, so got a bit complicated. We can split them into `merge_request` and `make_merge_request` at some point if we want that kind of reuse eventually.

tests/functional/api/test_merge_requests.py
nejch: This is just to fetch the latest actual state from the server with all the attributes after performing the merge I'd say.

tests/functional/api/test_merge_requests.py
nejch: Yes, my understanding of the pep is it recommends against this when comparing against a known boolean type, and here the API explicitly returns a boolean. But maybe we shouldn't waste time on this :)

tests/functional/conftest.py
nejch: @lmilbaum I think this makes sense if we use plain fixtures but this one is meant to be used as a function inside the test. I've created https://github.com/python-gitlab/python-gitlab/issues/2353 to tackle this for all fixtures consistently.

nejch (approved):
Thanks @JohnVillalovos for the quick fix and @lmilbaum for the review, let's tackle the follow-up separately :)

lmilbaum (approved):

PR #2357, opened 2022-11-01
feat: add remote import s3

### LINKED ISSUE #2356: Add support from remote project import from S3
Current import project capabilities do not support the import from s3 option mentioned here.
https://docs.gitlab.com/ee/api/project_import_export.html#import-a-file-from-aws-s3 

## Expected Behavior
`gl.projects.remote-import-s3()` is supported by the module with appropriate parameters.

## Actual Behavior
No method support available currently

## Specifications

  - python-gitlab version: 3.11.0
  - Gitlab server version (or gitlab.com): >= 15.10

abhiandthetruth:
Closes #2356

### DIFF
--- a/gitlab/v4/objects/projects.py
+++ b/gitlab/v4/objects/projects.py
@@ -894,6 +894,65 @@ class ProjectManager(CRUDMixin, RESTManager):
             "/projects/remote-import", post_data=data, **kwargs
         )
 
+    def remote_import_s3(
+        self,
+        region: str,
+        bucket_name: str,
+        file_key: str,
+        access_key_id: str,
+        secret_access_key: str,
+        path: str,
+        name: Optional[str] = None,
+        namespace: Optional[str] = None,
+        overwrite: bool = False,
+        override_params: Optional[Dict[str, Any]] = None,
+        **kwargs: Any,
+    ) -> Union[Dict[str, Any], requests.Response]:
+        """Import a project from an archive file stored on AWS S3.
+
+        Args:
+            region: AWS S3 region name where the file is stored
+            bucket_name: AWS S3 bucket name where the file is stored
+            file_key: AWS S3 file key to identify the file.
+            access_key_id: AWS S3 access key ID.
+            secret_access_key: AWS S3 secret access key.
+            path: Name and path for the new project
+            name: The name of the project to import. If not provided,
+                defaults to the path of the project.
+            namespace: The ID or path of the namespace that the project
+                will be imported to
+            overwrite: If True overwrite an existing project with the
+                same path
+            override_params: Set the specific settings for the project
+            **kwargs: Extra options to send to the server (e.g. sudo)
+
+        Raises:
+            GitlabAuthenticationError: If authentication is not correct
+            GitlabListError: If the server failed to perform the request
+
+        Returns:
+            A representation of the import status.
+        """
+        data = {
+            "region": region,
+            "bucket_name": bucket_name,
+            "file_key": file_key,
+            "access_key_id": access_key_id,
+            "secret_access_key": secret_access_key,
+            "path": path,
+            "overwrite": str(overwrite),
+        }
+        if override_params:
+            for k, v in override_params.items():
+                data[f"override_params[{k}]"] = v
+        if name is not None:
+            data["name"] = name
+        if namespace:
+            data["namespace"] = namespace
+        return self.gitlab.http_post(
+            "/projects/remote-import-s3", post_data=data, **kwargs
+        )
+
     def import_bitbucket_server(
         self,
         bitbucket_server_url: str,
--- a/gitlab/v4/objects/projects.py
+++ b/gitlab/v4/objects/projects.py
@@ -807,6 +807,7 @@ class ProjectManager(CRUDMixin, RESTManager):
     def get(self, id: Union[str, int], lazy: bool = False, **kwargs: Any) -> Project:
         return cast(Project, super().get(id=id, lazy=lazy, **kwargs))
 
+    @exc.on_http_error(exc.GitlabImportError)
     def import_project(
         self,
         file: str,
@@ -833,7 +834,7 @@ class ProjectManager(CRUDMixin, RESTManager):
 
         Raises:
             GitlabAuthenticationError: If authentication is not correct
-            GitlabListError: If the server failed to perform the request
+            GitlabImportError: If the server failed to perform the request
 
         Returns:
             A representation of the import status.
@@ -851,6 +852,7 @@ class ProjectManager(CRUDMixin, RESTManager):
             "/projects/import", post_data=data, files=files, **kwargs
         )
 
+    @exc.on_http_error(exc.GitlabImportError)
     def remote_import(
         self,
         url: str,
@@ -877,7 +879,7 @@ class ProjectManager(CRUDMixin, RESTManager):
 
         Raises:
             GitlabAuthenticationError: If authentication is not correct
-            GitlabListError: If the server failed to perform the request
+            GitlabImportError: If the server failed to perform the request
 
         Returns:
             A representation of the import status.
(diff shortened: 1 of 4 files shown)


abhiandthetruth:
@max-wittig @nejch can you have review this PR please? Pretty similar to https://github.com/python-gitlab/python-gitlab/pull/2348 
There is again a bit of replicated code from import_project function. Can't find a way to refactor it because it is quite small and creating any common entity just feals like overkill. Let me know if anyone has a way to go over it, else I think this much replication might be tolerable.

gitlab/v4/objects/projects.py
```diff
@@ -894,6 +894,65 @@ def remote_import(
...
+        region: str,
+        bucket_name: str,
+        file_key: str,
+        access_key_id: str,
+        secret_access_key: str,
+        path: str,
+        name: Optional[str] = None,
+        namespace: Optional[str] = None,
+        overwrite: bool = False,
+        override_params: Optional[Dict[str, Any]] = None,
```
nejch, line 908: Should we maybe have the path at the top here? Will be following the upstream docs for required attributes closer, visually.

```suggestion
        path: str,
        region: str,
        bucket_name: str,
        file_key: str,
        access_key_id: str,
        secret_access_key: str,
        name: Optional[str] = None,
        namespace: Optional[str] = None,
        overwrite: bool = False,
        override_params: Optional[Dict[str, Any]] = None,
```

gitlab/v4/objects/projects.py
```diff
@@ -894,6 +894,65 @@ def remote_import(
...
+            namespace: The ID or path of the namespace that the project
+                will be imported to
+            overwrite: If True overwrite an existing project with the
+                same path
+            override_params: Set the specific settings for the project
+            **kwargs: Extra options to send to the server (e.g. sudo)
+
+        Raises:
+            GitlabAuthenticationError: If authentication is not correct
+            GitlabListError: If the server failed to perform the request
```
nejch, line 931: Before we forget - this would technically be the way to go, but would also require a decorator `@exc.on_http_error(exc.GitlabImportError)`:

```suggestion
            GitlabImportError: If the server failed to perform the request
```

tests/unit/objects/test_project_import_export.py
```diff
@@ -53,6 +53,30 @@ def resp_remote_import():
...
+    content = {
+        "id": 1,
+        "description": None,
+        "name": "remote-project-s3",
+        "name_with_namespace": "Administrator / remote-project-s3",
+        "path": "remote-project-s3",
+        "path_with_namespace": "root/remote-project-s3",
+        "created_at": "2018-02-13T09:05:58.023Z",
+        "import_status": "scheduled",
+    }
```
nejch, line 67: I just realized that, at least in the upstream examples, all import status responses are the same, so we could just reuse a module-level `import_content` in all the fixtures IMO. But not a blocker here.

tests/functional/api/test_import_export.py
```diff
@@ -78,3 +78,23 @@ def test_project_remote_import(gl):
...
+            "secret-access-key",
+            "remote-project",
+            "remote-project",
+            "root",
+        )
+    assert err_info.value.response_code == 400
+    assert (
+        "Failed to open 'aws-file-key' in 'aws-bucket-name'"
+        in err_info.value.error_message
+    )
```
nejch, line 100: To answer your question from the previous PR, I think `match` works like inclusion tests for plain strings so no need for regex. But actually this looks fine :)

nejch (changes requested):
Looks great @abhiandthetruth thanks for the quick work again. Just a few quick notes :)

> There is again a bit of replicated code from import_project function. Can't find a way to refactor it because it is quite small and creating any common entity just feals like overkill. Let me know if anyone has a way to go over it, else I think this much replication might be tolerable.

I agree, no need yet. We have plenty of other custom methods where this is an issue as well.

I think it would need a lot of rework but another decorator with required/optional payload attributes might make these methods more declarative instead of parsing args inside every method. Something like https://github.com/python-gitlab/python-gitlab/pull/1999 if we revive that at some point.

gitlab/v4/objects/projects.py
abhiandthetruth: Oh right, will do. I was following https://gitlab.com/gitlab-org/gitlab/-/blob/master/lib/api/project_import.rb#L161 
Actually the docs do not mention the parameters `overrid_params` and `overwrite` in the list of params. In description they do mention `override_params`. Was confused, hence checked the code out :). But people are going to read from spec, so this is the right way. Thanks!

gitlab/v4/objects/projects.py
abhiandthetruth: Yup can do, hopefully `GItlabImportError` is defined beforehand

tests/unit/objects/test_project_import_export.py
abhiandthetruth: Yeah have observed that as well. But I don't really try to refactor test code xD. Will try

gitlab/v4/objects/projects.py
abhiandthetruth: but according to this, shouldn't name, namespace be at the top? 
https://docs.gitlab.com/ee/api/project_import_export.html#import-a-file-from-aws-s3

abhiandthetruth: lol sorry they are optional

abhiandthetruth: Done, updated at other places accordingly. Pls resolve if correct

gitlab/v4/objects/projects.py
abhiandthetruth: Done, updated tests to check for this error as well

tests/unit/objects/test_project_import_export.py
abhiandthetruth: Not done ;_;. Maybe we can tackle this in a separate  refactor ticket. I don't want to lose uniformity of each test having its own response

abhiandthetruth:
@nejch implemented the suggested changes.

tests/unit/objects/test_project_import_export.py
nejch: Makes sense :)

tests/functional/api/test_import_export.py
```diff
@@ -78,3 +78,23 @@ def test_project_remote_import(gl):
...
+        gl.projects.remote_import_s3(
+            "remote-project",
+            "aws-region",
+            "aws-bucket-name",
+            "aws-file-key",
+            "aws-access-key-id",
+            "secret-access-key",
+            "remote-project",
+            "root",
+        )
```
nejch, line 95: Something I just realized here is that this will trigger real requests (from the docker-compose GitLab instance I assume) to the actual AWS API. This is something we probably don't want to have long-term. I don't want to block this PR though but if you have any ideas, for example by blocking networking in a certain way in our docker-compose setup, that might be a good follow-up

abhiandthetruth: Hmm it will not probably be a simple config change, will have to update the exiswting s3 test case too as the error response might change now since we never go to s3

nejch: Ah true. It would be better if GitLab supported other instances such as minio, then we would maybe just configure that. Not urgent then at all, thanks!

abhiandthetruth: Ok. Will keep this in my mind. Will contribute if I find something!

nejch: I see that the Ruby client supports an `endpoint` param, this would probably work, but needs an upstream contribution first.

https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/S3/Client.html

abhiandthetruth: They do have configurable endpoints for other scopes https://gitlab.com/gitlab-org/omnibus-gitlab/blob/master/files/gitlab-config-template/gitlab.rb.template?_gl=1%2a1vxyfkb%2a_ga%2aMzg2MTk2NTY4LjE2NTk4OTgxMzU.%2a_ga_ENFH3X7M5Y%2aMTY2NzMzMDYyNC4xMS4xLjE2NjczMzE5MjguMC4wLjA.#L379.

abhiandthetruth: but this might require the param at api level,since the client is probably generated at runtime for each api call

PR #2362, opened 2022-11-03
feat(ci): Re-Run Tests on PR Comment workflow

lmilbaum:
#2359

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -24,6 +24,17 @@ message guidelines described below:
 
 Please provide your patches as GitHub pull requests. Thanks!
 
+Rerunning failed CI workflows
+-----------------------------
+
+* Ask the maintainers to add the ``ok-to-test`` label on the PR
+* Post a comment in the PR
+   ``/rerun-all`` - rerun all failed workflows
+
+   ``/rerun-workflow <workflow name>`` - rerun a specific failed workflow
+
+The functionality is provided by ``rerun-action <https://github.com/marketplace/actions/rerun-actions>``
+
 Commit message guidelines
 -------------------------
 
--- a/.github/workflows/rerun-test.yml
+++ b/.github/workflows/rerun-test.yml
@@ -1,0 +1,16 @@
+name: 'Rerun failed workflows'
+
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
+  rerun_pr_tests:
+    name: rerun_pr_tests
+    if: ${{ github.event.issue.pull_request }}
+    runs-on: ubuntu-20.04
+    steps:
+    - uses: estroz/rerun-actions@main
+      with:
+        repo_token: ${{ secrets.GITHUB_TOKEN }}
+        comment_id: ${{ github.event.comment.id }}


lmilbaum:
@nejch @JohnVillalovos Just my luck :-) The failed test has nothing to do with the changes in this PR. It is probably a flaky test. I am unable to re-run it. Would you mind pushing the rerun button?

CONTRIBUTING.rst
```diff
@@ -24,6 +24,15 @@ message guidelines described below:
 
 Please provide your patches as GitHub pull requests. Thanks!
 
+Rerunning failed CI workflows
+-------------------------
+
+* Ask the maintainers to add the ``ok-to-test`` label on the PR
```
nejch, line 30: Is there a way to avoid this? Not sure if it's hardcoded in the GH Action, I haven't looked into it in depth. We already have to approve workflow runs from first-time contributors, so if that was enough it'd be great. For regular contributors we can assume we trust them.

CONTRIBUTING.rst
```diff
@@ -24,6 +24,15 @@ message guidelines described below:
 
 Please provide your patches as GitHub pull requests. Thanks!
 
+Rerunning failed CI workflows
+-------------------------
```
nejch, line 28: Maybe we can add a sentence here before the list, currently the rendering is a bit odd: 
https://github.com/python-gitlab/python-gitlab/blob/6d98c7588447a7423d7506c6deafb665c642b955/CONTRIBUTING.rst#rerunning-failed-ci-workflows

CONTRIBUTING.rst
```diff
@@ -24,6 +24,15 @@ message guidelines described below:
...
+
+* Ask the maintainers to add the ``ok-to-test`` label on the PR
+* Post a comment in the PR 
+   ``/rerun-all`` - rerun all failed workflows
+   ``/rerun-workflow <workflow name>`` - rerun a specific failed workflow
+The functionality is provide by ``rerun-action <https://github.com/marketplace/actions/rerun-actions>``
```
nejch, line 34:
```suggestion
The functionality is provided by ``rerun-action <https://github.com/marketplace/actions/rerun-actions>``.
```

nejch (changes requested):
That looks like a great quality-of-life improvement for contributors and something I would already expect natively, coming from GitLab. Just a question about the label :)

CONTRIBUTING.rst
lmilbaum: Good point. I believe there is a way to automate the process. Let's examine it as a follow up. WDYT?

CONTRIBUTING.rst
lmilbaum: Apparently there is a cool vscode plugin which helped me realize what was wrong. Thanks for pushing towards it.

CONTRIBUTING.rst
lmilbaum: Fixed

lmilbaum:
@nejch need your help again to rerun the failed test (Docs / sphinx)

CONTRIBUTING.rst
lmilbaum: Yeah. I've noticed it as well.

nejch (approved):
Thanks @lmilbaum let's see how it goes in the wild :)

Issue #447, opened 2018-03-04
GPLv3?


tardyp:
I just realized python-gitlab is license under GPLv3.

GPLv3 license is arguably not suitable for a library. 
For example, I cannot use it in https://github.com/buildbot/buildbot which is GPLv2.
I can only use it for programs in GPLv3 which are very sparse nowadays. In 2015: 8% of projects in github are GPLv3 https://blog.github.com/2015-03-09-open-source-license-usage-on-github-com/

I believe that python-gitlab has good value and is well designed. Please consider putting it in a less restrictive license so that it can be more widely used.

gpocentek:
Actually the license is LGPL v3 (lesser GPL), which is much more permissive than GPL. You can "link" python-gitlab from any other application, even proprietary. I don't think the license will prevent you from using python-gitlab in your own project.

I am no expert on the subject though, so let me know if you think I misunderstood.

tardyp:
Sorry about that I could swear I triple checked it wasn't LGPL.
LGPL is perfectly fine!

thanks

PR #2363, opened 2022-11-03
Use SPDX license expression in project metadata

RazerM:
As a downstream user it is desirable to be able to programmatically determine the precise licenses used by our dependencies. The emerging convention for this is the [SPDX License List](https://spdx.org/licenses/). SPDX License Expressions are already used in the JavaScript (npm) and Rust (crates.io) ecosystems, for example.

In Python, there is [PEP 639](https://peps.python.org/pep-0639/) which would add a standard 'License-Expression' metadata field. The discussion around this PEP seems to have gone stale in 2021.

This merge request is in lieu of that PEP coming soon. Unless you as project maintainers think the proposed change in this PR has downsides (which you are entitled to!), I would ask that you accept this change so that it's possible for users to attempt to parse your license metadata as an SPDX License Expression.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,7 +25,7 @@ setup(
     long_description_content_type="text/x-rst",
     author="Gauvain Pocentek",
     author_email="gauvain@pocentek.net",
-    license="LGPLv3",
+    license="LGPL-3.0-or-later",
     url="https://github.com/python-gitlab/python-gitlab",
     packages=find_packages(exclude=["docs*", "tests*"]),
     install_requires=["requests>=2.25.0", "requests-toolbelt>=0.9.1"],


nejch:
Thanks for this @RazerM, I know the pain with unparseable licenses for SBOMs/OSS notices and SPDX makes this more sane, just missed this :)

We were actually going to switch to PEP 621 already (draft here https://github.com/python-gitlab/python-gitlab/blob/8d45db78b4e848b0ed41cc1ce338457b0ebc81bc/pyproject.toml#L39) and that leaves this intentionally a bit open by allowing either file or text keys, and we went for file there. Let me just check what the best approach for pep621 is. Otherwise if we go this way this'll just need a little `git commit --amend` as we use conventional commits (see CONTRIBUTING.rst).

RazerM:
Thanks for taking this into consideration

I totally get the PEP 621 switch — I've been doing it myself for several projects and it's a joy to finally be using this nice file format :-)

The part of that PEP which worries me a bit around this license stuff is

> A practical string value for the license key has been purposefully left out to allow for a > future PEP to specify support for [SPDX](https://spdx.dev/) expressions

Because it means you might have projects which provide a license file only (because the PEP disallows `file` *and* `text`) and there's no way for downstream users to find out which license it is programmatically without trying to parse the file. (Until PEP 639 arrives, at least).

In practice though, things are not so bad for setuptools, which you are still using in the linked pyproject.toml. I can confirm that if you use `license = { text = "LGPL-3.0-only" }`, setuptools still installs your `COPYING` file which I think might be the best of both worlds.

If you try it out you should see e.g. `lib/python3.10/site-packages/python_gitlab-3.11.0.dist-info/COPYING` exists.

An alternative I could imagine that would still let you use `license = { file = "COPYING" }` would be to add `SPDX-License-Identifier: LGPL-3.0-only` somewhere in the `COPYING` file itself?

nejch:
That's ok, it's still closer to the current `setup.py` text in any case :) I guess some of the reluctance in PEP 639 comes from this uncertainty - here as well, the current maintainers are not the authors so to me, we're basically disambiguating it right now in this PR and interpreting that the original author explicitly wanted `-only` and not `-or-later` when choosing the classifier, it's not entirely clear to me from reading https://github.com/python-gitlab/python-gitlab/issues/447 :)

Could you just amend your commit message according to https://www.conventionalcommits.org/en/v1.0.0/#summary? we use it for our cahngelogs. This will need a force-push after as well.

RazerM:
Will do, would you consider this a `chore`?

nejch:
I think chore is fine here yes unless you'd prefer to have this explicitly in the changelog as a fix.

setup.py
```diff
@@ -25,7 +25,7 @@ def get_version() -> str:
     long_description_content_type="text/x-rst",
     author="Gauvain Pocentek",
     author_email="gauvain@pocentek.net",
-    license="LGPLv3",
+    license="LGPL-3.0-only",
```
nejch, line 28: @max-wittig WDYT before I merge this, we don't have a `This project is licensed under the terms of the <something> license.` so from what I can tell it's not even clear (even from Gauvain's initial comments and text), and we should decide right here on `LGPL-3.0-only` vs `LGPL-3.0-or-later`.

The proposed PEP even doesn't want to disambiguate this automatically (https://peps.python.org/pep-0639/#mapping-license-classifiers-to-spdx-identifiers).

max-wittig: I guess or later is fine?

nejch: Works for me as well.

```suggestion
    license="LGPL-3.0-or-later",
```

nejch: @RazerM are you still interested and would like to apply the suggested change above?

RazerM: Done

nejch (approved):

setup.py
nejch: Thanks for the PR and the discussion above!

Issue #2368, opened 2022-11-08
group.pushrules.get() should return None not 404.


dreaquil:
## Description of the problem, including code/CLI snippet

According to the documentation, if pushrules are missing from a group, the function `group.pushrules.get()` should return None. 
## Expected Behavior

According to the documentation, if pushrules are missing from a group, the function `group.pushrules.get()` should return None. 
## Actual Behavior

It is raising the following error: 

`gitlab.exceptions.GitlabHttpError: 404: 404 Not Found`
## Specifications

  - python-gitlab version: 3.11.0
  - API version you are using (v3/v4): v4
  - Gitlab server version (or gitlab.com): gitlab.com

JohnVillalovos:
Thanks for the report. I have submitted a documentation update for this.

PR #2369, opened 2022-11-08
docs(api): pushrules remove saying `None` is returned when not found

JohnVillalovos:
In `groups.pushrules.get()`, GitLab does not return `None` when no rules are found. GitLab returns a 404.

Update docs to not say it will return `None`

Closes: #2368


docs/gl_objects/groups.rst
```diff
@@ -402,7 +402,7 @@ Create group push rules (at least one rule is necessary)::
 
     group.pushrules.create({'deny_delete_tag': True})
 
-Get group push rules (returns ``None`` if there are no push rules)::
+Get group push rules::
```
nejch, line 405: We have the same statement for projects in https://github.com/python-gitlab/python-gitlab/blob/main/docs/gl_objects/projects.rst#examples-10. Do you think we need to remove that too?

Likely this was copied from project push rules. Maybe there was a change in the API at some point.

nejch (commented):
Thanks @JohnVillalovos, just a small question :)

docs/gl_objects/groups.rst
JohnVillalovos: I honestly don't know. I haven't checked it  🤷‍♂️

JohnVillalovos: But I guess it doesn't hurt to remove it. Even if it does return `None`. Do we need to document that?

nejch (approved):

PR #2370, opened 2022-11-08
feat: implement secure files API

benjamb:


docs/gl_objects/project_secure_files.rst
```diff
@@ -0,0 +1,47 @@
...
+====================
+
+References
+----------
+
+* v4 API:
+
+  + :class:`gitlab.v4.objects.ProjectSecureFile`
+  + :class:`gitlab.v4.objects.ProjectSecureFileManager`
+  + :attr:`gitlab.Gitlab.project_secure_files`
```
nejch, line 15:
```suggestion
  + :attr:`gitlab.v4.objects.Project.secure_files`
```

(I realize now we need to also fix our project access token docs)

docs/api-objects.rst
```diff
@@ -44,6 +44,7 @@ API examples
    gl_objects/pipelines_and_jobs
    gl_objects/projects
    gl_objects/project_access_tokens
+   gl_objects/project_secure_files
```
nejch, line 47: Perhaps we could just have `secure_files` here to follow upstream https://docs.gitlab.com/ee/api/secure_files.html, even though they're only implemented at project level now. Project access tokens were a bit different as they have their own page upstream for some reason :grin:

docs/gl_objects/project_secure_files.rst
```diff
@@ -0,0 +1,47 @@
+####################
+Project Secure Files
+####################
```
nejch, line 3: Same as above - we can be generic here and just use project in the sub heading:

```suggestion
##########
Secure Files
##########
```

nejch (changes requested):
Amazing, thanks for all the work here @benjamb! I just have a few tiny questions. :)

docs/api-objects.rst
benjamb: Sure thing. Do you want me to make similar changes to the code paths/code itself?

nejch: Oh yes, that'd make sense so we don't introduce breaking changes later, thanks!

docs/gl_objects/project_secure_files.rst
```diff
@@ -0,0 +1,47 @@
...
+    print(secure_files[0].name)
+
+Create project secure file::
+
+    secure_file = gl.projects.get(1).secure_files.create({"name": "test", "file": "secure.txt"})
+
+Download a project secure file::
+
+    content = secure_file.download()
+    print(content)
```
lmilbaum, line 39: Shouldn't you use logging instead of the print command?

lmilbaum (changes requested):

docs/gl_objects/project_secure_files.rst
benjamb: This is consistent with the rest of the documentation, and only serves as a reference. Documenting the use of appropriate logging mechanisms seems out of scope.

lmilbaum: Apologies for my oversight. I was under the impression that logging is already being used in the project and we want to avoid using the `print` command.  Replacing the `print` commands with logging is probably not in the scope of this PR.

docs/api-objects.rst
benjamb: Done.

tests/unit/objects/test_secure_files.py
```diff
@@ -0,0 +1,112 @@
...
+    secure_file = gl.projects.get(1, lazy=True).secure_files.get(1)
+    secure_content = secure_file.download()
+    assert isinstance(secure_file, SecureFile)
+    assert secure_content == binary_content
+
+
+def test_remove_secure_file(gl, resp_remove_secure_file):
+    gl.projects.get(1, lazy=True).secure_files.delete(1)
+    secure_files = gl.projects.get(1, lazy=True).secure_files.list()
+    assert len(secure_files) == 0
```
nejch, line 112: Since we're mocking the effect entirely, I'm not sure we need the extra list here. We could just remove `assert_all_requests_are_fired=False` from the fixture for DELETE to ensure the request is correctly called, would also keep the fixture smaller.

nejch (commented):
Thanks again @benjamb, just made another pass and noticed some minor things we can change to keep tests shorter :)

gitlab/v4/objects/secure_files.py
```diff
@@ -0,0 +1,75 @@
...
+            path, streamed=streamed, raw=True, **kwargs
+        )
+        if TYPE_CHECKING:
+            assert isinstance(result, requests.Response)
+        return utils.response_content(
+            result, streamed, action, chunk_size, iterator=iterator
+        )
+
+
+class SecureFileManager(CreateMixin, GetMixin, ListMixin, DeleteMixin, RESTManager):
```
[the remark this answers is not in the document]
benjamb: Ah, that's handy!

docs/gl_objects/secure_files.rst
```diff
@@ -0,0 +1,47 @@
...
+============
+
+References
+----------
+
+* v4 API:
+
+  + :class:`gitlab.v4.objects.SecureFile`
+  + :class:`gitlab.v4.objects.SecureFileManager`
+  + :attr:`gitlab.v4.objects.secure_files`
```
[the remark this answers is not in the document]
benjamb: My bad.

nejch (approved):
LGTM :)

benjamb:
@nejch Thanks!

nejch:
I just discovered this feels a lot less beta than the official https://gitlab.com/gitlab-org/incubation-engineering/mobile-devops/download-secure-files, hehe thanks again @benjamb!

PR #2367, opened 2022-11-07
feat: add saml group links

abhiandthetruth:
CLOSES #2270

PR #2371, opened 2022-11-09
feat(groups): add a list_ldap_group_links to go along with the pre ex…

rayisbadat:
feat(groups): add a list_ldap_group_links to go along with the pre existing add_ldap_group_link and delete_ldap_group_link

### DIFF
--- a/gitlab/v4/objects/groups.py
+++ b/gitlab/v4/objects/groups.py
@@ -206,7 +206,7 @@ class Group(SaveMixin, ObjectDeleteMixin, RESTObject):
             GitlabAuthenticationError: If authentication is not correct
             GitlabGetError: If the server cannot perform the request
         """
-        path = "/groups/%s/ldap_group_links" % self.get_id()
+        path = f"/groups/{self.encoded_id}/ldap_group_links"
         return self.manager.gitlab.http_get(path, **kwargs)
 
     @cli.register_custom_action("Group")
--- a/gitlab/v4/objects/groups.py
+++ b/gitlab/v4/objects/groups.py
@@ -209,7 +209,7 @@ class Group(SaveMixin, ObjectDeleteMixin, RESTObject):
             GitlabGetError: If the server cannot perform the request
         """
         path = f"/groups/{self.encoded_id}/ldap_group_links"
-        return self.manager.gitlab.http_get(path, **kwargs)
+        return self.manager.gitlab.http_list(path, **kwargs)
 
     @cli.register_custom_action("Group")
     @exc.on_http_error(exc.GitlabCreateError)
--- a/gitlab/v4/objects/groups.py
+++ b/gitlab/v4/objects/groups.py
@@ -197,7 +197,7 @@ class Group(SaveMixin, ObjectDeleteMixin, RESTObject):
     @cli.register_custom_action("Group")
     @exc.on_http_error(exc.GitlabGetError)
     def list_ldap_group_links(self, **kwargs):
-        """Add an LDAP group link.
+        """List LDAP group links.
 
         Args:
             **kwargs: Extra options to send to the server (e.g. sudo)
--- a/gitlab/v4/objects/groups.py
+++ b/gitlab/v4/objects/groups.py
@@ -196,7 +196,9 @@ class Group(SaveMixin, ObjectDeleteMixin, RESTObject):
 
     @cli.register_custom_action("Group")
     @exc.on_http_error(exc.GitlabGetError)
-    def list_ldap_group_links(self, **kwargs):
+    def list_ldap_group_links(
+        self, **kwargs: Any
+    ) -> Union[gitlab.GitlabList, List[Dict[str, Any]]]:
         """List LDAP group links.
 
         Args:
--- a/gitlab/v4/objects/groups.py
+++ b/gitlab/v4/objects/groups.py
@@ -195,6 +195,21 @@ class Group(SaveMixin, ObjectDeleteMixin, RESTObject):
         self.manager.gitlab.http_delete(path, **kwargs)
 
     @cli.register_custom_action("Group")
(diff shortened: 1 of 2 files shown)


gitlab/v4/objects/groups.py
```diff
@@ -194,6 +194,21 @@ def delete_ldap_group_link(
         path += f"/{cn}"
         self.manager.gitlab.http_delete(path, **kwargs)
 
+    @cli.register_custom_action("Group")
+    @exc.on_http_error(exc.GitlabGetError)
+    def list_ldap_group_links(self, **kwargs):
+        """Add an LDAP group link.
```
JohnVillalovos, line 200: This seems incorrect.

gitlab/v4/objects/groups.py
```diff
@@ -194,6 +194,21 @@ def delete_ldap_group_link(
...
+
+        Args:
+            **kwargs: Extra options to send to the server (e.g. sudo)
+
+        Raises:
+            GitlabAuthenticationError: If authentication is not correct
+            GitlabGetError: If the server cannot perform the request
+        """
+        path = "/groups/%s/ldap_group_links" % self.get_id()
+        return self.manager.gitlab.http_get(path, **kwargs)
```
JohnVillalovos, line 210: As a note (for other reviewers), this is probably correct. I initially thought should probably use `http_list()`, but upstream docs have zero information on what is returned ☹️ With no docs I guess `http_get()` is probably correct.

@rayisbadat would be interesting if you could provide the return value from the API call.

JohnVillalovos (changes requested):

gitlab/v4/objects/groups.py
rayisbadat: If you curl the API directly It returns an array of dicts as show below (I redacted out our ldap info)
```
[{"cn":null,"group_access":40,"provider":"ldapmain","filter":"(memberOf=cn=XXXXX,ou=groups,ou=XXXXX,dc=XXXXX,dc=XXXXX,dc=XXXXX)"}]                              `
```

The output doesnt seem to change if i call http_get or http_list in the code.  Both look like
```
[{'cn': None,
  'filter': '(memberOf=cn=XXXXX,ou=groups,ou=XXXXX,dc=XXXXX,dc=XXXXX,dc=XXXXX)',
  'group_access': 40,
  'provider': 'ldapmain'}]
<class 'list'>
```

gitlab/v4/objects/groups.py
rayisbadat: oops, corrected.

gitlab/v4/objects/groups.py
JohnVillalovos: @nejch  What do you think, `http_get` or `http_list`. After seeing the data returned I'm leaning towards `http_list`, but not sure.

gitlab/v4/objects/groups.py
```diff
@@ -194,6 +194,21 @@ def delete_ldap_group_link(
...
+        """List LDAP group links.
+
+        Args:
+            **kwargs: Extra options to send to the server (e.g. sudo)
+
+        Raises:
+            GitlabAuthenticationError: If authentication is not correct
+            GitlabGetError: If the server cannot perform the request
+        """
+        path = "/groups/%s/ldap_group_links" % self.get_id()
```
lmilbaum, line 209: @JohnVillalovos @nejch A general question. Would it make sense to change the syntax such that f-string is being used instead of a string?

JohnVillalovos: Yes it would, since @nejch did a big PR to switch everything to f-strings 😊

JohnVillalovos: As a note, the `pylint` check should flag non f-string usage.

https://pylint.pycqa.org/en/latest/user_guide/checkers/features.html   consider-using-f-string (C0209)

gitlab/v4/objects/groups.py
nejch: Looking at https://github.com/python-gitlab/python-gitlab/pull/2367 which adds similar functionality, I'd even go for a full manager with a `ListMixin`, since the number of custom methods here is growing. Both links endpoints seem like 2 very similar things. But maybe outside the scope of this PR - I'd go for `http_list` just in case though, for pagination.

gitlab/v4/objects/groups.py
nejch: Agreed, CI will take care of this :)

rayisbadat: rebased off main, replaced with f string, and did the function annotation thing.

lmilbaum (approved):
LGTM

lmilbaum (changes requested):
Could you please add a unit test?

nejch (commented):
> > @lmilbaum  Could you please add a unit test?
> 
> I added a unit test and tox passed. But this my first time doing such a thing. So if its not right , I might need a little help on whats needed.

@rayisbadat  it's a great start! You've added a test fixture, which you can now use to also create a test case, for example `def test_list_ldap_group_links(group, resp_list_ldap_group_links):`.

It will look a bit like this existing test, but without the `isinstance` assert probably:

https://github.com/python-gitlab/python-gitlab/blob/373b46c5d16f1ae26d0c39ceb34fb7e2c9bb456e/tests/unit/objects/test_groups.py#L248-L251

Just a small tip, we usually put all the fixtures (things decorated with `@pytest.fixture`) at the top of the module, and test cases (`def test_*`) after them.

nejch (approved):

nejch:
@JohnVillalovos I think you wanted to have another look at this?

JohnVillalovos (approved):
@nejch Looks good to me, though would like it to be squashed when it is merged in.

PR #2373, opened 2022-11-10
chore: correct website for pylint

JohnVillalovos:
Use https://github.com/PyCQA/pylint as the website for pylint.


nejch (approved):
Thanks @JohnVillalovos, seems like they also have https://pylint.pycqa.org/ now, but it's all the same :)

PR #2379, opened 2022-11-17
feat(groups): add LDAP link manager and deprecate old API endpoints

nejch:
This builds on https://github.com/python-gitlab/python-gitlab/pull/2371.

I realized https://github.com/python-gitlab/python-gitlab/pull/1612 already did something similar and the `list()` `create()` `delete()` interface is more intuitive for the user.
Additionally, https://github.com/python-gitlab/python-gitlab/pull/2367 will bring in a similar interface for SAML.
And finally, the delete endpoints used in the current method are actually deprecated upstream, so good to move away from that on our side

https://docs.gitlab.com/ee/api/groups.html#delete-ldap-group-link (deprecated)
https://docs.gitlab.com/ee/api/groups.html#delete-ldap-group-link-with-cn-or-filter (new endpoint)

### DIFF
--- a/gitlab/v4/objects/groups.py
+++ b/gitlab/v4/objects/groups.py
@@ -5,9 +5,16 @@ import requests
 import gitlab
 from gitlab import cli
 from gitlab import exceptions as exc
-from gitlab import types
+from gitlab import types, utils
 from gitlab.base import RESTManager, RESTObject
-from gitlab.mixins import CRUDMixin, ListMixin, ObjectDeleteMixin, SaveMixin
+from gitlab.mixins import (
+    CreateMixin,
+    CRUDMixin,
+    DeleteMixin,
+    ListMixin,
+    ObjectDeleteMixin,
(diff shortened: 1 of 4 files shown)


gitlab/v4/objects/groups.py
```diff
@@ -188,29 +205,19 @@ def delete_ldap_group_link(
...
-
-        Args:
-            **kwargs: Extra options to send to the server (e.g. sudo)
-
-        Raises:
-            GitlabAuthenticationError: If authentication is not correct
-            GitlabGetError: If the server cannot perform the request
-        """
-        path = f"/groups/{self.encoded_id}/ldap_group_links"
-        return self.manager.gitlab.http_list(path, **kwargs)
```
nejch, line 212: No deprecation needed as we haven't released it yet.

JohnVillalovos (approved):
Thanks @nejch 🙂

PR #2381, opened 2022-11-17
docs: Use the term "log file" for getting a job log file

JohnVillalovos:
The GitLab docs refer to it as a log file:
https://docs.gitlab.com/ee/api/jobs.html#get-a-log-file

"trace" is the endpoint name but not a common term people will think of for a "log file"


nejch:
That makes sense, thanks! :)

nejch (approved):

PR #2382, opened 2022-11-19
chore: validate httpx package is not installed by default

lmilbaum:


lmilbaum:
@nejch @JohnVillalovos Is it possible the `Test / functional (cli_func_v4)`workflow is flaky? This is not the first time my PRs are hitting a similar failure. A rerun usually ends ok.

JohnVillalovos:
> @nejch @JohnVillalovos Is it possible the `Test / functional (cli_func_v4)`workflow is flaky? This is not the first time my PRs are hitting a similar failure. A rerun usually ends ok.

That new test seems to be flaky. Now who is responsible for that? /me checks `git blame` and sees he wrote it 😳

So looks like I need to spend some more time making `test_cancel_merge_when_pipeline_succeeds` more reliable. Created issue #2383

JohnVillalovos:
Can this be done using `tox`? So that these tests can be run locally instead of only being able to be done in the CI?

lmilbaum:
> Can this be done using `tox`? So that these tests can be run locally instead of only being able to be done in the CI?
> 
> Doesn't have to be a default test that is run in `tox` but would be good if it can be run locally.

Yup. I forgot to handle it via fox. Thanks for your feedback.

tests/install/test_install.py
```diff
@@ -0,0 +1,6 @@
+import pytest
+
+
+def test_install() -> None:
+    with pytest.raises(ImportError):
+        import httpx  # type: ignore # noqa
```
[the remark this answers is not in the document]
lmilbaum: I've tried to combine this test via `sandbox` with no avail.  See the discussion pypa/setuptools#3646.  I didn't think that refactoring the smoke tests is in the scope of this PR.

JohnVillalovos:
@lmilbaum I probably won't have time to check this until next or the following week. Sorry. Holidays this week in the states and my work is keeping me busy too. I just wanted to let you know that I haven't forgotten.

@nejch Feel free to go ahead without me if it you like it. No need to wait for my approval.

.github/workflows/install.yml
```diff
@@ -0,0 +1,52 @@
+name: "Install"
```
nejch, line 1: Could we put these 2 jobs into https://github.com/python-gitlab/python-gitlab/blob/main/.github/workflows/test.yml?

We define some additional workflow config there, so this way we make sure the changes apply to the jobs consistetly.

.github/workflows/install.yml
```diff
@@ -0,0 +1,52 @@
...
+        name: dist
+        path: dist
+
+  test:
+    runs-on: ubuntu-latest
+    needs: [dist]
+    strategy:
+      matrix:
+        python-version: [ '3.7', '3.8', '3.9', '3.10' ]
+        install-from: ["dist/*.whl"]
```
nejch, line 36: We don't run functional tests in a matrix, so I'd say it'd be fine to run on a single version here as well (3.11). This will also reduce the number of jobs.

tests/install/test_install.py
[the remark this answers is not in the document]
nejch: Oh, ok this is going really really in-depth for what we are trying to check here IMO. But thanks a lot for all the research - in that case I'd keep this here but I just have a small comment above to simplify a bit.

nejch (changes requested):

.github/workflows/install.yml
lmilbaum: ack

nejch (approved):
Thanks a lot for digging into this @lmilbaum, this took a bit longer than we all expected but will be useful later when we probably use it with some parametrized request clients :)

Issue #2383, opened 2022-11-19
API test `test_cancel_merge_when_pipeline_succeeds` is flaky


JohnVillalovos:
Seeing issues where the test is failing.

For example: https://github.com/python-gitlab/python-gitlab/actions/runs/3502081004/jobs/5867796535

PR #2385, opened 2022-11-19
test(api): fix flaky test `test_cancel_merge_when_pipeline_succeeds`

JohnVillalovos:
This is an attempt to fix the flaky test
`test_cancel_merge_when_pipeline_succeeds`.
Were seeing a: 405 Method Not Allowed error when setting the MR to
merge_when_pipeline_succeeds.

Closes: #2383

### DIFF
--- a/tests/functional/api/test_merge_requests.py
+++ b/tests/functional/api/test_merge_requests.py
@@ -183,6 +183,7 @@ def test_merge_request_reset_approvals(gitlab_url, project, wait_for_sidekiq):
 
 def test_cancel_merge_when_pipeline_succeeds(project, merge_request, wait_for_sidekiq):
     mr = merge_request(source_branch="test_merge_request_merge", create_pipeline=True)
+    wait_for_sidekiq(timeout=60)
     # Set to merge when the pipeline succeeds, which should never happen
     mr.merge(merge_when_pipeline_succeeds=True)
     wait_for_sidekiq(timeout=60)


tests/functional/api/test_merge_requests.py
```diff
@@ -183,6 +183,7 @@ def test_merge_request_reset_approvals(gitlab_url, project, wait_for_sidekiq):
 
 def test_cancel_merge_when_pipeline_succeeds(project, merge_request, wait_for_sidekiq):
     mr = merge_request(source_branch="test_merge_request_merge", create_pipeline=True)
+    wait_for_sidekiq(timeout=60)
```
lmilbaum, line 186: Not sure I know what the sidekiq so maybe a dumb question. Does it check that the merge requested created, the pipeline triggered and finished running?

JohnVillalovos: I'm not an expert but I believe sidekiq is what takes care of handling the tasks that get run asynchronously. So by waiting we are hoping that all pending tasks will be done before we continue.

nejch: Yes the original idea was to wait for all pending/asynchronous jobs to complete, with the assumption the server is done with whatever we want it to :) probably not the best way to do it, but kind of better than the random  hardcoded guesswork with `time.sleep()` we had before.

nejch (approved):
Thanks @JohnVillalovos let's see if this helps!

PR #2387, opened 2022-11-21
fix(cli): Enable debug before doing auth

em-:
Authentication issues are currently hard to debug since `--debug` only has effect after `gl.auth()` has been called.

For example, a 401 error is printed without any details about the actual HTTP request being sent:

    $ gitlab --debug --server-url https://gitlab.com current-user get
    401: 401 Unauthorized

By moving the call to `gl.enable_debug()` the usual debug logs get printed before the final error message.


nejch (approved):
Ah that makes sense. Our CLI still definitely needs a little more love, thanks for looking into this! :)

PR #2391, opened 2022-11-24
feat(client): bootstrap the http backends concept

lmilbaum:
Initial implementation of http engine concept. The long-term target is to shift all requests/response processing from the client to the http engine. When that is done for requests then it would be easy enough to add another engine, like httpx.

### DIFF
--- a/gitlab/client.py
+++ b/gitlab/client.py
@@ -3,7 +3,7 @@
 import os
 import re
 import time
-from typing import Any, cast, Dict, List, Optional, Tuple, TYPE_CHECKING, Union
+from typing import Any, cast, Dict, List, Optional, Tuple, Type, TYPE_CHECKING, Union
 from urllib import parse
 
 import requests
@@ -14,7 +14,7 @@ import gitlab
 import gitlab.config
 import gitlab.const
 import gitlab.exceptions
-from gitlab import utils
+from gitlab import http_backends, utils
 
 REDIRECT_MSG = (
     "python-gitlab detected a {status_code} ({reason!r}) redirection. You must update "
@@ -32,6 +32,7 @@ _PAGINATION_URL = (
 
 
 class Gitlab:
+
     """Represents a GitLab server connection.
 
     Args:
@@ -53,6 +54,10 @@ class Gitlab:
             or 52x responses. Defaults to False.
         keep_base_url: keep user-provided base URL for pagination if it
             differs from response headers
+
+    Keyward Args:
+        requests.Session session: Http Requests Session
+        RequestsBackend http_backend: Backend that will be used to make http requests
     """
 
     def __init__(
@@ -66,15 +71,14 @@ class Gitlab:
         http_password: Optional[str] = None,
         timeout: Optional[float] = None,
         api_version: str = "4",
-        session: Optional[requests.Session] = None,
         per_page: Optional[int] = None,
         pagination: Optional[str] = None,
         order_by: Optional[str] = None,
         user_agent: str = gitlab.const.USER_AGENT,
         retry_transient_errors: bool = False,
         keep_base_url: bool = False,
+        **kwargs: Any,
     ) -> None:
-
         self._api_version = str(api_version)
         self._server_version: Optional[str] = None
         self._server_revision: Optional[str] = None
@@ -98,7 +102,11 @@ class Gitlab:
         self._set_auth_info()
 
         #: Create a session object for requests
-        self.session = session or requests.Session()
+        http_backend: Type[http_backends.DefaultBackend] = kwargs.pop(
+            "http_backend", http_backends.DefaultBackend
+        )
+        self.http_backend = http_backend(**kwargs)
+        self.session = self.http_backend.client
 
         self.per_page = per_page
         self.pagination = pagination
--- a/gitlab/http_backends/requests_backend.py
+++ b/gitlab/http_backends/requests_backend.py
@@ -1,0 +1,12 @@
+from typing import Optional
+
+import requests
+
+
+class RequestsBackend:
+    def __init__(self, session: Optional[requests.Session] = None) -> None:
+        self._client: requests.Session = session or requests.Session()
+
+    @property
+    def client(self) -> requests.Session:
+        return self._client
--- a/gitlab/http_backends/__init__.py
+++ b/gitlab/http_backends/__init__.py
@@ -1,0 +1,7 @@
+"""
+Defines http backends for processing http requests
+"""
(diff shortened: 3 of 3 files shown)


gitlab/client.py
```diff
@@ -66,14 +71,20 @@ def __init__(
...
         retry_transient_errors: bool = False,
         keep_base_url: bool = False,
+        **kwargs: Any,
     ) -> None:
+        """
+        kwargs:-
+        :param requests.Session session: (optional). Http Requests Session.
+        :param RequestsEngine http_engine: (optional). Engine that will be used
+               to make http requests.
+        """
```
nejch, line 87: Quick note that we use the google style docstrings in this project (https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html). See the class docstring above.

Also, should this be placed along with the class docstring? It looks a bit out of place here on the dunder method. 
See this discussion on SO:
https://stackoverflow.com/questions/1137161/what-is-the-correct-way-to-document-a-kwargs-parameter

gitlab/http_engines/requests_engine.py
```diff
@@ -0,0 +1,15 @@
...
+
+import requests
+
+
+class RequestsEngine:
+
+    _client: requests.Session
+
+    def __init__(self, **kwargs: Any) -> None:
+        self._client = kwargs.pop("session", requests.Session())
```
nejch, line 11: I'm thinking that on the specific implementation, it can be a proper keyword argument and not popped from `kwargs`, right? Or am I missing something?

gitlab/client.py
```diff
@@ -98,7 +109,9 @@ def __init__(
         self._set_auth_info()
 
         #: Create a session object for requests
-        self.session = session or requests.Session()
+        http_engine = kwargs.pop("http_engine", http_engines.DefaultEngine)
+        self.http_engine = http_engine(**kwargs)
+        self.session = self.http_engine.client
```
nejch, line 114: note to self, Most likely we will want to get rid of this `self.session` and abstract it in the engine/backend code so each implementation implements `close()` internally But we can leave it here for now.

nejch (commented):
Thanks for breaking this up @lmilbaum! Just a few minor questions.

We used the redminelib approach here IIRC, after poking around github a bit more I think also calling these `http backends`, e.g. `from gitlab import backends` might make sense, but just a thought.

gitlab/client.py
lmilbaum: Thanks @nejch for clarification. I would expected the CI to catch that one. But, not in the scope of this PR.  Pushed a fix. LMK if that is still not in the correct format.

gitlab/http_engines/requests_engine.py
lmilbaum: Makes totally sense. Pushed a change.

lmilbaum:
I would prefer that #2392 lands before this one. It is needed for testing this PR.

nejch:
> I would prefer that #2392 lands before this one. It is needed for testing this PR.

@lmilbaum it shouldn't block if you cherry-pick the commit from https://github.com/python-gitlab/python-gitlab/pull/2397 here (feel free to cherry pick already), it seems like the tests in that PR are really meant for this one, then we can really test the method we are changing (`__init__`) and not indirectly through another method.

(Our use of `Gitlab.from_config()` in tests is a bit odd, leading to problems like this one. I think we should move away from it for default tests. Then we can look at `from_config()` independently :) )

lmilbaum:
@nejch could you please take a look. I would like to move forward. Thanks.

gitlab/client.py
```diff
@@ -98,7 +108,9 @@ def __init__(
         self._set_auth_info()
 
         #: Create a session object for requests
-        self.session = session or requests.Session()
+        http_backend = kwargs.pop("http_backend", http_backends.DefaultBackend)
```
nejch, line 111: Based on the comment above, this should disambiguate the session object as well as any other potential backend-related typing (needs a `Type` import at the top):

```suggestion
        http_backend: Type[http_backends.DefaultBackend] = kwargs.pop(
            "http_backend", http_backends.DefaultBackend
        )
```

gitlab/http_backends/requests_backend.py
```diff
@@ -0,0 +1,15 @@
...
+
+import requests
+
+
+class RequestsBackend:
+
+    _client: requests.Session
+
+    def __init__(self, session: Optional[requests.Session] = None) -> None:
+        self._client = session or requests.Session()
```
nejch, line 11: Same question as above, I believe we can add the hint directly in the assignment? mypy seems to be happy with the move.

```suggestion
    def __init__(self, session: Optional[requests.Session] = None) -> None:
        self._client: requests.Session = session or requests.Session()
```

gitlab/client.py
```diff
@@ -32,6 +32,13 @@
 
 
 class Gitlab:
+
+    # The session object is defined here as long as the RequestsBackend is not
+    # fully implemented. The long-term goal is to wrap all the http
+    # requests/responses by the RequestsBackend class
+
+    session: requests.Session
+
```
nejch, line 41: For other variables in `__init__` we add type hints along with the variable assignment directly, I think we could be consistent here WDYT? The ambiguity comes from `http_backend`, so I would add the default hint there.

```suggestion
```

nejch (changes requested):
Sorry for the delays @lmilbaum, was out for a week. Looks great, just some tiny questions on the typing approach.

gitlab/client.py
lmilbaum: Sorry for my ignorance. Not that I mind adding this change but which concern is this change addressing for the current change? I would prefer to delay it to a later stage if that is related to the follow up changes.

gitlab/http_backends/requests_backend.py
lmilbaum: Same comment as above

gitlab/client.py
```diff
@@ -66,15 +71,14 @@ def __init__(
         http_password: Optional[str] = None,
         timeout: Optional[float] = None,
         api_version: str = "4",
-        session: Optional[requests.Session] = None,
```
nejch, line 69: /cc @JohnVillalovos technically this is a small breaking change, but it's buried so deep into keyword args I can't imagine it being a problem, as people cannot provide multiple tokens in any case. So these would likely be provided as keyword anyway and the order would not be affected.

nejch (approved):
Thanks again for all the work here @lmilbaum!

PR #2392, opened 2022-11-24
feat: allow passing kwargs to Gitlab class when instantiating with `from_config`

lmilbaum:
This PR is needed for the validation of #2391

### DIFF
--- a/tests/functional/api/test_gitlab.py
+++ b/tests/functional/api/test_gitlab.py
@@ -1,4 +1,5 @@
 import pytest
+import requests
 
 import gitlab
 
@@ -21,6 +22,24 @@ def test_auth_from_config(gl, temp_dir):
     )
     test_gitlab.auth()
     assert isinstance(test_gitlab.user, gitlab.v4.objects.CurrentUser)
+
+
+def test_no_custom_session(gl, temp_dir):
+    """Test no custom session"""
+    custom_session = requests.Session()
+    test_gitlab = gitlab.Gitlab.from_config(
+        config_files=[temp_dir / "python-gitlab.cfg"]
+    )
+    assert test_gitlab.session != custom_session
+
+
+def test_custom_session(gl, temp_dir):
+    """Test custom session"""
+    custom_session = requests.Session()
+    test_gitlab = gitlab.Gitlab.from_config(
+        config_files=[temp_dir / "python-gitlab.cfg"], session=custom_session
+    )
+    assert test_gitlab.session == custom_session
 
 
 def test_broadcast_messages(gl, get_all_kwargs):
--- a/docs/api-usage-advanced.rst
+++ b/docs/api-usage-advanced.rst
@@ -181,3 +181,15 @@ on your own, such as for nested API responses and ``Union`` return types. For ex
 
    if TYPE_CHECKING:
       assert isinstance(license["plan"], str)
+
+Custom session (Bring your own Session) 
+---------------------------------------
+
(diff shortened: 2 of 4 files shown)


tests/functional/api/test_gitlab.py
```diff
@@ -23,6 +24,24 @@ def test_auth_from_config(gl, temp_dir):
...
+    assert test_gitlab.session != custom_session
+
+
+def test_custom_session(gl, temp_dir):
+    """Test custom session"""
+    custom_session = requests.Session()
+    test_gitlab = gitlab.Gitlab.from_config(
+        config_files=[temp_dir / "python-gitlab.cfg"], session=custom_session
+    )
+    assert test_gitlab.session == custom_session
```
nejch, line 42: Since we're not making any real requests, I think these could be in unit tests, and done on `gitlab.Gitlab()` where the session logic is handled.

nejch (changes requested):

docs/api-usage.rst
```diff
@@ -62,6 +62,13 @@ You can also use configuration files to create ``gitlab.Gitlab`` objects:
 
    gl = gitlab.Gitlab.from_config('somewhere', ['/tmp/gl.cfg'])
 
+With custom session
+
+.. code-block:: python
+
+   gl = gitlab.Gitlab.from_config('somewhere', ['/tmp/gl.cfg'], session=custom_session)
+
+
```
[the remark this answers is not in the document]
lmilbaum: ok

tests/functional/api/test_gitlab.py
lmilbaum: Could you please clarify how a real request can make a difference? 
I could add unit tests as well if you think it makes sense.

nejch: Sorry for the confusion, what I meant to say is that here we're essentially testing the `__init__()` method of the `Gitlab` class, so ideally we would put these tests in https://github.com/python-gitlab/python-gitlab/blob/main/tests/unit/test_gitlab.py. Probably somewhere in here:

https://github.com/python-gitlab/python-gitlab/blob/0ecf3bbe28c92fd26a7d132bf7f5ae9481cbad30/tests/unit/test_gitlab.py#L76-L96

lmilbaum: I've tried the direction you are suggestion. It complicates the implementation while I don't fully understand the value. In my point of view, testing the `__init__` function functionality is valuable.

lmilbaum: After re-reviewing the tests, those functions tests the `from_config` function

lmilbaum: I've decided to implement both unit tests and functional tests. It might be redundant for now but still worth keeping both of them.

nejch: Thanks! I think this can go in as is, but just to clarify: my point regarding unit/functional was more that this test can easily be done in isolation and doesn't need to live in the functional test suite where we fire up a real server to make requests against :)

I guess there was some confusion about `__init__`, my suggestion would be to simply change the test to assert after `gitlab.Gitlab()` is instantiated with a custom session, as that is where the logic is. In any case, https://github.com/python-gitlab/python-gitlab/pull/2397 should help in the future, as I assume you are testing `from_config` mostly because we use it everywhere in our functional tests.

lmilbaum: TBH, I'm not following :-)

PR #2396, opened 2022-11-26
feat(files): allow decoding project files directly to string

nejch:
Provides a slightly less clunky interface so people don't need to do `f.decode.().decode("utf-8")` in user code.

### DIFF
--- a/docs/gl_objects/projects.rst
+++ b/docs/gl_objects/projects.rst
@@ -434,8 +434,11 @@ Get a file::
     # get the base64 encoded content
     print(f.content)
 
-    # get the decoded content
+    # get the decoded content as bytes
     print(f.decode())
+
+    # get the decoded content as a string
+    print(f.decode("utf-8"))
 
 Get file details from headers, without fetching its entire content::
 
--- a/gitlab/v4/objects/files.py
+++ b/gitlab/v4/objects/files.py
@@ -7,6 +7,7 @@ from typing import (
     Iterator,
     List,
     Optional,
+    overload,
     TYPE_CHECKING,
     Union,
 )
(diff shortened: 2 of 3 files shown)


gdubicki:
Hi! Sorry if I am oversharing here, especially as I don't know this project and its conventions well (yet!).

While this is definitely an improvement from the previous syntax I feel that this could and maybe should be even simpler.

Because people mostly store source code in git then the most popular use-case here will probably be getting the content of the file as `UTF-8` text.

So I propose to add a method `f.text(encoding='UTF-8')` (similar to [Requests](https://requests.readthedocs.io/en/latest/#)'s `t.text`) so you could change the encoding but most users could just do `f.text()`. 

For bytes I would personally prefer having `f.bytes()` methods to avoid using `decode` keyword as for me it's almost never clear what it does as it depends on the context. Especially in GitLab Files API, where we have double encoding, first with base64 and then with a text encoding like `UTF-8`.

For completeness I would add `f.base64()`. This way you can have a single call to any form of the underlying file.

PS I've never used `encoding = 'text'` in GitLab Files API so I don't know - doesn't it cause the file to be stored without base64? If so, we should make sure that our helper functions deal with those 2 cases, base64-encoded files and not.

nejch:
@gdubicki thanks for the feedback! I'll take another look at what wording makes the most sense, especially when compared to `requests`/`httpx` and similar underlying libraries.

For context, this `decode()` method is one of the few public methods in python-gitlab that are completely custom (i.e. not implementing an API endpoint) and here for the user's convenience rather than to mirror API functionality (this one introduced by the original author in https://github.com/python-gitlab/python-gitlab/commit/2eac07139eb288cda2dd2d22d191ab3fc1053437). 

In most other cases we do not add more higher-level convenience and instead keep it a thin wrapper, so I'd be wary of introducing more methods but if we do already provide some decoding then it probably makes sense to make it _actually_ convenient for the user ;)

nejch:
Maybe an additional `.text()` really is more explicit. It's 2 totally different kind of `decode`s we're calling so maybe we shouldn't mix them. I'll rework this a bit :)

PR #2404, opened 2022-12-02
fix: Use the ProjectIterationManager within the Project object

### LINKED ISSUE #2403: project iterations are broken
project objects are using the `GroupIterationManager` instead of the `ProjectIterationManager` resulting in a `404 Group Not Found` error received from the server when no project ID matches a group ID. I haven't come across it, but I could see this returning the matching group IDs iterations instead of the project ID's iterations if a matching group ID were to exist.

hollerdawg:
The Project object was previously using the GroupIterationManager resulting in the incorrect API endpoint being used. Utilize the correct ProjectIterationManager instead.

Resolves #2403


nejch (approved):
Thanks for the catch @hollerdawg!

Note to self, this was probably an oversight when I added it initially because "project iterations" don't really exist anymore and associated group iterations are returned instead. So we might need to revisit this a bit if there's any kind of iteration-level endpoint later.

PR #2406, opened 2022-12-03
fix(api): Make description optional for releases

lovetox:
For the Release API Endpoint `description` is not a required attribute, see https://docs.gitlab.com/14.10/ee/api/releases/#create-a-release


nejch (approved):
Thanks for the catch and for your PR @lovetox, not sure if this changed at some point or if we simply missed this.

lovetox:
I checked docs back to Gitlab Version 14 and it was als optional, so maybe it changed but not recently.

Issue #2263, opened 2022-08-30
Deprecation warning due to requests-toolbelt dependency


robsan00:
## Description of the problem, including code/CLI snippet
(I thought it is not a bad idea to open a ticket here as this is a pretty active project while requests-toolbelt is everything but, i.e. it is not impossible that requests will kick out the deprecated module before requests-toolbelt will update, i.e. this project might become unusable.)
If you use python-gitlab with the latest urllib3 version (and pyopenSSL and cryptography installed), you get a deprecation warning.

Virtualenv:
```
(gitlab_req) C:\work\dev\venv>pip freeze
certifi==2022.6.15
cffi==1.15.1
charset-normalizer==2.1.1
cryptography==37.0.4
idna==3.3
pycparser==2.21
pyOpenSSL==22.0.0
python-gitlab==3.9.0
requests==2.28.1
requests-toolbelt==0.9.1
urllib3==1.26.12
```

Weirdly, the warning seems to be swallowed somewhere along the way, but if I try to use gitlab within pytest for example, I get the message. Consider the following dummy test:
```
import gitlab

def test_hugo():
    pass
```
Executing this test results in the warning:
```
(gitlab_req) C:\work>pytest test.py -v -s
================================================= test session starts =================================================
platform win32 -- Python 3.8.8, pytest-7.1.2, pluggy-1.0.0 -- c:\work\dev\venv\gitlab_req\scripts\python.exe
cachedir: .pytest_cache
rootdir: C:\work
collecting ... 1
collected 1 item

test.py::test_hugo PASSED

================================================== warnings summary ===================================================
dev\venv\gitlab_req\lib\site-packages\requests_toolbelt\_compat.py:56
  c:\work\dev\venv\gitlab_req\lib\site-packages\requests_toolbelt\_compat.py:56: DeprecationWarning: 'urllib3.contrib.pyopenssl' module is deprecated and will be removed in a future release of urllib3 2.x. Read more in this issue: https://github.com/urllib3/urllib3/issues/2680
    from requests.packages.urllib3.contrib.pyopenssl \

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
============================================ 1 passed, 1 warning in 0.20s =============================================
```

Here the related issue at requests-toolbelt:
https://github.com/requests/toolbelt/issues/331

## Expected Behavior

no deprecation warning shown

## Actual Behavior

deprecation warning shown

## Specifications

[...]

PR #2408, opened 2022-12-04
fix(deps): bump requests-toolbelt to fix deprecation warning

nejch:
Closes https://github.com/python-gitlab/python-gitlab/issues/2263

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,7 +28,7 @@ setup(
     license="LGPL-3.0-or-later",
     url="https://github.com/python-gitlab/python-gitlab",
     packages=find_packages(exclude=["docs*", "tests*"]),
-    install_requires=["requests>=2.25.0", "requests-toolbelt>=0.9.1"],
+    install_requires=["requests>=2.25.0", "requests-toolbelt>=0.10.1"],
     package_data={
         "gitlab": ["py.typed"],
     },


lmilbaum:
@nejch Does it make sense to add a test that verifies the warning is no longer present?

nejch:
> @nejch Does it make sense to add a test that verifies the warning is no longer present?

@lmilbaum the warning was reported here due to upstream inactivity, but actually coming from a transitive dependency (urllib3) via another dependency (requests-toolbelt). It was fixed in that release, so I think a test should be upstream in toolbelt rather than in the scope of our library (maybe this was already done in https://github.com/requests/toolbelt/pull/336). Unless I misunderstood and you meant a more general test for deprecation warnings (not sure if that's feasible, deprecation warnings do happen).

In any case this is a very minor issue, we do not restrict a maximum version of requests-toolbelt, so users are free to upgrade already on their own (and if they don't pin transitive dependencies in lockfiles then they probably are already on the latest version). Hope that makes sense :)

lmilbaum:
> > @nejch Does it make sense to add a test that verifies the warning is no longer present?
> 
> @lmilbaum the warning was reported here due to upstream inactivity, but actually coming from a transitive dependency (urllib3) via another dependency (requests-toolbelt). It was fixed in that release, so I think a test should be upstream in toolbelt rather than in the scope of our library (maybe this was already done in [requests/toolbelt#336](https://github.com/requests/toolbelt/pull/336)). Unless I misunderstood and you meant a more general test for deprecation warnings (not sure if that's feasible, deprecation warnings do happen).
> 
> In any case this is a very minor issue, we do not restrict a maximum version of requests-toolbelt, so users are free to upgrade already on their own (and if they don't pin transitive dependencies in lockfiles then they probably are already on the latest version). Hope that makes sense :)

Totally makes sense. Thanks for the clarification.

PR #2411, opened 2022-12-05
refactor: migrating timeout from client to http_backend

lmilbaum:
@nejch The PR is ready for review but I'm not sure the functional tests cover this change. LMK if you think that adding functional tests might be of help.

### DIFF
--- a/gitlab/client.py
+++ b/gitlab/client.py
@@ -43,7 +43,6 @@ class Gitlab:
         ssl_verify: Whether SSL certificates should be validated. If
             the value is a string, it is the path to a CA file used for
             certificate validation.
-        timeout: Timeout to use for requests to the GitLab server.
         http_username: Username for HTTP authentication
         http_password: Password for HTTP authentication
         api_version: Gitlab API version to use (support for 4 only)
@@ -57,6 +56,7 @@ class Gitlab:
 
     Keyward Args:
         requests.Session session: Http Requests Session
+        float timeout: Timeout to use for requests to the GitLab server.
         RequestsBackend http_backend: Backend that will be used to make http requests
     """
 
@@ -69,7 +69,6 @@ class Gitlab:
         ssl_verify: Union[bool, str] = True,
         http_username: Optional[str] = None,
         http_password: Optional[str] = None,
-        timeout: Optional[float] = None,
         api_version: str = "4",
         per_page: Optional[int] = None,
         pagination: Optional[str] = None,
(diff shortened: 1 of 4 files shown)


nejch:
Thanks a lot for the PR @lmilbaum! Could you give me a bit of context on why this change is necessary? As it's just an `int`, I don't see from this PR how this would be dependent on a specific backend implementation. I'd assume we can just pass it on to the backend(s) as well as the custom methods (as in `projects.py`), and keep a minimum of changes to the `Gitlab` client itself.

Just asking as I think we need to keep breaking changes to a minimum. We are already introducing a slight breaking behavior with our `session` migration (though I think that's much more backend-specific and needed).

lmilbaum:
> Thanks a lot for the PR @lmilbaum! Could you give me a bit of context on why this change is necessary? As it's just a `float`, I don't see from this PR how this would be dependent on a specific backend implementation. I'd assume we can just pass it on to the backend(s) as well as the custom methods (as in `projects.py`), and keep a minimum of changes to the `Gitlab` client itself.
> 
> Just asking as I think we need to keep breaking changes to a minimum. We are already introducing a slight breaking behavior with our `session` migration (though I think that's much more backend-specific and needed).

What makes sense to me is to wrap all the related business logic by the http backend. `timeout` is part of the related business logic.
I can close the PR and move on to other functions if you find it risky.

nejch:
> What makes sense to me is to wrap all the related business logic by the http backend. `timeout` is part of the related business logic. I can close the PR and move on to other functions if you find it risky.

I may be wrong here but there will always be some things relevant to the client itself and likely even across different clients in the future (REST/GraphQL): base URL, auth params, SSL verify, retry params, timeout, user agent etc (of course backends may process these differently internally depending on implementation). And we can also see with the custom `import_*` methods that reuse it outside of backend code, it may be useful like that sometimes.

By keeping these as explicit arguments in the public interface, IMO we are also helping the end user as it's very clear in type hints, IDE auto-completion and in auto-generated docs.

lmilbaum:
> > What makes sense to me is to wrap all the related business logic by the http backend. `timeout` is part of the related business logic. I can close the PR and move on to other functions if you find it risky.
> 
> I may be wrong here but there will always be some things relevant to the client itself and likely even across different clients in the future (REST/GraphQL): base URL, auth params, SSL verify, retry params, timeout, user agent etc (of course backends may process these differently internally depending on implementation). And we can also see with the custom `import_*` methods that reuse it outside of backend code, it may be useful like that sometimes.
> 
> By keeping these as explicit arguments in the public interface, IMO we are also helping the end user as it's very clear in type hints, IDE auto-completion and in auto-generated docs.

Makes sense. Closing the PR.

PR #2413, opened 2022-12-06
refactor: move the request call to the backend

lmilbaum:


gitlab/http_backends/requests_backend.py
```diff
@@ -10,3 +11,40 @@ def __init__(self, session: Optional[requests.Session] = None) -> None:
     @property
     def client(self) -> requests.Session:
         return self._client
+
+    def http_request(
+        self,
+        method: Union[str, bytes],
```
nejch, line 17: This can only ever be a string, right? It can only really be one of `get`,`post`,`put`,`delete` etc.

```suggestion
        method: str,
```

gitlab/http_backends/requests_backend.py
```diff
@@ -10,3 +11,40 @@ def __init__(self, session: Optional[requests.Session] = None) -> None:
...
+        url: str,
+        json: Optional[Union[Dict[str, Any], bytes]] = None,
```
nejch, line 19: The `bytes` here seems incorrect, I think we need to fix the typing in our `_prepare_send_data` which is probably the reason you added it here. I will have a look.

Also just thinking, since `_prepare_send_data` involves backend-specific code (at least the `MultipartEncoder` is requests/toolbelt specific), I assume we will potentially move (some of) that code into the backend as a follow-up.

nejch (changes requested):
Thanks @lmilbaum, just a few initial questions on the typing here.

gitlab/http_backends/requests_backend.py
lmilbaum: Thanks @nejch for the review. Do you want to hold this PR before `_prepare_send_data` is amended or we can continue with this one and provide a fix later on?

nejch: Had a quick look through requests/urllib3/json and seems like `bytes` is totally fine for json as well, although I don't expect it coming from our code. Maybe we can leave it as is, for now. Sorry for the detour here.

gitlab/http_backends/requests_backend.py
```diff
@@ -10,3 +11,40 @@ def __init__(self, session: Optional[requests.Session] = None) -> None:
...
+    ) -> requests.Response:
+        """Make HTTP request
+
+        Args:
+            method: The HTTP method to call ('get', 'post', 'put', 'delete')
+            url: Path or full URL
+            stream: Whether the data should be streamed
+            files: The files to send to the server
+            timeout: The timeout, in seconds, for the request
+            json: Data to send in the body in json by default
```
nejch, line 35: Just one more nit:

```suggestion
            method: The HTTP method to call ('get', 'post', 'put', 'delete', etc.)
            url: The full URL
            stream: Whether the data should be streamed
            data: The data to send to the server in the body of the request
            timeout: The timeout, in seconds, for the request
            verify: Whether SSL certificates should be validated. If
                the value is a string, it is the path to a CA file used for
                certificate validation.
            json: Data to send in the body in json by default
```

nejch (approved):

PR #2419, opened 2022-12-09
feat: allow filtering pipelines by source

nickbroon:
See: https://docs.gitlab.com/ee/api/pipelines.html#list-project-pipelines

Added in GitLab 14.3


nejch (approved):
Thanks @nickbroon for keeping things tidy here! :) Will be available on the 28th in the CLI. With the API, you should be able to use it already.

PR #2420, opened 2022-12-10
refactor: move Response object to backends

lmilbaum:


gitlab/client.py
```diff
@@ -774,47 +774,51 @@ def http_request(
 
                 raise
 
-            self._check_redirects(result)
+            self._check_redirects(result.response)
 
-            if 200 <= result.status_code < 300:
-                return result
+            if 200 <= result.response.status_code < 300:
+                return result.response
```
nejch, line 780: This is getting a bit nested here, could we maybe just define the variables so we don't repeat the usage of `result.response.attribute`? I think this way we might also be able to hide the backend details a bit later.

```python
headers = result.response.headers
status_code = result.response.status_code
content = result.response.content # replaces error_message = result.content
```

lmilbaum: Good idea. I've implemented it a bit differently. WDYT?

lmilbaum:
@nejch Kindest reminder

gitlab/http_backends/requests_backend.py
```diff
@@ -4,6 +4,30 @@
...
+    @property
+    def response(self) -> requests.Response:
+        return self._response
+
+    @property
+    def status_code(self) -> int:
+        return self._response.status_code
+
+    @property
+    def headers(self) -> Any:
```
nejch, line 20: Should this be requests' `CaseInsensitiveDict` here?

```suggestion
    def headers(self) -> CaseInsensitiveDict[str]:
```

gitlab/client.py
nejch: Thanks, this works for me although I think we will need to add more Response attributes later, for history and reason phrases at least.

nejch (changes requested):
Sorry for the delay @lmilbaum, I've looked at this a few times now and looks ok to me, I'm just not 100% sure how we're going to split this later for streaming responses. Let's try this, just one more question.

gitlab/http_backends/requests_backend.py
lmilbaum: Originally I used it but it caused some errors which I didn't know how to fix. Could you please take a look?

lmilbaum: I've also filed an issue psf/requests#6311

nejch: Ah ok, I'll take a look a bit later, thanks. I don't think the requests project can help with this as they don't provide any typing at all themselves, this would be in the typeshed project most likely in case there is an upstream issue.

lmilbaum: The `requests` project maintainer provided me a hint on the root cause. Pushed a change which removed the errors.

PR #2435, opened 2022-12-19
refactor: Replacing http_requests return type hint

lmilbaum:
http_request() will be deprecated in future release in favor of backend_request(). That is because of the return type change and the need to warn users of the change.

### DIFF
--- a/gitlab/client.py
+++ b/gitlab/client.py
@@ -639,7 +639,7 @@ class Gitlab:
                 )
             )
 
-    def http_request(
+    def backend_request(
         self,
         verb: str,
         path: str,
@@ -653,7 +653,7 @@ class Gitlab:
         retry_transient_errors: Optional[bool] = None,
         max_retries: int = 10,
         **kwargs: Any,
-    ) -> requests.Response:
+    ) -> _backends.DefaultResponse:
         """Make an HTTP request to the Gitlab server.
 
         Args:
@@ -722,7 +722,7 @@ class Gitlab:
         cur_retries = 0
         while True:
             try:
-                result = self._backend.http_request(
+                backend_response = self._backend.http_request(
                     method=verb,
                     url=url,
                     json=send_data.json,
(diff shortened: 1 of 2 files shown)


gitlab/client.py
```diff
@@ -749,7 +749,7 @@ def http_request(
             self._check_redirects(result.response)
 
             if 200 <= result.status_code < 300:
-                return result.response
+                return result
```
nejch, line 752: This is a breaking change, so this would need to be at least communicated to users via a breaking change trailer (and triger 4.0.0). But also I would like to see how this affects our streamed responses and how we deal with `download()` when `iterator=True` is passed to it, where people iterate over chunks on their own.

I'd like to do some local testing for this to see what behavior we get.

lmilbaum: Make sense. How can I help?

nejch: @lmilbaum I like your idea with protocols (or maybe abstract classes) in the other PR, but I would maybe try to apply it here for backends instead of the client, if that makes sense. I actually had a similar WIP locally, where I started with some ABC work on the backends rather than the client itself. Have you tried that as well?

lmilbaum: @nejch This is a great idea. Updated the other PR. Take a look.

gitlab/client.py
```diff
@@ -653,7 +653,7 @@ def http_request(
         retry_transient_errors: Optional[bool] = None,
         max_retries: int = 10,
         **kwargs: Any,
-    ) -> requests.Response:
+    ) -> _backends.DefaultResponse:
```
nejch, line 656: I'll wait for https://github.com/python-gitlab/python-gitlab/pull/2442 to get merged and I think we can then do some magic with generic types perhaps.:mage:

lmilbaum:
@nejch after moving the protocols to the backend, this change will pass because the change is for the front end (client) http_request call. The whole point of the protocol is to identify this kind of change.

gitlab/client.py
lmilbaum: The end goal is to move the usage of the requests package to the backend. Using generic types like Union, can work for a transition period, but eventually, it will have to be removed.

nejch: I had in mind more something to do with `TypeVar` to define acceptable types or an upper bound, so we could include both requests and httpx. But this works for me for now.

lmilbaum:
@nejch @JohnVillalovos Can you please review this PR?

gitlab/client.py
```diff
@@ -769,36 +775,74 @@ def should_retry() -> bool:
...
+        obey_rate_limit: bool = True,
+        retry_transient_errors: Optional[bool] = None,
+        max_retries: int = 10,
+        **kwargs: Any,
+    ) -> requests.Response:
+        utils.warn(
+            "`http_request()` is deprecated and will be removed in a future version.\n"
+            "Please use `backend_request()` instead.",
+            category=DeprecationWarning,
+        )
```
nejch, line 830: I'm a bit wary of this, it would feel wrong deprecating this after we already had a similar discussion in https://github.com/python-gitlab/python-gitlab/pull/1842. What would be the issue with having a `http_request` in the public client?

Keeping in mind that the user does not care about the backends generally.

lmilbaum: >This is a breaking change, so this would need to be at least communicated to users via a breaking change trailer (and triger 4.0.0).

I've thought it would make sense using this method and not to trigger 4.0.0 for now.

nejch: @lmilbaum what I meant was what would be wrong with simply keeping `http_backend` on the client? It's quite consistent with the existing `http_*` methods (there is some history in the other MR, where we discussed we'd keep it in the public client as it is).

lmilbaum: I am not following. `http_backend` is kept on the client while having backend-specific implementation in the backend package.
I was just trying to address the valid concern you have raised regarding the breaking change.

PR #2438, opened 2022-12-20
ci: Complete all unit tests even if one has failed

lmilbaum:
There are cases where unit tests are failing for a specific python version but not for all


nejch (commented):
Thanks @lmilbaum!

PR #2439, opened 2022-12-20
refactor: add reason property to RequestsResponse

lmilbaum:
main is broken. This PR is fixing it.


nejch (approved):
Thanks @lmilbaum. I'll reword this to a refactor or chore commit message as this has not yet been released and will not be a user-facing change for the changelog.

PR #2442, opened 2022-12-26
feat(backends): use PEP544 protocols for structural subtyping

lmilbaum:
The purpose of this change is to track [api](https://github.com/python-gitlab/python-gitlab/blob/main/docs/api-levels.rst) changes. For example: package versioning and breaking change announcement in case of protocol change.
This is MVP implementation to be used by #2435 
Haven't figured out yet how to implement unit tests for a protocol and its value.

### DIFF
--- a/gitlab/_backends/protocol.py
+++ b/gitlab/_backends/protocol.py
@@ -1,0 +1,34 @@
+import abc
+import sys
+from typing import Any, Dict, Optional, Union
+
+import requests
+from requests_toolbelt.multipart.encoder import MultipartEncoder  # type: ignore
+
+if sys.version_info >= (3, 8):
+    from typing import Protocol
+else:
+    from typing_extensions import Protocol
+
+
+class BackendResponse(Protocol):
+    @abc.abstractmethod
+    def __init__(self, response: requests.Response) -> None:
+        ...
+
+
+class Backend(Protocol):
+    @abc.abstractmethod
+    def http_request(
+        self,
+        method: str,
+        url: str,
+        json: Optional[Union[Dict[str, Any], bytes]],
+        data: Optional[Union[Dict[str, Any], MultipartEncoder]],
+        params: Optional[Any],
+        timeout: Optional[float],
+        verify: Optional[Union[bool, str]],
+        stream: Optional[bool],
+        **kwargs: Any,
+    ) -> BackendResponse:
+        ...
--- a/gitlab/_backends/requests_backend.py
+++ b/gitlab/_backends/requests_backend.py
@@ -6,8 +6,10 @@ import requests
 from requests.structures import CaseInsensitiveDict
 from requests_toolbelt.multipart.encoder import MultipartEncoder  # type: ignore
 
+from . import protocol
 
-class RequestsResponse:
+
+class RequestsResponse(protocol.BackendResponse):
     def __init__(self, response: requests.Response) -> None:
         self._response: requests.Response = response
 
@@ -35,7 +37,7 @@ class RequestsResponse:
         return self._response.json()
 
 
-class RequestsBackend:
+class RequestsBackend(protocol.Backend):
     def __init__(self, session: Optional[requests.Session] = None) -> None:
         self._client: requests.Session = session or requests.Session()
 


.github/workflows/protocol.yml
```diff
@@ -0,0 +1,20 @@
...
+      - main
+    paths:
+      - gitlab/protocols/**
+
+jobs:
+  protocol:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Fail
+      run: /bin/false
```
nejch, line 20: Perhaps we can use one of our existing test suites if we write tests for this.

See also my comment on your other PR - if it maybe makes sense to do this at the backend level, and hide some of this from the client.

lmilbaum: I can't figure out how to implement a test which will identify the breaking change if a contributor would update the protocol.  The maintainer will have to manually watch for protocol change and be alerted. That is what this PR was about. Do you have ideas how to set it up?

gitlab/protocols/__init__.py
```diff
@@ -0,0 +1,3 @@
+"""
+Defines protocols
+"""
```
nejch, line 3: Same as above: rather than having a separate filepath for this I think this belongs along with the backends in `gitlab/_backends`. Probably both in something like `gitlab/_backends/base.py` to keep it simple for now, also simplifies the imports :)

nejch (changes requested):
Thanks @lmilbaum, I'm quite slow with the reviews here. Just a small note on keeping things simple for now.

gitlab/protocols/http_backend.py
```diff
@@ -0,0 +1,29 @@
+import sys
+from abc import abstractmethod
```
JohnVillalovos, line 2: I personally (I think @nejch may have a different opinion) am not a fan of importing methods/classes.

I think just `import abc` is good and then easier when reading the code below to know where it comes from.

This is also recommended by Google's Python Style Guide, though I am not suggesting we adopt Google's Python Style Guide 🙂 And yes I realize that we are breaking this guideline in our current code 😥

https://google.github.io/styleguide/pyguide.html

2.2 Imports

Use import statements for packages and modules only, not for individual classes or functions.

gitlab/protocols/http_backend_response.py
```diff
@@ -0,0 +1,15 @@
+import sys
+from abc import abstractmethod
```
JohnVillalovos, line 2: I personally (I think @nejch may have a different opinion) am not a fan of importing methods/classes.

I think just `import abc` is good and then easier when reading the code below to know where it comes from.

JohnVillalovos (commented):
Thanks. Can the commit message have some details on the "why" of this PR?

JohnVillalovos:
> Thanks. Can the commit message have some details on the "why" of this PR?

I still would like to see the description be put into the commit message.

At the moment all I see is:
```
$ git show --no-patch
commit 958d35d3c3d1755edf651cb15f5d1397ca571cc6 (HEAD -> refs/heads/protocol, refs/remotes/origin/protocol)
Author: Liora Milbaum <liora@lmb.co.il>
Date:   Mon Dec 26 06:39:58 2022 +0200

    feat: PEP 544 – Protocols: Structural subtyping (static duck typing)

```

gitlab/_backends/protocol.py
```diff
@@ -0,0 +1,34 @@
...
+
+if sys.version_info >= (3, 8):
+    from typing import Protocol
+else:
+    from typing_extensions import Protocol
+
+
+class BackendResponse(Protocol):
+    @abc.abstractmethod
+    def __init__(self, response: requests.Response) -> None:
```
JohnVillalovos, line 16: Has this idea been tested? I mean if we want to use `httpx` instead of `requests` how will this look?

I'm assuming that the idea is that we will be able to use `BackendResponse` and `Backend` in place of both `httpx` and `requests`. So wondering how that will look. Just wanting to make sure that this will work for that case.

lmilbaum: This represent the existing interface which is `requests` only based. The plan is to change the interface as you described as soon as `httpx` would be introduced. Hoping that this answers your concern.

nejch: Let's cross that bridge when we get there, we can still change the approach if it doesn't work - otherwise if we wanted to test with httpx already, we'd have a chunky PR that would also be hard to review :sweat_smile:

nejch: But I think if the protocol defines all the methods and attributes we might be ok. It will be a problem where httpx and requests differ, for example `iter_content` vs `iter_bytes` etc. But we can maybe have little wrappers for that.

nejch:
> > Thanks. Can the commit message have some details on the "why" of this PR?
> 
> I still would like to see the description be put into the commit message.
> 
> At the moment all I see is:

As an alternative we can squash-merge and it will link to the PR which has much more context than we would usually put in commit messages. WDYT @JohnVillalovos

JohnVillalovos:
> As an alternative we can squash-merge and it will link to the PR which has much more context than we would usually put in commit messages. WDYT @JohnVillalovos

That is okay if it is in the commit message. I want the details in the Git log. Not only on GitHub. What if the project gets moved to GitLab in the future. That information can disappear if it isn't in the git repository.

Seems relatively easy to take the description above and put it in the commit message. But maybe I am missing something?

nejch:
Sure, we can maybe copy it during the squash merge in the description field and it'll appear in the body. I've already sneakily edited the PR title a bit :smiley_cat: 

GitLab's importer is actually quite good these days so it would preserve the commit-to-MR associations, I personally like  that over long commit bodies as it includes the review context but that's subjective of course.

I'll approve for now, and see if there's anything else to cover :)

nejch (approved):

gitlab/_backends/protocol.py
```diff
@@ -0,0 +1,34 @@
...
+import sys
+from typing import Any, Dict, Optional, Union
+
+import requests
+from requests_toolbelt.multipart.encoder import MultipartEncoder  # type: ignore
+
+if sys.version_info >= (3, 8):
+    from typing import Protocol
+else:
+    from typing_extensions import Protocol
```
nejch, line 11: Actually, just one more comment. This won't work on 3.7 without a change in `setup.py`. But I'll open a follow-up for that.

gitlab/_backends/protocol.py
nejch: Just to get a bit of confidence in the direction here, I created an httpx backend locally to take this for a spin. It passes almost all functional tests except a few things:
* httpx does not coerce query params to strings, which requests seems to do, so for example Enums have to explicitly provide the value
* `response_content` is backend-specific, and shouldn't live in utils
* streamed responses dont work out of the box, since it's handled differently
* multipart is also handled differently, so it doesn't work
* I completely skirted the typing issue, that will probably be a headache.
* we cannot run unit tests on both backends unless we do some wrapping of `responses`/`respx` libraries based on the backend.

Simple requests already work though. /cc @JohnVillalovos @lmilbaum

nejch:
> > As an alternative we can squash-merge and it will link to the PR which has much more context than we would usually put in commit messages. WDYT @JohnVillalovos
> 
> That is okay if it is in the commit message. I want the details in the Git log. Not only on GitHub. What if the project gets moved to GitLab in the future. That information can disappear if it isn't in the git repository.
> 
> Seems relatively easy to take the description above and put it in the commit message. But maybe I am missing something?

I added the description in the body and reworded a bit during merge, https://github.com/python-gitlab/python-gitlab/commit/4afeaff0361a966254a7fbf0120e93583d460361.

PR #2445, opened 2023-01-07
chore: relative links in markup files

lmilbaum:
Following [github guidelines](https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/setting-guidelines-for-repository-contributors)


nejch:
Thanks for this PR @lmilbaum! I agree relative links would be best practice, but in this case I'm not sure it's currently the best idea as this link can show up in a multitude of places where this might break the URL:
* https://pypi.org/project/python-gitlab/#contributing (readme is collected with the package itself)
* https://python-gitlab.readthedocs.io/en/stable/#contributing (we include our readme from our sphinx docs)
* our mirror at https://gitlab.com/python-gitlab/python-gitlab (this would still work, but with `.github/` we lose the contributing button I think, due to the regex in https://gitlab.com/gitlab-org/gitlab/-/blob/master/lib/gitlab/file_detector.rb#L14)
* third-party package analyzers like https://snyk.io/advisor/python/python-gitlab#package-footer

WDYT, is there some other reason to consider this that I missed?

lmilbaum:
> Thanks for this PR @lmilbaum! I agree relative links would be best practice, but in this case I'm not sure it's currently the best idea as this link can show up in a multitude of places where this might break the URL:
> 
> * https://pypi.org/project/python-gitlab/#contributing (readme is collected with the package itself)
> * https://python-gitlab.readthedocs.io/en/stable/#contributing (we include our readme from our sphinx docs)
> * our mirror at https://gitlab.com/python-gitlab/python-gitlab (this would still work, but with `.github/` we lose the contributing button I think, due to the regex in https://gitlab.com/gitlab-org/gitlab/-/blob/master/lib/gitlab/file_detector.rb#L14)
> * third-party package analyzers like https://snyk.io/advisor/python/python-gitlab#package-footer
> 
> WDYT, is there some other reason to consider this that I missed?

ohh... wow... I wasn't aware of those links. I will close the PR. Thanks.

PR #2456, opened 2023-01-21
chore: remove tox `envdir` values

JohnVillalovos:
tox > 4 no longer will re-use the tox directory :(  What this means is that with the previous config if you ran:
    $ tox -e mypy; tox -e isort; tox -e mypy
It would recreate the tox environment each time :(

By removing the `envdir` values it will have the tox environments in separate directories and not recreate them.

The have an FAQ entry about this:
https://tox.wiki/en/latest/upgrading.html#re-use-of-environments

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -28,42 +28,36 @@ commands =
 
 [testenv:black]
 basepython = python3
-envdir={toxworkdir}/lint
 deps = -r{toxinidir}/requirements-lint.txt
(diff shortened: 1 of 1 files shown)


nejch (approved):
Thanks a lot @JohnVillalovos! Noticed this as well since I usually run the whole `tox` suite but was too lazy to fix :sweat_smile:

PR #2457, opened 2023-01-23
fix(client): regression - do not automatically get_next if page=# and iterator=True/as_list=False are used

helen-fornazier:
This fix a regression introduced on commit
https://github.com/python-gitlab/python-gitlab/commit/1339d645ce58a2e1198b898b9549ba5917b1ff12
(cc @JohnVillalovos)

If page is used, then get_next should be false.

This was found on the mesa ci project, after upgrading the python-gitlab version, the script that monitors the ci was getting killed by consuming too much memory.

### DIFF
--- a/gitlab/client.py
+++ b/gitlab/client.py
@@ -929,7 +929,7 @@ class Gitlab:
 
         page = kwargs.get("page")
 
-        if iterator:
+        if iterator and page is None:
             # Generator requested
             return GitlabList(self, url, query_data, **kwargs)
 
--- a/docs/api-usage.rst
+++ b/docs/api-usage.rst
@@ -393,6 +393,9 @@ The generator exposes extra listing information as received from the server:
    Prior to python-gitlab 3.6.0 the argument ``as_list`` was used instead of
    ``iterator``.  ``as_list=False`` is the equivalent of ``iterator=True``.
 
+.. note::
+   If ``page`` and ``iterator=True`` are used together, the latter is ignored.
+
 Sudo
 ====
 
--- a/tests/unit/test_gitlab_http_methods.py
+++ b/tests/unit/test_gitlab_http_methods.py
@@ -542,6 +542,18 @@ def test_list_request(gl):
     assert responses.assert_call_count(url, 3) is True
 
 
+@responses.activate
+def test_list_request_page_and_iterator(gl):
+    response_dict = copy.deepcopy(large_list_response)
+    response_dict["match"] = [responses.matchers.query_param_matcher({"page": "1"})]
+    responses.add(**response_dict)
+
+    result = gl.http_list("/projects", iterator=True, page=1)
+    assert isinstance(result, list)
+    assert len(result) == 20
+    assert len(responses.calls) == 1
+
+
 large_list_response = {
     "method": responses.GET,
     "url": "http://localhost/api/v4/projects",


nejch:
Thanks a lot for the fix @helen-fornazier! I guess we didn't anticipate that edge case with supplying both

It'd be great if you could add a little test to reproduce this just so we avoid future regressions. Maybe something similar to this (but more targeted to your specific use case - this test case below is a bit too broad maybe)?

https://github.com/python-gitlab/python-gitlab/blob/9322db663ecdaecf399e3192810d973c6a9a4020/tests/unit/test_gitlab_http_methods.py#L517-L542

helen-fornazier:
> Thanks a lot for the fix @helen-fornazier! I guess we didn't anticipate that edge case with supplying both
> 
> It'd be great if you could add a little test to reproduce this just so we avoid future regressions. Maybe something similar to this (but more targeted to your specific use case - this test case below is a bit too broad maybe)?
> 
> https://github.com/python-gitlab/python-gitlab/blob/9322db663ecdaecf399e3192810d973c6a9a4020/tests/unit/test_gitlab_http_methods.py#L517-L542

Sure! Since I'm not very familiar with this code, I need to study it a bit, but if you could give me some pointers on how to do it, it would be great.

nejch:
> Sure! Since I'm not very familiar with this code, I need to study it a bit, but if you could give me some pointers on how to do it, it would be great.

I'll post a few links in a bit. Just a question, what's the use case for using both `iterator=True` and `page=xyz` together? Just curious, if you have a reference from your project!

helen-fornazier:
> I'll post a few links in a bit. Just a question, what's the use case for using both `iterator=True` and `page=xyz` together? Just curious, if you have a reference from your project!

Thanks!

This script https://gitlab.freedesktop.org/gfx-ci/mesa-performance-tracking/-/blob/master/mesa-ci-status/mesa_ci_status.py
generates reports like this: https://gitlab.freedesktop.org/mesa/mesa/-/issues/7984#note_1726881

We parallelized the handling of pages and other requests with `ThreadPoolExecutor`. The threads fetch pipelines, jobs, their logs as well so we can classify errors.

When we started this, I don't recall there was an option that would fetch the next page automatically from the gitlab api with the generator (maybe it was already there and I didn't see it), so we went to the path to deal with the pages and parallelize things (since it was taking an annoying time to execute).

We thought that with `as_list=False` it would be faster, but I just noticed on this script that it uses list to filter jobs (yeah, a few things can be improved there for sure heh).

gitlab/client.py
```diff
@@ -931,7 +931,7 @@ def http_list(
 
         if iterator:
```
JohnVillalovos, line 932: To fix the regression and replicate as closely as things were before the change. I think this should be the change. But maybe I'm wrong.

```suggestion
        if iterator and page is None:
```

Agree with @nejch on adding a unit-test for this.

gitlab/client.py
```diff
@@ -931,7 +931,7 @@ def http_list(
...
             # Generator requested
-            return GitlabList(self, url, query_data, **kwargs)
+            return GitlabList(self, url, query_data, get_next=(not page), **kwargs)
```
JohnVillalovos, line 934: This can go away if suggestion for line 932 is correct.

JohnVillalovos (changes requested):

JohnVillalovos:
As a note. Previously when `as_list=False` and `page=X` where passed, it would ignore `as_list=False` and return a list. As only fetching one page and that data was returned as a list.

So realistically the fix on the Mesa CI side is to not use `as_list=False`. But I'm for this fix here as we should honor the `page=X` parameter.

gitlab/client.py
helen-fornazier: I see. Indeed, the previous code would return a list if `page` is used, even if `is_false` is `True`, but should it return a generator instead?

helen-fornazier:
> As a note. Previously when `as_list=False` and `page=X` where passed, it would ignore `as_list=False` and return a list. As only fetching one page and that data was returned as a list.
> 
> So realistically the fix on the Mesa CI side is to not use `as_list=False`. But I'm for this fix here as we should honor the `page=X` parameter.

I see. Please check comment on https://github.com/python-gitlab/python-gitlab/pull/2457#discussion_r1086637918 and let me know what you think

nejch:
Btw @helen-fornazier completely off-topic, I see you use marge-bot, you might be interested in the discussions over at https://github.com/smarkets/marge-bot/issues/295 :sweat_smile:

gitlab/client.py
helen-fornazier: I see. Let's restore to the old behavior than. I'll update the MR and add a note on the docs.

JohnVillalovos: @nejch I agree. Thanks.

helen-fornazier:
> Btw @helen-fornazier completely off-topic, I see you use marge-bot, you might be interested in the discussions over at [smarkets/marge-bot#295](https://github.com/smarkets/marge-bot/issues/295) sweat_smile

thanks for letting me know, I'll take a look :)

JohnVillalovos (approved):
Thanks. Looks good to me.

PR #2462, opened 2023-01-25
chore: add a UserWarning if both `iterator=True` and `page=X` are used

JohnVillalovos:
If a caller calls a `list()` method with both `iterator=True` (or `as_list=False`) and `page=X` then emit a `UserWarning` as the options are mutually exclusive.

### DIFF
--- a/gitlab/client.py
+++ b/gitlab/client.py
@@ -929,6 +929,19 @@ class Gitlab:
 
         page = kwargs.get("page")
 
+        if iterator and page is not None:
+            arg_used_message = f"iterator={iterator}"
+            if as_list is not None:
+                arg_used_message = f"as_list={as_list}"
+            utils.warn(
+                message=(
+                    f"`{arg_used_message}` and `page={page}` were both specified. "
+                    f"`{arg_used_message}` will be ignored and a `list` will be "
+                    f"returned."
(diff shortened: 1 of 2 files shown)


gitlab/client.py
```diff
@@ -929,6 +929,18 @@ def http_list(
...
         page = kwargs.get("page")
 
+        if iterator and page is not None:
+            arg_used_message = f"iterator={iterator}"
+            if as_list is not None:
+                arg_used_message = f"as_list={as_list}"
+            utils.warn(
+                message=(
+                    f"`{arg_used_message}` and `page={page}` were both specified. "
+                    f"`{arg_used_message}` will be ignored and a `list` will returned."
```
nejch, line 939:
```suggestion
                    f"`{arg_used_message}` will be ignored and a `list` will be returned."
```

nejch (commented):
Thanks @JohnVillalovos just a little typo there!

Hopefully we can remove `as_list` for a major bump at some point and simplify these checks :grin:

gitlab/client.py
JohnVillalovos: Thanks for catching that! Updated.

nejch (approved):

Issue #2466, opened 2023-02-02
RFE: support for `allow_force_push` in  ProjectProtectedBranchManager


nickbroon:
## Description of the problem, including code/CLI snippet

Allow setting the `allow_force_push` boolean attribute when creating protected branches.

Attribute is documented here:
https://docs.gitlab.com/ee/api/protected_branches.html#protect-repository-branches

And is missing from here:
https://github.com/python-gitlab/python-gitlab/blob/main/gitlab/v4/objects/branches.py#L35

## Specifications

  - python-gitlab version: 3.13.0
  - API version you are using (v3/v4): v4
  - Gitlab server version (or gitlab.com): gitlab.com

JohnVillalovos:
As a note, this is only an issue for the CLI. Users of the API can pass this parameter in the call.

PR #2468, opened 2023-02-03
feat(cli): add setting of `allow_force_push` for protected branch

JohnVillalovos:
For the CLI: add `allow_force_push` as an optional argument for creating a protected branch.

API reference:
https://docs.gitlab.com/ee/api/protected_branches.html#protect-repository-branches

Closes: #2466


nejch (approved):
Thanks for the quick response! :)

PR #2470, opened 2023-02-03
chore: remove `pre-commit` as a default `tox` environment

JohnVillalovos:
For users who use `tox` having `pre-commit` as part of the default
environment list is redundant as it will run the same tests again that
are being run in other environments. For example: black, flake8,
pylint, and more.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -2,7 +2,7 @@
 minversion = 1.6
 skipsdist = True
 skip_missing_interpreters = True
-envlist = py310,py39,py38,py37,flake8,black,twine-check,mypy,isort,cz,pylint,pre-commit
+envlist = py310,py39,py38,py37,flake8,black,twine-check,mypy,isort,cz,pylint
 
 [testenv]
 passenv =


lmilbaum:
Wouldn't it make more sense to remove the other environments and to keep pre-commit? 
pre-commit performs validations which are not covered by the granular ones.

JohnVillalovos:
> Wouldn't it make more sense to remove the other environments and to keep pre-commit? 
> pre-commit performs validations which are not covered by the granular ones.

Not for me. I don't want to use pre-commit.

nejch:
> > Wouldn't it make more sense to remove the other environments and to keep pre-commit?
> > pre-commit performs validations which are not covered by the granular ones.
> 
> Not for me. I don't want to use pre-commit.

John has some very strong opinions on pre-commit :grinning: 

I also usually run a full `tox` locally before pushing a new PR, and also prefer the native python dependency management of that over pre-commit, at least for tests etc. I also get the full duplicate suite run now which takes a while (it used to be ~15-20 seconds IIRC).

We discussed this in https://github.com/python-gitlab/python-gitlab/discussions/2321. I think we agreed we'd try to deduplicate this (see https://github.com/python-gitlab/python-gitlab/discussions/2321#discussioncomment-3905645), so I'd say the best way would be to have a tox environment that runs the missing checks only and that can be added to the default. But I'm ok to do that as a follow-up and merge this

lmilbaum:
> > > Wouldn't it make more sense to remove the other environments and to keep pre-commit?
> > > pre-commit performs validations which are not covered by the granular ones.
> > 
> > 
> > Not for me. I don't want to use pre-commit.
> 
> John has some very strong opinions on pre-commit 😀
> 
> I also usually run a full `tox` locally before pushing a new PR, and also prefer the native python dependency management of that over pre-commit, at least for tests etc. I also get the full duplicate suite run now which takes a while (it used to be ~15-20 seconds IIRC).
> 
> We discussed this in #2321. I think we agreed we'd try to deduplicate this (see [#2321 (reply in thread)](https://github.com/python-gitlab/python-gitlab/discussions/2321#discussioncomment-3905645)), so I'd say the best way would be to have a tox environment that runs the missing checks only and that can be added to the default. But I'm ok to do that as a follow-up and merge this

I got that @JohnVillalovos dislikes pre-commit. :-) I can't bit that with my engineering arguments.

nejch:
> I got that @JohnVillalovos dislikes pre-commit. :-) I can't bit that with my engineering arguments.

There's another issue with this I just realized. pre-commit keeps re-initializing on almost every run because there's always an outdated hook, which can take a while, even if the tox dependencies are otherwise up-to-date. Takes away from quick "shift left" hooks IMO, maybe let's remove it for now and find another way later.

nejch (approved):

Issue #2469, opened 2023-02-03
RFE: Add support for `PATCH` requests. Create a `http_patch()` method


JohnVillalovos:
There are now at least two different API calls which are using the HTTP `PATCH` method:

  * https://docs.gitlab.com/ee/api/users.html#set-user-status
  * https://docs.gitlab.com/ee/api/protected_branches.html#update-a-protected-branch

Need to create a `http_patch()` method in order to support these endpoints.

Issue #2390, opened 2022-11-23
Allow update of protected branches


hoerup:
In gitlab 15.6 gitlab finally added api support to update protected branch settings, so ProjectProtectedBranch should be updated accordingly

https://gitlab.com/gitlab-org/gitlab/-/issues/20229/

nejch:
Sadly a premium feature and we'll need to add a `http_patch` method for this as well.

JohnVillalovos:
Link to the API docs:

https://docs.gitlab.com/ee/api/protected_branches.html#update-a-protected-branch

Hokwang:
I think free version will have this endpoint. We know gitlab developers, right?
How about provide `.save()` or `.update()` anyway?

PR #2471, opened 2023-02-03
feat(client): add `http_patch` method

JohnVillalovos:
In order to support some new API calls we need to support the HTTP `PATCH` method.

Closes: #2469

### DIFF
--- a/docs/api-levels.rst
+++ b/docs/api-levels.rst
@@ -60,6 +60,7 @@ The available methods are:
 * ``http_get()``
 * ``http_post()``
 * ``http_put()``
+* ``http_patch()``
 * ``http_delete()``
 * ``http_list()`` (a wrapper around ``http_get`` handling pagination, including with lazy generators)
 * ``http_head()`` (only returns the header dictionary)
--- a/gitlab/client.py
+++ b/gitlab/client.py
@@ -1094,6 +1094,54 @@ class Gitlab:
                 error_message="Failed to parse the server message"
             ) from e
 
+    def http_patch(
+        self,
+        path: str,
+        *,
+        query_data: Optional[Dict[str, Any]] = None,
+        post_data: Optional[Union[Dict[str, Any], bytes]] = None,
+        raw: bool = False,
+        **kwargs: Any,
+    ) -> Union[Dict[str, Any], requests.Response]:
+        """Make a PATCH request to the Gitlab server.
+
+        Args:
+            path: Path or full URL to query ('/projects' or
+                        'http://whatever/v4/api/projecs')
+            query_data: Data to send as query parameters
+            post_data: Data to send in the body (will be converted to
+                              json by default)
+            raw: If True, do not convert post_data to json
+            **kwargs: Extra options to send to the server (e.g. sudo)
+
+        Returns:
+            The parsed json returned by the server.
+
+        Raises:
+            GitlabHttpError: When the return code is not 2xx
+            GitlabParsingError: If the json data could not be parsed
+        """
+        query_data = query_data or {}
+        post_data = post_data or {}
+
+        result = self.http_request(
+            "patch",
+            path,
+            query_data=query_data,
+            post_data=post_data,
+            raw=raw,
+            **kwargs,
+        )
+        try:
+            json_result = result.json()
+            if TYPE_CHECKING:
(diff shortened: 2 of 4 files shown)


JohnVillalovos:
@nejch Just wanted to get the ball rolling on this. Not sure how much should be in the PR.

Some unit tests are going to be wanted. Not sure about functional tests.

lmilbaum:
How about placing the function in the backend?

JohnVillalovos:
> How about placing the function in the backend?

Since I don't know much about the backend, can you explain why? I'm doing it like all the other `http_*` methods which currently exist.

lmilbaum:
> > How about placing the function in the backend?
> 
> Since I don't know much about the backend, can you explain why? I'm doing it like all the other `http_*` methods which currently exist.

Moving the http_* functions to the backend has already started. It will save a future work of moving this one as well and there will be no need to change the function signature.

JohnVillalovos:
> Moving the http_* functions to the backend has already started. It will save a future work of moving this one as well and there will be no need to change the function signature.

Sorry. I haven't looked into the back-end code. Can you explain how that will work? I don't really see much of anything in the back-end. I don't see any of the `http_*` methods there.

If all the other `http_*` methods need to be moved then I'm not sure how this one extra function is going to add much to the work required.

lmilbaum:
> > Moving the http_* functions to the backend has already started. It will save a future work of moving this one as well and there will be no need to change the function signature.
> 
> Sorry. I haven't looked into the back-end code. Can you explain how that will work? I don't really see much of anything in the back-end. I don't see any of the `http_*` methods there. EDIT: I see `http_request`
> 
> If all the other `http_*` methods need to be moved then I'm not sure how this one extra function is going to add much to the work required.

You might want to take a look at #2435

nejch:
@JohnVillalovos, I think this could be a `feat` as people can already use it for unimplemented endpoints after merge. 

> > > Moving the http_* functions to the backend has already started. It will save a future work of moving this one as well and there will be no need to change the function signature.
> > 
> > 
> > Sorry. I haven't looked into the back-end code. Can you explain how that will work? I don't really see much of anything in the back-end. I don't see any of the `http_*` methods there. EDIT: I see `http_request`
> > If all the other `http_*` methods need to be moved then I'm not sure how this one extra function is going to add much to the work required.
> 
> You might want to take a look at #2435

@lmilbaum in any case we will have these user-facing helper methods on the client, preferably backend-agnostic (https://python-gitlab.readthedocs.io/en/stable/api-levels.html#lower-level-api-http-methods).

I hope if we get this right only `http_request` and the associated typing will need a significant backend-specific implementation and the rest can wrap around it like we do now.

nejch (changes requested):

lmilbaum:
> @JohnVillalovos, I think this could be a `feat` as people can already use it for unimplemented endpoints after merge.
> 
> > > > Moving the http_* functions to the backend has already started. It will save a future work of moving this one as well and there will be no need to change the function signature.
> > > 
> > > 
> > > Sorry. I haven't looked into the back-end code. Can you explain how that will work? I don't really see much of anything in the back-end. I don't see any of the `http_*` methods there. EDIT: I see `http_request`
> > > If all the other `http_*` methods need to be moved then I'm not sure how this one extra function is going to add much to the work required.
> > 
> > 
> > You might want to take a look at #2435
> 
> @lmilbaum in any case we will have these user-facing helper methods on the client, preferably backend-agnostic (https://python-gitlab.readthedocs.io/en/stable/api-levels.html#lower-level-api-http-methods). They help users work around endpoints we haven't implemented in the library.
> 
> I hope if we get this right only `http_request` and the associated typing will need a significant backend-specific implementation and the rest can wrap around it like we do now.

I am not following what you are trying to say. In any case, maybe switching the return type in a similar way I've done in the PR #2435.

nejch:
> I am not following what you are trying to say. In any case, maybe switching the return type in a similar way I've done in the PR #2435.

@lmilbaum since this PR is a separate concern, I think we can do that after, I can also rebase your PR with `http_patch` to avoid more rebase issues on your end. I think we can go ahead with this one just to unblock users who want to update protected branches in https://github.com/python-gitlab/python-gitlab/issues/2390.

nejch (approved):

PR #2472, opened 2023-02-04
WIP: DO NOT MERGE

JohnVillalovos:
Playing around with https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html


setup.py
```diff
@@ -1,57 +1,3 @@
...
-        "Programming Language :: Python :: 3.9",
-        "Programming Language :: Python :: 3.10",
-        "Programming Language :: Python :: 3.11",
-    ],
-    extras_require={
-        "autocompletion": ["argcomplete>=1.10.0,<3"],
-        "yaml": ["PyYaml>=5.2"],
-    },
-)
+setup()
```
lmilbaum, line 3: From my experience, setup.py is not needed a all.  I like the approach of moving to pyproject.toml file.

nejch: It's mostly for backwards compatibility, maybe not needed anymore now that it's been a while. But we already have a working config in https://github.com/python-gitlab/python-gitlab/commit/8d45db78b4e848b0ed41cc1ce338457b0ebc81bc for this.

Issue #2448, opened 2023-01-13
project-merge-request commits return no data


ramarro123:
hello,

executing
gitlab --private-token XXX -o json project-merge-request commits --project-id 63 --iid 2875
return no data (not even a json, just empty)

the same call via curl
curl --location --request GET 'YYY/api/v4/projects/63/merge_requests/2875/commits' \
--header 'PRIVATE-TOKEN: XXX'

return the full json of data.

am i missing anything in that?

nejch:
Thanks again for the report @ramarro123, I can reproduce this.

Again, it's because we have custom code that returns a `RESTObjectList` rather than a list. So we get data from the server but never print it out properly:

https://github.com/python-gitlab/python-gitlab/blob/3f86d36218d80b293b346b37f8be5efa6455d10c/gitlab/v4/objects/merge_requests.py#L240-L244

We could potentially handle this in the CLI, but also move away from custom methods.
Note to self, this quick hack fixed it:

```diff
diff --git a/gitlab/v4/cli.py b/gitlab/v4/cli.py
index 41ffa1e9..91f2e0db 100644
--- a/gitlab/v4/cli.py
+++ b/gitlab/v4/cli.py
@@ -547,6 +547,8 @@ def run(
         printer.display(data, verbose=True, obj=data)
     elif isinstance(data, list):
         printer.display_list(data, fields, verbose=verbose)
+    elif isinstance(data, gitlab.base.RESTObjectList):
+        printer.display_list([item for item in data], fields, verbose=verbose)
     elif isinstance(data, gitlab.base.RESTObject):
         printer.display(get_dict(data, fields), verbose=verbose, obj=data)
     elif isinstance(data, str):
```

PR #2475, opened 2023-02-05
fix(cli): display items when iterator is returned

nejch:
Closes https://github.com/python-gitlab/python-gitlab/issues/2448.

Separately, fce790328df81ac61679f40c0142d70e58049703 renames the fixture factory as is done in https://docs.pytest.org/en/7.1.x/how-to/fixtures.html#factories-as-fixtures and adds a fixture that can be reused directly in tests.


JohnVillalovos (approved):
Thanks! Looks good to me :)

PR #2488, opened 2023-02-12
refactor: move response_content into backend code

nejch:
Needed in order to later add other backends that handle streaming responses differently. Just a first step, later maybe we could get rid of `response_content` completely from individual methods and just handle this in the request code :thinking:

### DIFF
--- a/gitlab/_backends/protocol.py
+++ b/gitlab/_backends/protocol.py
@@ -1,6 +1,6 @@
 import abc
 import sys
-from typing import Any, Dict, Optional, Union
+from typing import Any, Callable, Dict, Iterator, Optional, Union
 
 import requests
 from requests_toolbelt.multipart.encoder import MultipartEncoder  # type: ignore
@@ -17,6 +17,17 @@ class BackendResponse(Protocol):
 
 
 class Backend(Protocol):
+    @staticmethod
+    @abc.abstractmethod
+    def response_content(
+        response: requests.Response,
+        streamed: bool,
+        action: Optional[Callable[[bytes], None]],
+        chunk_size: int,
+        *,
+        iterator: bool,
+    ) -> Optional[Union[bytes, Iterator[Any]]]: ...
+
     @abc.abstractmethod
     def http_request(
         self,
--- a/gitlab/_backends/requests_backend.py
+++ b/gitlab/_backends/requests_backend.py
@@ -1,7 +1,16 @@
 from __future__ import annotations
 
 import dataclasses
-from typing import Any, BinaryIO, Dict, Optional, TYPE_CHECKING, Union
+from typing import (
+    Any,
+    BinaryIO,
+    Callable,
+    Dict,
+    Iterator,
+    Optional,
+    TYPE_CHECKING,
+    Union,
+)
 
 import requests
 from requests import PreparedRequest
@@ -53,6 +62,11 @@ class SendData:
                 f"`json` and `data` are mutually exclusive. Only one can be set. "
                 f"json={self.json!r}  data={self.data!r}"
             )
+
+
+class _StdoutStream:
+    def __call__(self, chunk: Any) -> None:
+        print(chunk)
 
 
 class RequestsResponse(protocol.BackendResponse):
@@ -125,6 +139,29 @@ class RequestsBackend(protocol.Backend):
             assert not isinstance(post_data, BinaryIO)
 
         return SendData(json=post_data, content_type="application/json")
+
+    @staticmethod
+    def response_content(
+        response: requests.Response,
+        streamed: bool,
+        action: Optional[Callable[[bytes], None]],
+        chunk_size: int,
+        *,
+        iterator: bool,
+    ) -> Optional[Union[bytes, Iterator[Any]]]:
+        if iterator:
+            return response.iter_content(chunk_size=chunk_size)
+
+        if streamed is False:
+            return response.content
+
+        if action is None:
+            action = _StdoutStream()
+
+        for chunk in response.iter_content(chunk_size=chunk_size):
+            if chunk:
+                action(chunk)
+        return None
 
     def http_request(
         self,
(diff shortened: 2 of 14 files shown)


gitlab/_backends/requests_backend.py
```diff
@@ -9,6 +9,11 @@
 from . import protocol
 
 
+class _StdoutStream:
+    def __call__(self, chunk: Any) -> None:
+        print(chunk)
```
nejch, line 14: TBH I don't know why we have this. Just doing `action = print` and then `action(chunk)` would do in the function itself. Maybe the author had bigger plans for this.

lmilbaum: It worth adding it as a comment in the code so that contributors could peek it up and change it

gitlab/_backends/requests_backend.py
```diff
@@ -117,3 +122,26 @@ def http_request(
             **kwargs,
         )
         return RequestsResponse(response=response)
+
+    def response_content(
```
lmilbaum, line 126: The prefix "response_" is redundant

lmilbaum (changes requested):

gitlab/_backends/requests_backend.py
nejch: Keep in mind this method is in `RequestsBackend`, not in `RequestsResponse` as the GitHub diff suggests. Did you still mean this prefix is redundant in the backend class?

gitlab/_backends/requests_backend.py
nejch: Thanks, I'll open a follow-up issue to track this. Maybe that's a bit cleaner than having comments in the code.

nejch: Opened https://github.com/python-gitlab/python-gitlab/issues/2496

gitlab/_backends/requests_backend.py
lmilbaum: Thanks for the clarification. I totally missed it.  In that case, what is the connection between the `response_content` function and the class `RequestsBackend`?

nejch: The implementation is backend-specific, so it can't live in utils anymore if we want to add other backends. For example requests has `iter_content`, whereas httpx has `iter_bytes` and `iter_text`.

Putting it into the backend class was an easy minimal copy/paste, eventually it would probably make sense to remove this method and handle it in the response itself I assume as I say above:

> Needed in order to later add other backends that handle streaming responses differently. Just a first step, later maybe we could get rid of response_content completely from individual methods and just handle this in the request code :thinking:  

Hope that makes sense:) I can push a branch with an httpx backend to show the different implemenations if needed.

lmilbaum: Thanks again for the clarification. I am totally fine with going forward with this approach as long as we agree that it might be refactored in the future.
Only thing I would add the staticmethod decorator which will eliminate the need of instantiating the class.

nejch: I agree with that, my thinking is as I said,`eventually it would probably make sense to remove this method and handle it in the response itself`, but maybe there are other ways :)

Ah good idea, will add that, completely missed that while copy pasting.

nejch: Done, thanks @lmilbaum!

nejch:
@JohnVillalovos I've rebased and reworked this a bit based on previous comments. Would you be able to have a look at this one?

IMO it's also a bit cleaner like this to have the response logic be part of the client/backend than a separate util function since we've split out the client code (and more recently the backend code) from the monolith `__init__` module.

JohnVillalovos (approved):
Thanks @nejch 

Overall looks good to me. Only a couple of nits on my end. But I would be okay with merging as is. So approved by me.

gitlab/_backends/protocol.py
```diff
@@ -17,6 +17,17 @@ def __init__(self, response: requests.Response) -> None: ...
 
 
 class Backend(Protocol):
+    @staticmethod
+    @abc.abstractmethod
+    def response_content(
```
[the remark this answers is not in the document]
nejch: I believe this would be a breaking change as we still expose it in utils as a non-private function. In that case I'd maybe drop the wrapper entirely and not just deprecate it, as then we don't need to worry about maintaining the signature twice (https://github.com/python-gitlab/python-gitlab/pull/2488#discussion_r1633344488).

Might be worth grouping breaking changes and wait to merge them all together though for a major release.

gitlab/utils.py
```diff
@@ -50,26 +43,14 @@ def format(self, record: logging.LogRecord) -> str:
 
 
 def response_content(
-    response: requests.Response,
-    streamed: bool,
-    action: Optional[Callable[[bytes], None]],
-    chunk_size: int,
-    *,
-    iterator: bool,
+    *args: Any, **kwargs: Any
```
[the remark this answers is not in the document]
nejch: My initial idea was to just leave a simple wrapper for backward compatibility but maybe we can just drop it (see https://github.com/python-gitlab/python-gitlab/pull/2488#discussion_r1634514002).

gitlab/_backends/protocol.py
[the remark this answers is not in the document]
JohnVillalovos: Well I didn't think it would be a breaking change if this is the new function and then the wrapper (with the deprecation) would call this one with keyword args.

gitlab/utils.py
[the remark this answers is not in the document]
JohnVillalovos: My thought was that this one would maintain the same exact signature as before, and then call the new function with keyword arguments.

But! It isn't that important and perfectly fine with me if you want to merge this in as is 👍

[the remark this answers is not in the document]
nejch: Ahh I misunderstood your comment I think, I thought you wanted to match the signatures. I'll add the changes!

PR #2490, opened 2023-02-12
chore(github): add default pull request template

nejch:


CONTRIBUTING.rst
```diff
@@ -12,50 +12,44 @@ You can contribute to the project in multiple ways:
 Development workflow
 --------------------
 
-Before contributing, please make sure you have `pre-commit <https://pre-commit.com>`_
-installed and configured. This will help automate adhering to code style and commit
-message guidelines described below:
+Before contributing, install `tox <https://tox.wiki/>`_ and `pre-commit <https://pre-commit.com>`_:
```
JohnVillalovos, line 15: I guess I would like it if `pre-commit` was listed as optional.

nejch: Ok, I added a note below for people who don't want to use it. This is targeted more at new contributors. Just wanted to keep it prominent here, as first-time contributors using it properly would probably save us 90% of the review work we have to do for small PRs.

nejch: Does that work for you @JohnVillalovos? Or did you have something else in mind?

JohnVillalovos (approved):

PR #2498, opened 2023-02-18
chore: bump version of GitLab to 15.4.6-ee.0

JohnVillalovos:
Increase version of GitLab to latest major version that works without
changing tests.


nejch:
We're now on v16 (soon 17) so I'll close this one, I think we need some follow-up to get rid of the sleeps though.

Issue #2506, opened 2023-03-04
Group import fails when parent_id is passed as int


vlizanae:
## Description of the problem, including code/CLI snippet

Testing group export/import between different Gitlab instances. Used to the path of the group working as the group id, I tried the following:

```python
gl.groups.import_group(buffer, path='group_slug', name='Group name', parent_id='path/to/parent')
```
which yields
```
gitlab.exceptions.GitlabImportError: 400: parent_id is invalid
```

Not very descriptive but reading gitlab's api docs I find out:
> `parent_id`, integer: ID of a parent group to import the group into. Defaults to the current user’s namespace if not provided.

So I try the numerical id of the group:

```python
gl.groups.import_group(buffer, path='group_slug', name='Group name', parent_id=42)
```
even worse:
```
AttributeError: 'int' object has no attribute 'encode'
```

## Expected Behavior

Because the path works as an id in several calls (e.g. `groups.get`) I think the first version _should_ work. The second version should definitely work.

## Actual Behavior

In the end, after digging a bit on the code, what works is:
```python
gl.groups.import_group(buffer, path='group_slug', name='Group name', parent_id='42')
```

## Specifications

  - python-gitlab version: 3.13.0
  - API version you are using (v3/v4): v4
  - Gitlab server version (or gitlab.com): 15.8.1-ee

nejch:
Thanks for the report @vlizanae.

Regarding the confusing naming, this is really GitLab's behavior so i think we will want to keep it as is (we're just passing this to the API, so it's up to GitLab to accept IDs or url-encoded paths).

For your proposal to work, we would have to add another GET call inside the method, to find out the group's ID from the path, but as a library we really want to just be a thin wrapper around existing endpoints. If you want this behavior I'd write a little wrapper function to do so. Does that work for yo.u?

From experience, when `<something>_id` is passed as a parameter, it usually is only the ID, whereas as part of the URL itself it usually accepts both.  As we state in the docs (https://python-gitlab.readthedocs.io/en/stable/api-usage.html), we encourage users to check upstream docs for attribute specifics, it would be too much work for us to try and mirror the upstream docs here reliably.

But the integer encoding issue is definitely a bug, we'll have to fix that, thanks again for the catch!

vlizanae:
Hmm I didn't realize that passing the path as id was part of gitlab's API, I thought it was a feature of the library. I agree with just replicating the behavior of the API so it's ok having just the numerical id.

[...]

PR #2507, opened 2023-03-04
fix: support int for `parent_id` in `import_group`

JohnVillalovos:
This will also fix other use cases where an integer is passed in to
MultipartEncoder.

Added unit tests to show it works.

Closes: #2506

### DIFF
--- a/gitlab/_backends/requests_backend.py
+++ b/gitlab/_backends/requests_backend.py
@@ -1,12 +1,27 @@
 from __future__ import annotations
 
-from typing import Any, Dict, Optional, Tuple, TYPE_CHECKING, Union
+import dataclasses
+from typing import Any, Dict, Optional, TYPE_CHECKING, Union
 
 import requests
 from requests.structures import CaseInsensitiveDict
 from requests_toolbelt.multipart.encoder import MultipartEncoder  # type: ignore
 
 from . import protocol
+
+
+@dataclasses.dataclass
+class SendData:
+    content_type: str
+    data: Optional[Union[Dict[str, Any], MultipartEncoder]] = None
+    json: Optional[Union[Dict[str, Any], bytes]] = None
+
+    def __post_init__(self) -> None:
+        if self.json is not None and self.data is not None:
+            raise ValueError(
+                f"`json` and `data` are mutually exclusive. Only one can be set. "
+                f"json={self.json!r}  data={self.data!r}"
+            )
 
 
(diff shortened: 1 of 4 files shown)


gitlab/_backends/requests_backend.py
```diff
@@ -64,18 +75,18 @@ def prepare_send_data(
                 if TYPE_CHECKING:
                     assert isinstance(post_data, dict)
                 for k, v in post_data.items():
-                    if isinstance(v, bool):
+                    if isinstance(v, (bool, int)):
```
nejch, line 78: This is the real fix here, but maybe we could already future-proof it a bit, WDYT?

```suggestion
                    if isinstance(v, (bool, complex, float, int)):
```

So we don't get similar issues in the future. These should probably be enough according to https://en.wikipedia.org/wiki/Python_(programming_language)#/media/File:Python_3._The_standard_type_hierarchy-en.svg

nejch (commented):
Thanks @JohnVillalovos, just a quick initial question on the fix, the rest seems to me an unrelated refactor so I'll look at it a bit later as I'm a bit tired right now :sweat_smile:

gitlab/_backends/requests_backend.py
JohnVillalovos: But below we convert it to an int and then a str... And the only time we go through this path is when there is a file.

So I am unsure. What do you think?

JohnVillalovos: @nejch I did what you suggested. Well a bit more than you suggested 🙂

nejch: Ah yes, makes sense like that, missed it. TBH I checked https://github.com/python-gitlab/python-gitlab/pull/1404 and don't see why we should really be converting the boolean to int, but maybe we can keep it for now.

Yeah, seems like it can happen whenever someone adds a file-type attribute but then affects other attributes as we switch to the multipart encoder for everything.

nejch:
Sorry for the delay here, I just realized the testing conflicts a bit with those added in https://github.com/python-gitlab/python-gitlab/pull/2488/files#diff-fc2667e8a1b0a974c8447eee8fca931262283245469f31693874cad25651c8f1, I'll get back to this on the weekend :)

nejch (approved):
Thanks @JohnVillalovos, I had some thoughts on the testing structure but will do a follow-up as we have a big mix now from older tests anyway.

PR #2515, opened 2023-03-12
fix(client): properly parse content-type when charset is present

### LINKED ISSUE #2503: Client not parsing json response when "; charset=utf-8 " in `Content-Type` header

When performing a graphql query, with `gitlab.client.http_post` , the result is not parsed as json.

I believe this is because the content type we receive is 
```
      Content-Type:
      - application/json; charset=utf-8
```
while the client check for equality with `application/json` in https://github.com/python-gitlab/python-gitlab/blob/main/gitlab/client.py#L1036

## Expected Behavior

The result is parsed as json

## Actual Behavior

The raw response is returned.

## Specifications

  - python-gitlab version: `python-gitlab==3.13.0`
  - API version you are using (v3/v4): graphql api
  - Gitlab server version (or gitlab.com): 14.10

nejch:
Closes #2503

### DIFF
--- a/gitlab/client.py
+++ b/gitlab/client.py
@@ -828,12 +828,9 @@ class Gitlab:
         result = self.http_request(
             "get", path, query_data=query_data, streamed=streamed, **kwargs
         )
+        content_type = utils.get_content_type(result.headers.get("Content-Type"))
 
-        if (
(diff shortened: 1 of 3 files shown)


JohnVillalovos (approved):
Thanks @nejch

PR #2516, opened 2023-03-12
chore(.github): actually make PR template the default

nejch:
:facepalm: the folder I used was for multiple PR templates. To have the default it needs to be in `.github/` it turns out. Unlike our issue template config.


JohnVillalovos (approved):
Thanks @nejch

PR #2518, opened 2023-03-14
chore(pre-commit): Bumping versions

lmilbaum:
## Changes

Bumping versions for mypy pre-commit hook and fixing renovate config file to pick up
mypy repo versions

### Documentation and testing

### DIFF
--- a/.renovaterc.json
+++ b/.renovaterc.json
@@ -43,6 +43,15 @@
       ],
       "automerge": true,
       "groupName": "GitLab"
+    },
+    {
+      "matchPackageNames": [
+        "pre-commit/mirrors-mypy"
+      ],
+      "matchManagers": [
+        "pre-commit"
+      ],
+      "versioning": "pep440"
     }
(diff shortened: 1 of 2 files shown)


.renovaterc.json
```diff
@@ -43,6 +43,15 @@
...
       "groupName": "GitLab"
+    },
+    {
+      "matchPackageNames": [
+        "pre-commit/mirrors-mypy"
+      ],
+      "matchManagers": [
+        "pre-commit"
+      ],
+      "versioning": "pep440"
```
nejch, line 54: @lmilbaum it seems like since v1.0.0, mypy including the mirror, follow semver so it might not be necessary:

https://github.com/pre-commit/mirrors-mypy/tags

It likely wasn't picked up before due to the weird 0.xxx versioning.

But this won't hurt for sure.

nejch (approved):
Thanks for catching this @lmilbaum!

PR #2527, opened 2023-03-23
feat: add resource_weight_event for ProjectIssue

paulfariello:
## Changes

Add support for resource_weight_event in ProjectIssue see https://docs.gitlab.com/ee/api/resource_weight_events.html

### Documentation and testing

### DIFF
--- a/docs/gl_objects/issues.rst
+++ b/docs/gl_objects/issues.rst
@@ -207,6 +207,10 @@ Get the list of iteration events::
 
     iteration_events = issue.resource_iteration_events.list()
 
+Get the list of weight events::
+
+    weight_events = issue.resource_weight_events.list()
+
 Issue links
 ===========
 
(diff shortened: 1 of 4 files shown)


nejch:
Thanks @paulfariello, looks good! Would you mind adding just some basic functional tests for this, it should only be a few lines of code I believe, if you model it after similar tests like here:

https://github.com/python-gitlab/python-gitlab/blob/e973729e007f664aa4fde873654ef68c21be03c8/tests/functional/api/test_issues.py#L69-L74

I think should be enough, the inital weight and `issue.save()` is there just so we ensure we have something to work with:

```python
def test_issue_weight_events(issue):
    issue.weight = 1
    issue.save()

    events = issue.resource_weight_events.list() 
    assert isinstance(events, list) 
 
    event = issue.resource_weight_events.get(events[0].id) 
    assert isinstance(event, gitlab.v4.objects.ProjectIssueResourceWeightEvent) 
```

nejch (approved):
Thanks for the quick rework @paulfariello, LGTM!

PR #2529, opened 2023-03-23
feat(api): add field shared_with_groups to groups

leospinola:
## Changes

feat(api): add property shared_with_groups for groups

### Documentation and testing


nejch:
Thanks for the contribution @leospinola! I don't see `shared_with_groups` in the create group API parameters - https://docs.gitlab.com/ee/api/groups.html#new-group, so not sure this will have the desired effect.

Could you give us a bit more context on what you're trying to achieve? Thanks a lot!

leospinola:
You right, thanks! Was a crossed change on my repo.

PR #2538, opened 2023-04-07
feat(objects): support fetching PATs via id or `self` endpoint

### LINKED ISSUE #2537: Need support to [...]/personal_access_tokens/self Gitlab API
So far, I'm unable to get scopes information from the 'personal access token' used to authenticate to Gitlab API.
Ofc, it is easy to get the list of the personal access token of the authenticated user, but I see no way to get the link with the one really used to authenticate.

It seems the dedicated Gitlab API entrypoint for that is [...]/personal_access_tokens/self which is not actually supported by python-gitlab.

Context: the idea is to check with efficiency some required scopes/permissions/rights, instead of trying an operation (like download repository tarball) because it could be time consuming although a simple check could be immediate.

## Expected Behavior

Get scopes (and roles ?) information of the 'personal access token' currently used to authenticate to Gitlab API.

## Actual Behavior

Unable to get this information.

## Specifications

  - python-gitlab version: 3.13.0

nejch:
## Changes

Closes #2537

### Documentation and testing

### DIFF
--- a/tests/unit/objects/test_personal_access_tokens.py
+++ b/tests/unit/objects/test_personal_access_tokens.py
@@ -13,6 +13,7 @@ token_name = "Test Token"
 
 token_url = "http://localhost/api/v4/personal_access_tokens"
 single_token_url = f"{token_url}/{token_id}"
+self_token_url = f"{token_url}/self"
 user_token_url = f"http://localhost/api/v4/users/{user_id}/personal_access_tokens"
 
 content = {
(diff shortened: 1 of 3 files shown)


tests/unit/objects/test_personal_access_tokens.py
```diff
@@ -41,15 +42,47 @@ def resp_create_user_personal_access_token():
...
+def resp_get_personal_access_token_self():
+    with responses.RequestsMock() as rsps:
+        rsps.add(
+            method=responses.GET,
+            url=self_token_url,
+            json=content,
+            content_type="application/json",
+            status=200,
+        )
+        yield rsps
```
nejch, line 80: A bit repetitive here but I think that's ok for tests as we would parametrize only 2 variants.

JohnVillalovos (approved):

PR #2541, opened 2023-04-07
feat(projects): allow importing additional items from GitHub

nejch:


gitlab/v4/objects/projects.py
```diff
@@ -1152,13 +1159,9 @@ def import_github(
...
         """
-        data = {
-            "personal_access_token": personal_access_token,
-            "repo_id": repo_id,
-            "target_namespace": target_namespace,
-        }
-        if new_name:
-            data["new_name"] = new_name
+        attributes = copy.deepcopy(locals())
+        data = utils.remove_none_from_dict(attributes, ignore_keys=("self", "kwargs"))
```
nejch, line 1163: Not sure about this but I was getting a bit tired of the custom code we need when we add arguments. Seems cleaner this way while preserving the method signature with typing.

gitlab/utils.py
```diff
@@ -141,8 +141,10 @@ def __new__(cls, value: Union[str, int, "EncodedId"]) -> "EncodedId":
         return super().__new__(cls, value)
 
 
-def remove_none_from_dict(data: Dict[str, Any]) -> Dict[str, Any]:
-    return {k: v for k, v in data.items() if v is not None}
+def remove_none_from_dict(
+    data: Dict[str, Any], ignore_keys: Tuple[str, ...] = ()
```
JohnVillalovos, line 145: Just feels wrong to have that `ignore_keys = ()` but I guess it is okay since it is a tuple. I know it can be an issue with dicts and lists.

gitlab/v4/objects/projects.py
JohnVillalovos: Yeah. I'm not sure about this either. I think it makes the code harder to understand. It is cool 😎.  But I think the other way is easier to understand.

So basically I don't think we should do it this way.

JohnVillalovos (changes requested):

gitlab/v4/objects/projects.py
JohnVillalovos: What about something like this?

```suggestion
        data = {
            k: v
            for k, v in [
                ("personal_access_token", personal_access_token),
                ("repo_id", repo_id),
                ("target_namespace", target_namespace),
                ("new_name", new_name),
                ("github_hostname", github_hostname),
                ("optional_stages", optional_stages),
            ]
            if v is not None
        }
```

gitlab/utils.py
nejch: I think that's more an issue for mutable defaults yes.

gitlab/v4/objects/projects.py
nejch: This type of duplication is what I was trying to avoid.. Ok, I'll take another look at what options we might have for handling this in a generic way :) Since we also have other custom methods where the explicit arguments lag behind GitLab's implementation.

nejch (commented):
Thanks for taking a look @JohnVillalovos! I'll take another stab at this.

gitlab/v4/objects/projects.py
nejch: For now I reverted the fancy stuff but reused the helper function. :P

JohnVillalovos (approved):
Thanks @nejch . Looks good. I think it will be easy to update in the future if/when needed.

Issue #2522, opened 2023-03-21
gitlab project-push-rules get --project-id returns empty output


w7089:
## Description of the problem, including code/CLI snippet

`gitlab project-push-rules get --project-id [id]` returns empty output while there are push rules set in UI.

## Expected Behavior

`gitlab project-push-rules get --project-id [id]` should return existing push rules or specific ones similar to `gitlab project-push-rules update` behaviour

## Actual Behavior

empty output returned 

## Specifications

  - python-gitlab version: `3.13.0`
  - API version you are using (v3/v4): `v4`
  - Gitlab server version (or gitlab.com): `15.8.1-ee`

w7089:
yes, it worked, if that's the expected behaviour, i'll close it

w7089:
by the way, selecting fields doesn't work as well:

`gitlab -f id project-push-rules get --project-id [id]` shows empty output

nejch:
Sorry, the `--fields` flag doesn't work with the default formatter, you'll need to use json or yaml (`-o json`).

https://python-gitlab.readthedocs.io/en/stable/cli-usage.html#optional-arguments

PR #2542, opened 2023-04-07
fix(cli): warn user when no fields are displayed

nejch:
Closes https://github.com/python-gitlab/python-gitlab/issues/2522


tests/unit/test_cli.py
```diff
@@ -146,3 +149,23 @@ def test_v4_parser():
     )
     actions = user_subparsers.choices["create"]._option_string_actions
     assert actions["--name"].required
+
+
+@pytest.mark.skipif(sys.version_info < (3, 8), reason="added in 3.8")
+def test_legacy_display_without_fields_warns(fake_object_no_id):
+    printer = v4_cli.LegacyPrinter()
+
+    with mock.patch("builtins.print") as mocked:
```
JohnVillalovos, line 158: As a note `pytest` has `monkeypatch`. I'm fine with this but if you wanted to be all pytest all the time...

JohnVillalovos (approved):
@nejch Approved please merge this. I didn't merge just in case you wanted to use `monkeypatch`.

Thanks!

tests/unit/test_cli.py
nejch: Oh right, I forgot we already use that all over the place. Will change it!

PR #2544, opened 2023-04-10
chore(ci): wait for all coverage reports in CI status

nejch:
This has been annoying me - with the new integration, we get annotations and status checks, but they seem to get notified early, so CI looks like it's failing until all coverage reports are fully uploaded.

https://docs.codecov.com/docs/codecovyml-reference


JohnVillalovos (approved):
Thanks @nejch

PR #2551, opened 2023-04-19
feat: usernames support for MR approvals

nickbroon:
This can be used instead of 'user_ids'

See: https://docs.gitlab.com/ee/api/merge_request_approvals.html#create-project-level-rule


nejch (approved):
Thanks @nickbroon, looks good! 

Btw, I chuckled there for a second, I guess it's pretty recent :sweat_smile: :
> `usernames` property was [introduced](https://gitlab.com/gitlab-org/gitlab/-/merge_requests/) in GitLab 1x.x.

PR #2559, opened 2023-05-01
chore(ci): use OIDC trusted publishing for pypi.org

### LINKED ISSUE #2554: Replace static PyPI credentials with OIDC
https://blog.pypi.org/posts/2023-04-20-introducing-trusted-publishers/

nejch:
Closes #2554 

Tested on test.pypi.org, I've removed the static PYPI_TOKEN already. We can also remove the deploy token member now IMO

### DIFF
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -7,3 +7,5 @@ pytest-github-actions-annotate-failures==0.1.8
 pytest==7.2.2
 PyYaml==5.4.1
 responses==0.23.1
+setuptools==67.7.2
+wheel==0.40.0
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,13 +9,27 @@ jobs:
   release:
     if: github.repository == 'python-gitlab/python-gitlab'
     runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+    environment: pypi.org
     steps:
     - uses: actions/checkout@v3.5.0
       with:
         fetch-depth: 0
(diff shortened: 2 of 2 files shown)


requirements-test.txt
```diff
@@ -7,3 +7,5 @@ pytest-github-actions-annotate-failures==0.1.8
 pytest==7.2.2
 PyYaml==5.4.1
 responses==0.23.1
+setuptools==67.7.2
+wheel==0.40.0
```
nejch, line 11: Added separately in 5b549c58d2309f0b3a4056e0b0a7d4bc874405ba to keep CI happy (unrelated failures on python 3.12). I'll refactor the build to declarative so we can move away from `setup.py` and we hopefully don't need these anymore.

nejch:
@max-wittig could you take a look? Related settings are in
https://github.com/python-gitlab/python-gitlab/settings/environments, and
https://pypi.org/manage/project/python-gitlab/settings/publishing/.

The script is mostly from https://docs.pypi.org/trusted-publishers/using-a-publisher/#the-manual-way with a small fix, rather than using the whole extra action, so we can keep the current release flow.

nejch:
@max-wittig could you take a look? Related settings are in
https://github.com/python-gitlab/python-gitlab/settings/environments, and
https://pypi.org/manage/project/python-gitlab/settings/publishing/.

The script is mostly from https://docs.pypi.org/trusted-publishers/using-a-publisher/#the-manual-way with a small fix, rather than using the whole extra action, so we can keep the current release flow.

PR #2578, opened 2023-05-28
chore: update sphinx from 5.3.0 to 6.2.1

JohnVillalovos:
### Documentation and testing


nejch (approved):
I think we're good in terms of breaking changes, though I didn't check the built pages but we can always revert from `/latest`. :+1:

PR #2588, opened 2023-05-31
fix(cli): add _from_parent_attrs to user-project manager (#2558)

### LINKED ISSUE #2558: CLI: gitlab user-project list doesn't work

`gitlab user-project list --user-id  42` fails.

A similar issue has previously been reported in  #881 .

## Expected Behavior

List of projects owned by the current user are shown.

## Actual Behavior

`Impossible to list objects ('user_id')`  is printed.

## Specifications

  - python-gitlab version: 3.14.0
  - Gitlab server version (or gitlab.com): GitLab Community Edition [15.11.0](https://gitlab.com/gitlab-org/gitlab-foss/-/tags/v15.11.0)

## Notes

So I solved my particular problem by hard-coding my user-id in gitlab/v4/objects/user.py l. 641 (in the UserProjectManager list method). Obviously not a general solution, but maybe it helps to track down the error or someone else who comes across this. :)

Furthermore, what I really don't understand is why no KeyError is raised when the format string is processed.

w7089:
## Changes

setting `_from_parent_attrs` on `UserProjectManager` and using it in its `list` method resolves #2558 and allows specific user's projects to be returned successfully. 

The inspiration for the fix came from `_from_parent_attrs = {"user_id": "id"}` line of `UserProjectManager`.

Tested by running below script with arguments: `user-project list --user-id [user_id]`

```
import gitlab
gl = gitlab.GitLab(...)
gl.auth()
import gitlab.cli as cli
cli.main()
```

### DIFF
--- a/gitlab/v4/cli.py
+++ b/gitlab/v4/cli.py
@@ -46,7 +46,7 @@ class GitlabCLI:
 
         self.mgr_cls._path = self.mgr_cls._path.format(**self.parent_args)
         self.mgr = self.mgr_cls(gl)
-
+        self.mgr._from_parent_attrs = self.parent_args
         if self.mgr_cls._types:
             for attr_name, type_cls in self.mgr_cls._types.items():
                 if attr_name in self.args.keys():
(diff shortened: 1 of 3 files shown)


nejch (approved):
Thanks @w7089 and sorry for the delay here. I've pushed a functional test so we can make sure this works in the future.

Issue #2425, opened 2022-12-12
netrc auth overrides OAuth bearer token header


MHLut:
## Description of the problem, including code/CLI snippet

When using the OAuth integration while having a `.netrc` file on the filesystem, the `.netrc` authentication overrides the OAuth bearer token.

Use the following code to trigger an error:

```python
user_token = "long oauth token here"
client = gitlab.Gitlab(url=settings.GITLAB_BASE_URL, oauth_token=user_token)

# Trigger an HTTP 401 response (GitlabAuthenticationError):
client.auth()
```

_(Any call using the Client's `http_xxx()` method should fail; `auth()` is the easiest to use, though)_

## Expected Behavior

Every request to the GitLab API via the `Gitlab` object should authenticate using the OAuth token provided by the user when initializing the client. It should use the `Authorization` header containing the OAuth token as a bearer.

## Actual Behavior

Instead of using the bearer token from the passed `headers` kwarg, `requests` falls back to basic authentication using credentials from the `netrc` file.

### Troubleshooting

So far, I've found out the following:

- The header override occurs in [requests.sessions.prepare_request()](https://github.com/psf/requests/blob/v2.28.1/requests/sessions.py#L457)
- At line `481`, since `request.auth` is empty, auth is filled using `get_netrc_auth()`
  - At line `494`, the `auth` containing netrc values is passed to the prepared request
- At line `490`, the python-gitlab authorization header get overridden by requests' basic auth one

See also this relevant quote from [Requests netrc documentation](https://docs.python-requests.org/en/latest/user/authentication/?highlight=oauth#netrc-authentication):

> If no authentication method is given with the auth argument, Requests will attempt to get the authentication credentials for the URL’s hostname from the user’s netrc file. The netrc file overrides raw HTTP authentication headers set with headers=.
>
> If credentials for the hostname are found, the request is sent with HTTP Basic Auth.

This might be fixed by explicitly passing an `auth` kwarg when using `requests` to make requests.

## Specifications

  - python-gitlab version: `3.12.0`
  - requests version: `2.28.1`
  - API version you are using (v3/v4): `v4`
  - Gitlab server version (or gitlab.com): `15.6.2-ee`

nejch:
[...]

PR #2595, opened 2023-06-16
feat(client): Use requests AuthBase classes

uda:
## Changes

Use Private / Job / OAuth token auth classes based on requests' `AuthBase` class to provide an encapsulated mechanism for passing authentication credentials.

Additionally this helps with #2425 since we always pass the `auth` argument, so unless hitting redirects requests should always prefer the passed token over netrc credentials.

This does however break the behavior when credentials are set using environment variables and a user wants to set `netrc` to override that (username / password should be discouraged, but users have better ideas...), for example, running in the CI the `CI_JOB_TOKEN` is set it will prevent the netrc creds from being loaded, instead of the old API logic (as can seen in the tests changes)

### Documentation and testing

This changes internal implementation by leveraging standard `requests` mechanisms, if the tests cover this part, they should succeed or fail the same way they did before.

Update: I misunderstood the logic re. token precedence and assumed, after rewriting the tests I got the logic

### DIFF
--- a/gitlab/_backends/requests_backend.py
+++ b/gitlab/_backends/requests_backend.py
@@ -4,10 +4,41 @@ import dataclasses
 from typing import Any, Dict, Optional, TYPE_CHECKING, Union
 
 import requests
+from requests import PreparedRequest
+from requests.auth import AuthBase
 from requests.structures import CaseInsensitiveDict
 from requests_toolbelt.multipart.encoder import MultipartEncoder  # type: ignore
 
 from . import protocol
+
+
+class TokenAuth:
+    def __init__(self, token: str):
+        self.token = token
+
+
+class OAuthTokenAuth(TokenAuth, AuthBase):
+    def __call__(self, r: PreparedRequest) -> PreparedRequest:
+        r.headers["Authorization"] = f"Bearer {self.token}"
+        r.headers.pop("PRIVATE-TOKEN", None)
+        r.headers.pop("JOB-TOKEN", None)
+        return r
+
+
+class PrivateTokenAuth(TokenAuth, AuthBase):
+    def __call__(self, r: PreparedRequest) -> PreparedRequest:
+        r.headers["PRIVATE-TOKEN"] = self.token
+        r.headers.pop("JOB-TOKEN", None)
+        r.headers.pop("Authorization", None)
+        return r
+
+
+class JobTokenAuth(TokenAuth, AuthBase):
+    def __call__(self, r: PreparedRequest) -> PreparedRequest:
+        r.headers["JOB-TOKEN"] = self.token
+        r.headers.pop("PRIVATE-TOKEN", None)
+        r.headers.pop("Authorization", None)
+        return r
 
 
 @dataclasses.dataclass
--- a/gitlab/client.py
+++ b/gitlab/client.py
@@ -485,30 +485,24 @@ class Gitlab:
             not self.http_username and self.http_password
         ):
             raise ValueError("Both http_username and http_password should be defined")
-        if self.oauth_token and self.http_username:
+        if tokens and self.http_username:
             raise ValueError(
-                "Only one of oauth authentication or http "
+                "Only one of token authentications or http "
                 "authentication should be defined"
             )
 
-        self._http_auth = None
+        self._auth: Optional[requests.auth.AuthBase] = None
         if self.private_token:
-            self.headers.pop("Authorization", None)
-            self.headers["PRIVATE-TOKEN"] = self.private_token
-            self.headers.pop("JOB-TOKEN", None)
+            self._auth = _backends.PrivateTokenAuth(self.private_token)
 
         if self.oauth_token:
-            self.headers["Authorization"] = f"Bearer {self.oauth_token}"
-            self.headers.pop("PRIVATE-TOKEN", None)
-            self.headers.pop("JOB-TOKEN", None)
+            self._auth = _backends.OAuthTokenAuth(self.oauth_token)
 
         if self.job_token:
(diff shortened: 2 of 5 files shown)


uda:
I hate some of the changes I had to make, the tension between the mypy requirements, pylint and flake8 feel a bit clunky, can this be revisited?

nejch:
> I hate some of the changes I had to make, the tension between the mypy requirements, pylint and flake8 feel a bit clunky, can this be revisited?

If it helps get things done, sure!

We'll just have to make sure later we get the same behavior once we refactor if needed.

gitlab/client.py
```diff
@@ -511,27 +511,21 @@ def _set_auth_info(self) -> None:
...
+            self._http_auth = _backends.HTTPOAuthTokenAuth(self.oauth_token)
 
         if self.job_token:
-            self.headers.pop("Authorization", None)
-            self.headers.pop("PRIVATE-TOKEN", None)
-            self.headers["JOB-TOKEN"] = self.job_token
+            self._http_auth = _backends.HTTPJobTokenAuth(self.job_token)
 
         if self.http_username and self.http_password:
             self._http_auth = requests.auth.HTTPBasicAuth(
```
nejch, line 531: Similar to the above - now that we encapsulate all types of auth here, I would maybe just call it `self._auth` for brevity.

In fact, the current `_http_auth` based on `HTTPBasicAuth` is dead code because GitLab hasn't supported basic auth with user/password for years, so we can remove it in a follow-up.

tests/unit/test_gitlab_auth.py
```diff
@@ -39,51 +41,64 @@ def test_invalid_auth_args():
...
+    p = PreparedRequest()
+    p.prepare(url=gl.url, auth=gl._http_auth)
     assert gl.private_token is None
     assert gl.oauth_token == "oauth_token"
     assert gl.job_token is None
-    assert gl._http_auth is None
-    assert gl.headers["Authorization"] == "Bearer oauth_token"
-    assert "PRIVATE-TOKEN" not in gl.headers
-    assert "JOB-TOKEN" not in gl.headers
+    assert isinstance(gl._http_auth, HTTPOAuthTokenAuth)
```
nejch, line 63:
```suggestion
    assert isinstance(gl._http_auth, OAuthTokenAuth)
```

tests/unit/test_gitlab_auth.py
```diff
@@ -39,51 +41,64 @@ def test_invalid_auth_args():
...
+    p = PreparedRequest()
+    p.prepare(url=gl.url, auth=gl._http_auth)
     assert gl.private_token is None
     assert gl.oauth_token is None
     assert gl.job_token == "CI_JOB_TOKEN"
-    assert gl._http_auth is None
-    assert "Authorization" not in gl.headers
-    assert "PRIVATE-TOKEN" not in gl.headers
-    assert gl.headers["JOB-TOKEN"] == "CI_JOB_TOKEN"
+    assert isinstance(gl._http_auth, HTTPJobTokenAuth)
```
nejch, line 77:
```suggestion
    assert isinstance(gl._http_auth, JobTokenAuth)
```

nejch (changes requested):
Thanks a lot for looking into this @uda!

I had another look and had a few ideas for the inheritance/typing issues (as well as a silly naming idea).

> This does however break the behavior when credentials are set using environment variables and a user wants to set netrc to override that (username / password should be discouraged, but users have better ideas...), for example, running in the CI the CI_JOB_TOKEN is set it will prevent the netrc creds from being loaded, instead of the old API logic (as can seen in the tests changes)

Just to clarify, people cannot use netrc with passwords for the GitLab API for basic auth, as that's been disabled for years (see my other comment). They might be using `.netrc` with tokens as a way to have technology-agnostic auth stored locally (e.g. for API, git over https and packages). But I agree that's still plain-text on a machine. I've seen these used in CI for this purpose though, so maybe we can add a breaking change trailer for this once we're done, or document and test it.

gitlab/_backends/requests_backend.py
```diff
@@ -4,12 +4,46 @@
...
+class HTTPGitLabTokenAuthBase(AuthBase):
+    def __init__(self, token: str):
+        self.token = token
+
+    def __call__(self, r: Request) -> Request:  # type: ignore[override]
+        raise NotImplementedError("Auth hooks must be callable.")
+
+
+class HTTPOAuthTokenAuth(HTTPGitLabTokenAuthBase):
+    def __call__(self, r: Request) -> Request:  # type: ignore[override]
```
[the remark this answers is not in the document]
uda: No reason, except for me misinterpreting the mypy message when there was no type... fixing

gitlab/client.py
uda: Makes sense, implemented

uda:
Let me know when you want to merge and I'll rebase, so I don't chase my tail rebasing :smile:

nejch:
Sorry for keeping this open for so long @uda. I can maybe add a final push here so you don't need to go through the back and forth, but I basically want to check the new netrc behavior and maybe test it. From what I can tell from requests code, when explicitly providing auth, netrc is completely ignored, so we can document this as a breaking change. We'll get this merged though :)

nejch (approved):

nejch:
> Sorry for keeping this open for so long @uda. I can maybe add a final push here so you don't need to go through the back and forth, but I basically want to check the new netrc behavior and maybe test it. From what I can tell from requests code, when explicitly providing auth, netrc is completely ignored, so we can document this as a breaking change. We'll get this merged though :)

Added some tests and docs, as the behavior has changed.

PR #2600, opened 2023-06-23
docs(files): fix minor typo in variable declaration

Julian-Sz:
This should be an instant merge

You can see the documentation creating wrong variables here: https://python-gitlab.readthedocs.io/en/stable/api/gitlab.v4.html#gitlab.v4.objects.ProjectFileManager.raw


nejch (approved):
Nice catch, thanks @Julian-Sz. I just amended the commit message to make our linter happy and will merge.

PR #2629, opened 2023-07-31
feat(api): Add iteration to issue and group filters

cristianocasella:
## Changes

This MR allow to filter issues from groups and projects by iteration.

### Documentation and testing

This change doesn't include new tests and documentation as they don't exist for the other filters.  

Side note: The iteration_id filter is actually supported via API call but not documented yet on public gitlab documentation: https://docs.gitlab.com/ee/api/issues.html


gitlab/v4/objects/issues.py
```diff
@@ -55,8 +55,10 @@ class IssueManager(RetrieveMixin, RESTManager):
         "state",
         "labels",
         "milestone",
+        "iteration_id",
         "scope",
         "author_id",
+        "iteration_id",
```
JohnVillalovos, line 61: Looks like this has been done twice. L58 above already does this.

JohnVillalovos (changes requested):
Thanks for the PR. One minor issue.

gitlab/v4/objects/issues.py
cristianocasella: whops, thanks for finding it!

cristianocasella: I just removed the duplication!

JohnVillalovos (approved):

JohnVillalovos:
Thanks for the PR!

cristianocasella:
Thanks for the support and review @JohnVillalovos!

Issue #1815, opened 2022-01-08
Allow upload generic package from bytes


dAnjou:
Hi,

```python
gl = Gitlab()
project = gl.projects.get(project_id)
```
Project file upload allows both, path or bytes:
```python
file = project.upload(
    filename="my_file.txt",
    filepath="/path/to/my_file.txt",
    # or
    filedata=b"content",
)
```
But project generic package upload only allows path:
```python
package = project.generic_packages.upload(
    package_name="my_package",
    package_version="0.1",
    file_name="my_file.txt",
    path="/path/to/my_file.txt",
    # no bytes allowed :(
)
```
---
I'm happy to work on this, but I'd like to discuss with you first.

I might be too pedantic here but I like consistent argument naming. Both methods already have inconsistent argument names, `filename` and `file_name`, but I guess there's not too much you can do about it without a breaking change, unless you'd be okay with having a weird function signature that allows both for some time while deprecating the old names.

Then there are `filepath` and `path`, so considering there is `filedata` in project file upload, maybe `data` could be used in project generic package upload?

Ideally though, a much cleaner approach would be to accept a file-like object exclusively, considering how easy it is to create one: `pathlib.Path("my_file.txt").open("rb")` or `io.BytesIO(b"content")`. This would result in simpler code, less stuff to test and maintain, and arguably more flexibility for users.

What do you think?

nejch:
Hi @dAnjou thanks a lot for starting this discussion! Since I added that initially, I can try to explain the context for the inconsistencies as I also like consistent arguments, and with some features added years ago we do have to make breaking changes to get there.

I think the idea was actually to _be_ consistent - at least with the upstream API, and I think I was deciding about the naming convention for a while when adding it. You can probably see that as I initially started off basing it on the file upload method and I still have a typo in there (a sneaky `filepath` in the docstring). I even wanted an `UploadMixin` initially but for some reason thought it would be too messy in the end.

So the reason for `file_name` was to follow the [API description](https://docs.gitlab.com/ee/user/packages/generic_packages/). It also seemed more consistent with the `package_*` params there. Why I then went for `path` I can't remember exactly :( I believe I wanted to distinguish it from `new_file_path` which is supposedly returned by the API, but means something else (the path on the server).

[...]

PR #2633, opened 2023-08-03
feat: Allow uploading packages from files and bytes

ernestask:
## Changes

This commit adds a keyword argument to GenericPackageManager.upload() to allow uploading bytes and file-like objects to the generic package registry. That necessitates changing file path to be a keyword argument as well, which then cascades into a whole slew of checks to not allow passing both and to not allow uploading file-like objects as JSON data.

Closes https://github.com/python-gitlab/python-gitlab/issues/1815

### Documentation and testing

### DIFF
--- a/tests/functional/api/test_packages.py
+++ b/tests/functional/api/test_packages.py
@@ -38,7 +38,7 @@ def test_upload_generic_package(tmp_path, project):
     assert package.message == "201 Created"
 
 
-def test_download_generic_package_bytes(tmp_path, project):
+def test_upload_generic_package_as_bytes(tmp_path, project):
     path = tmp_path / file_name
 
     path.write_text(file_content)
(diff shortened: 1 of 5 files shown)


tests/functional/api/test_packages.py
```diff
@@ -38,6 +38,38 @@ def test_upload_generic_package(tmp_path, project):
...
+        package_version=package_version,
+        file_name=file_name,
+        data=path.read_bytes(),
+    )
+
+    assert isinstance(package, GenericPackage)
+    assert package.message == "201 Created"
+
+
+def test_download_generic_package_file(tmp_path, project):
```
nejch, line 57:
```suggestion
def test_upload_generic_package_as_file(tmp_path, project):
```

nejch (changes requested):

nejch:
Thanks @ernestask and sorry to keep you waiting, I pushed a small commit to fix test names so it's not blocking you any longer. We could probably document this a bit as well, but not urgent.

ernestask:
Memory fails me now, but I did want to write some docs. Something stopped me, though… Probably just wasn’t sure where and what to put.

PR #2635, opened 2023-08-08
feat(api): Add support for new runner creation API

### LINKED ISSUE #2606: Support User Runner Registration

As part of Gitlab's runner registration changes, they have introduced a new [runner registration api](https://docs.gitlab.com/ee/api/users.html#create-a-runner) tied to the authenticated current user. This is similar to the group/project runner create (which are being deprecated), except that the registration token is no longer used.

Runner registration tokens will be removed in 16.6.

## Specifications

  - python-gitlab version:
  - Gitlab server version (or gitlab.com): 16+

lgatellier:
## Changes

This PR adds support for GitLab's new runner creation API `POST /user/runners`.
Closes #2606 

### Documentation and testing

- [X] [Unit tests](https://github.com/python-gitlab/python-gitlab/tree/main/tests/unit)

### DIFF
--- a/docs/gl_objects/runners.rst
+++ b/docs/gl_objects/runners.rst
@@ -71,6 +71,14 @@ Register a new runner::
 
     runner = gl.runners.create({'token': secret_token})
 
+.. note::
+
+   A new runner registration workflow has been introduced since GitLab 16.0. This new
+   workflow comes with a new API endpoint to create runner, which does not use
+   registration tokens.
+
+   The new endpoint can be called using ``user.runners.create()``.
+
 Update a runner::
 
     runner = gl.runners.get(runner_id)
--- a/docs/gl_objects/runners.rst
+++ b/docs/gl_objects/runners.rst
@@ -74,10 +74,10 @@ Register a new runner::
 .. note::
 
    A new runner registration workflow has been introduced since GitLab 16.0. This new
-   workflow comes with a new API endpoint to create runner, which does not use
+   workflow comes with a new API endpoint to create runners, which does not use
    registration tokens.
 
-   The new endpoint can be called using ``user.runners.create()``.
+   The new endpoint can be called using ``gl.user.runners.create() after authenticating with `gl.auth()```.
 
 Update a runner::
 
--- a/docs/gl_objects/users.rst
+++ b/docs/gl_objects/users.rst
@@ -456,3 +456,60 @@ Get the users activities::
         query_parameters={'from': '2018-07-01'},
         get_all=True,
     )
+
+Create new runner
+=================
+
+References
+----------
+
+* New runner registration API endpoint (see `Migrating to the new runner registration workflow <https://docs.gitlab.com/ee/ci/runners/new_creation_workflow.html#creating-runners-programmatically>`_)
+
+* v4 API:
+
+  + :class:`gitlab.v4.objects.CurrentUserRunner`
+  + :class:`gitlab.v4.objects.CurrentUserRunnerManager`
+  + :attr:`gitlab.Gitlab.user.runners`
+
+* GitLab API : https://docs.gitlab.com/ee/api/users.html#create-a-runner
+
+Examples
+--------
+
+Create an instance-wide runner::
+
+    runner = gl.user.runners.create({
+        "runner_type": "instance_type",
+        "description": "My brand new runner",
+        "paused": True,
+        "locked": False,
+        "run_untagged": True
+        "tag_list": ["linux", "docker", "testing"],
+        "access_level": "not_protected"
+    })
+
+Create a group runner::
+
+    runner = gl.user.runners.create({
+        "runner_type": "group_type",
+        "group_id": 12345678,
+        "description": "My brand new runner",
+        "paused": True,
+        "locked": False,
+        "run_untagged": True
+        "tag_list": ["linux", "docker", "testing"],
+        "access_level": "not_protected"
+    })
+
+Create a project runner::
+
+    runner = gl.user.runners.create({
+        "runner_type": "project_type",
+        "project_id": 987564321,
+        "description": "My brand new runner",
+        "paused": True,
+        "locked": False,
+        "run_untagged": True
+        "tag_list": ["linux", "docker", "testing"],
+        "access_level": "not_protected"
+    })
--- a/gitlab/v4/objects/users.py
+++ b/gitlab/v4/objects/users.py
@@ -36,6 +36,8 @@ __all__ = [
     "CurrentUserGPGKeyManager",
     "CurrentUserKey",
     "CurrentUserKeyManager",
+    "CurrentUserRunner",
+    "CurrentUserRunnerManager",
     "CurrentUserStatus",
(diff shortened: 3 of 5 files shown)


lgatellier:
Not sure if unit tests fits into `test_runners.py`. Could also be into ` test_users.py`.

Do not hesitate to tell me if I should move them.

JohnVillalovos:
> Not sure if unit tests fits into `test_runners.py`. Could also be into `test_users.py`.
> 
> Please tell me if I should move them.

In my opinion it should be in the `test_users.py` file since that is where the code is that it is testing.

But otherwise this looks good to me.

lgatellier:
Hi @JohnVillalovos, and thanks for your feedback !

I moved the unit test into ` test_users.py` , and fixed a little doc issue in my project runner creation example (which used `group_id` instead of `project_id`).

gitlab/client.py
```diff
@@ -189,6 +189,8 @@ def __init__(
         """See :class:`~gitlab.v4.objects.PagesDomainManager`"""
         self.user_activities = objects.UserActivitiesManager(self)
         """See :class:`~gitlab.v4.objects.UserActivitiesManager`"""
+        self.user_runners = objects.UserRunnerManager(self)
```
JohnVillalovos, line 192: So I don't think this should be added here. I think it should be done how `UserEmail` is done.

gitlab/v4/objects/users.py
```diff
@@ -678,3 +680,28 @@ class UserFollowingManager(ListMixin, RESTManager):
     _path = "/users/{user_id}/following"
     _obj_cls = User
     _from_parent_attrs = {"user_id": "id"}
+
```
JohnVillalovos, line 683: In `class User()` should add an attribute `runners` that is of type `UserRunnerManager`

Then the call would be:  `user.runners.create(...)`

tests/unit/objects/test_users.py
```diff
@@ -304,3 +318,9 @@ def test_list_starred_projects(user, resp_starred_projects):
     projects = user.starred_projects.list()
     assert isinstance(projects[0], StarredProject)
     assert projects[0].id == project_content["id"]
+
+
+def test_create_user_runner(gl: gitlab.Gitlab, resp_runner_create):
```
JohnVillalovos, line 323: Probably the test should be something like.

```
runner = user.runners.create({"runner_type": "instance_type"})
```

docs/gl_objects/runners.rst
```diff
@@ -71,6 +71,14 @@ Register a new runner::
 
     runner = gl.runners.create({'token': secret_token})
 
+.. note::
+
+   A new runner registration workflow has been introduced since GitLab 16.0. This new
+   workflow comes with a new API endpoint to create runner, which does not use
+   registration tokens.
+
+   The new endpoint can be called using ``user_runners.create()``.
```
JohnVillalovos, line 80: Change it to user.runners.create()

docs/gl_objects/users.rst
```diff
@@ -456,3 +456,60 @@ Get the users activities::
...
+  + :attr:`gitlab.Gitlab.user_runners`
+
+* GitLab API : https://docs.gitlab.com/ee/api/users.html#create-a-runner
+
+Examples
+--------
+
+Create an instance-wide runner::
+
+    runner = gl.user_runners.create({
```
JohnVillalovos, line 481: Update all of these to be `user.runners.create`

JohnVillalovos (changes requested):

JohnVillalovos:
@lgatellier Thanks for the PR!

Left  comment to have it work like the other endpoints that use `/user/*`. Please let me know if you disagree with my thoughts.

gitlab/v4/objects/users.py
nejch: @JohnVillalovos just to be sure: I think you meant `CurrentUser` (i.e. all endpoints relating to `/user/` not `/users/`).

gitlab/client.py
nejch: See https://github.com/python-gitlab/python-gitlab/pull/2635#discussion_r1299316773 - @lgatellier I think John here meant `CurrentUserEmail` and other endpoints related to the authenticated user:

https://github.com/python-gitlab/python-gitlab/blob/d054917ccb3bbcc9973914409b9e34ba9301663a/gitlab/v4/objects/users.py#L128-L135

nejch: @lgatellier I tend to agree with John here we like to group our manages and objects according to how GitLab structures its REST API endpoints.

But I understand this is frustrating because it becomes a 2-way step now (create the runner via the user endpoint, then lazily get the created runner for manipulation via the root endpoint). I'm also unsure why the endpoint is done like that upstream as it will be inconsistent with the other runner requests now :confused:

gitlab/v4/objects/users.py
JohnVillalovos: Ah. I think you are correct. Thanks @nejch !

gitlab/client.py
JohnVillalovos: What @nejch said 🙂

lgatellier: Thanks @nejch  @JohnVillalovos  for your feedbacks !

I applied your suggestions. :wink:

nejch (approved):

docs/gl_objects/runners.rst
```diff
@@ -71,6 +71,14 @@ Register a new runner::
 
     runner = gl.runners.create({'token': secret_token})
 
+.. note::
+
+   A new runner registration workflow has been introduced since GitLab 16.0. This new
+   workflow comes with a new API endpoint to create runner, which does not use
```
nejch, line 77:
```suggestion
   workflow comes with a new API endpoint to create runners, which does not use
```

docs/gl_objects/runners.rst
```diff
@@ -71,6 +71,14 @@ Register a new runner::
...
+   registration tokens.
+
+   The new endpoint can be called using ``user.runners.create()``.
```
nejch, line 80:
```suggestion
   The new endpoint can be called using ``gl.user.runners.create() after authenticating with `gl.auth()```.
```

tests/unit/objects/test_users.py
```diff
@@ -241,6 +241,19 @@ def resp_starred_projects():
...
 
+@pytest.fixture
+def resp_runner_create():
+    with responses.RequestsMock() as rsps:
+        rsps.add(
+            method=responses.POST,
+            url="http://localhost/api/v4/user/runners",
+            json={"id": "6", "token": "6337ff461c94fd3fa32ba3b1ff4125"},
+            content_type="application/json",
+            status=200,
```
nejch, line 252:
```suggestion
            status=201,
```

nejch:
Thanks a lot @lgatellier @JohnVillalovos I just pushed a minor typo suggestion so that we can get this in and continue with the tasks before the next scheduled release. If you see anything else just open a follow-up :)

PR #2637, opened 2023-08-16
fix: Fix 404 not found error for saving approval rules 

robinson96:
save() was getting id of the merge request instead of id of the approval_rule, setting _id_attr to approval_rule_id addresses this issue.


JohnVillalovos:
I'm going to close this as I think it has already been fixed in #2773 . And the latest published version of python-gitlab has the fix.

If you disagree please re-open.

PR #2639, opened 2023-08-19
feat: remove support for Python 3.7, require 3.8 or higher

JohnVillalovos:
Python 3.8 is End-of-Life (EOL) as of 2023-06-27 as stated in https://devguide.python.org/versions/ and
https://peps.python.org/pep-0537/

By dropping support for Python 3.7 and requiring Python 3.8 or higher it allows python-gitlab to take advantage of new features in Python 3.8, which are documented at:
https://docs.python.org/3/whatsnew/3.8.html

BREAKING CHANGE: As of python-gitlab 4.0.0, Python 3.7 is no longer
supported. Python 3.8 or higher is required.

### Documentation and testing

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -26,8 +26,6 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         python:
-          - version: "3.7"
-            toxenv: py37,smoke
           - version: "3.8"
             toxenv: py38,smoke
           - version: "3.9"
--- a/.renovaterc.json
+++ b/.renovaterc.json
@@ -38,11 +38,6 @@
       "enabled": false
     },
     {
-      "description": "Pin pytest-console-scripts until we drop 3.7 support",
-      "packageName": "pytest-console-scripts",
-      "allowedVersions": "<1.4.1"
-    },
-    {
       "packagePatterns": [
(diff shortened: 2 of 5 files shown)


JohnVillalovos:
@nejch Not sure if you wanted this to be a BREAKING CHANGE. I don't think it needs to be. People using Python 3.7 can still use v3.15.0 so not sure we have to bump to 4.0.0. But we can...

nejch:
Thanks @JohnVillalovos was just going to start something similar :sweat_smile: 

Personally I'd do a breaking change commit (and we can then add other similar ones), just so people don't get dependency resolution surprises like we do here https://github.com/python-gitlab/python-gitlab/actions/runs/5909806739/job/16030754729?pr=2628. WDYT?

JohnVillalovos:
> Thanks @JohnVillalovos was just going to start something similar 😅
> 
> Personally I'd do a breaking change commit (and we can then add other similar ones), just so people don't get dependency resolution surprises like we do here https://github.com/python-gitlab/python-gitlab/actions/runs/5909806739/job/16030754729?pr=2628. WDYT?

Sounds good to me. I just updated it.

nejch (approved):
Perfect, thanks!

Issue #2535, opened 2023-04-03
the `--all` option is non-functional and should be retired (correct scripts use `--get-all` instead)


d-hat:
Since [this change in v3.7.0](https://github.com/python-gitlab/python-gitlab/commit/7c71d5db11991#diff-cfa2b05f7b559e334405867e2c02564c88032f0dee928f61cba4b84d3d680886L887), the `--all` option has no effect.  It was already marked as deprecated.

The `if get_all is None` test a few lines below is intended to keep `--all` as a synonym for `--get-all`, but it fails because `kwargs['get_all']` is either `True` or `False` - never `None`.  This can be observed by inserting some print statements.

Since the `--all` flag has been non-functional since v3.7.0 and marked as deprecated for much longer, I propose removing it altogether.  This should just be a matter of removing those two lines from `client.py` and the `sub_parser.add_argument("--all", ...)` statement from `v4/cli.py`.

nejch:
Thanks for the report @d-hat, odd no one reported this earlier but I guess as it only affects the CLI people just adapted.

We never really deprecated `all` on the client as it's used so often, but we should remove it from the CLI I agree.

PR #2641, opened 2023-08-20
fix(cli): remove deprecated `--all` option in favor of `--get-all`

nejch:
BREAKING CHANGE: The `--all` option is no longer available in the CLI. Use `--get-all` instead.

Closes https://github.com/python-gitlab/python-gitlab/issues/2535


JohnVillalovos (approved):
Thanks @nejch

PR #2645, opened 2023-08-20
refactor(const): remove deprecated global constant import

nejch:
BREAKING CHANGE: Constants defined in `gitlab.const` can no longer be imported globally from `gitlab`. Import them from `gitlab.const` instead.


gitlab/const.py
```diff
@@ -138,5 +102,33 @@ class SearchScope(GitlabEnum):
...
+    "SEARCH_SCOPE_MERGE_REQUESTS",
+    "SEARCH_SCOPE_MILESTONES",
+    "SEARCH_SCOPE_PROJECT_NOTES",
+    "SEARCH_SCOPE_PROJECTS",
+    "SEARCH_SCOPE_USERS",
+    "SEARCH_SCOPE_WIKI_BLOBS",
+    "USER_AGENT",
+    "VISIBILITY_INTERNAL",
+    "VISIBILITY_PRIVATE",
+    "VISIBILITY_PUBLIC",
```
nejch, line 133: I'd maybe also consider removing some of these entirely as a follow-up and only provide the Enums directly. :thinking:

nejch:
We might also need to drop the dynamic `__all__`:
https://github.com/microsoft/pyright/blob/main/docs/typed-libraries.md#library-interface
(see https://github.com/microsoft/pylance-release/issues/289#issuecomment-1103264263).

JohnVillalovos (approved):
Thanks @nejch. Glad you figured it out 👍

PR #2648, opened 2023-08-20
refactor(list): `as_list` support is removed.

JohnVillalovos:
In `list()` calls support for the `as_list` argument has been removed.
`as_list` was previously deprecated and now the use of `iterator` will
be required if wanting to have same functionality as using `as_list`

BREAKING CHANGE: Support for the deprecated `as_list` argument in
`list()` calls has been removed. Use `iterator` instead.

### DIFF
--- a/gitlab/client.py
+++ b/gitlab/client.py
@@ -843,7 +843,6 @@ class Gitlab:
         path: str,
         query_data: Optional[Dict[str, Any]] = None,
         *,
-        as_list: Optional[bool] = None,  # Deprecated in favor of `iterator`
         iterator: Optional[bool] = None,
         **kwargs: Any,
     ) -> Union["GitlabList", List[Dict[str, Any]]]:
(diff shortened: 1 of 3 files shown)


gitlab/client.py
```diff
@@ -870,21 +870,11 @@ def http_list(
...
             iterator = not as_list
-            utils.warn(
-                message=(
-                    f"`as_list={as_list}` is deprecated and will be removed in a "
-                    f"future version. Use `iterator={iterator}` instead."
-                ),
-                category=DeprecationWarning,
+            raise ValueError(
+                f"`as_list` was previously deprecated. Usage of `as_list` is no "
+                f"longer allowed. Use `iterator={iterator}` instead."
```
nejch, line 877: Since it's been over 15 months of people getting warnings, how about we completely remove all the codebase related to `as_list`? :partying_face:

nejch (changes requested):

gitlab/client.py
JohnVillalovos: I worry that someone ignored `as_list` warnings that things will sort of silently break for them if we just remove it. And then they will not understand why things are breaking.

This way BOOM it is broken and they have to fix it.

JohnVillalovos: So thinking about it more I guess the breakage for them would most likely be that things run longer than expected?  If they are doing a list of something which is 5,000 items it will now be returned as a list and take probably a long time to initially return back to them. But I guess it probably won't break them but slow things down. Especially if they have a `break` statement in their `for` loop.

JohnVillalovos: So basically I am okay just removing it. Since worst case for them is it runs slower and it is passing `as_list` to GitLab in the API call. What do you think @nejch ?

nejch: I would also say it's fine @JohnVillalovos, I'm also not for big sudden breaking changes but this has been deprecated for over a year, and we are still following semver this way so I don't think it's inconsiderate to users if we add a clear changelog and maybe make some release notes like we did for v3. It would allow us to simplify the code as well, I think. :)

JohnVillalovos: Done

nejch (approved):
Thanks @JohnVillalovos!

PR #2683, opened 2023-10-09
feat(api): add support for job token scope settings

nejch:
Depends on #2682

### DIFF
--- a/docs/gl_objects/job_token_scope.rst
+++ b/docs/gl_objects/job_token_scope.rst
@@ -1,0 +1,51 @@
+#####################
+CI/CD job token scope
+#####################
+
+Reference
+---------
+
+* v4 API:
+
+  + :class:`gitlab.v4.objects.ProjectJobTokenScope`
+  + :class:`gitlab.v4.objects.ProjectJobTokenScopeManager`
+  + :attr:`gitlab.v4.objects.Project.job_token_scope`
+
+* GitLab API: https://docs.gitlab.com/ee/api/project_job_token_scopes.html
+
+Examples
+--------
+
+.. warning::
+
+   The GitLab API does **not** return any data when saving or updating
+   the job token scope settings. You need to call ``refresh()`` (or ``get()``
+   a new object) as shown below to get the latest state.
+
+Get a project's CI/CD job token access settings::
+
+    scope = project.job_token_scope.get()
+    print(scope.inbound_enabled)
+    # True
+
+Update the job token scope settings::
+
+    scope.enabled = False
+    scope.save()
+
+.. warning::
+
+   As you can see above, the attributes you receive from and send to the GitLab API
+   are not consistent. GitLab returns `inbound_enabled` and `outbound_enabled`,
+   but expects `enabled`, which only refers to the inbound scope. This is important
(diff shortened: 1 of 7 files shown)


gitlab/mixins.py
```diff
@@ -324,6 +325,8 @@ def _get_update_method(
         """
         if self._update_uses_post:
             http_method = self.gitlab.http_post
+        elif self._update_uses_patch:
+            http_method = self.gitlab.http_patch
```
nejch, line 329: I will refactor this into a proper mapping so we just return `self._update_method`, defaulting to PUT. But for now, this was a smaller change

docs/gl_objects/job_token_scope.rst
```diff
@@ -0,0 +1,51 @@
...
+
+    scope.enabled = False
+    scope.save()
+
+.. warning::
+
+   As you can see above, the attributes you receive from and send to the GitLab API
+   are not consistent. GitLab returns `inbound_enabled` and `outbound_enabled`,
+   but expects `enabled`, which only refers to the inbound scope. This is important
+   when accessing and updating these attributes.
```
nejch, line 41: I didn't try to reconcile this with custom code - if the GitLab API is inconsistent I guess we should follow it. Who knows how this will change in the coming months as the outbound scope gets deprecated ¯\_(ツ)_/¯

JohnVillalovos: Makes sense to me.

gitlab/mixins.py
```diff
@@ -312,6 +312,7 @@ class UpdateMixin(_RestManagerBase):
     _parent_attrs: Dict[str, Any]
     _path: Optional[str]
     _update_uses_post: bool = False
+    _update_uses_patch: bool = False
```
JohnVillalovos, line 315: Should we use an `enum` or something for this?

Or check that only one is set to `True`?

JohnVillalovos: I put together a PR here: https://github.com/python-gitlab/python-gitlab/pull/2685

If you like it you can merge it or incorporate it into this PR.

nejch: Makes sense, but would be cool if we could use that, or whatever we go with, to get rid of the helper method and its if/elif/else clauses. Haven't thought about it too much though.

gitlab/v4/objects/job_token_scope.py
```diff
@@ -0,0 +1,23 @@
...
+
+class ProjectJobTokenScope(RefreshMixin, SaveMixin, RESTObject):
+    _id_attr = None
+
+
+class ProjectJobTokenScopeManager(GetWithoutIdMixin, UpdateMixin, RESTManager):
+    _path = "/projects/{project_id}/job_token_scope"
+    _obj_cls = ProjectJobTokenScope
+    _from_parent_attrs = {"project_id": "id"}
+    _update_uses_patch = True
```
JohnVillalovos, line 20: I think this needs to be updated.

nejch: Sorry that was a lazy rebase :grin: fixed.

JohnVillalovos (approved):

PR #2684, opened 2023-10-10
feat: officially support Python 3.12

nejch:


.github/workflows/test.yml
```diff
@@ -34,17 +34,19 @@ jobs:
             toxenv: py310,smoke
           - version: "3.11"
             toxenv: py311,smoke
-          - version: '3.12.0-alpha - 3.12' # SemVer's version range syntax
+          - version: "3.12"
             toxenv: py312,smoke
+          - version: '3.13.0-alpha - 3.13' # SemVer's version range syntax
+            toxenv: py313,smoke
```
nejch, line 40: A bit too eager - 3.13 doesn't exist on here yet:
https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json

nejch:
Let me know if you'd like to take a look already @JohnVillalovos. The pre-commit failure is a bit cryptic and I think it's to do with all the magic it does in installing its own deps, I wouldn't take that too seriously as our pylint job works well in tox. It's not the first time the pre-commit job is being weird in CI

JohnVillalovos:
> Let me know if you'd like to take a look already @JohnVillalovos. The pre-commit failure is a bit cryptic and I think it's to do with all the magic it does in installing its own deps, I wouldn't take that too seriously as our pylint job works well in tox. It's not the first time the pre-commit job is being weird in CI

Change pre-commit.yml to use Python 3.11 and we don't have an issue. Looks like a bug in pre-commit.

I tested here: https://github.com/python-gitlab/python-gitlab/pull/2687

nejch:
> > Let me know if you'd like to take a look already @JohnVillalovos. The pre-commit failure is a bit cryptic and I think it's to do with all the magic it does in installing its own deps, I wouldn't take that too seriously as our pylint job works well in tox. It's not the first time the pre-commit job is being weird in CI
> 
> Change pre-commit.yml to use Python 3.11 and we don't have an issue. Looks like a bug in pre-commit.
> 
> I tested here: #2687

Ah lol nice, that's much easier than trying to wrestle with that. Thanks! Done.

JohnVillalovos:
> > > Let me know if you'd like to take a look already @JohnVillalovos. The pre-commit failure is a bit cryptic and I think it's to do with all the magic it does in installing its own deps, I wouldn't take that too seriously as our pylint job works well in tox. It's not the first time the pre-commit job is being weird in CI
> > 
> > 
> > Change pre-commit.yml to use Python 3.11 and we don't have an issue. Looks like a bug in pre-commit.
> > I tested here: #2687
> 
> Ah lol nice, that's much easier than trying to wrestle with that. Thanks! Done.

I'm not sure what the real issue but I feel this would be a deep rabbit hole to go down to figure it out. So I prefer to leave it at 3.11 and later in a month or so it will probably be fixed and we can move to 3.12.

Approved and thanks.

nejch:
Yes, I have the same feeling and didn't have the energy 😅 thanks!

PR #2685, opened 2023-10-11
chore: use an `enum` for `_update_uses` so we can support `PATCH`

JohnVillalovos:
Use an `enum` to indicate what method the `update` action will use. At the moment it supports `POST` and `PUT`. But can in the future support `PATCH`.

### Documentation and testing


gitlab/mixins.py
```diff
@@ -304,14 +305,20 @@ def create(
         return self._obj_cls(self, server_data)
 
 
+@enum.unique
+class UpdateUses(enum.IntEnum):
```
JohnVillalovos, line 309: Maybe this should be called `UpdateMethod` ? Or a different name. Or this name?

nejch: `UpdateMethod` makes sense to me.

I was thinking of something that could entirely replace our `_get_update_method` and just define the `http_*` method inside the Enum, but not sure there's an elegant way to do it as it's only available after instantiating the manager :thinking:

JohnVillalovos: > `UpdateMethod` makes sense to me.

Done

> I was thinking of something that could entirely replace our `_get_update_method` and just define the `http_*` method inside the Enum, but not sure there's an elegant way to do it as it's only available after instantiating the manager 🤔

Not sure how to do that off the top of my head. So going with this simpler method now.

PR #2686, opened 2023-10-11
refactor(build): build project using PEP 621

nejch:
BREAKING CHANGE: python-gitlab now stores metadata in pyproject.toml
as per PEP 621, with setup.py removed. pip version v21.1 or higher is
required if you want to perform an editable install.

---

Reviving https://github.com/python-gitlab/python-gitlab/pull/2224, I can't seem to reopen that PR now that I've force-pushed to the branch. Only kept a subset of changes here.

This is related to #2684, might be good to drop setup.py completely here now, as 3.12 drops setuptools and it complicates things.


gitlab/v4/objects/appearance.py
```diff
@@ -39,7 +39,7 @@ def update(
         self,
         id: Optional[Union[str, int]] = None,
         new_data: Optional[Dict[str, Any]] = None,
-        **kwargs: Any
+        **kwargs: Any,
```
nejch, line 42: I didn't really dig into why this is now suddenly an issue, maybe the removal of deps freed up some newer versions of transitive deps that require this change now.

nejch:
@JohnVillalovos since 3.12 doesn't ship with setuptools anymore, I thought I'd bring #2224 back here. The change in config brought in some unrelated linter errors :sweat_smile:

pyproject.toml
```diff
@@ -1,3 +1,63 @@
...
+    "License :: OSI Approved :: GNU Lesser General Public License v3 (LGPLv3)",
+    "Natural Language :: English",
+    "Operating System :: POSIX",
+    "Operating System :: Microsoft :: Windows",
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3.8",
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11"
```
JohnVillalovos, line 36: Probably add Python 3.12, but not too important.

JohnVillalovos (approved):

JohnVillalovos:
Thanks @nejch !

pyproject.toml
nejch: Yeah, adding that in the next one :)

PR #2692, opened 2023-10-16
chore(ci): update release build for python-semantic-release v8

nejch:
Fixes failure in https://github.com/python-gitlab/python-gitlab/actions/runs/6530439704/job/17729726207.

python-semantic-release uses a very similar config:
https://github.com/python-semantic-release/python-semantic-release/blob/master/pyproject.toml#L329-L331

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -88,11 +88,11 @@ disable_error_code = ["no-untyped-def"]
 
 [tool.semantic_release]
 branch = "main"
+build_command = "python -m build ."
 version_variables = [
     "gitlab/_version.py:__version__",
 ]
-commit_subject = "chore: release v{version}"
-commit_message = ""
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -88,11 +88,11 @@ disable_error_code = ["no-untyped-def"]
 
 [tool.semantic_release]
 branch = "main"
+build_command = "python -m build ."
 version_variables = [
     "gitlab/_version.py:__version__",
 ]
-commit_subject = "chore: release v{version}"
-commit_message = ""
+commit_message = "chore: release v{version}"
```
nejch, line 95: `commit_subject` is gone from v8 so it was failing. and we have to explicitly add `build_command` now otherwise it won't automatically build the wheel / source.

PR #2707, opened 2023-10-27
docs(users): add missing comma in v4 API create runner examples

anon6789:
The examples which show usage of new runner registration api endpoint are missing commas. This change adds the missing commas.

## Changes

Fix examples in v4 runner registration API docs.


nejch (approved):
Nice catch @anon6789, looks good!

Issue #2714, opened 2023-11-03
Boolean values parsed on the cli by argument or enviroment are not parsed correctly


TheDJVG:
## Description of the problem, including code/CLI snippet
Some arguments expect a boolean value but these are not actually parsed as the Config Parser only looks at the sections in the config file and expect the value

Trying to disable ssl-verify by argument:
```
# docker run --rm -it registry.gitlab.com/python-gitlab/python-gitlab:latest --ssl-verify false --debug user list 
Impossible to list objects (Could not find a suitable TLS CA certificate bundle, invalid path: false)
```
By env. var:
```
# docker run --rm -it -e GITLAB_SSL_VERIFY=false registry.gitlab.com/python-gitlab/python-gitlab:latest user list
Impossible to list objects (Could not find a suitable TLS CA certificate bundle, invalid path: false)
```

## Expected Behavior

Expected SSL/TLS cert. verification to be turned off.

## Actual Behavior

The string "false" is used as a path.

## Specifications

  - python-gitlab version: 4.1.0
  - API version you are using (v3/v4): not applicable
  - Gitlab server version (or gitlab.com): not applicable

nejch:
Thanks for the report @TheDJVG. It's an unfortunate consequence of accepting both boolean and strings for this option which is not ideal. 

For now you should be able to configure this using a configuration file. But we might need some strtobool logic for the CLI.

JohnVillalovos:
Does `0` work instead of `false`?

JohnVillalovos:
> Thanks for the report @TheDJVG. It's an unfortunate consequence of accepting both boolean and strings for this option which is not ideal.

Maybe in the CLI could use an `action="store_true"` or `store_false`.  Might have to use a mutually exclusive group and have two options (`--ssl-verify` and `--no-ssl-verify`). There is a `BooleanOptionalAction` in `argparse` but wasn't added until 3.9.

TheDJVG:
I think additionally the challenge is that if it does not evaluate to true or false it should be the path to CA certificates.

JohnVillalovos:
> I think additionally the challenge is that if it does not evaluate to true or false it should be the path to CA certificates.

#2717 makes it so can specify `--no-ssl-verify` and will set it to `False`  With no arguments then by default `ssl_verify=True` and if specify `--ssl-verify` then it will be set to the string specified which should be a path to CA certificates.  `--no-ssl-verify` and `--ssl-verify` are mutually exclusive.

PR #2717, opened 2023-11-04
fix(cli): add ability to disable SSL verification

JohnVillalovos:
Add a `--no-ssl-verify` option to disable SSL verification

Closes: #2714

### Documentation and testing

### DIFF
--- a/gitlab/cli.py
+++ b/gitlab/cli.py
@@ -194,14 +194,25 @@ def _get_base_parser(add_help: bool = True) -> argparse.ArgumentParser:
         required=False,
         default=os.getenv("GITLAB_URL"),
     )
-    parser.add_argument(
+
+    ssl_verify_group = parser.add_mutually_exclusive_group()
+    ssl_verify_group.add_argument(
         "--ssl-verify",
         help=(
-            "Whether SSL certificates should be validated. [env var: GITLAB_SSL_VERIFY]"
+            "Path to a CA_BUNDLE file or directory with certificates of trusted CAs. "
+            "[env var: GITLAB_SSL_VERIFY]"
         ),
         required=False,
         default=os.getenv("GITLAB_SSL_VERIFY"),
     )
+    ssl_verify_group.add_argument(
+        "--no-ssl-verify",
+        help="Disable SSL verification",
+        required=False,
+        dest="ssl_verify",
+        action="store_false",
+    )
+
     parser.add_argument(
         "--timeout",
         help=(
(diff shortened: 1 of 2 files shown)


gitlab/cli.py
```diff
@@ -194,14 +194,25 @@ def _get_base_parser(add_help: bool = True) -> argparse.ArgumentParser:
...
         default=os.getenv("GITLAB_SSL_VERIFY"),
     )
+    ssl_verify_group.add_argument(
+        "--no-ssl-verify",
+        help="Disable SSL verification",
+        required=False,
+        dest="ssl_verify",
+        action="store_false",
+    )
+
```
nejch, line 215: I like the simplicity of this, but it doesn't fully fix #2714. We still cannot use environment variables (this is true for all of our booleans on the CLI). We'd need some kind of strtobool logic for those, and that maybe also falls back to strings for the `ssl_verify` arg. Also, the new flag doesn't have an env var equivalent this way like the others do. WDYT?

nejch (changes requested):
Thanks a lot @JohnVillalovos! This is nice and simple but I think we might need a more general fix, WDYT?

JohnVillalovos:
> Thanks a lot @JohnVillalovos! This is nice and simple but I think we might need a more general fix, WDYT?

I don't disagree that it could be better but it does handle the issue brought up in #2714  As it will allow them to disable `ssl_verify`

At the moment I'm not sure if I personally have the time to start down the path that will be required to make all environment variables work.

If someone wants to attempt it then the strtobool could be done similar to this: https://github.com/python/cpython/blob/f55cb44359821e71c29903f2152b4658509dac0d/Lib/configparser.py#L1092-L1097

nejch:
> > Thanks a lot @JohnVillalovos! This is nice and simple but I think we might need a more general fix, WDYT?
> 
> I don't disagree that it could be better but it does handle the issue brought up in #2714 As it will allow them to disable `ssl_verify`
> 
> At the moment I'm not sure if I personally have the time to start down the path that will be required to make all environment variables work.
> 
> If someone wants to attempt it then the strtobool could be done similar to this: https://github.com/python/cpython/blob/f55cb44359821e71c29903f2152b4658509dac0d/Lib/configparser.py#L1092-L1097

Sorry for the delay here, I was just thinking about whether we should introduce something that would potentially be deprecated again if we have a strtobool solution later, maybe we can go ahead with this if I don't get to it before the 28th :grin:

PR #2722, opened 2023-11-15
feat(api): add support for wiki attachments

TigreModerata:
## Changes

UploadMixin added, inherits from RESTObject;
UploadMixin used in projects.py and wikis.py to upload files; 
Test for upload to wiki api added in tests/functional/api/test_wikis.
Documentation and unit tests not yet added - waiting for corrections/objections 

### Documentation and testing

### DIFF
--- a/docs/gl_objects/wikis.rst
+++ b/docs/gl_objects/wikis.rst
@@ -54,3 +54,39 @@ Update a wiki page::
 Delete a wiki page::
 
     page.delete()
+
+
+File uploads
+============
+
+Reference
+---------
+
+* v4 API:
+
+  + :attr:`gitlab.v4.objects.ProjectWiki.upload`
+  + :attr:`gitlab.v4.objects.GrouptWiki.upload`
+
+
+* Gitlab API for Projects: https://docs.gitlab.com/ee/api/wikis.html#upload-an-attachment-to-the-wiki-repository
+* Gitlab API for Groups: https://docs.gitlab.com/ee/api/group_wikis.html#upload-an-attachment-to-the-wiki-repository
+
+Examples
+--------
+
+Upload a file into a project wiki using a filesystem path::
+
+    page.upload("filename.txt", filepath="/some/path/filename.txt")
+
+Upload a file into a project wiki without a filesystem path::
+
+    page.upload("filename.txt", filedata="Raw data")
+
+Upload a file into a project wiki using a filesystem path::
+
+    page.upload("filename.txt", filepath="/some/path/filename.txt")
+
+Upload a file into a project wiki without a filesystem path::
+
+    page.upload("filename.txt", filedata="Raw data")
+
--- a/docs/gl_objects/wikis.rst
+++ b/docs/gl_objects/wikis.rst
@@ -76,17 +76,19 @@ Examples
 
 Upload a file into a project wiki using a filesystem path::
 
+    page = project.wikis.get(page_slug)
     page.upload("filename.txt", filepath="/some/path/filename.txt")
 
-Upload a file into a project wiki without a filesystem path::
+Upload a file into a project wiki with raw data::
 
     page.upload("filename.txt", filedata="Raw data")
 
-Upload a file into a project wiki using a filesystem path::
+Upload a file into a group wiki using a filesystem path::
 
+    page = group.wikis.get(page_slug)
     page.upload("filename.txt", filepath="/some/path/filename.txt")
 
-Upload a file into a project wiki without a filesystem path::
+Upload a file into a group wiki using raw data::
 
     page.upload("filename.txt", filedata="Raw data")
 
--- a/gitlab/mixins.py
+++ b/gitlab/mixins.py
@@ -944,3 +944,74 @@ class PromoteMixin(_RestObjectBase):
         if TYPE_CHECKING:
             assert not isinstance(result, requests.Response)
         return result
+
+
+class UploadMixin(_RestObjectBase):
+    _id_attr: Optional[str]
+    _attrs: Dict[str, Any]
+    _module: ModuleType
+    _parent_attrs: Dict[str, Any]
+    _updated_attrs: Dict[str, Any]
+    _upload_path: str
+    manager: base.RESTManager
+
+    def _get_upload_path(self) -> str:
+        """Formats _upload_path with object attributes.
+
+        Returns:
+            The upload path
+        """
+        if TYPE_CHECKING:
+            assert isinstance(self._upload_path, str)
+        data = self.attributes
+        return self._upload_path.format(**data)
+
+    @cli.register_custom_action(("Project", "ProjectWiki"), ("filename", "filepath"))
+    @exc.on_http_error(exc.GitlabUploadError)
+    def upload(
+        self,
+        filename: str,
+        filedata: Optional[bytes] = None,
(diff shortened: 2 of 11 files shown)


nejch (changes requested):
Thanks a lot @TigreModerata, I like the direction here but just had a question about the exception raising. This would also potentially address the coverage drop I think. Let me know what you think!

TigreModerata:
I removed the exception. Changed _upload_path to _upload_string in Project, ProjectWiki and UploadMixin.
I keep messing up the commit message it seems...

If you are happy with the overall approach, maybe I should add the tests for the mixin and other unit tests (and update the docs)?

TigreModerata:
Sorry again about the commit messages, I'll fix it in the next one, have to run!

nejch:
> I removed the exception. Changed _upload_path to _upload_string in Project, ProjectWiki and UploadMixin.
> I keep messing up the commit message it seems...
> 
> If you are happy with the overall approach, maybe I should add the tests for the mixin and other unit tests (and update the docs)?

My suggestion was actually in the other direction (to use `_upload_path` everywhere), as that's consistent with our `_path` and `_computed_path` attributes. Otherwise I agree, we can go ahead with this and update the tests/docs, yes! Thanks so much!

nejch:
> Sorry again about the commit messages, I'll fix it in the next one, have to run!

@TigreModerata no worries, it's best to just squash the commits into one when you're finished with the fixups.

TigreModerata:
OK, I managed an ok commit :) 
I added a section to the wikis documentation about the file upload and a unit test for the UploadMixin.

I only dealt with ProjectWiki because that was mentioned in the issue, but for completeness I should probably do GroupWiki too (not sure if there are more objects that support upload). I'm just not sure if it should be on the same branch/issue, or a new one?

docs/gl_objects/wikis.rst
```diff
@@ -54,3 +54,39 @@ Update a wiki page::
...
+
+    page.upload("filename.txt", filedata="Raw data")
+
+Upload a file into a project wiki using a filesystem path::
+
+    page.upload("filename.txt", filepath="/some/path/filename.txt")
+
+Upload a file into a project wiki without a filesystem path::
+
+    page.upload("filename.txt", filedata="Raw data")
```
nejch, line 91: This is a bit duplicated now, but I guess you wanted to add docs for both group/project wikis. Maybe we can do something like this to make the distinction:

```suggestion
Upload a file into a project wiki using a filesystem path::

    page = project.wikis.get(page_slug)
    page.upload("filename.txt", filepath="/some/path/filename.txt")

Upload a file into a project wiki with raw data::

    page.upload("filename.txt", filedata="Raw data")

Upload a file into a group wiki using a filesystem path::

    page = group.wikis.get(page_slug)
    page.upload("filename.txt", filepath="/some/path/filename.txt")

Upload a file into a group wiki using raw data::

    page.upload("filename.txt", filedata="Raw data")
```

gitlab/mixins.py
```diff
@@ -944,3 +944,74 @@ def promote(self, **kwargs: Any) -> Dict[str, Any]:
...
+        **kwargs: Any,
+    ) -> Dict[str, Any]:
+        """Upload the specified file.
+
+        .. note::
+
+            Either ``filedata`` or ``filepath`` *MUST* be specified.
+
+        Args:
+            path: api endpoint where file is to be posted
```
nejch, line 985: Just noticed this is no longer relevant:
```suggestion
```

tests/unit/objects/test_projects.py
```diff
@@ -682,29 +661,6 @@ def test_delete_project(project, resp_delete_project):
...
-        exceptions.GitlabUploadError, match="No file contents or path specified"
-    ):
-        project.upload("filename.png")
-
-
-def test_upload_file_with_filepath_and_filedata(project):
-    with pytest.raises(
-        exceptions.GitlabUploadError, match="File contents and file path specified"
-    ):
-        project.upload("filename.png", "filedata", "/filepath")
```
nejch, line 705: Could we perhaps keep these unit tests? That way we know we don't have any regressions later on. If we need any shared fixtures we can put them into our `conftest.py`.

tests/unit/mixins/test_mixin_methods.py
```diff
@@ -502,3 +505,44 @@ class M(SetMixin, FakeManager):
...
+    assert res_only_data["file_content"] == "testing contents"
+    assert responses.assert_call_count(url, 1) is True
+
+    with patch("builtins.open", mock_open(read_data="raw\nfile\ndata")):
+        res_only_path = obj.upload("test.txt", None, "/filepath")
+    assert obj._get_upload_path() == "/tests/42/uploads"
+    assert isinstance(res_only_path, dict)
+    assert res_only_path["file_name"] == "test.txt"
+    assert res_only_path["file_content"] == "testing contents"
+    assert responses.assert_call_count(url, 2) is True
```
nejch, line 548: Ideally these different types of asserts would be in individual unit tests IMO, like we had in the file upload tests, but if it's too much hassle we can keep it this way.

nejch (changes requested):
Thanks again @TigreModerata, I had a few more comments here, let me know what you think!

tests/unit/objects/test_projects.py
TigreModerata: Sure, makes sense!
I wasn't sure if we also wanted unit tests for wiki uploads - it seems redundant (everything seems to get tested as is), but what do you think?

TigreModerata:
Separated unit tests for upload mixin and re-added the file upload unit tests for projects

requirements-lint.txt
```diff
@@ -1,13 +1,13 @@
...
-mypy==1.6.1
+mypy==1.7.0
 pylint==3.0.2
 pytest==7.4.3
-responses==0.24.0
+responses==0.24.1
 types-PyYAML==6.0.12.12
 types-requests==2.31.0.10
-types-setuptools==68.2.0.0
+types-setuptools==68.2.0.1
```
nejch, line 13: I guess these dependency changes were added here by mistake, but since it's not too drastic and it's green, let's keep this.

nejch (approved):

nejch:
Thanks a lot for your patience @TigreModerata! I squashed the commits in the end as there were 2 identical commit messages :bow:

TigreModerata:
> Thanks a lot for your patience @TigreModerata! I squashed the commits in the end as there were 2 identical commit messages 🙇

OK, l have no idea what happened (I did many wrong resets, to squash commits, and I never hit the right point it seems) - I will learn one day. Thank YOU for your patience, are you kidding! I do realize that my "help" is more of a chore, but you guys' help and patience  is the only way I make any progress at all! 
Speaking of, if there are any other old issues that would be useful to tackle, I'm always happy to try (and commit properly maybe).

nejch:
No worries @TigreModerata reviewing can be easier than starting things sometimes :sweat_smile: 

Most open issues could use some more love but some are definitely lower-hanging fruit - I'd say things that are labelled `feature` are the most similar to what you've done here, i.e. if it's about implementing endpoints that we're missing here:

https://github.com/python-gitlab/python-gitlab/issues?q=is%3Aissue+is%3Aopen+label%3Afeature

PR #2729, opened 2023-11-23
feat: add pipeline status as Enum

nickbroon:
https://docs.gitlab.com/ee/api/pipelines.html

Would be nice to extend the [ProjectPipeline object](https://python-gitlab.readthedocs.io/en/stable/api/gitlab.v4.html#gitlab.v4.objects.ProjectPipeline) with an api to get the status that returns this, but 100% how do that.


nejch (approved):
Great idea, thanks a lot @nickbroon!

Issue #2733, opened 2023-12-02
Missing api call to get single merge request reviewers


kesmik:
## Description of the problem, including code/CLI snippet
Gitlab has an option to fetch reviewers list for single MR, as it is described here: https://docs.gitlab.com/ee/api/merge_requests.html#get-single-merge-request-reviewers

It actually returns a little bit more information than information in MR itself: "state" and "created_at" attributes are returned

## Expected Behavior
merge requests should contain reviewers() method which would call:
`GET /projects/:id/merge_requests/:merge_request_iid/reviewers` gitlab rest api call

## Actual Behavior
method to get reviewers list for single mr does not exist

## Specifications

  - python-gitlab version: 4.2.0
  - API version you are using (v3/v4): v4
  - Gitlab server version (or gitlab.com): 15.4.3-ee

kesmik:
Probably ProjectMergeRequest class should be extended to have `reviewers()` method, but since I am not familiar with codebase I may be wrong.

as a quick and dirty way I tested this way:
```python
class CustProjectMergeRequest(ProjectMergeRequest):
    def reviewers(self, **kwargs: Any) -> RESTObjectList:

        path = f"{self.manager.path}/{self.encoded_id}/reviewers"
        result = self.manager.gitlab.http_get(path, **kwargs)
        if TYPE_CHECKING:
            assert not isinstance(result, requests.Response)
        return result


class CustProjectMergeRequestManager(ProjectMergeRequestManager):
    _obj_cls = CustProjectMergeRequest

    def get(
        self, id: Union[str, int], lazy: bool = False, **kwargs: Any
    ) -> CustProjectMergeRequest:
        return cast(CustProjectMergeRequest, super().get(id=id, lazy=lazy, **kwargs))


class CustProject(Project):
    mergerequests: CustProjectMergeRequestManager


class CustProjectManager(ProjectManager):
    _obj_cls = CustProject

    def get(
        self, id: Union[str, int], lazy: bool = False, **kwargs: Any
    ) -> CustProject:
        return cast(CustProject, super().get(id=id, lazy=lazy, **kwargs))
```

Also I have changed default projects class
```
and gitlab.Gitlab().projects = CustProjectManager(self.gl_api)
```

If someone could quickly advice if updating `ProjectMergeRequest` in original library is enough to be updated I can submit a PR.

PR #2734, opened 2023-12-03
feat(api): add get_reviewers call for mergrequest to get reviewers of…

kesmik:
Those changes implements 'GET /projects/:id/merge_requests/:merge_request_iid/reviewers' gitlab API call. Naming for call is not reviewers because reviewers atribute already present in merge request class.
https://docs.gitlab.com/ee/api/merge_requests.html#get-single-merge-request-reviewers

Closes: https://github.com/python-gitlab/python-gitlab/issues/2733

### Documentation and testing

### DIFF
--- a/gitlab/v4/objects/merge_requests.py
+++ b/gitlab/v4/objects/merge_requests.py
@@ -42,6 +42,7 @@ from .merge_request_approvals import (  # noqa: F401
 )
 from .notes import ProjectMergeRequestNoteManager  # noqa: F401
 from .pipelines import ProjectMergeRequestPipelineManager  # noqa: F401
+from .reviewers import ProjectMergeRequestReviewerDetailManager
 
 __all__ = [
     "MergeRequest",
@@ -164,6 +165,7 @@ class ProjectMergeRequest(
     resourcelabelevents: ProjectMergeRequestResourceLabelEventManager
     resourcemilestoneevents: ProjectMergeRequestResourceMilestoneEventManager
     resourcestateevents: ProjectMergeRequestResourceStateEventManager
+    reviewer_details: ProjectMergeRequestReviewerDetailManager
 
     @cli.register_custom_action("ProjectMergeRequest")
     @exc.on_http_error(exc.GitlabMROnBuildSuccessError)
--- a/gitlab/v4/objects/reviewers.py
+++ b/gitlab/v4/objects/reviewers.py
@@ -1,0 +1,17 @@
+from gitlab.base import RESTManager, RESTObject
+from gitlab.mixins import ListMixin
+
+__all__ = [
+    "ProjectMergeRequestReviewerDetail",
+    "ProjectMergeRequestReviewerDetailManager",
+]
+
+
+class ProjectMergeRequestReviewerDetail(RESTObject):
+    pass
+
+
+class ProjectMergeRequestReviewerDetailManager(ListMixin, RESTManager):
+    _path = "/projects/{project_id}/merge_requests/{mr_iid}/reviewers"
+    _obj_cls = ProjectMergeRequestReviewerDetail
+    _from_parent_attrs = {"project_id": "project_id", "mr_iid": "iid"}
--- a/docs/gl_objects/merge_requests.rst
+++ b/docs/gl_objects/merge_requests.rst
@@ -94,6 +94,10 @@ Get a single MR::
 
     mr = project.mergerequests.get(mr_iid)
 
+Get MR reviewer details::
+    mr = project.mergerequests.get(mr_iid)
+    reviewers = mr.reviewer_details.list()
+
 Create a MR::
 
     mr = project.mergerequests.create({'source_branch': 'cool_feature',
--- a/gitlab/v4/objects/__init__.py
+++ b/gitlab/v4/objects/__init__.py
@@ -56,6 +56,7 @@ from .push_rules import *
 from .releases import *
 from .repositories import *
 from .resource_groups import *
+from .reviewers import *
 from .runners import *
 from .secure_files import *
 from .settings import *
--- a/tests/unit/objects/test_merge_requests.py
+++ b/tests/unit/objects/test_merge_requests.py
@@ -8,7 +8,11 @@ import re
 import pytest
 import responses
 
-from gitlab.v4.objects import ProjectDeploymentMergeRequest, ProjectMergeRequest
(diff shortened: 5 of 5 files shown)


gitlab/v4/objects/merge_requests.py
```diff
@@ -409,6 +409,30 @@ def merge(
...
+
+        Returns:
+            The list of reviewers
+        """
+
+        path = f"{self.manager.path}/{self.encoded_id}/reviewers"
+        result = self.manager.gitlab.http_get(path, **kwargs)
+        if TYPE_CHECKING:
+            assert not isinstance(result, requests.Response)
+        return result
```
nejch, line 434: Thanks a lot for your contribution @kesmik! 

The reviewers endpoint presumably returns a paginated response (though I admit we'll rarely see more than 20 reviewers, but still). This means we should use `http_list` for this endpoint.

We usually don't write custom `get_*` methods for this, but rather add a manager uses the `ListMixin` as a base. Then people can call `mr.reviewers.list()`, and we have a few other convenience attributes there to control list filters etc. I admit the MR manager violates this with a few custom methods here already.

A recent example of a similar addition was for MR draft notes which you can also list:
https://github.com/python-gitlab/python-gitlab/pull/2728/files

Would you be able to rework your PR to use the manager instead @kesmik? Thanks a lot!

nejch (changes requested):

gitlab/v4/objects/merge_requests.py
kesmik: Hello, thanks for feedback I will check and try to rewrite as you recommending, it seems better way than mine. Initially I thought its not possible since projects already has mergerequest atttibute ;D

kesmik: Ok so I have refactored and moved api call to reviewers manager. I still have a problem for "reviewers" attribute naming.
merge request api (https://docs.gitlab.com/ee/api/merge_requests.html#get-single-mr) returns reviewers in JSON:
```json
"reviewers": [{
    "name": "Miss Monserrate Beier",
    "username": "axel.block",
    "id": 12,
    "state": "active",
    "avatar_url": "http://www.gravatar.com/avatar/46f6f7dc858ada7be1853f7fb96e81da?s=80&d=identicon",
    "web_url": "https://gitlab.example.com/axel.block"
  }],
```

That leads that I can not have manager named "reviewers",  because it will override original reviewers attribute which is supposed to be list.

Also I can not somehow override it with a response of ../reviewers, because format differs..
https://docs.gitlab.com/ee/api/merge_requests.html#get-single-merge-request-reviewers
```json
[
  {
    "user": {
      "id": 1,
      "name": "John Doe1",
      "username": "user1",
      "state": "active",
      "avatar_url": "http://www.gravatar.com/avatar/c922747a93b40d1ea88262bf1aebee62?s=80&d=identicon",
      "web_url": "http://localhost/user1"
    },
    "state": "unreviewed",
    "created_at": "2022-07-27T17:03:27.684Z"
  }
```

Is there anythong I can solve in order to achieve mr.reviewers.list() idea so that:
mr = project.mergerequests.get(1)
reviewers = mr.reviewers.list()
would satisfy:
assert mr.reviewers[0]["name"] == reviewers[0]["user"]["name"]

I think it is not possible with python, and I still need to call that manager somehow diffrently? 

Any help would be appreciated, I would be able to clean this PR

nejch: @kesmik that's a tricky one yes. We have something similar in https://github.com/python-gitlab/python-gitlab/blob/main/gitlab/mixins.py#L749-L761, but since the responses between this endpoint and the attribute are not consistent, I'd say it might make sense to rename the manager attribute.

`mr_reviewers` looks a bit off to me because of the abbreviation, I'd rather expand that to `merge_request_reviewers` in that case. But since we don't actually get just a list of users but also other attributes, I like your other suggestions more.. How about something like `mr.reviewer_details.list()`? Then if GitLab expands the endpoint with other attributes, we're still accurately describing what's being fetched.

But it's definitely a weird endpoint. The `/participants` endpoint is right there as well and does not nest the user details.

kesmik: Yep it is really weird for me as well. So I have rebased all these changes and presuming it is ready to be merged. If anything is needed to be changed let me know I will try my best!

gitlab/v4/objects/reviewers.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class ProjectMergeRequestReviewers(RESTObject):
+    pass
+
+
+class ProjectMergeRequestReviewersManager(ListMixin, RESTManager):
+    _path = "/projects/{project_id}/merge_requests/{mr_iid}/reviewers"
+    _obj_cls = ProjectMergeRequestReviewers
+    _from_parent_attrs = {"project_id": "project_id", "mr_iid": "iid"}
```
nejch, line 17: Just one more nit @kesmik. We tend to use the singular for our objects/managers, so maybe like this:

```suggestion
class ProjectMergeRequestReviewerDetail(RESTObject):
    pass


class ProjectMergeRequestReviewerDetailManager(ListMixin, RESTManager):
    _path = "/projects/{project_id}/merge_requests/{mr_iid}/reviewers"
    _obj_cls = ProjectMergeRequestReviewers
    _from_parent_attrs = {"project_id": "project_id", "mr_iid": "iid"}
```

gitlab/v4/objects/merge_requests.py
nejch: Thanks a lot for the rework, just went through this one last time and just had a small note that I forgot, based on our previous rework, let me know what you think and then we should be good to go!

nejch (changes requested):

gitlab/v4/objects/reviewers.py
kesmik: Hello, @nejch i have reworked as you suggested, one thing that api functional tests failed. I can not reproduce that on my local setup (at least cosnantly), should I worry about that? (it was same with my previous commit, but rerun of piepile helped_

nejch: @kesmik our functional tests are a bit flaky because of resource limitations on GitHub runners, sorry about that. I've restarted the failing job, thanks for your patience! LGTM otherwise:)

nejch (approved):

Issue #2752, opened 2023-12-27
CLI: Reading values from files does not support binary files


JohnVillalovos:
When using the `@` prefix to specify a file that should be read, it will read the file as "utf-8" text. This fails with binary files such as a PNG.

This was discussed in #2751 as pointed out by @trafford-tbe 

Code is: https://github.com/python-gitlab/python-gitlab/blob/d7bdb0257a5587455c3722f65c4a632f24d395be/gitlab/cli.py#L299-L307

PR #2753, opened 2023-12-28
fix(cli): support binary files with `@` notation

JohnVillalovos:
Support binary files being used in the CLI with arguments using the
`@` notation. For example `--avatar @/path/to/avatar.png`

Also explicitly catch the common FileNotFoundError and PermissionError
exceptions.

Remove the bare exception handling. We would rather have the full
traceback of any exceptions that we don't know about and add them
later if needed.

Closes: #2752

### DIFF
--- a/gitlab/cli.py
+++ b/gitlab/cli.py
@@ -1,6 +1,7 @@
 import argparse
 import functools
 import os
+import pathlib
 import re
 import sys
 import textwrap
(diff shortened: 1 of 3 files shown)


gitlab/cli.py
```diff
@@ -299,11 +300,15 @@ def _parse_value(v: Any) -> Any:
...
         try:
-            with open(v[1:], encoding="utf-8") as f:
+            with open(filepath, encoding="utf-8") as f:
                 return f.read()
-        except Exception as e:
-            sys.stderr.write(f"{e}\n")
+        except UnicodeDecodeError:
+            with open(filepath, "rb") as f:
+                return f.read()
+        except (FileNotFoundError, PermissionError) as exc:
```
nejch, line 310: Maybe we should also add `IsADirectoryError` or use the parent OSError here?

gitlab/cli.py
```diff
@@ -299,11 +300,15 @@ def _parse_value(v: Any) -> Any:
     if isinstance(v, str) and v.startswith("@"):
         # If the user-provided value starts with @, we try to read the file
         # path provided after @ as the real value. Exit on any error.
```
nejch, line 302: Not sure if we should keep this now as we handle specific ones:
```suggestion
        # path provided after @ as the real value.
```

nejch (approved):
@JohnVillalovos thanks a lot, looks nice! Maybe we can just add that one additional I/O error that I think we can expect to happen with OS paths?

gitlab/cli.py
JohnVillalovos: Done using `OSError`

nejch (approved):

gitlab/cli.py
nejch: Thanks for the rework!

Issue #2746, opened 2023-12-20
RFE: support the personal_access_tokens/:id/rotate endpoint


svenXY:
## Description of the problem, including code/CLI snippet

gitlab from 16.6 supports to rotate (i.e. refresh lifetime) for personal access tokens:
https://docs.gitlab.com/ee/api/personal_access_tokens.html#rotate-a-personal-access-token

## Expected Behavior

Tokens can be rotated programmatically

## Actual Behavior

the rotate endpoint is not yet supported by python-gitlab

Issue #2589, opened 2023-06-02
Support for the rotate API endpoint (Project Access Token, Personal Access Token, Group Access Token)


mrLarbi:
## Description of the problem, including code/CLI snippet

Since Gitlab 16.0, Gitlab introduced an API endpoint to rotate tokens. Here are the endpoints :

Project Access Token : https://docs.gitlab.com/ee/api/project_access_tokens.html#rotate-a-project-access-token
Personal Access Token : https://docs.gitlab.com/ee/api/personal_access_tokens.html#rotate-a-personal-access-token
Group Access Token : https://docs.gitlab.com/ee/api/group_access_tokens.html#rotate-a-group-access-token

As of now (Apologies if I am mistaken), it seems that these calls are not supported yet by python-gitlab. It would be great to be able to rotate these tokens using python-gitlab.

## Expected Behavior

The ability to rotate Project Access Tokens, Personal Access Tokens and Group Access Tokens with python-gitlab

## Actual Behavior

None.

## Specifications

  - python-gitlab version: latest
  - API version you are using (v3/v4): v4
  - Gitlab server version (or gitlab.com): 16.0 (gitlab.com)

nejch:
Thanks for the report @mrLarbi, this is definitely on my radar as I also want to use it, so will start working on it soon

mrLarbi:
Hello @nejch oh that's nice. Thanks for the info and for the feature

PR #2756, opened 2024-01-03
feat(api): support access token rotate API

nejch:
## Changes

Adds a `RotateMixin` and `ObjectRotateMixin` similar to the delete mixins.

Closes https://github.com/python-gitlab/python-gitlab/issues/2589
Closes https://github.com/python-gitlab/python-gitlab/issues/2746

### Documentation and testing

### DIFF
--- a/docs/gl_objects/personal_access_tokens.rst
+++ b/docs/gl_objects/personal_access_tokens.rst
@@ -52,6 +52,15 @@ Revoke the personal access token currently used::
 
     gl.personal_access_tokens.delete("self")
 
+Rotate a personal access token and retrieve its new value::
+
+    token = gl.personal_access_tokens.get(42, lazy=True)
+    token.rotate()
+    print(token.token)
+    # or directly using a token ID
+    new_token = gl.personal_access_tokens.rotate(42)
+    print(new_token.token)
+
 Create a personal access token for a user (admin only)::
 
     user = gl.users.get(25, lazy=True)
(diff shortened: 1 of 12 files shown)


docs/gl_objects/personal_access_tokens.rst
```diff
@@ -52,6 +52,13 @@ Revoke the personal access token currently used::
 
     gl.personal_access_tokens.delete("self")
 
+Rotate a personal access token and retrieve its new value::
+
+    gl.personal_access_tokens.rotate(42)
```
max-wittig, line 57: isn't that a bit confusing.

In my mind it should be `gl.personal_access_tokens.get(42, lazy=True).rotate()`

nejch: Done by switching the docs to show the object method first.

gitlab/mixins.py
```diff
@@ -653,6 +653,62 @@ def download(
...
+
+        Args:
+            id: ID of the token to rotate
+            **kwargs: Extra options to send to the server (e.g. sudo)
+
+        Raises:
+            GitlabAuthenticationError: If authentication is not correct
+            GitlabRotateError: If the server cannot perform the request
+        """
+        data = {"expires_at": expires_at}
```
JohnVillalovos, line 679: Is it okay for this to be set to `None`?

nejch: I tested this before making the PR and it works, but maybe we shouldn't send null parameters if the user doesn't send them just to keep requests cleaner. I changed this in https://github.com/python-gitlab/python-gitlab/commit/4f27e365de83e92804dbbfcc075f4b3c8a7a15dd, thanks!

Issue #2762, opened 2024-01-11
Support /projects/:id/job_token_scope/allowlist


Sjord:
I would like to access the job token allow list through the Gitlab API. 

Docs: 
- [Get a project’s CI/CD job token inbound allowlist](https://docs.gitlab.com/ee/api/project_job_token_scopes.html#get-a-projects-cicd-job-token-inbound-allowlist)
- [Allow access to your project with a job token](https://docs.gitlab.com/ee/ci/jobs/ci_job_token.html#allow-access-to-your-project-with-a-job-token)

API endpoint: 
```
GET/POST/DELETE /projects/:id/job_token_scope/allowlist
```

I currently have the following manager to access the job token scope allow list:

```
class ProjectJobTokenScopeAllowlistManager(
    gitlab.mixins.ListMixin,
    gitlab.mixins.CreateMixin,
    gitlab.mixins.DeleteMixin,
    gitlab.base.RESTManager,
):
    _path = "/projects/{project_id}/job_token_scope/allowlist"
    _obj_cls = gitlab.v4.objects.Project
    _from_parent_attrs = {"project_id": "id"}
    _create_attrs = gitlab.types.RequiredOptional(required=("target_project_id",))
```

PR #2767, opened 2024-01-19
feat(job_token_scope): support job token access allowlist API

nejch:
## Changes

Closes https://github.com/python-gitlab/python-gitlab/issues/2762.
Adds https://docs.gitlab.com/ee/api/project_job_token_scopes.html#get-a-projects-cicd-job-token-inbound-allowlist

### Documentation and testing

### DIFF
--- a/docs/gl_objects/job_token_scope.rst
+++ b/docs/gl_objects/job_token_scope.rst
@@ -49,3 +49,23 @@ Refresh the current state of job token scope::
     scope.refresh()
     print(scope.inbound_enabled)
     # False
+
+Get a project's CI/CD job token inbound allowlist::
+
+    allowlist = scope.allowlist.list()
+
+Add a project to the project's inbound allowlist::
+
+    allowed_project = scope.allowlist.create({"target_project_id": 42})
+
+Remove a project from the project's inbound allowlist::
+
+    allowed_project.delete()
+    # or directly using a project ID
+    scope.allowlist.delete(42)
+
+.. warning::
+
+   Similar to above, the ID attributes you receive from the create and list
+   APIs are not consistent. To safely retrieve the ID of the allowlisted project
+   regardless of how the object was created, always use its ``.get_id()`` method.
--- a/gitlab/v4/objects/job_token_scope.py
+++ b/gitlab/v4/objects/job_token_scope.py
@@ -2,12 +2,18 @@ from typing import Any, cast
 
 from gitlab.base import RESTManager, RESTObject
 from gitlab.mixins import (
+    CreateMixin,
+    DeleteMixin,
     GetWithoutIdMixin,
+    ListMixin,
+    ObjectDeleteMixin,
     RefreshMixin,
     SaveMixin,
     UpdateMethod,
     UpdateMixin,
 )
+from gitlab.types import RequiredOptional
+
 
 __all__ = [
     "ProjectJobTokenScope",
@@ -18,6 +24,8 @@ __all__ = [
 class ProjectJobTokenScope(RefreshMixin, SaveMixin, RESTObject):
     _id_attr = None
 
+    allowlist: "AllowlistedProjectManager"
+
 
 class ProjectJobTokenScopeManager(GetWithoutIdMixin, UpdateMixin, RESTManager):
     _path = "/projects/{project_id}/job_token_scope"
@@ -27,3 +35,23 @@ class ProjectJobTokenScopeManager(GetWithoutIdMixin, UpdateMixin, RESTManager):
 
     def get(self, **kwargs: Any) -> ProjectJobTokenScope:
         return cast(ProjectJobTokenScope, super().get(**kwargs))
+
+
+class AllowlistedProject(ObjectDeleteMixin, RESTObject):
+    _id_attr = "target_project_id"  # note: only true for create endpoint
+
+    def get_id(self) -> int:
+        """Returns the id of the resource. This override deals with
+        the fact that either an `id` or a `target_project_id` attribute
+        is returned by the server depending on the endpoint called."""
+        try:
+            return cast(int, getattr(self, self._id_attr))
+        except AttributeError:
+            return cast(int, getattr(self, "id"))
+
+
+class AllowlistedProjectManager(ListMixin, CreateMixin, DeleteMixin, RESTManager):
+    _path = "/projects/{project_id}/job_token_scope/allowlist"
+    _obj_cls = AllowlistedProject
+    _from_parent_attrs = {"project_id": "project_id"}
+    _create_attrs = RequiredOptional(required=("target_project_id",))


docs/gl_objects/job_token_scope.rst
```diff
@@ -49,3 +49,24 @@ Refresh the current state of job token scope::
     scope.refresh()
     print(scope.inbound_enabled)
     # False
+
+Get a projec's CI/CD job token inbound allowlist::
+
+    allowlist = scope.allowlist.list()
```
nejch, line 55: Not sure yet about this naming here, but it follows both the endpoint URL and our convention :confused:

gitlab/v4/objects/job_token_scope.py
```diff
@@ -27,3 +35,23 @@ class ProjectJobTokenScopeManager(GetWithoutIdMixin, UpdateMixin, RESTManager):
...
+    _id_attr = "target_project_id"  # note: only true for create endpoint
+
+    def get_id(self) -> int:
+        """Returns the id of the resource. This override deals with
+        the fact that either an `id` or a `target_project_id` attribute
+        is returned by the server depending on the endpoint called."""
+        try:
+            return cast(int, getattr(self, self._id_attr))
+        except AttributeError:
+            return cast(int, getattr(self, "id"))
```
nejch, line 50: GitLab at it again - create response:

```json
{
  "source_project_id": 1,
  "target_project_id": 2
}
```

list response:
```json
{
  "id": 4,
  "description": null,
  "name": "Diaspora Client",
  "name_with_namespace": "Diaspora / Diaspora Client",
  "path": "diaspora-client",
  ...
}
```

JohnVillalovos: I'm kind of thinking for the create to not return anything. As the caller appears to use all the values when creating that will be returned. Not sure if that would simplify things or not.

Sjord:
This currently assumes a job token scope as the parent of the allowlist, but I don't think that is correct. I would expect the allowlist to work without retrieving the job token scope first:

```
# Works, retrieves job_token_scope:
print(project.job_token_scope.get().allowlist.list())

# Doesn't work without job_token_scope:
print(project.job_token_scope.allowlist.list())
```

Test could look something like this:

```
"""
GitLab API: https://docs.gitlab.com/ee/api/project_job_token_scopes.html#get-a-projects-cicd-job-token-inbound-allowlist
"""

import pytest
import responses
from gitlab.v4.objects.job_token_scope import AllowlistedProjectManager

allowlist_small = {"source_project_id":1,"target_project_id":2}

allowlist_large = {
    "id": 2,
    "description": None,
    "name": "testproject",
    "name_with_namespace": "testgroup / testproject",
    "path": "testproject",
    "path_with_namespace": "testgroup/testproject",
    "created_at": "2024-01-21T19:18:38.829Z",
    "default_branch": "main",
    "tag_list": [],
    "topics": [],
    "ssh_url_to_repo": "git@localhost:testgroup/testproject.git",
    "http_url_to_repo": "http://localhost/testgroup/testproject.git",
    "web_url": "http://localhost/testgroup/testproject",
    "readme_url": "http://localhost/testgroup/testproject/-/blob/main/README.md",
    "forks_count": 0,
    "avatar_url": None,
    "star_count": 0,
    "last_activity_at": "2024-01-23T18:22:16.321Z",
    "namespace": {
        "id": 3,
        "name": "testgroup",
        "path": "testgroup",
        "kind": "group",
        "full_path": "testgroup",
        "parent_id": None,
        "avatar_url": None,
        "web_url": "http://localhost/groups/testgroup",
    },
}


@pytest.fixture
def resp_get_allowlist():
    with responses.RequestsMock(assert_all_requests_are_fired=False) as rsps:
        rsps.add(
            method=responses.GET,
            url="http://localhost/api/v4/projects/1/job_token_scope/allowlist",
            json=[allowlist_large],
            content_type="application/json",
            status=200,
        )
        yield rsps


@pytest.fixture
def resp_post_allowlist():
    with responses.RequestsMock(assert_all_requests_are_fired=False) as rsps:
        rsps.add(
            method=responses.POST,
            url="http://localhost/api/v4/projects/1/job_token_scope/allowlist",
            json=allowlist_small,
            content_type="application/json",
            status=201,
        )
        yield rsps


@pytest.fixture
def allowlist(project):
    return project.job_token_scope.allowlist

def test_get_allowlist(allowlist, resp_get_allowlist):
    projects = allowlist.list()
    assert projects[0].get_id() == 2

def test_create_allowlist(allowlist, resp_post_allowlist):
    resp = allowlist.create({"target_project_id": 2})
    assert resp.get_id() == 2
```

TimKnight-DWP:
@nejch FYI if you haven't seen:

Gitlab have just merged an update to GraphQL endpoint to add support for adding groups to the `CI_JOB_TOKEN` allowlist: https://gitlab.com/gitlab-org/gitlab/-/merge_requests/143132

And hopefully will be adding the same support to the REST API: https://gitlab.com/gitlab-org/gitlab/-/issues/435903#note_1761348177

`POST /projects/:id/job_token_scope/allowlist`

nejch:
> @nejch FYI if you haven't seen:
> 
> Gitlab have just merged an update to GraphQL endpoint to add support for adding groups to the `CI_JOB_TOKEN` allowlist: https://gitlab.com/gitlab-org/gitlab/-/merge_requests/143132
> 
> And hopefully will be adding the same support to the REST API: https://gitlab.com/gitlab-org/gitlab/-/issues/435903#note_1761348177
> 
> `POST /projects/:id/job_token_scope/allowlist`

Thanks @TimKnight-DWP! :bow:  I'm pretty sure they will, because their terraform provider relies on REST. I'll get back to this PR this week.

TimKnight-DWP:
Cool - let me know if there's anything I can do to help with this MR @nejch , it's something that's high on our agenda, adding this here and then using from gitlabform to manage our state

I'll take a look through Gitlabs MR and see how they intend to handle "Groups" whether it's taking group id (consistent with how projects are added) or something else. And feedback here.

gitlab/v4/objects/job_token_scope.py
TimKnight-DWP: Convention in REST and so far in python-gitlab would be to return the object that has been created. Means that in future if Gitlab add optional things into the contract which get returned back in the POST, user's would get insight into that

TimKnight-DWP:
Group Allowlist REST API -> https://gitlab.com/gitlab-org/gitlab/-/merge_requests/145069

docs/gl_objects/job_token_scope.rst
TimKnight-DWP: yeah it would definetly be clearer if it were `scope.projects_allowlist` since gitlab have added `groups_allowlist` https://docs.gitlab.com/ee/api/project_job_token_scopes.html#get-a-projects-cicd-job-token-allowlist-of-groups

But `scope.allowlist` does absolutely match the API design doc so it is probably best overall to be consistent with their docs 🤷

nejch:
> @Sjord very true, I would at least want to add support for `lazy=True` for GetWithoutIDMixin first so that this wouldn't require 2 API calls.
> 
> The reason I did it this way was I wanted to follow GitLab's REST API path segment hierarchy. But lately it's making less and less sense to me, so we could also flatten this potentially - I'm not sure, also part of why this is still in draft 😅 thanks for the feedback!

Sorry @Sjord this was already done in another PR as is, but I think we could look into adding a lazy get at least, so we don't require another API call to use the endpoint.

PR #2770, opened 2024-01-21
fix(cli): allow exclusive arguments as optional

### LINKED ISSUE #2769: Exclusive arguments not allowed in CLI
In [gitlab/v4/cli.py](https://github.com/python-gitlab/python-gitlab/blob/72e1aa7f2c2aad739db2f5dd4394165a5b5f2432/gitlab/v4/cli.py#L256-L264), required and optional arguments are added to the argument parser:

```
if action_name == "create":
    for x in mgr_cls._create_attrs.required:
        sub_parser_action.add_argument(
            f"--{x.replace('_', '-')}", required=True
        )
    for x in mgr_cls._create_attrs.optional:
        sub_parser_action.add_argument(
            f"--{x.replace('_', '-')}", required=False
        )
```

Besides *required* and *optional*, the RequiredOptional class can also have an *exclusive* field, and many objects use this to list options not in *required* or *optional*. From [gitlab/v4/objects/invitations.py](https://github.com/python-gitlab/python-gitlab/blob/72e1aa7f2c2aad739db2f5dd4394165a5b5f2432/gitlab/v4/objects/invitations.py#L34-L43):

```
class ProjectInvitationManager(InvitationMixin, RESTManager):
    _path = "/projects/{project_id}/invitations"
    _obj_cls = ProjectInvitation
    _from_parent_attrs = {"project_id": "id"}
        required=("access_level",),
        optional=(
            "expires_at",
            "invite_source",
            "tasks_to_be_done",
            "tasks_project_id",
        ),
        exclusive=("email", "user_id"),
    )
```

Here *email* is an argument defined in *exclusive*, and not allowed on the CLI:

```
$ python -m gitlab --server-url http://172.16.38.146/ project-invitation create --access-level 10 --email myemail --project-id 1
...
__main__.py: error: unrecognized arguments: --email myemail
```

Issue #2738 describes a similar problem.

Sjord:
The CLI takes its arguments from the RequiredOptional, which has three fields: required, optional, and exclusive. In practice, the exclusive options are not defined as either required or optional, and would not be allowed in the CLI. This changes that, so that exclusive options are also added to the argument parser.

Closes #2769

### Documentation and testing

### DIFF
--- a/gitlab/v4/cli.py
+++ b/gitlab/v4/cli.py
@@ -284,9 +284,9 @@ def _populate_sub_parser_by_class(
                         f"--{x.replace('_', '-')}", required=False
                     )
 
-            if mgr_cls._create_attrs.exclusive:
+            if mgr_cls._update_attrs.exclusive:
                 group = sub_parser_action.add_mutually_exclusive_group()
-                for x in mgr_cls._create_attrs.exclusive:
+                for x in mgr_cls._update_attrs.exclusive:
                     group.add_argument(f"--{x.replace('_', '-')}")
 
     if cls.__name__ in cli.custom_actions:
--- a/gitlab/v4/cli.py
+++ b/gitlab/v4/cli.py
@@ -258,10 +258,14 @@ def _populate_sub_parser_by_class(
                 sub_parser_action.add_argument(
                     f"--{x.replace('_', '-')}", required=True
                 )
-            for x in mgr_cls._create_attrs.optional + mgr_cls._create_attrs.exclusive:
+            for x in mgr_cls._create_attrs.optional:
                 sub_parser_action.add_argument(
                     f"--{x.replace('_', '-')}", required=False
                 )
+            if mgr_cls._create_attrs.exclusive:
+                group = sub_parser_action.add_mutually_exclusive_group()
+                for x in mgr_cls._create_attrs.exclusive:
+                    group.add_argument(f"--{x.replace('_', '-')}")
 
         if action_name == "update":
             if cls._id_attr is not None:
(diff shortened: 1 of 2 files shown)


JohnVillalovos:
Would it make more sense to create a mutually exclusive group and add the exclusive args to that?

https://docs.python.org/3/library/argparse.html#mutual-exclusion

Sjord:
I've changed the code to create a mutually exclusive group.

I think the data model of RequiredOptional is not ideal. Whether options are exclusive should be perpendicular to whether they are required or optional. Also, it would be possible to have multiple exclusive groups. Currently this is not reflected in the RequiredOptional.exclusive property. It only allows one group, and it is not clear whether this group is required or not.

Ideally, you would have something lilke this:

```
RequiredOptional(
    required=("foo", exclusive("user_id", "user_email"), exclusive("project_slug", "project_id")),
    optional=("bar", exclusive("cn", "filter"))
)
```

But adding these exclusive options to the CLI argument parser at least makes it functioning again, if not perfect.

gitlab/v4/cli.py
```diff
@@ -280,6 +284,11 @@ def _populate_sub_parser_by_class(
                         f"--{x.replace('_', '-')}", required=False
                     )
 
+            if mgr_cls._create_attrs.exclusive:
+                group = sub_parser_action.add_mutually_exclusive_group()
+                for x in mgr_cls._create_attrs.exclusive:
```
JohnVillalovos, line 289: Should this be '_update_attrs`?

Also could the commits be squashed down to one commit?

Sjord: Indeed. This was a copy/paste error. I bit the bullet and added a unit test to prevent mistakes like this.

JohnVillalovos (approved):
Thank you very much for this.

PR #2771, opened 2024-01-21
feat(api): allow updating protected branches

### LINKED ISSUE #2390: Allow update of protected branches

Sjord:
Closes #2390

## Changes

    p_b.allow_force_push = True
    p_b.save()

### Documentation and testing

### DIFF
--- a/docs/gl_objects/protected_branches.rst
+++ b/docs/gl_objects/protected_branches.rst
@@ -27,6 +27,11 @@ Get a single protected branch::
 
     p_branch = project.protectedbranches.get('main')
 
+Update a protected branch:
+
+    p_branch.allow_force_push = True
+    p_branch.save()
+
 Create a protected branch::
 
     p_branch = project.protectedbranches.create({
--- a/gitlab/v4/objects/branches.py
+++ b/gitlab/v4/objects/branches.py
@@ -1,7 +1,13 @@
 from typing import Any, cast, Union
 
 from gitlab.base import RESTManager, RESTObject
-from gitlab.mixins import NoUpdateMixin, ObjectDeleteMixin
+from gitlab.mixins import (
+    CRUDMixin,
+    NoUpdateMixin,
+    ObjectDeleteMixin,
+    SaveMixin,
+    UpdateMethod,
+)
 from gitlab.types import RequiredOptional
 
 __all__ = [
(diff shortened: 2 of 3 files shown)


Sjord:
I don't understand why the integration test fails. It works on my local instance, and the code seems to make sense:

```
p_b.allow_force_push = True
p_b.save()

p_b = project.protectedbranches.get("*-stable")
assert p_b.allow_force_push
```

JohnVillalovos:
> I don't understand why the integration test fails. It works on my local instance, and the code seems to make sense:
> 
> ```
> p_b.allow_force_push = True
> p_b.save()
> 
> p_b = project.protectedbranches.get("*-stable")
> assert p_b.allow_force_push
> ```

Just a guess, but it may need to use the `wait_for_sidekiq()` function. As the GitLab server may still be processing the `save()` asynchronously while doing the `get()`.

`functional/api/test_merge_requests.py` has usage of `wait_for_sidekiq()`

nejch:
Sadly our functional tests still run against 15.4 (yes, I know!) because we had issues with flaky and failing tests when upgrading. This was introduced in 15.6 so this is sort of expected. We need to take the time to fix that soon I guess.

I would suggest adding a skip or xfail (I forget if we still have a decorator for gitlab versions - that would be ideal. At least we have a `gitlab_version` fixture so you could conditionally test those few lines as well.

nejch (approved):
Thanks @Sjord and sorry for the delays! I'll go ahead and merge this now, but I'll squash the commits if that's ok with you as the other commits aren't user-facing fixes. Thanks again!

nickbroon:
Is this expected to also work with `protectedbranch.allowed_to_merge` or `protectedbranch.merge_access_levels` ?

Sjord:
I made this to support `allow_force_push`, and if anything else worked that would be a bonus. I expect `allowed_to_push` to work.

It is indeed a pain that the API expects different information than it returns. I have been bitten by this before, and I think I saw an issue about this in some other context, but I can't find it right now. Unfortunately, there is not really a clear solution for this.

Perhaps it's better to create a new issue for your problem. This pull request is already closed so comments here will get out of sight.

PR #2779, opened 2024-01-28
ci: prevent merge-commits in a PR

JohnVillalovos:
Have the CI prevent merge-commits being part of a PR

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -41,3 +41,14 @@ jobs:
         run: tox -e isort -- --check
       - name: Run pylint Python code static checker (https://github.com/PyCQA/pylint)
         run: tox -e pylint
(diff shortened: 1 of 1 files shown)


JohnVillalovos:
Goal of this is to block PRs like: https://github.com/python-gitlab/python-gitlab/pull/2780

JohnVillalovos:
This should not be required as we have "Prevent Linear History" set in the settings for the protected branch `main`

nejch:
Thanks @JohnVillalovos that was my thinking as well. I guess that was merged before we enabled the feature right?

JohnVillalovos:
> Thanks @JohnVillalovos that was my thinking as well. I guess that was merged before we enabled the feature right?

I guess so. There is no history (that I know of) where we can see when settings where changed for the project on GitHub.

PR #2784, opened 2024-02-04
fix(api): update manual job status when playing it

1ace:
## Changes

When `.play()`ing a `ProjectJob`, its status would not be updated from `manual` to `pending`/`running`/etc.

Fix this by updating the object with the data returned by the api.

### Documentation and testing

- [x] Documentation in the matching [docs section](https://github.com/python-gitlab/python-gitlab/tree/main/docs): not needed IMO, this is just a fix to make the behaviour of this method match the others and match the user expectation
- [ ] [Unit tests](https://github.com/python-gitlab/python-gitlab/tree/main/tests/unit) and/or [functional tests](https://github.com/python-gitlab/python-gitlab/tree/main/tests/functional): not added, it might be good but I wanted to post this first, and I'll add the tests if this change is accepted and tests are requested :)

### DIFF
--- a/gitlab/v4/objects/jobs.py
+++ b/gitlab/v4/objects/jobs.py
@@ -65,7 +65,10 @@ class ProjectJob(RefreshMixin, RESTObject):
             GitlabJobPlayError: If the job could not be triggered
         """
         path = f"{self.manager.path}/{self.encoded_id}/play"
-        self.manager.gitlab.http_post(path, **kwargs)
+        result = self.manager.gitlab.http_post(path, **kwargs)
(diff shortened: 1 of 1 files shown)


nejch (approved):
Thanks a lot @1ace looks good and this totally makes sense!

The coverage complains but this is really because we didn't have existing tests for the method. If you like, you could add a unit test similar to what we have in https://github.com/python-gitlab/python-gitlab/blob/main/tests/unit/objects/test_pipeline_schedules.py#L95-L99.

Otherwise I'm happy to merge this as-is as well.

PR #2494, opened 2023-02-13
test: adapt functional tests for GitLab 15.8

JohnVillalovos:

PR #2790, opened 2024-02-08
test: update integration tests to run using Gitlab 16

TimKnight-DWP:
- Updates tests to use Gitlab 16 image
- Updates test to handle changes to the underlying APIs
- Adds programmatic date times to Access Token tests

### DIFF
--- a/tests/functional/api/test_boards.py
+++ b/tests/functional/api/test_boards.py
@@ -5,7 +5,6 @@ def test_project_boards(project):
     board = project.boards.get(board.id)
 
     project.boards.delete(board.id)
-    assert not project.boards.list()
 
 
 def test_group_boards(group):
@@ -15,4 +14,3 @@ def test_group_boards(group):
     board = group.boards.get(board.id)
 
     group.boards.delete(board.id)
-    assert not group.boards.list()
--- a/tests/functional/api/test_deploy_tokens.py
+++ b/tests/functional/api/test_deploy_tokens.py
@@ -1,8 +1,8 @@
-from datetime import date
+import datetime
 
 
 def test_project_deploy_tokens(gl, project):
-    today = date.today().isoformat()
+    today = datetime.date.today().isoformat()
     deploy_token = project.deploytokens.create(
         {
             "name": "foo",
@@ -21,8 +21,6 @@ def test_project_deploy_tokens(gl, project):
     assert deploy_token.username == "bar"
 
     deploy_token.delete()
-    assert deploy_token not in project.deploytokens.list()
-    assert deploy_token not in gl.deploytokens.list()
 
 
 def test_group_deploy_tokens(gl, group):
@@ -41,5 +39,3 @@ def test_group_deploy_tokens(gl, group):
     assert deploy_token.scopes == ["read_registry"]
 
     deploy_token.delete()
-    assert deploy_token not in group.deploytokens.list()
-    assert deploy_token not in gl.deploytokens.list()
--- a/tests/functional/api/test_deploy_tokens.py
+++ b/tests/functional/api/test_deploy_tokens.py
@@ -1,9 +1,13 @@
+from datetime import date
+
+
 def test_project_deploy_tokens(gl, project):
+    today = date.today().isoformat()
     deploy_token = project.deploytokens.create(
         {
             "name": "foo",
             "username": "bar",
-            "expires_at": "2022-01-01",
+            "expires_at": today,
             "scopes": ["read_registry"],
         }
     )
@@ -12,7 +16,7 @@ def test_project_deploy_tokens(gl, project):
 
     deploy_token = project.deploytokens.get(deploy_token.id)
     assert deploy_token.name == "foo"
-    assert deploy_token.expires_at == "2022-01-01T00:00:00.000Z"
+    assert deploy_token.expires_at == today + "T00:00:00.000Z"
     assert deploy_token.scopes == ["read_registry"]
     assert deploy_token.username == "bar"
 
--- a/tests/functional/api/test_deploy_tokens.py
+++ b/tests/functional/api/test_deploy_tokens.py
@@ -16,7 +16,7 @@ def test_project_deploy_tokens(gl, project):
 
     deploy_token = project.deploytokens.get(deploy_token.id)
     assert deploy_token.name == "foo"
-    assert deploy_token.expires_at == today + "T00:00:00.000Z"
+    assert deploy_token.expires_at == f"{today}T00:00:00.000Z"
     assert deploy_token.scopes == ["read_registry"]
     assert deploy_token.username == "bar"
 
--- a/tests/functional/api/test_merge_requests.py
+++ b/tests/functional/api/test_merge_requests.py
@@ -1,3 +1,4 @@
+import datetime
 import time
 
 import pytest
@@ -72,8 +73,6 @@ def test_merge_request_discussion(project):
     assert discussion.attributes["notes"][-1]["body"] == "updated body"
 
     note_from_get.delete()
-    discussion = mr.discussions.get(discussion.id)
-    assert len(discussion.attributes["notes"]) == 1
 
 
 def test_merge_request_labels(project):
@@ -164,27 +163,33 @@ def test_project_merge_request_approval_rules(group, project):
     assert approval_rules[0].approvals_required == 2
 
     approval_rules[0].delete()
-    ars = project.approvalrules.list(get_all=True)
-    assert len(ars) == 0
 
 
-def test_merge_request_reset_approvals(gitlab_url, project, wait_for_sidekiq):
-    bot = project.access_tokens.create({"name": "bot", "scopes": ["api"]})
+def test_merge_request_reset_approvals(gitlab_url, project):
+    today = datetime.date.today()
+    future_date = today + datetime.timedelta(days=4)
+    bot = project.access_tokens.create(
+        {"name": "bot", "scopes": ["api"], "expires_at": future_date.isoformat()}
+    )
+
     bot_gitlab = gitlab.Gitlab(gitlab_url, private_token=bot.token)
     bot_project = bot_gitlab.projects.get(project.id, lazy=True)
 
-    wait_for_sidekiq(timeout=60)
+    # Pause to let GL catch up (happens on hosted too, sometimes takes a while for server to be ready to merge)
+    time.sleep(5)
+
     mr = bot_project.mergerequests.list()[0]  # type: ignore[index]
+
     assert mr.reset_approvals()
 
 
-def test_cancel_merge_when_pipeline_succeeds(
-    project, merge_request_with_pipeline, wait_for_sidekiq
-):
-    wait_for_sidekiq(timeout=60)
+def test_cancel_merge_when_pipeline_succeeds(project, merge_request_with_pipeline):
+    # Pause to let GL catch up (happens on hosted too, sometimes takes a while for server to be ready to merge)
+    time.sleep(5)
     # Set to merge when the pipeline succeeds, which should never happen
     merge_request_with_pipeline.merge(merge_when_pipeline_succeeds=True)
-    wait_for_sidekiq(timeout=60)
+    # Pause to let GL catch up (happens on hosted too, sometimes takes a while for server to be ready to merge)
+    time.sleep(5)
 
     mr = project.mergerequests.get(merge_request_with_pipeline.iid)
     assert mr.merged_at is None
@@ -193,9 +198,10 @@ def test_cancel_merge_when_pipeline_succeeds(
     assert cancel == {"status": "success"}
 
 
-def test_merge_request_merge(project, merge_request, wait_for_sidekiq):
+def test_merge_request_merge(project, merge_request):
     merge_request.merge()
-    wait_for_sidekiq(timeout=60)
+    # Pause to let GL catch up (happens on hosted too, sometimes takes a while for server to be ready to merge)
+    time.sleep(5)
 
     mr = project.mergerequests.get(merge_request.iid)
     assert mr.merged_at is not None
@@ -205,15 +211,14 @@ def test_merge_request_merge(project, merge_request, wait_for_sidekiq):
         mr.merge()
 
 
-def test_merge_request_should_remove_source_branch(
-    project, merge_request, wait_for_sidekiq
-) -> None:
+def test_merge_request_should_remove_source_branch(project, merge_request) -> None:
     """Test to ensure
     https://github.com/python-gitlab/python-gitlab/issues/1120 is fixed.
     Bug reported that they could not use 'should_remove_source_branch' in
     mr.merge() call"""
     merge_request.merge(should_remove_source_branch=True)
-    wait_for_sidekiq(timeout=60)
+    # Pause to let GL catch up (happens on hosted too, sometimes takes a while for server to be ready to merge)
+    time.sleep(5)
 
     # Wait until it is merged
     mr = None
@@ -227,7 +232,8 @@ def test_merge_request_should_remove_source_branch(
     assert mr is not None
     assert mr.merged_at is not None
     time.sleep(0.5)
-    wait_for_sidekiq(timeout=60)
+    # Pause to let GL catch up (happens on hosted too, sometimes takes a while for server to be ready to merge)
+    time.sleep(5)
 
     # Ensure we can NOT get the MR branch
     with pytest.raises(gitlab.exceptions.GitlabGetError):
@@ -240,9 +246,7 @@ def test_merge_request_should_remove_source_branch(
         print("result:", pprint.pformat(result))
 
 
-def test_merge_request_large_commit_message(
-    project, merge_request, wait_for_sidekiq
-) -> None:
+def test_merge_request_large_commit_message(project, merge_request) -> None:
     """Test to ensure https://github.com/python-gitlab/python-gitlab/issues/1452
     is fixed.
     Bug reported that very long 'merge_commit_message' in mr.merge() would
@@ -255,7 +259,8 @@ def test_merge_request_large_commit_message(
         merge_commit_message=merge_commit_message, should_remove_source_branch=False
     )
 
-    wait_for_sidekiq(timeout=60)
+    # Pause to let GL catch up (happens on hosted too, sometimes takes a while for server to be ready to merge)
+    time.sleep(5)
 
     # Wait until it is merged
     mr = None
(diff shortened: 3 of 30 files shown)


tests/functional/conftest.py
```diff
@@ -212,21 +211,32 @@ def _check(
...
     """
-    Return a helper function to wait until there are no busy sidekiq processes.
+    Return a helper function to wait until there are spare sidekiq processes.
+
+    Because not all Groups can be deleted immediately in GL 16, we will likely have busy Sidekiq jobs
+    set aside for their deletion in 1 days time.
 
     Use this with asserts for slow tasks (group/project/user creation/deletion).
+
+    { ..., 'labels': [], 'concurrency': 20, 'busy': 3}
```
nejch, line 221: @TimKnight-DWP I think I introduced this sidekiq helper years ago thinking it's a bit smarter than simple sleep calls. But I wonder if we should just delete this and instead have simpler helpers that wait for a resource to be in a certain state. E.g. wait for branch to exist/not exist, wait for MR to have certain status etc.

I'm just not sure how much work it would take, but it might be simpler.

nejch (commented):
Thanks a lot for working on this @TimKnight-DWP!

I think this is based on some initial work by @JohnVillalovos so maybe cherry-picking or adding a co-authored-by trailer to include him might be good here. :+1: 

Also I started something related to bulk imports in https://github.com/python-gitlab/python-gitlab/pull/2494, so if you think you're close to getting this done, just mark the import/export tests as xfail and we can tackle them separately as a follow-up, just to unblock this upgrade. Let me know if that works!

tests/functional/conftest.py
TimKnight-DWP: I'll have a look, yeah the ones it mainly seems to matter for now are Merging MRs and I'm having to put more specific logic

I have John's MR open in a tab to guide me when I start on the `api` tests 👍

TimKnight-DWP:
@nejch - something has clearly changed in user_agent which was causing failures when we did snipper.user_agent_details()

I can't see exactly what value that has in our tests, I will add one as a separate test marked as skipped - as I can't see anything in the changelogs as to what changed

nejch:
@TimKnight-DWP if it only affects an individual test feel free to also skip it and we deal with it separately. GitLab often has subtle breaking changes in the API that are not announced - hence also some of the failures here.

TimKnight-DWP:
Have done, separated it out into a distinct test so the feature can be validated when there's chance to come bacl. I also see a fgew open MRs around which look related to some of the remaining API failures

nejch:
@TimKnight-DWP also if we are religiously asserting on something that is maybe not so relevant, we could change the tests a bit to really only check what matters. Especially with deleting resources, maybe we just test something simpler if the async deletions are a pain.

TimKnight-DWP:
Sounds like a good shout, deletions definitely seem to be one of the biggest remaining pains, and always need a bit of time to sleep. 

I'm currently putting some logging in to see if there's an obvious "Scheduled for deletion" flag we can use rather than asserting thing is not there (maybe an if there -> assert scheduled for deletion, else assert has been deleted)

TimKnight-DWP:
I think it may also. be worth validating the response from Delete APIs such as: https://docs.gitlab.com/ee/api/users.html#user-deletion

If GL returns a 204, we don't need to check that the thing has been immediately deleted, the api has told us yes, if it takes a while thats Gitlab functionality rather than python-gitlab functionaity

nejch:
> I think it may also. be worth validating the response from Delete APIs such as: https://docs.gitlab.com/ee/api/users.html#user-deletion
> 
> If GL returns a 204, we don't need to check that the thing has been immediately deleted, the api has told us yes, if it takes a while thats Gitlab functionality rather than python-gitlab functionaity.
> 
> If my hunch is correct now, I'll do that for these failing users tests and then take a look through and swap over

Ah that's a great idea @TimKnight-DWP. We shouldn't be testing GitLab behavior, just that it receives the right requests. :+1: we could probably do that with a lot of other tests too, now that I think of it.

TimKnight-DWP:
@nejch I think with this last push that should be tests Green 🤞 

I'll go through this aft and update all delete tests at least to just verify the API responded 2xx (i.e. didn't throw an exception) and then squash my commits

Should be good to go then 👍

tests/functional/api/test_boards.py
```diff
@@ -1,11 +1,18 @@
...
     board = project.boards.get(board.id)
 
-    project.boards.delete(board.id)
-    assert not project.boards.list()
+    # Raises:
+    #   GitlabDeleteError: If the server cannot perform the request (the return code is not 2xx)
+    try:
+        project.boards.delete(board.id)
+    except GitlabDeleteError as error:
+        assert False, f"project.boards.delete raise error {error}"
```
nejch, line 15: @TimKnight-DWP I wonder if we really even need to wrap and re-raise this. The traceback we get from exposing the error might be more helpful than just a failing assert, since the error would be unexpected. WDYT? Then we just simplify and call the delete method.

This would apply to most other deletes as well, and the diff here would be a lot smaller too and it'll be easier for me to review. Sorry for the back and forth :bow:

tests/functional/api/test_merge_requests.py
```diff
@@ -163,28 +169,43 @@ def test_project_merge_request_approval_rules(group, project):
...
+    future_date = today + timedelta(days=4)
+    bot = project.access_tokens.create(
+        {"name": "bot", "scopes": ["api"], "expires_at": future_date.isoformat()}
+    )
+    logging.info("Created token for Bot")
 
-def test_merge_request_reset_approvals(gitlab_url, project, wait_for_sidekiq):
-    bot = project.access_tokens.create({"name": "bot", "scopes": ["api"]})
     bot_gitlab = gitlab.Gitlab(gitlab_url, private_token=bot.token)
+    logging.info("Created bot")
```
nejch, line 189: I think we can delete the logs now if they were mostly used for debugging - same for any other logging added inside the actual test cases.

```suggestion
```

tests/functional/api/test_boards.py
TimKnight-DWP: @nejch 🤔 if we did that it'd come up as an "Error" rather than a "Fail" 

I'm happy to go through and do that

tests/functional/api/test_merge_requests.py
TimKnight-DWP: Will go through my changelist and reduce logging and any other unnecessary changes

TimKnight-DWP:
@nejch - would you be able to give this another review please? I think it'll also unblock adding integration tests for the CI Job Token PR

TimKnight-DWP:
Showing as failing because -0.04% tests 😿

nejch:
> Showing as failing because -0.04% tests 😿

That's ok @TimKnight-DWP I can override that (and fix the coverage in the code that is the root cause :sweat_smile:). I was focusing on something else at the moment, sorry!

TimKnight-DWP:
@JohnVillalovos @nejch @max-wittig @bufferoverflow - this one is good to go in and then will be able to look at getting changes to modify the `job_token_scope` via python gitlab in, as GL have now released the APIs in a main build 👍

tests/functional/api/test_deploy_tokens.py
```diff
@@ -1,9 +1,13 @@
+from datetime import date
```
JohnVillalovos, line 1: nit: I'm personally not a fan of this type of import.

I prefer: `import datetime` and then just use `datetime.date.today()`

tests/functional/cli/test_cli_resource_access_tokens.py
```diff
@@ -1,4 +1,4 @@
-import pytest
+from datetime import date
```
JohnVillalovos, line 1: Same nit as above.

tests/functional/cli/test_cli_users.py
```diff
@@ -1,3 +1,6 @@
+from datetime import date
```
JohnVillalovos, line 1: Same nit as above

tests/functional/cli/test_cli_v4.py
```diff
@@ -1,5 +1,9 @@
+import logging
 import os
 import time
+from datetime import date
```
JohnVillalovos, line 3: Same nit as above

tests/functional/conftest.py
```diff
@@ -4,6 +4,7 @@
 import tempfile
 import time
 import uuid
+from datetime import date
```
JohnVillalovos, line 7: Same nit as above

tests/functional/conftest.py
```diff
@@ -212,21 +211,32 @@ def _check(
...
+            total_concurrency = 0
             for process in processes:
-                if process["busy"]:
-                    busy = True
-            if not busy:
+                busy_processes += process["busy"]
+                total_concurrency += process["concurrency"]
+
+            logging.info(f"Busy processes {busy_processes}")
+            #  If we have space concurrency in the process, continue
```
JohnVillalovos: nit: s/space/spare/

tests/functional/conftest.py
```diff
@@ -212,21 +211,32 @@ def _check(
...
+            if busy_processes < total_concurrency:
```
JohnVillalovos: I'll be honest this looks concerning to me. As I think this will increase our chances of things randomly breaking because the process we wanted to finish is not finished.

tests/functional/helpers.py
```diff
@@ -40,15 +36,37 @@ def safe_delete(
         except gitlab.exceptions.GitlabGetError:
             return
 
+        logging.info(f"Deleting object type {type(object)}")
+
         if index:
             logging.info(f"Attempt {index + 1} to delete {object!r}.")
         try:
-            if hard_delete:
+            if type(object) is gitlab.v4.objects.User:
```
JohnVillalovos, line 42: nit: Any reason not to use `if isinstance(object, gitlab.v4.objects.User)` ?

tests/functional/helpers.py
```diff
@@ -40,15 +36,37 @@ def safe_delete(
...
+                # You can't use this option if the selected user is the sole owner of any groups
+                # Use `hard_delete=True` or a 'Ghost User' may be created.
+                # https://docs.gitlab.com/ee/api/users.html#user-deletion
                 object.delete(hard_delete=True)
+                if index > 1:
+                    # If User is the sole owner of any group it won't be deleted,
+                    # which combined with parents group never immediately deleting in GL 16
+                    # we shouldn't cause test to fail if it still exists
+                    return
+            elif type(object) is gitlab.v4.objects.Project:
```
JohnVillalovos, line 52: nit: Any reason not to use `if isinstance(object, gitlab.v4.objects.Project)` ?

tests/functional/helpers.py
```diff
@@ -40,15 +36,37 @@ def safe_delete(
...
+                # Immediately delete rather than waiting for at least 1day
+                # https://docs.gitlab.com/ee/api/projects.html#delete-project
+                object.delete(permanently_remove=True)
+                pass
             else:
+                # We only attempt to delete parent groups to prevent dangling sub-groups
+                # However parent groups can only be deleted on a delay in Gl 16
+                # https://docs.gitlab.com/ee/api/projects.html#delete-group
```
JohnVillalovos, line 60: This link doesn't seem correct.

tests/functional/helpers.py
```diff
@@ -40,15 +36,37 @@ def safe_delete(
...
                 object.delete()
         except gitlab.exceptions.GitlabDeleteError:
-            logging.info(f"{object!r} already deleted.")
+            logging.info(f"{object!r} already deleted or scheduled for deletion.")
+            if type(object) is gitlab.v4.objects.Group:
```
JohnVillalovos, line 64: nit: Any reason not to use `if isinstance(object, gitlab.v4.objects.Group)` ?

tests/functional/api/test_merge_requests.py
```diff
@@ -1,4 +1,5 @@
 import time
+from datetime import date, timedelta
```
JohnVillalovos, line 2: Same nit as before.

JohnVillalovos (commented):
Left some comments and some nits.

I'm not sure I like the changes to `wait_for_sidekiq` and also how it isn't called anymore in a lot of places and just replaced with `sleep`. I worry that this will make the tests more likely to have random failures.

Also seems like we no longer test that `delete` actually deletes what we want deleted, we are testing that we can call delete and no error is raised though.

But on the positive we now have tests working with Gitlab 16, which is great.

So overall I'm not sure. I would like to hear what @nejch thinks.

TimKnight-DWP:
With regards to no longer calling "delete", gitlab have significantly changed how they handle deletes, it happens way more asynchronously, and thus is really hard to actually account for in the tests without making them even more brittle (in the case of some deletes it happens after 1 day regardless). So all we can do is assert we successfully called the delete, and we have to trust Gitlab will act on the DELETE Rest request, rather than also try and test their functionality

tests/functional/helpers.py
TimKnight-DWP: nope - my lack of embedded python knowledge when I first did the MR 👍

TimKnight-DWP:
It's been a month, but if I recall `wait_for_sidekiq` did a lot of checking assuming one free worker means we're good, but with the delete changes, gitlab stores those deletes in a Q, which is represented by checking how much concurrency sidekiq has used vs total. But when doing so I discovered we never got close to the limits due to threading on the container, so the check became somewhat redundant. The worker always showed as "busy" because it was waiting to perform deletes, but it had spare concurrency to perform other tasks.

The sleeps could probably go away to be honest, they are mainly there to help out the stability, sometimes the container/worker are a bit slower than expected so can randomly fail a test

TimKnight-DWP:
@JohnVillalovos @nejch - any additional thoughts on this PR?

I think it's important for us to get the tests updated to 16, but also that unlocks some later PRs to add additional, new functionality only present in 16+

TimKnight-DWP:
Hi @nejch @JohnVillalovos @max-wittig @bufferoverflow - sorry to tag all of you and to nudge this along, but with Gitlab 17 now around the corner, I'd like to get this in so the testing suite isn't > 2 versions behind the release, and to unblock some of the JOB_TOKEN_SCOPE proposed PRs, as people may need that functionality when GitLab start to deprecate the existing Token logic in 17.x

JohnVillalovos:
I'm traveling at the moment, so don't have time to do the review.

tests/functional/cli/test_cli_v4.py
```diff
@@ -252,28 +230,42 @@ def test_create_merge_request(gitlab_cli, project):
...
     }
     project.files.create(file_data)
-    time.sleep(2)
-    wait_for_sidekiq(timeout=60)
-
-    cmd = [
+    # Pause to let GL catch up (happens on hosted too, sometimes takes a while for server to be ready to merge)
+    time.sleep(30)
+
+    if mr.detailed_merge_status == "not_approved":
```
nejch, line 246: Just so we're in sync here: this doesn't check the current approved status at this point in time, but whatever the state was at the time you created the MR object when fetching the data from the server with `list()`.

But maybe that's expected, as I'm not entirely sure what the purpose of this `if` check is? Perhaps dealing with pre-existing MRs if there's some race conditions in the project tests.

tests/functional/conftest.py
```diff
@@ -413,6 +388,10 @@ def _make_merge_request(*, source_branch: str, create_pipeline: bool = False):
                 "commit_message": "New commit in new branch",
             }
         )
+
+        # Helps with Debugging why MRs fail to merge resulting in 405 from downstream tests
+        project.approvalrules.list()
```
nejch, line 393: This doesn't actually print anything out though, right? Feels a bit wrong to just leave it here, unless I missed something. I think we can remove it if it was used to debug during this PR. or do we want to assert here that we really have approval rules?

nejch (commented):
Thanks again @TimKnight-DWP for all the work and patience. I have a few more last comments and then I'd say we can get this merged just to unblock thins, and improve on the hardcoded sleeps later.

If you could also reword the commit messages to only say `chore` / `test` as this is not really user-facing changes that'd be great, but if it's too much hassle we can just squash the commits here in the end.

nejch:
> Left some comments and some nits.
> 
> I'm not sure I like the changes to `wait_for_sidekiq` and also how it isn't called anymore in a lot of places and just replaced with `sleep`. I worry that this will make the tests more likely to have random failures.
> 
> Also seems like we no longer test that `delete` actually deletes what we want deleted, we are testing that we can call delete and no error is raised though.
> 
> But on the positive we now have tests working with Gitlab 16, which is great.
> 
> So overall I'm not sure. I would like to hear what @nejch thinks.

@JohnVillalovos I'm also not sure how I feel about removing _all_ the deletes, but in the end it won't be very reliable with current GitLab versions if we keep them. We might lose some breaking changes in the API this way, but this would be an upstream issue.

As for `wait_for_sidekiq`, I also think that was maybe not very reliable. If we find a more generic way, we can also point these tests against external gitlab instances in the future.

So I'd say lets try and get this done finally :sweat_smile: we can still open follow-up issues to improve on it.

tests/functional/api/test_deploy_tokens.py
```diff
@@ -12,13 +16,11 @@ def test_project_deploy_tokens(gl, project):
 
     deploy_token = project.deploytokens.get(deploy_token.id)
     assert deploy_token.name == "foo"
-    assert deploy_token.expires_at == "2022-01-01T00:00:00.000Z"
+    assert deploy_token.expires_at == today + "T00:00:00.000Z"
```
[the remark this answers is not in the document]
TimKnight-DWP: @nejch yeah and/or I'll happily squash down to one prior to merge 👍

TimKnight-DWP:
I can reproduce that smoke test failure locally on `main` branch too, will attempt to determine what change has caused it

TimKnight-DWP:
@nejch updated with your comments, but something odd seems to be happening with the github-actions, and sudden unexpected failures in the smoke tests

nejch:
@TimKnight-DWP the new smoke test failures seem unrelated as they also appear in https://github.com/python-gitlab/python-gitlab/pull/2833, might be something new in the build dependencies that change how the package is built, will need to check.

nejch:
> Thanks @TimKnight-DWP! The new smoke test failures seem unrelated as they also appear in #2833, might be something new in the build dependencies that change how the package is built, will need to check.

The culprit seems to be https://github.com/pypa/build/issues/770. I'll adapt the tests.

nejch (approved):
Thanks again @TimKnight-DWP and @JohnVillalovos - I'll keep the commits as I see you've amended the messages and it's a pretty big PR :+1: 

Let's see if we can improve the `sleep()` calls in a follow-up. And thank for your patience :sweat_smile:

PR #2791, opened 2024-02-08
Fix argument docstring in artifacts

maxnoe:
### Documentation and testing

simple docs change


nejch (approved):
Nice catch @maxnoe, thanks for this!

Issue #2795, opened 2024-02-12
Missing parameters in iteration search


cristianocasella:
## Description of the problem, including code/CLI snippet

In https://github.com/python-gitlab/python-gitlab/blob/main/gitlab/v4/objects/iterations.py the iteration implementation let you to filter the list with a search parameter.

According to the API doc (https://docs.gitlab.com/ee/api/iterations.html) by default this parameter filter the title of the iteration.

If the iteration is created by a cadence, title and description will be empty, so you will not be able to search for a specific iteration.

## Expected Behavior

I would like to pass the additional parameter, documented in the API, "in" to select where the search will be performed, and search in "cadence_title" for cadence derivated iterations.

## Actual Behavior

The search is only performed on the iteration title, that is empty when built from a cadence.

## Specifications

  - python-gitlab version: 4.4.0
  - API version you are using (v3/v4): v4
  - Gitlab server version (or gitlab.com): 16.8.1 EE

## Update

After a first iteration we decided to extend the scope to all the new parameters available

nejch:
Thanks for the report and the PR @cristianocasella! 

Note that the list filter attribute is only used for the CLI. You can pass arbitrary arguments to the list methods when using the API already.

The problem with the `in` keyword is that it collides with python's `in` keyword so you'll always get a SyntaxError. For this reason, we have to use the `query_parameters` argument:

```
group.iterations.list(query_parameters={"in": "title"}) 
```

cristianocasella:
> Thanks for the report and the PR @cristianocasella!
> 
> Note that the list filter attribute is only used for the CLI. You can pass arbitrary arguments to the list methods when using the API already.
> 
> The problem with the `in` keyword is that it collides with python's `in` keyword so you'll always get a SyntaxError. For this reason, we have to use the `query_parameters` argument:
> 
> ```
> group.iterations.list(query_parameters={"in": "title"}) 
> ```
> 
> See https://python-gitlab.readthedocs.io/en/stable/api-usage.html#managers for more details.

Really thanks for your support and hint.

I decided to extend the PR to all the parameters currently available, to save other people time I also added your help above into the project documentation.

Let me know if it works for you

nejch:
Done in https://github.com/python-gitlab/python-gitlab/pull/2796.

PR #2796, opened 2024-02-12
feat(api):  add additional parameter to project/group iteration search

cristianocasella:
Relates to https://github.com/python-gitlab/python-gitlab/issues/2795

## Changes

Added "in" parameter to define where te search should be performed in a iteration list.

### Documentation and testing

- [X] [Unit tests](https://github.com/python-gitlab/python-gitlab/tree/main/tests/unit) and/or [functional tests](https://github.com/python-gitlab/python-gitlab/tree/main/tests/functional)

### DIFF
--- a/docs/gl_objects/iterations.rst
+++ b/docs/gl_objects/iterations.rst
@@ -31,3 +31,13 @@ List iterations for a project's ancestor groups::
 List iterations for a group::
 
     iterations = group.iterations.list()
+
+Unavailable filters or keyword conflicts::
+    
+    In case you are trying to pass a parameter that collides with a python
+    keyword (i.e. `in`) or with python-gitlab's internal arguments, you'll have
+    to use the `query_parameters` argument:
+
+    ```
+    group.iterations.list(query_parameters={"in": "title"}) 
+    ```
--- a/gitlab/v4/objects/iterations.py
+++ b/gitlab/v4/objects/iterations.py
@@ -16,11 +16,11 @@ class GroupIterationManager(ListMixin, RESTManager):
     _path = "/groups/{group_id}/iterations"
     _obj_cls = GroupIteration
     _from_parent_attrs = {"group_id": "id"}
-    _list_filters = ("state", "search", "include_ancestors")
+    _list_filters = ("state", "search", "in", "include_ancestors")
 
 
 class ProjectIterationManager(ListMixin, RESTManager):
     _path = "/projects/{project_id}/iterations"
     _obj_cls = GroupIteration
     _from_parent_attrs = {"project_id": "id"}
-    _list_filters = ("state", "search", "include_ancestors")
+    _list_filters = ("state", "search", "in", "include_ancestors")
--- a/gitlab/v4/objects/iterations.py
+++ b/gitlab/v4/objects/iterations.py
@@ -1,3 +1,4 @@
+from gitlab import types
 from gitlab.base import RESTManager, RESTObject
 from gitlab.mixins import ListMixin
 
@@ -16,11 +17,37 @@ class GroupIterationManager(ListMixin, RESTManager):
     _path = "/groups/{group_id}/iterations"
     _obj_cls = GroupIteration
     _from_parent_attrs = {"group_id": "id"}
-    _list_filters = ("state", "search", "in", "include_ancestors")
+    # When using the API, the "in" keyword collides with python's "in" keyword
+    # raising a SyntaxError.
+    # For this reason, we have to use the query_parameters argument:
+    # group.iterations.list(query_parameters={"in": "title"})
+    _list_filters = (
+        "include_ancestors",
+        "include_descendants",
+        "in",
+        "search",
+        "state",
+        "updated_after",
+        "updated_before",
+    )
+    _types = {"scopes": types.ArrayAttribute}
 
 
 class ProjectIterationManager(ListMixin, RESTManager):
     _path = "/projects/{project_id}/iterations"
     _obj_cls = GroupIteration
     _from_parent_attrs = {"project_id": "id"}
-    _list_filters = ("state", "search", "in", "include_ancestors")
+    # When using the API, the "in" keyword collides with python's "in" keyword
+    # raising a SyntaxError.
+    # For this reason, we have to use the query_parameters argument:
+    # project.iterations.list(query_parameters={"in": "title"})
+    _list_filters = (
+        "include_ancestors",
+        "include_descendants",
+        "in",
+        "search",
+        "state",
+        "updated_after",
+        "updated_before",
+    )
+    _types = {"in": types.ArrayAttribute}
--- a/gitlab/v4/objects/iterations.py
+++ b/gitlab/v4/objects/iterations.py
@@ -30,7 +30,7 @@ class GroupIterationManager(ListMixin, RESTManager):
         "updated_after",
         "updated_before",
     )
-    _types = {"scopes": types.ArrayAttribute}
+    _types = {"in": types.ArrayAttribute}
 
 
 class ProjectIterationManager(ListMixin, RESTManager):


gitlab/v4/objects/iterations.py
```diff
@@ -16,11 +16,11 @@ class GroupIterationManager(ListMixin, RESTManager):
     _path = "/groups/{group_id}/iterations"
     _obj_cls = GroupIteration
     _from_parent_attrs = {"group_id": "id"}
-    _list_filters = ("state", "search", "include_ancestors")
+    _list_filters = ("state", "search", "in", "include_ancestors")
```
nejch, line 19: LGTM @cristianocasella, but while we're doing this, maybe it makes sense to add the remaining missing filters? See https://docs.gitlab.com/ee/api/iterations.html.

nejch (changes requested):

gitlab/v4/objects/iterations.py
cristianocasella: that's a great idea, let me update this

docs/gl_objects/iterations.rst
```diff
@@ -31,3 +31,13 @@ List iterations for a project's ancestor groups::
 List iterations for a group::
 
     iterations = group.iterations.list()
+
+Unavailable filters or filters conflicts::
+    
+    In case you are trying to pass a parameter that is not supported or
+    collides with a python keyword (i.e. `in`) you have to use the
+    `query_parameters` argument:
```
nejch, line 39: There aren't really "unsupported" filters (since we just pass whatever the user provides to the API), only those that clash with other keywords. I'd maybe phrase it like this:
```suggestion
    In case you are trying to pass a parameter that collides with a python
    keyword (i.e. `in`) or with python-gitlab's internal arguments, you'll have
    to use the `query_parameters` argument:
```

gitlab/v4/objects/iterations.py
```diff
@@ -16,11 +16,31 @@ class GroupIterationManager(ListMixin, RESTManager):
     _path = "/groups/{group_id}/iterations"
     _obj_cls = GroupIteration
     _from_parent_attrs = {"group_id": "id"}
-    _list_filters = ("state", "search", "include_ancestors")
+    # The "in" keyword collides with python's "in" keyword raising a SyntaxError.
+    # For this reason, we have to use the query_parameters argument:
+    # group.iterations.list(query_parameters={"in": "title"})
+    _list_filters = (
+        "include_ancestors",
+        "include_descendants",
```
nejch, line 24:
```suggestion
        "include_descendants",
        "in",
```

gitlab/v4/objects/iterations.py
```diff
@@ -16,11 +16,31 @@ class GroupIterationManager(ListMixin, RESTManager):
...
 
 
 class ProjectIterationManager(ListMixin, RESTManager):
     _path = "/projects/{project_id}/iterations"
     _obj_cls = GroupIteration
     _from_parent_attrs = {"project_id": "id"}
-    _list_filters = ("state", "search", "include_ancestors")
+    # The "in" keyword collides with python's "in" keyword raising a SyntaxError.
+    # For this reason, we have to use the query_parameters argument:
+    # group.iterations.list(query_parameters={"in": "title"})
```
nejch, line 38:
```suggestion
    # project.iterations.list(query_parameters={"in": "title"})
```

gitlab/v4/objects/iterations.py
```diff
@@ -16,11 +16,31 @@ class GroupIterationManager(ListMixin, RESTManager):
...
+    _list_filters = (
+        "include_ancestors",
+        "include_descendants",
```
nejch, line 41:
```suggestion
        "include_descendants",
        "in",
```

gitlab/v4/objects/iterations.py
```diff
@@ -16,11 +16,31 @@ class GroupIterationManager(ListMixin, RESTManager):
     _path = "/groups/{group_id}/iterations"
     _obj_cls = GroupIteration
     _from_parent_attrs = {"group_id": "id"}
-    _list_filters = ("state", "search", "include_ancestors")
+    # The "in" keyword collides with python's "in" keyword raising a SyntaxError.
```
nejch, line 19: I should probably clarify this - this is only an issue with direct API usage. We supply the `in` argument differently to the CLI. So this actually works for the CLI and we can add the `in` keyword here. Sorry for the confusion.

```suggestion
    # When using the API, the "in" keyword collides with python's "in" keyword raising a SyntaxError.
```

gitlab/v4/objects/iterations.py
```diff
@@ -16,11 +16,31 @@ class GroupIterationManager(ListMixin, RESTManager):
...
+        "updated_before",
+    )
 
 
 class ProjectIterationManager(ListMixin, RESTManager):
     _path = "/projects/{project_id}/iterations"
     _obj_cls = GroupIteration
     _from_parent_attrs = {"project_id": "id"}
-    _list_filters = ("state", "search", "include_ancestors")
+    # The "in" keyword collides with python's "in" keyword raising a SyntaxError.
```
nejch, line 36:
```suggestion
    # When using the API, the "in" keyword collides with python's "in" keyword raising a SyntaxError.
```

nejch (changes requested):
Thanks a lot for the rework and documenting this @cristianocasella! Just a few more nits from my side, let me know if you have any questions.

gitlab/v4/objects/iterations.py
cristianocasella: from the test that I've conducted on my istance, the right query here would be:

```suggestion
    # project.iterations.list(query_parameters={"in[]": "title"})
```

does it sounds good for you?

cristianocasella: @nejch thanks for all the feedback, sorry, I totally forgot the CLI ^^

a last point here and then I can make a final squash before the merge

cristianocasella: I'll lint also in that step

nejch: > from the test that I've conducted on my istance, the right query here would be:
> 
> does it sounds good for you?

@cristianocasella I think we shouldn't manually add brackets- instead we should make `in` an `ArrayAttribute` and python-gitlab will do that on its own already. But we can do this in a follow-up MR as well.

docs/gl_objects/iterations.rst
```diff
@@ -31,3 +31,13 @@ List iterations for a project's ancestor groups::
 List iterations for a group::
 
     iterations = group.iterations.list()
+
+Unavailable filters or filters conflicts::
```
nejch, line 35:
```suggestion
Unavailable filters or keyword conflicts::
```

gitlab/v4/objects/iterations.py
```diff
@@ -16,11 +16,35 @@ class GroupIterationManager(ListMixin, RESTManager):
...
+    # group.iterations.list(query_parameters={"in": "title"})
+    _list_filters = (
+        "include_ancestors",
+        "include_descendants",
+        "in",
+        "search",
+        "state",
+        "updated_after",
+        "updated_before",
+    )
```
nejch, line 31: Same as the other one, will also need `from gitlab import types import`

```suggestion
    )
    _types = {"scopes": types.ArrayAttribute}
```

gitlab/v4/objects/iterations.py
```diff
@@ -16,11 +16,35 @@ class GroupIterationManager(ListMixin, RESTManager):
...
```
nejch, line 50: This will also require `from gitlab import types` at the top:

```suggestion
    )
    _types = {"in": types.ArrayAttribute}
```

nejch (commented):
@cristianocasella I've left one final round of comments/nits that also need an additional import based on your latest comment. Let me know and if you can update this then we can go ahead and merge!

cristianocasella:
Going for a final rebase to include all the last adds

gitlab/v4/objects/iterations.py
```diff
@@ -16,11 +17,37 @@ class GroupIterationManager(ListMixin, RESTManager):
...
+    _types = {"scopes": types.ArrayAttribute}
```
nejch, line 33: Sorry, I did a copy/paste derp while suggesting this. I'll apply this now and squash the PR.

```suggestion
    _types = {"in": types.ArrayAttribute}
```

nejch (approved):

Issue #2799, opened 2024-02-15
refs call of commit causes wrong warning


cweber-dbs:
## Description of the problem, including code/CLI snippet

Fetching the refs for a commit causes a warning. The warning is missleading due to the fact, that it provides some information for the list() method.

```
    def _fetch_commit_tag(self, commit: ProjectCommit) -> str:
        tags = commit.refs(type="tag")
```

## Expected Behavior

The warning should inform about the refs() method and not the list() method.

## Actual Behavior

The following warning is shown:
```
UserWarning: Calling a list() method without specifying get_all=True or iterator=True will return a maximum of 20 items. Your query returned 20 of xxx items. See https://python-gitlab.readthedocs.io/en/v4.4.0/api-usage.html#pagination for more details. If this was done intentionally, then this warning can be supressed by adding the argumentget_all=Falseto thelist()call.
```

## Specifications

python-gitlab version: 4.4.0
API version you are using (v3/v4): ?
Gitlab server version (or gitlab.com): v16.8.1-ee

cweber-dbs:
Using `tags = commit.refs("tag", get_all=True)` works fine.

The documentation is misleading as well due to the fact, that only the first 20 matches are returned.
<img width="841" alt="image" src="https://github.com/python-gitlab/python-gitlab/assets/38488152/85c81198-43bb-43e5-9b5f-d71f5834cef6">

nejch:
> Using `tags = commit.refs("tag", get_all=True)` works fine.
> 
> The documentation is misleading as well due to the fact, that only the first 20 matches are returned.

@cweber-dbs this documentation refers to all types of references (both tags and branches), not necessarily all pages returned by the API, which is covered by the pagination docs. But this is also why we have this runtime warning as it's often confusing in other endpoints.

@JohnVillalovos I guess we could tweak this a bit to also supply the calling method instead if it calls `http_list` internally, but sometimes I would almost prefer the interface on this to use a manager like we do with most endpoints, so people would call `commit.refs.list()`. WDYT?

JohnVillalovos:
Not an answer to this. Did the warning not indicate the file and line number in the user code that caused the warning? I thought it was supposed to do that.

cweber-dbs:
Nope, only the reference to utils.py line 199 is included:

[...]

PR #2802, opened 2024-02-16
fix: user.warn() to show correct filename of issue

JohnVillalovos:
Previously would only go to the 2nd level of the stack for determining the offending filename and line number. When it should be showing the first filename outside of the python-gitlab source code. As we want it to show the warning for the user of the libraries code.

Update test to show it works as expected.

This is sort of related to #2799

### DIFF
--- a/gitlab/utils.py
+++ b/gitlab/utils.py
@@ -191,8 +191,7 @@ def warn(
     stacklevel = 1
     warning_from = ""
     for stacklevel, frame in enumerate(reversed(stack), start=1):
-        if stacklevel == 2:
-            warning_from = f" (python-gitlab: {frame.filename}:{frame.lineno})"
+        warning_from = f" (python-gitlab: {frame.filename}:{frame.lineno})"
         frame_dir = str(pathlib.Path(frame.filename).parent.resolve())
         if not frame_dir.startswith(str(pg_dir)):
             break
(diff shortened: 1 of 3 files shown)


gitlab/utils.py
```diff
@@ -191,8 +191,7 @@ def warn(
     stacklevel = 1
     warning_from = ""
     for stacklevel, frame in enumerate(reversed(stack), start=1):
-        if stacklevel == 2:
-            warning_from = f" (python-gitlab: {frame.filename}:{frame.lineno})"
+        warning_from = f" (python-gitlab: {frame.filename}:{frame.lineno})"
```
nejch, line 194: Should we also adapt the message passed in here from https://github.com/python-gitlab/python-gitlab/blob/1e37b8fb1109fbfe01b908c4663971876e31c18c/gitlab/client.py#L957-L962?

Since this happens in `http_list` and when called form custom methods like `refs()`, the `list()` method is never involved so I'm wondering if we need something a bit more generic to avoid the confusion reported.

But I would also still need to see what OP's error looked like, maybe there is already a hint as well.

nejch (changes requested):
Thanks @JohnVillalovos looks good! One question about the method name in the message.

gitlab/utils.py
JohnVillalovos: @nejch That is a separate issue I think. As that is from the caller of this method. I'm just trying to fix this so the correct user's calling filename appears.

Agree that we should figure out if we want a better message.

nejch (approved):

PR #2803, opened 2024-02-17
chore: add py312 & py313 to tox environment list

JohnVillalovos:
Even though there isn't a Python 3.13 at this time, this is done for
the future.  tox is already configured to just warn about missing
Python versions, but not fail if they don't exist.


nejch:
@JohnVillalovos sorry I merged the other PR which has this as well, do we still want to add anything here?

JohnVillalovos:
No need. Other pr had it

PR #2814, opened 2024-03-01
fix: Consider scope an ArrayAttribute in PipelineJobManager

gagallo7:
## Changes

List query params like 'scope' were not being handled correctly. This change ensures multiple values are appended with '[]', resulting in the correct URL structure.

---

Background:
If one queries for pipeline jobs with `scope=["failed", "success"]`

One gets:
```
GET /api/v4/projects/176/pipelines/1113028/jobs?scope=success&scope=failed
```

But it is supposed to get:
```
GET /api/v4/projects/176/pipelines/1113028/jobs?scope[]=success&scope[]=failed
```

The current version only considers the last element of the list argument.

### Documentation and testing

### DIFF
--- a/gitlab/v4/objects/pipelines.py
+++ b/gitlab/v4/objects/pipelines.py
@@ -17,7 +17,7 @@ from gitlab.mixins import (
     SaveMixin,
     UpdateMixin,
 )
-from gitlab.types import RequiredOptional
+from gitlab.types import ArrayAttribute, RequiredOptional
 
 __all__ = [
     "ProjectMergeRequestPipeline",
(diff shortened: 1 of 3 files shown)


nejch:
Thanks for the PR @gagallo7!  

I was under the impression we were solving this for specific endpoints that expect this format using our ArrayAttribute (https://github.com/python-gitlab/python-gitlab/blob/main/gitlab/types.py#L72-L82). Has GitLab switched to using this on all list attributes? Because that was not always the case, sometimes they also expect comma-separated values.

gagallo7:
Sorry for the delay, I was sick.

Thanks for your comments, actually the solution would be to consider `scope` an `ArrayAttribute` for the `pipeline/jobs` endpoint.

JohnVillalovos (approved):
This looks good to me. Would like @nejc to do a quick double check on it.

From my understanding if a single value is provided for `scope` it will still work. And the `ArrayAttribute` class has specific support for a single string value.

Issue #2819, opened 2024-03-12
const module not exposed


valentingregoire:
## Description of the problem, including code/CLI snippet

The `const` module is not exported, therefore, one cannot access any const properly.

## Expected Behavior

One can simply access the consts

## Actual Behavior

Right now, it is possible to access the consts with `gitlab.const.XYZ`, but this is not the most ideal situation I believe.

## Specifications

  - python-gitlab version: 4.4.0
  - API version you are using (v3/v4): v4
  - Gitlab server version (or gitlab.com): most recent

nejch:
As discussed in #2820 we'll keep this as is - it's expected that the constants aren't exposed at the root package level.

PR #2820, opened 2024-03-12
fix: Added const module to base namespace.

valentingregoire:
### Documentation and testing

### DIFF
--- a/gitlab/__init__.py
+++ b/gitlab/__init__.py
@@ -32,7 +32,6 @@ from gitlab.exceptions import *  # noqa: F401,F403
 
 warnings.filterwarnings("default", category=DeprecationWarning, module="^gitlab")
 
-
 __all__ = [
     "__author__",
     "__copyright__",
@@ -44,3 +43,4 @@ __all__ = [
     "GitlabList",
 ]
 __all__.extend(gitlab.exceptions.__all__)
+__all__.extend(gitlab.const.__all__)


valentingregoire:
See #2819.

nejch:
Thanks for the PR @valentingregoire! AFAIK this was an intentional change at some point to enforce namespacing the constants, as the number of package-level exports kept growing. Or am I forgetting something, do you remember @JohnVillalovos?

In any case I don't mind this but perhaps we should avoid exporting the deprecated strings that have been replaced by enums.

JohnVillalovos:
That is what I remember @nejch

valentingregoire:
@nejch, @JohnVillalovos, I see that this could be the intention, indeed. Would it then make sense to add the `const` module to the root `__init__` [here](https://github.com/python-gitlab/python-gitlab/blob/main/gitlab/__init__.py#L36)? Else it is not properly exported, no?

nejch:
> @nejch, @JohnVillalovos, I see that this could be the intention, indeed. Would it then make sense to add the `const` module to the root `__init__` [here](https://github.com/python-gitlab/python-gitlab/blob/main/gitlab/__init__.py#L36)? Else it is not properly exported, no?

@valentingregoire modules are not exported in `__all__` like that at least to my knowledge. Either we re-export them again at top-level or keep the current namespaced imports.

You simply do one of the folllowing:
```python
# module import with full namespace
import gitlab.const
gitlab.const.AccessLevel.ADMIN
<AccessLevel.ADMIN: 60>

# module import
from gitlab import const
const.AccessLevel.ADMIN
<AccessLevel.ADMIN: 60>

# class import
from gitlab.const import AccessLevel
AccessLevel.ADMIN
<AccessLevel.ADMIN: 60>
```

Does this not work for you?

valentingregoire:
@nejch, yes that works perfectly fine. It's just that it's a bit inconsistent with the exceptions that are added to the global namespace, where the constants are not. For example:

```python
from gitlab import GitlabGetError
from gitlab.const import MAINTAINER_ACCESS
```

This feels a bit weird and inconsistent. Maybe the correct solution is either to leave everything as is, or to put the exceptions out of the global namespace. This would be another major change though...

nejch:
> @nejch, yes that works perfectly fine. It's just that it's a bit inconsistent with the exceptions that are added to the global namespace, where the constants are not. For example:
> 
> ```python
> from gitlab import GitlabGetError
> from gitlab.const import MAINTAINER_ACCESS
> ```
> 
> This feels a bit weird and inconsistent. Maybe the correct solution is either to leave everything as is, or to put the exceptions out of the global namespace. This would be another major change though...

@valentingregoire a lot of python libraries seem to export exceptions at the root package level as a convention, but true if we want to be really consistent here we would leave them out in just `gitlab.exceptions` the same way that e.g. `boto3` does. If we do this I would first go with a deprecation like we did for the constants. I'll leave it to you to decide as you were the first to report this :)

nejch:
@valentingregoire I'm leaning towards leaving this as is, please consider https://github.com/python-gitlab/python-gitlab/pull/2820#issuecomment-2120020809 if you disagree and would like to go ahead and reopen this PR :bow:

valentingregoire:
I agree, we can leave it as is 👍.

Op wo 29 mei 2024 22:20 schreef Nejc Habjan ***@***.***>:

PR #2829, opened 2024-03-25
chore: update `mypy` to 1.9.0 and resolve one issue

JohnVillalovos:
mypy 1.9.0 flagged one issue in the code. Resolve the issue. Current unit tests already check that a `None` value returns `text/plain`. So function is still working as expected.

### Documentation and testing


nejch (approved):
Thanks for taking care of this @JohnVillalovos!

PR #2834, opened 2024-04-01
build: Add "--no-cache-dir" to pip commands in Dockerfile

xuhdev:
This would not leave cache files in the built docker image.

Additionally, also only build the wheel in the build phase.

On my machine, before this PR, size is 74845395; after this PR, size is
72617713.


nejch (approved):
Thanks @xuhdev that makes sense!

Issue #2812, opened 2024-02-29
Support group-level service-accounts


sven-hergenhahn-dmtech:
Since 16.1, the gitlab-api supports the creation of group-level service-accounts and group-level service-account tokens (They are not the same as group access tokens!)
So far they can not yet be created via the Web UI.

It would be great to have this feature added to python-gitlab.

The gitlab API documentation is here: https://docs.gitlab.com/ee/api/groups.html#service-accounts

nejch:
Service accounts were added in #2851 - we can track adding the tokens here though.

svenXY:
I don't think that service-account-tokens have been added meanwhile, so this is still something I'd like to see in python-gitlab

svenXY:
I don't think that service-account-tokens have been added meanwhile, so this is still something I'd like to see in python-gitlab

svenXY:
I don't think that service-account-tokens have been added meanwhile, so this is still something I'd like to see in python-gitlab

syphernl:
There is [an open PR](https://github.com/python-gitlab/python-gitlab/pull/3109) that adds this functionality, but it has been stalled for a long time now..

sven-hergenhahn-dmtech:
I don't think that service-account-tokens have been added meanwhile, so this is still something I'd like to see in python-gitlab

syphernl:
There has been some activity in the PR 2 months ago but since then nothing happened: https://github.com/python-gitlab/python-gitlab/pull/3109

While this issue may be "stale", this functionality is still something multiple users would like to see.

syphernl:
There is still a PR waiting for review..

https://github.com/python-gitlab/python-gitlab/pull/3361

PR #2851, opened 2024-05-03
feat(api): add support for gitlab service account

SachinKSingh28:
Adds support for creating group service account: https://docs.gitlab.com/ee/api/groups.html#service-accounts

https://github.com/python-gitlab/python-gitlab/issues/2812


TimKnight-DWP (approved):

docs/gl_objects/groups.rst
```diff
@@ -458,3 +458,24 @@ Edit group push rules::
 Delete group push rules::
 
     pr.delete()
+
+Group Service Account
+======================
```
nejch, line 463: I think pre-commit will complain if we don't match the RST title here:

```suggestion
Group Service Account
=====================
```

gitlab/v4/objects/service_accounts.py
```diff
@@ -0,0 +1,24 @@
...
+from gitlab.types import RequiredOptional
+
+__all__ = [
+    "GroupServiceAccount",
+    "GroupServiceAccountManager",
+]
+
+
+class GroupServiceAccount(RESTObject):
+    _id_attr = None
```
nejch, line 12: Are you sure this should be `None`? The `id` is returned on create and if we end up supporting PAT endpoints for service accounts, the endpoint should support the `id` attribute I believe. Does that make sense?

https://docs.gitlab.com/ee/api/groups.html#rotate-a-personal-access-token-for-service-account-user

tests/functional/api/test_groups.py
```diff
@@ -308,3 +308,12 @@ def test_group_saml_group_links(group):
...
     )
+
+
+@pytest.mark.gitlab_premium
+def test_group_service_account(group):
+    hook = group.service_accounts.create(
+        {"name": "gitlab-service-account", "username": "gitlab-service-account"}
+    )
+    assert hook.name == "gitlab-service-account"
+    assert hook.username == "gitlab-service-account"
```
nejch, line 319: Not sure if this was intentional but this seems like it would be a better fit:

```suggestion
    account = group.service_accounts.create(
        {"name": "gitlab-service-account", "username": "gitlab-service-account"}
    )
    assert account.name == "gitlab-service-account"
    assert account.username == "gitlab-service-account"
```

tests/unit/objects/test_groups.py
```diff
@@ -108,6 +113,13 @@ def resp_groups():
             content_type="application/json",
             status=200,
         )
+        rsps.add(
+            method=responses.POST,
+            url="http://localhost/api/v4/groups/1/service_accounts",
+            json=service_account_content,
+            content_type="application/json",
+            status=200,
+        )
```
nejch, line 122: I don't think we need this one here - it should be fine to use the `resp_create_group_service_account` fixture you have below.

tests/unit/objects/test_groups.py
```diff
@@ -466,3 +491,9 @@ def test_delete_saml_group_link(group, resp_delete_saml_group_link):
 
 def test_group_restore(group, resp_restore_group):
     group.restore()
+
+
+def test_create_group_service_account(group, resp_create_group_service_account):
+    group.service_accounts.create(
+        {"name": "gitlab-service-account", "username": "gitlab-service-account"}
+    )
```
nejch, line 499: I know it's not testing much, but if we already mock the response content should we also assert on this? Same as in the `test_create_group` test above.

```suggestion
    account = group.service_accounts.create(
        {"name": "gitlab-service-account", "username": "gitlab-service-account"}
    )
    assert account.name == "gitlab-service-account"
    assert account.username == "gitlab-service-account"
```

nejch (changes requested):
Thanks a lot @SachinKSingh28! Looks good. I just have a few comments :bow:

docs/gl_objects/groups.rst
SachinKSingh28: Thanks! I sort it out. I did run pre-commit on this change and it didn't pick this issue. I'll verify that too.

gitlab/v4/objects/service_accounts.py
SachinKSingh28: will sort that.

tests/functional/api/test_groups.py
SachinKSingh28: Not intended. Copy pasta.

SachinKSingh28:
@nejch addressed your comments! Ready for a review again.

gitlab/v4/objects/service_accounts.py
```diff
@@ -0,0 +1,21 @@
+from gitlab.base import RESTManager, RESTObject
+from gitlab.mixins import CreateMixin
+from gitlab.types import RequiredOptional
+
+__all__ = ["GroupServiceAccount", "GroupServiceAccountManager"]
+
+
+class GroupServiceAccount(RESTObject):
+    _id_attr = "id"
```
nejch, line 9: We seem to still have the redundant id here:

```suggestion
    pass
```

gitlab/v4/objects/service_accounts.py
```diff
@@ -0,0 +1,21 @@
...
+
+class GroupServiceAccountManager(CreateMixin, RESTManager):
+    _path = "/groups/{group_id}/service_accounts"
+    _obj_cls = GroupServiceAccount
+    _from_parent_attrs = {"group_id": "id"}
+    _create_attrs = RequiredOptional(
+        optional=(
+            "name",
+            "username",
+        ),
```
nejch, line 20: Forgot to add this in the previous round - I think this fits in a single line for now:

```suggestion
        optional=("name", "username"),
```

nejch:
Thanks again @SachinKSingh28, I applied a nit left over from the first round myself so we don't block this anymore. I'll squash and merge :+1:

PR #2857, opened 2024-05-07
feat(api): more usernames support for MR approvals

jarodwilson:
I don't think commit a2b8c8ccfb5d went far enough to enable usernames support. We create and edit a lot of approval rules based on an external service (similar to CODE_OWNERS), but only have the usernames available, and currently, have to look up each user to get their user ID to populate user_ids for .set_approvers() calls. Would very much like to skip the lookup and just send the usernames, which this change should allow.

See: https://docs.gitlab.com/ee/api/merge_request_approvals.html#create-project-level-rule

## Changes

- Add a new approver_usernames argument to .set_approvers()
- Pass new arg to GitLab as a list of usernames via their usernames arg

### Documentation and testing

This is not yet tested or documented, wanted to make sure I was on the right track first.

### DIFF
--- a/gitlab/v4/objects/merge_request_approvals.py
+++ b/gitlab/v4/objects/merge_request_approvals.py
@@ -89,6 +89,8 @@ class ProjectMergeRequestApprovalManager(GetWithoutIdMixin, UpdateMixin, RESTMan
         approver_ids: Optional[List[int]] = None,
         approver_group_ids: Optional[List[int]] = None,
         approval_rule_name: str = "name",
+        *,
+        approver_usernames: Optional[List[str]] = None,
         **kwargs: Any,
     ) -> RESTObject:
         """Change MR-level allowed approvers and approver groups.
@@ -104,6 +106,7 @@ class ProjectMergeRequestApprovalManager(GetWithoutIdMixin, UpdateMixin, RESTMan
         """
         approver_ids = approver_ids or []
         approver_group_ids = approver_group_ids or []
+        approver_usernames = approver_usernames or []
 
         data = {
             "name": approval_rule_name,
@@ -111,6 +114,7 @@ class ProjectMergeRequestApprovalManager(GetWithoutIdMixin, UpdateMixin, RESTMan
             "rule_type": "regular",
             "user_ids": approver_ids,
             "group_ids": approver_group_ids,
+            "usernames": approver_usernames,
         }
         if TYPE_CHECKING:
             assert self._parent is not None
@@ -124,6 +128,7 @@ class ProjectMergeRequestApprovalManager(GetWithoutIdMixin, UpdateMixin, RESTMan
                 ar.user_ids = data["user_ids"]
                 ar.approvals_required = data["approvals_required"]
                 ar.group_ids = data["group_ids"]
+                ar.usernames = data["usernames"]
                 ar.save()
                 return ar
         # if there was no rule matching the rule name, create a new one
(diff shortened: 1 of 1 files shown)


gitlab/v4/objects/merge_request_approvals.py
```diff
@@ -87,6 +87,7 @@ def set_approvers(
         self,
         approvals_required: int,
         approver_ids: Optional[List[int]] = None,
+        approver_usernames: Optional[List[str]] = None,
```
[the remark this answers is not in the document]
jarodwilson: Just pushed what I think you're after -- not terribly familiar with keyword required arguments.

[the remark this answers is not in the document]
JohnVillalovos: > Just pushed what I think you're after -- not terribly familiar with keyword required arguments.

It just means that when the caller wants to use that argument they will have to do `set_approvers(approver_usernames=["foo", "blah"])`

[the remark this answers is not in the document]
jarodwilson: Trying again w/some slight tweaks since lint was unhappy

gitlab/v4/objects/merge_request_approvals.py
```diff
@@ -89,6 +89,8 @@ def set_approvers(
         approver_ids: Optional[List[int]] = None,
         approver_group_ids: Optional[List[int]] = None,
         approval_rule_name: str = "name",
+        *,
+        approver_usernames: List[str] = [],
```
JohnVillalovos, line 93: No, this is not correct Python usage. There are articles [1] out there why not use `[]` as a default value.

Do, like was done in the very first revision:

```
approver_usernames: Optional[List[str]] = None
```

[1] https://nikos7am.com/posts/mutable-default-arguments/

JohnVillalovos (changes requested):

gitlab/v4/objects/merge_request_approvals.py
jarodwilson: Done

gitlab/v4/objects/merge_request_approvals.py
```diff
@@ -111,6 +113,7 @@ def set_approvers(
             "rule_type": "regular",
             "user_ids": approver_ids,
             "group_ids": approver_group_ids,
+            "usernames": approver_usernames,
```
JohnVillalovos, line 116: Before around line 108 had:

```
approver_usernames = approver_usernames or []
```

We still need that. Otherwise `None` will be sent in the request.

JohnVillalovos (changes requested):

gitlab/v4/objects/merge_request_approvals.py
jarodwilson: Bah, right, adding that back -- had took it out thinking a default  of [] would cover it, forgot to add it back.

JohnVillalovos (approved):

JohnVillalovos:
Looks good to me, but hoping @nejch can take a quick look in case I missed something.

Thank you @jarodwilson

JohnVillalovos:
@jarodwilson Looks like this will need to be rebased and conflicts resolved after merging in that PR you wanted to get merged.

jarodwilson:
PR updated based on latest main w/the .save() changes merged.

JohnVillalovos (approved):

JohnVillalovos:
Thanks @jarodwilson

PR #2860, opened 2024-05-10
chore(cli): on the CLI help show the API endpoint of resources

JohnVillalovos:
This makes it easier for people to map CLI command names to the API.

Looks like this:
```
    $ gitlab --help
    <snip>
                            The GitLab resource to manipulate.
        application         API endpoint: /applications
        application-appearance
                            API endpoint: /application/appearance
        application-settings
                            API endpoint: /application/settings
        application-statistics
                            API endpoint: /application/statistics
    <snip>
```

### DIFF
--- a/tests/functional/cli/test_cli.py
+++ b/tests/functional/cli/test_cli.py
@@ -34,12 +34,6 @@ def test_config_error_with_help_prints_help(script_runner):
     assert ret.returncode == 0
 
 
-def test_global_help_prints_resources_vertically(script_runner):
-    ret = script_runner.run(["gitlab", "--help"])
-    assert """resource:\n  application\n  application-appearance\n""" in ret.stdout
-    assert ret.returncode == 0
-
-
 def test_resource_help_prints_actions_vertically(script_runner):
     ret = script_runner.run(["gitlab", "project", "--help"])
     assert """action:\n  list\n  get""" in ret.stdout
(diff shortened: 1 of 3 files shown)


TimKnight-DWP (approved):

JohnVillalovos:
@nejch If you have some time to review, it would be appreciated. Nothing too complicated, mostly just a quality of life improvements for the users (and me) 😊

tests/functional/cli/test_cli.py
```diff
@@ -34,12 +34,6 @@ def test_config_error_with_help_prints_help(script_runner):
     assert ret.returncode == 0
 
 
-def test_global_help_prints_resources_vertically(script_runner):
-    ret = script_runner.run(["gitlab", "--help"])
-    assert """resource:\n  application\n  application-appearance\n""" in ret.stdout
```
nejch, line 39: Should we check here that the endpoint is shown, maybe with a regex if a hard coded string is too brittle?

nejch (commented):
Thanks @JohnVillalovos, that's a really cool idea! Just one quick question about the deleted test.

tests/functional/cli/test_cli.py
JohnVillalovos: My plan in the long term is to get rid of the other test and the VerticalHelp printer. As if I add `help=None` to arguments it does the same thing.

This is one step in that direction.

JohnVillalovos: Ah. I didn't read that close enough. Check for the endpoints in the test?

We could. Not sure if we need to.

nejch: I just remember we've had some CLI regressions in the past, but if it's going away then it shouldn't matter :+1:

nejch (approved):

Issue #2825, opened 2024-03-19
ProjectMergeRequestApprovalManager.set_approvers cannot handle too many approval rules


liu15:
## Description of the problem, including code/CLI snippet
In v4/objects/merge_request_approvals.py ~ line 121

        # update any existing approval rule matching the name
        existing_approval_rules = approval_rules.list()
        for ar in existing_approval_rules:
            if ar.name == approval_rule_name:
                ar.user_ids = data["user_ids"]
                ar.approvals_required = data["approvals_required"]
                ar.group_ids = data["group_ids"]
                ar.save()
                return ar
        # if there was no rule matching the rule name, create a new one
        return approval_rules.create(data=data, **kwargs)

An existing approval rule will not be found if it exceeds the default pagination count for list().  It will instead attempt to create it (which will fail because it already exists)

## Expected Behavior
approval rule should be found in the list of existing approval rules and be modified.

## Actual Behavior
approval rule is not found and creation of new rule fails.

## Specifications

I believe this fix is to change

approval_rules.list() 

to

approval_rules.list(all=True)
  - python-gitlab version: 4.4.0
  - API version you are using (v3/v4): v4
  - Gitlab server version (or gitlab.com):

PR #2866, opened 2024-05-15
fix: handle large number of approval rules

JohnVillalovos:
Use `iterator=True` when going through the list of current approval rules. This allows it to handle more than the default of 20 approval rules.

Closes: #2825


nejch (approved):
Ah nice, thanks!

PR #2871, opened 2024-05-16
fix(projects): fix 'import_project' file argument type for typings

AdrianDC:
## Changes

Official documentations for the 'import_project' API use a file object.

```python
with open('/tmp/export.tgz', 'rb') as f:
    output = gl.projects.import_project(
        f, path='my_new_project', name='My New Project'
    )
```

However the `import_project` file argument is declared as `str`, breaking Python typings checks.

```bash
src/features/gitlab.py:344: error: Argument 1 to "import_project" of                        
"ProjectManager" has incompatible type "BufferedReader"; expected "str"                                                                                                                 
[arg-type]
                    file,
                    ^~~~
Found 1 error in 1 file (checked 22 source files)
```

### Documentation and testing

- Detected over `gitlab-projects-migrate` : https://gitlab.com/AdrianDC/gitlab-projects-migrate/-/jobs/6873996382

### DIFF
--- a/gitlab/v4/objects/projects.py
+++ b/gitlab/v4/objects/projects.py
@@ -3,6 +3,7 @@ GitLab API:
 https://docs.gitlab.com/ee/api/projects.html
 """
 
+import io
 from typing import (
     Any,
     Callable,
@@ -786,7 +787,7 @@ class ProjectManager(CRUDMixin, RESTManager):
     @exc.on_http_error(exc.GitlabImportError)
     def import_project(
         self,
-        file: str,
+        file: io.BufferedReader,
         path: str,
         name: Optional[str] = None,
         namespace: Optional[str] = None,


gitlab/v4/objects/projects.py
```diff
@@ -3,6 +3,8 @@
 https://docs.gitlab.com/ee/api/projects.html
 """
 
+from io import BufferedReader
```
JohnVillalovos, line 6: Please do

```
import io
```

And then use `io.BufferedReader`

AdrianDC: Done :+1:

JohnVillalovos (approved):
Thank you

AdrianDC:
In case you're interested @JohnVillalovos, using the API in both tools : 
- https://pypi.org/project/gitlab-projects-migrate/
- https://pypi.org/project/gitlab-projects-settings/

We new use the `--reset-features` mode a lot, and `--set-avatar` + `--update-description`  
automatically while migrating through self-managed GitLab instances.

JohnVillalovos:
> In case you're interested @JohnVillalovos, using the API in both tools :
> 
>     * https://pypi.org/project/gitlab-projects-migrate/
> 
>     * https://pypi.org/project/gitlab-projects-settings/
> 
> 
> We new use the `--reset-features` mode a lot, and `--set-avatar` + `--update-description` automatically while migrating through self-managed GitLab instances.

Cool! Great to see it being useful 😊

Thanks.

Issue #2876, opened 2024-05-21
Redirected `head()` requests raise `RedirectError`


EugenSusurrus:
## Description

When downloading artifacts from a GitLab job I'd like to implement a progress bar. For that I'd need to know the file size in advance. The file size can be inferred by issuing a HEAD request and is found in the `content-length` response header.

```python
job = project.jobs.get(job.id)

endpoint = f"{job.manager.path}/artifacts/{job.ref}/download"

response: dict = job.manager.gitlab.http_head(endpoint, query_data={"job": job.name})
file_size = int(response.get('content-length', 0)
cd: str = response.get("content-disposition", "filename=artifacts.zip")
file_name = cd.partition("filename=")[-1].strip('"')
```

## Expected Behavior

I'd expect that head redirects are allowed, judging by the comments in the `client.py`

```python
    @staticmethod
    def _check_redirects(result: requests.Response) -> None:
        # Check the requests history to detect 301/302 redirections.
        # If the initial verb is POST or PUT, the redirected request will use a
        # GET request, leading to unwanted behaviour.
        # If we detect a redirection with a POST or a PUT request, we
        # raise an exception with a useful error message.
        if not result.history:
            return

        for item in result.history:
            if item.status_code not in (301, 302):
                continue
            # GET methods can be redirected without issue
            if item.request.method == "GET":   # should be  if item.request.method in ("GET", "HEAD"):
                continue
            target = item.headers.get("location")
            raise gitlab.exceptions.RedirectError(
                REDIRECT_MSG.format(
                    status_code=item.status_code,
                    reason=item.reason,
                    source=item.url,
                    target=target,
                )
            )
```

## Actual Behavior

This snippet of code results in:

```
File "/Users/secret/Workspace/repo/env/lib/python3.10/site-packages/gitlab/client.py", line 632, in _check_redirects
    raise gitlab.exceptions.RedirectError(
gitlab.exceptions.RedirectError: python-gitlab detected a 302 ('Found') redirection. You must update your GitLab URL to the correct URL to avoid issues. The redirection was from: 'https://gitlab.com/api/v4/projects/1234567/jobs/artifacts/develop/download?job=mac%3Abuild' to 'https://cdn.artifacts.gitlab-static.net/secret'
```

## Specifications

  - python-gitlab version: 4.5.0
  - API version: v4
  - gitlab.com as SaaS

nejch:
[...]

PR #2877, opened 2024-05-21
fix: don't raise `RedirectError` for redirected `HEAD` requests

EugenSusurrus:
## Changes

Aims to fix the issue described in:
https://github.com/python-gitlab/python-gitlab/issues/2876


EugenSusurrus:
Thanks as well for your fast response @nejch!

PR #2880, opened 2024-05-22
fix(deps): update minimum dependency versions

JohnVillalovos:
Update the minimum versions of the dependencies in the pyproject.toml file.

This is related to PR #2878

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,8 +17,8 @@ maintainers = [
 ]
 requires-python = ">=3.8.0"
 dependencies = [
-    "requests>=2.25.0",
-    "requests-toolbelt>=0.10.1",
+    "requests>=2.32.0",
+    "requests-toolbelt>=1.0.0",
 ]
 classifiers = [
     "Development Status :: 5 - Production/Stable",


nejch (approved):
LGTM, I would just maybe change the commit type to at least `fix` as it's a user-facing change. Technically this could break for some people (e.g. if another dependency tracks `requests` with e.g. `<=2.23.0` or something like that) and the dependencies will fail to resolve.

I can do this as part of the squash merge later, but I'll wait a bit in case you have feedback :+1:

JohnVillalovos:
> LGTM, I would just maybe change the commit type to at least `fix` as it's a user-facing change. Technically this could break for some people (e.g. if another dependency tracks `requests` with e.g. `<=2.23.0` or something like that) and the dependencies will fail to resolve.
> 
> I can do this as part of the squash merge later, but I'll wait a bit in case you have feedback 👍

Done. I updated the commit message.

Thanks.

JohnVillalovos:
@nejch Since you had approved it, I went ahead and merged it.

PR #2887, opened 2024-05-28
feat: add `--no-mask-credentials` CLI argument

JohnVillalovos:
This gives the ability to not mask credentials when using the
`--debug` argument.

### DIFF
--- a/gitlab/cli.py
+++ b/gitlab/cli.py
@@ -399,25 +399,25 @@ def main() -> None:
     args_dict = vars(args)
     # Remove CLI behavior-related args
     for item in (
-        "gitlab",
+        "api_version",
         "config_file",
-        "verbose",
         "debug",
-        "gitlab_resource",
-        "resource_action",
-        "version",
-        "output",
         "fields",
+        "gitlab",
+        "gitlab_resource",
+        "job_token",
+        "oauth_token",
+        "output",
+        "pagination",
+        "private_token",
+        "resource_action",
         "server_url",
+        "skip_login",
         "ssl_verify",
         "timeout",
-        "api_version",
-        "pagination",
         "user_agent",
-        "private_token",
-        "oauth_token",
-        "job_token",
-        "skip_login",
+        "verbose",
+        "version",
     ):
         args_dict.pop(item)
     args_dict = {k: _parse_value(v) for k, v in args_dict.items() if v is not None}
--- a/gitlab/cli.py
+++ b/gitlab/cli.py
@@ -308,6 +308,12 @@ def _get_base_parser(add_help: bool = True) -> argparse.ArgumentParser:
         action="store_true",
         default=os.getenv("GITLAB_SKIP_LOGIN"),
     )
+    parser.add_argument(
+        "--no-mask-credentials",
+        help="Don't mask credentials in debug mode",
+        dest="mask_credentials",
+        action="store_false",
+    )
     return parser
 
 
@@ -395,6 +401,7 @@ def main() -> None:
     gitlab_resource = args.gitlab_resource
     resource_action = args.resource_action
     skip_login = args.skip_login
+    mask_credentials = args.mask_credentials
 
     args_dict = vars(args)
     # Remove CLI behavior-related args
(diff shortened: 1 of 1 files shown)


gitlab/cli.py
```diff
@@ -308,6 +308,11 @@ def _get_base_parser(add_help: bool = True) -> argparse.ArgumentParser:
         action="store_true",
         default=os.getenv("GITLAB_SKIP_LOGIN"),
     )
+    parser.add_argument(
+        "--unmask-credentials",
+        help="Don't mask credentials in debug mode",
+        action="store_true",
```
nejch, line 314: I think if we invert this the same way as `--no-ssl-verify`, we can make the the enable_debug call more elegant later?

```suggestion
        action="store_false",
```

gitlab/cli.py
```diff
@@ -395,37 +400,39 @@ def main() -> None:
...
+        "pagination",
+        "private_token",
+        "resource_action",
         "server_url",
+        "skip_login",
         "ssl_verify",
         "timeout",
-        "api_version",
-        "pagination",
+        "unmask_credentials",
```
nejch, line 424: (will need more sorting if that was your idea in this PR)

```suggestion
        "mask_credentials",
```

gitlab/cli.py
```diff
@@ -395,37 +400,39 @@ def main() -> None:
...
+        "version",
     ):
         args_dict.pop(item)
     args_dict = {k: _parse_value(v) for k, v in args_dict.items() if v is not None}
 
     try:
         gl = gitlab.Gitlab.merge_config(vars(options), gitlab_id, config_files)
         if debug:
-            gl.enable_debug()
+            gl.enable_debug(mask_credentials=not unmask_credentials)
```
nejch, line 435: To go with the above:

```suggestion
            gl.enable_debug(mask_credentials=mask_credentials)
```

gitlab/cli.py
```diff
@@ -395,37 +400,39 @@ def main() -> None:
     gitlab_resource = args.gitlab_resource
     resource_action = args.resource_action
     skip_login = args.skip_login
+    unmask_credentials = args.unmask_credentials
```
nejch, line 403:
```suggestion
    mask_credentials = args.mask_credentials
```

nejch (changes requested):
Thanks @JohnVillalovos! Just a few suggestions to make it consistent with our `ssl_verify` approach. :bow:

I didn't test my suggestions, but we have a test for ssl_verify so should be OK.

gitlab/cli.py
```diff
@@ -308,6 +308,11 @@ def _get_base_parser(add_help: bool = True) -> argparse.ArgumentParser:
         action="store_true",
         default=os.getenv("GITLAB_SKIP_LOGIN"),
     )
+    parser.add_argument(
+        "--unmask-credentials",
```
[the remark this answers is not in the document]
JohnVillalovos: Good idea. Done.

gitlab/cli.py
JohnVillalovos: 👍

gitlab/cli.py
JohnVillalovos: Done

gitlab/cli.py
JohnVillalovos: Done

JohnVillalovos:
@nejch Updated with your suggestions. Thanks for the review.

gitlab/cli.py
```diff
@@ -308,6 +308,12 @@ def _get_base_parser(add_help: bool = True) -> argparse.ArgumentParser:
         action="store_true",
         default=os.getenv("GITLAB_SKIP_LOGIN"),
     )
+    parser.add_argument(
+        "--no-mask-credentials",
+        help="Don't mask credentials in debug mode",
+        dest="mask_credentials",
+        action="store_false",
```
nejch, line 315: Follow-up could be a default with `os.getenv()` but we need our little strtobool helper first :+1:

nejch (approved):
Thanks @JohnVillalovos!

Issue #2900, opened 2024-06-17
CLI: For `list` actions the CLI defaults to `get_all=False` 


JohnVillalovos:
For `list` actions the CLI defaults to `get_all=False`

This causes a problem in that no warning is given to the user that the list is truncated if there are more than 20 items.

Maybe we should default to `None` and then figure out some way to warn the user if the list is being truncated.

PR #2901, opened 2024-06-22
fix(cli): generate UserWarning if `list` does not return all entries

JohnVillalovos:
Previously in the CLI, calls to `list()` would have `get_all=False` by
default. Therefore hiding the fact that not all items are being
returned if there were more than 20 items.

Added `--no-get-all` option to `list` actions. Along with the already
existing `--get-all`.

Closes: #2900

### DIFF
--- a/gitlab/client.py
+++ b/gitlab/client.py
@@ -869,6 +869,7 @@ class Gitlab:
         query_data: Optional[Dict[str, Any]] = None,
         *,
         iterator: Optional[bool] = None,
+        message_details: Optional[utils.WarnMessageData] = None,
         **kwargs: Any,
     ) -> Union["GitlabList", List[Dict[str, Any]]]:
         """Make a GET request to the Gitlab server for list-oriented queries.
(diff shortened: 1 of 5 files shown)


gitlab/client.py
```diff
@@ -952,16 +953,26 @@ def should_emit_warning() -> bool:
...
         # maximum items. This is a common cause of issues filed.
         total_items = "many" if gl_list.total is None else gl_list.total
-        utils.warn(
-            message=(
+        if cli:
+            message = (
+                f"Your query returned {len(items)} of {total_items}. To return all "
+                f"items use `--get-all`. To silence this warning use `--no-get-all`."
+            )
+            show_caller = False
```
nejch, line 961: Is there a way we can just pass the message specifics (e.g. `message_details`) here, rather than a CLI boolean. This way we

1. Keep the API client and CLI decoupled
2. Allow passing other messages from non-`list()` calls like we had in a previous discussion. For example, 

We can keep a small default message with the pagination link with `list()`, but allow specifying a custom message from other callers.

nejch (commented):
THanks a lot @JohnVillalovos, I've been thinking about this one for a bit because we reintroduce more CLI logic into the API client, and I wonder if there's a more generic way. Just a tiny idea below :bow:

JohnVillalovos:
> THanks a lot @JohnVillalovos, I've been thinking about this one for a bit because we reintroduce more CLI logic into the API client, and I wonder if there's a more generic way. Just a tiny idea below 🙇

@nejch thanks for the review. I have updated it to pass in a WarnMessageData object. So I think it will do what you want.

nejch (approved):

PR #2911, opened 2024-07-05
fix(files): make `ref` parameter optional in get raw file API

peteraisher:
The `ref` parameter for the raw file API was made optional in GitLab v13.11.0.

## Changes

- Make the `ref` parameter for the corresponding method on `ProjectFileManager` optional to match the API.
- Minor changes to the docstring for this method.

### Documentation and testing

### DIFF
--- a/gitlab/v4/objects/files.py
+++ b/gitlab/v4/objects/files.py
@@ -228,13 +228,14 @@ class ProjectFileManager(GetMixin, CreateMixin, UpdateMixin, DeleteMixin, RESTMa
         self.gitlab.http_delete(path, query_data=data, **kwargs)
 
     @cli.register_custom_action(
-        cls_names="ProjectFileManager", required=("file_path", "ref")
+        cls_names="ProjectFileManager",
+        required=("file_path",),
     )
     @exc.on_http_error(exc.GitlabGetError)
     def raw(
         self,
         file_path: str,
-        ref: str,
+        ref: Optional[str] = None,
         streamed: bool = False,
         action: Optional[Callable[..., Any]] = None,
         chunk_size: int = 1024,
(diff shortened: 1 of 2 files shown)


gitlab/v4/objects/files.py
```diff
@@ -266,7 +267,10 @@ def raw(
         """
         file_path = utils.EncodedId(file_path)
         path = f"{self.path}/{file_path}/raw"
-        query_data = {"ref": ref}
+        if ref:
```
nejch, line 270: I doubt people are going to call their branch `0`, but just in case:

```suggestion
        if ref is not None:
```

tests/functional/api/test_repository.py
```diff
@@ -46,7 +46,7 @@ def test_repository_files(project):
     blame = project.files.blame(file_path="README.rst", ref="main")
     assert blame
 
-    raw_file = project.files.raw(file_path="README.rst", ref="main")
+    raw_file = project.files.raw(file_path="README.rst")
     assert os.fsdecode(raw_file) == "Initial content"
```
nejch: As you can see we now don't cover that path, what if we just add a simple assert like this. Not sure if you want to be more elaborate in testing:

```suggestion
    raw_file = project.files.raw(file_path="README.rst")
    assert os.fsdecode(raw_file) == "Initial content"

    raw_file_from_ref = project.files.raw(file_path="README.rst", ref="main")
    assert raw_file == raw_file_from_ref
```

nejch (changes requested):
Thanks @peteraisher for your contribution! Looks good. I just had 2 little nits :bow: WDYT?

tests/functional/api/test_repository.py
nejch: Ok, I see you've already done something similar, never mind!

gitlab/v4/objects/files.py
peteraisher: Checking against `None` is better, yes.

I wanted to also catch the case where an empty string is passed, since that would be / lead to an error — maybe raising an exception would be appropriate in that case?

tests/functional/api/test_repository.py
peteraisher: Had exactly the same thought!

nejch:
Thanks again @peteraisher, I'll go ahead and merge - ignore thecodecov issue :)

nejch (approved):

peteraisher:
Thanks for the review @nejch :)

Issue #2913, opened 2024-07-07
Should the participants method use `http_list`?


JohnVillalovos:
Currently the `participants()` method is using `http_get()`.

Should it use `http_list()`? I'm thinking yes. @nejch What do you think?

https://github.com/python-gitlab/python-gitlab/blob/51d8f888aca469cff1c5ee5e158fb259d2862017/gitlab/mixins.py#L914-L935

Possibly related is #2912

PR #2914, opened 2024-07-07
fix: Have `participants()` method use `http_list()`

JohnVillalovos:
Previously it was using `http_get()` but the `participants` API returns a list of participants. Also by using this then we will warn if only a subset of the participants are returned.

Closes: #2913


nejch (approved):
Thanks @JohnVillalovos!

Yes makes sense IMO this is why I'm in favor of using managers with ListMixin everywhere in place of custom methods as they often get it wrong or can get outdated it seems.

PR #2921, opened 2024-07-10
chore(ci): stale: allow issues/PRs that have stale label to be closed

JohnVillalovos:
If a `stale` label is manually applied, allow the issue or PR to be closed by the stale job.

Previously it would require the `stale` label and to also have one of 'need info' or 'Waiting for response' labels added.


nejch (approved):
Makes sense, thanks @JohnVillalovos! I might reopen some if the first wave of closing is too eager and then we can try to tackle those issues.

Issue #2618, opened 2023-07-18
CLI cannot handle server response


twil69:
## Description of the problem, including code/CLI snippet
`gitlab -c ./gitlab-cli.cfg  project-merge-request get --iid 98765  --project-id 12345`

with config file content being
```
[global]
default = foo
ssl_verify = False
timeout = 5
api_version = 4

[foo]
url=https://some-project.url
private_token=some-token-here
```

results in 
```
Attempted to initialize RESTObject with a non-dictionary value: <Response [200]>
This likely indicates an incorrect or malformed server response.
```

## Expected Behavior
Since I can successfully interact with the same server programmatically from Python (i.e. I am getting a valid server JSON response) I suspect this is a problem of the CLI.
The server response is 200 this typically indicates OK. So it will have sent some payload back.
I expect identical behaviour of package and CLI.

## Specifications

  - python-gitlab version: 3.15.0
  - API version you are using (v3/v4): v4
  - Gitlab server version (or gitlab.com): 14.9.5

stdedos:
Apologies for the "necrobump" and the @nejch-bump :sweat_smile: 

I am having the same issue in `gitlab --version 4.4.0`.

```
$ gitlab --debug project-merge-request-discussion list --project-id ./. --mr-iid 14254 
DEBUG:urllib3.connectionpool:Starting new HTTPS connection (1): gitlab.....com:443
DEBUG:http.client:send: b'GET /-/profile/personal_access_tokens/api/v4/user HTTP/1.1\r\nHost: gitlab.....com\r\nUser-Agent: python-gitlab/4.4.0\r\nAccept-Encoding: gzip, deflate\r\nAccept: */*\r\nConnection: keep-alive\r\nContent-type: application/json\r\nPRIVATE-TOKEN: [MASKED]\r\n\r\n'
DEBUG:http.client:reply: 'HTTP/1.1 302 Found\r\n'
DEBUG:http.client:header: Server: nginx
DEBUG:http.client:header: Date: Wed, 06 Mar 2024 10:15:56 GMT
DEBUG:http.client:header: Content-Type: text/html; charset=utf-8
DEBUG:http.client:header: Content-Length: 113
DEBUG:http.client:header: Connection: keep-alive
DEBUG:http.client:header: Cache-Control: no-cache
DEBUG:http.client:header: Content-Security-Policy: 
DEBUG:http.client:header: Location: https://gitlab.....com/users/sign_in
[...]

PR #2925, opened 2024-07-12
chore(ci): specify name of "stale" label

JohnVillalovos:
Saw the following error in the log:
  [#2618] Removing the label "Stale" from this issue...
  ##[error][#2618] Error when removing the label: "Label does not exist"

My theory is that the case doesn't match ("Stale" != "stale") and that is why it failed.  Our label is "stale" so update this to match. Thought of changing the label name on GitHub but then would also require a change here to the "any-of-labels". So it seemed simpler to just change it here.

It is confusing though that it detected the label "stale", but then couldn't delete it.

### Documentation and testing

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -17,6 +17,9 @@ jobs:
     steps:
       - uses: actions/stale@v9.0.0
         with:
+          stale-issue-label: "stale"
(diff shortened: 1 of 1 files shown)


nejch (approved):
Weird, it might be doing some search/lookup or so. Thanks of looking into this!

Issue #2839, opened 2024-04-16
get_id() retruns name of label


cweber-dbs:
## Description of the problem, including code/CLI snippet

get_id() of a GroupLabel returns the name of the label and not the id.
In the case of a label inherited from a parent group, it is not possible to distinguish between two labels if they have the same name.

## Expected Behavior

I expect to get the value of `label.id`

## Actual Behavior

I receive the value of `label.name`

## Specifications

  - python-gitlab version: 3.8.1
  - API version you are using (v3/v4):
  - Gitlab server version (or gitlab.com): 16.9.2

PR #2926, opened 2024-07-13
fix: correct id attribute for GroupLabel

JohnVillalovos:
This makes it work like ProjectLabel

Closes: #2839


nejch:
@JohnVillalovos this seems to be a duplicate of https://github.com/python-gitlab/python-gitlab/pull/2811, has anything changed in the meantime? See my comments there, I think they both need updating but would make sense to keep it as repr_attr.

Issue #2835, opened 2024-04-04
update CI/CD Variables  with a specific env 


bungeetux:
## Description of the problem, including code/CLI snippet

When trying to update a variable with an environment set, the variable could not be updated as mention in the documentation 
https://python-gitlab.readthedocs.io/en/stable/gl_objects/variables.html
you need to add a filter as the gitlab api endpoint mention it in https://docs.gitlab.com/ee/api/project_level_variables.html#update-a-variable

It could be added to the documentation or fix in the code, as you could assume that you would never change the environment of a variable.

```
  for var in project.variables.list():  
    if var.environment_scope == env:
      if var.key == name:
        print('Variable found')
        print(var)
        var.value = value
        **var.filter = {'environment_scope': env}** # necessary to update a variable that has multiple environnements
        var.save()
        print('Variable updated')
```

## Expected Behavior

variable updated 

## Actual Behavior
```
gitlab.exceptions.GitlabUpdateError: 409: There are multiple variables with provided parameters. Please use 'filter[environment_scope]'
```

## Specifications

  - python-gitlab version: 4.2.0
  - API version you are using (v3/v4): v4
  - Gitlab server version (or gitlab.com): gitlab.com

Issue #1387, opened 2021-03-30
Wrong behaviour deleting variable


Gisson:
## Description of the problem, including code/CLI snippet

On a project where we have the same variable names in different environments, running the code below to delete a variable produces a scope error when running it.
```python
variable.delete()
```

## Expected Behavior

Successful variable deletion

## Actual Behavior

Produces the error below:
```
Traceback (most recent call last):
  File "/repo/lib/python3.8/site-packages/gitlab/exceptions.py", line 279, in wrapp
ed_f
    return f(*args, **kwargs)
  File "/repo/lib/python3.8/site-packages/gitlab/mixins.py", line 341, in delete
    self.gitlab.http_delete(path, query_data=query_data, **kwargs)
  File "/repo/lib/python3.8/site-packages/gitlab/__init__.py", line 744, in http_de
lete
    return self.http_request("delete", path, query_data=query_data, **kwargs)
  File "/repo/lib/python3.8/site-packages/gitlab/__init__.py", line 567, in http_request
    raise GitlabHttpError(
gitlab.exceptions.GitlabHttpError: 409: There are multiple variables with provided parameters. Please use 'filter[environment_scope]'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/repo/lib/python3.8/site-packages/gitlab/mixins.py", line 403, in delete
    self.manager.delete(self.get_id(), query_data)
  File "/repo/lib/python3.8/site-packages/gitlab/exceptions.py", line 281, in wrapp
ed_f
    raise error(e.error_message, e.response_code, e.response_body) from e
gitlab.exceptions.GitlabDeleteError: 409: There are multiple variables with provided parameters. Please use 'filter[environment_scope]'
```

## Specifications

  - python-gitlab version: 2.6.0
  - API version you are using (v3/v4): v4
  - Gitlab server version (or gitlab.com): 13.9

Gisson:
I've confirmed that this workaround does work. However this can delete unwanted variables if we don't make sure that the environment scope in the filter is not the same as the one on the variable we fetched.

Gisson:
But it is possible to alter the `http_delete` method to have query arguments and pass it the query arguments we need, that way the `variable.delete` method would work (tried it on a previous version).

acdha:
[...]

Issue #1125, opened 2020-06-22
Wrong behavior saving project variables when environment scopes are configured. 


bymarshall:
## Description of the problem, including code/CLI snippet
If environment scopes are used in project variables,  modifying and saving a project variable with a specific "environment_scope" is always altering the variable with the last defined scope.

Example: 

```
>>> variables = project.variables.list()
>>> for variable in variables:
...     print("key: %s scope: %s" % (variable.key, variable.environment_scope))
... 
key: SCOPE_VARIABLE scope: qa
key: SCOPE_VARIABLE scope: prod
key: SCOPE_VARIABLE scope: test
>>> for variable in variables:
...    if variable.environment_scope == "prod":
...        variable.value="new_value"
...        variable.save()
...        print(variable.attributes)
... 
{'variable_type': 'env_var', 'key': 'SCOPE_VARIABLE', 'value': 'new_value', 'protected': False, 'masked': False, 'environment_scope': 'test', 'project_id': 123}
```
The variable with scope "prod" has not been modified, instead the variable with scope "test" is altered. 
## Expected Behavior
The variable with the specified scope should be updated. 

## Actual Behavior
The last scope defined for the variable is always taking precedence over the scope of the current variable that is going to be updated..

## Specifications

  - python-gitlab version: 2.3.1
  - API version you are using (v3/v4): v4
  - Gitlab server version (or gitlab.com): 13.0.5

acdha:
This might have some relation to #1387 — the default manager seems to ignore the environment scopes and requires passing something like `filter={"environment_scope": i.environment_scope}` to `save()`, `delete()`, etc.

PR #2927, opened 2024-07-13
docs: variables: add note about `filter` for updating

JohnVillalovos:
Add a note about using `filter` when updating a variable.

Closes: #2835
Closes: #1387
Closes: #1125


nejch (approved):
Thanks @JohnVillalovos, that was a source of confusion in a lot of places! Including downstream projects from what I remember.

Issue #532, opened 2018-06-14
sudo should be remembered for SaveMixin objects


JonathonReinhart:
## Description of the problem, including code/CLI snippet
```python
grp = gl.groups.get('example-group')

notif = grp.notificationsettings.get(sudo=otheruser)
print("Old level:", notif.level)

notif.level = gitlab.const.NOTIFICATION_LEVEL_GLOBAL
notif.save()    # missing sudo=otheruser

notif = grp.notificationsettings.get(sudo=otheruser)
print("New level:", notif.level)
```

## Expected Behavior
I would expect this to change the notification level for the sudo'd user.

## Actual Behavior
When I call `notif.save()`, it has seemingly no effect, because I am missing `sudo`.

Instead, the changes were applied *to my admin account*!

Changing the code to `notif.save(sudo=otheruser)` works as expected.

## Specifications

  - python-gitlab version: **1.4.0**
  - API version you are using (v3/v4): **v4**
  - Gitlab server version (or gitlab.com): **10.6.4**

## Proposal
If an object is acquired using `sudo=user`, perhaps the library can *attach* that `sudo` user to the `SaveMixin` object, and automatically apply it when `.save()` is called without arguments.

I consider this a bug, and I can't imagine that anyone is *actually* relying on the current behavior.

gpocentek:
Hi @JonathonReinhart 

I agree that this is an unexpected behaviour. I'm still trying to find the proper way to handle this in the code but I'll come up with something (not in the 1.5.0 release though).

JonathonReinhart:
I believe this is still appropriate.

JohnVillalovos:
I'm not sure if I agree with the premise of this issue. As python-gitlab is a wrapper around the GitLab API. I'm not sure that I think it should be making decisions to save the `sudo` state. But that is just my opinion.

Also this issue was opened over six years ago and nobody has felt motivated enough to do the work to create a PR to close this issue, which makes me think this should be closed.

PR #2928, opened 2024-07-13
docs: document how to use `sudo` if modifying an object

JohnVillalovos:
Add a warning about using `sudo` when saving.

Give an example of how to `get` an object, modify it, and then `save` it using `sudo`

Closes: #532

### DIFF
--- a/docs/api-usage.rst
+++ b/docs/api-usage.rst
@@ -411,6 +411,27 @@ user. For example:
 
    p = gl.projects.create({'name': 'awesome_project'}, sudo='user1')
 
+.. warning::
+   When using ``sudo``, its usage is not remembered. If you use ``sudo`` to
+   retrieve an object and then later use ``save()`` to modify the object, it
+   will not use ``sudo``.  You should use ``save(sudo='user1')`` if you want to
+   perform subsequent actions as the  user.
+
+Updating with ``sudo``
+----------------------
+
+An example of how to ``get`` an object (using ``sudo``), modify the object, and
+then ``save`` the object (using ``sudo``):
+
+.. code-block:: python
+
+   group = gl.groups.get('example-group')
+   notification_setting = group.notificationsettings.get(sudo='user1')
+   notification_setting.level = gitlab.const.NOTIFICATION_LEVEL_GLOBAL
+   # Must use `sudo` again when doing the save.
+   notification_setting.save(sudo='user1')
+
+
 Logging
 =======
 


docs/api-usage.rst
```diff
@@ -411,6 +411,27 @@ user. For example:
 
    p = gl.projects.create({'name': 'awesome_project'}, sudo='user1')
 
+.. warning:: 
+   When using ``sudo``, its usage is not remembered. What this means is that if
+   you use ``sudo`` to retrieve an object and then later use ``save()`` to
+   modify the object it will not use ``sudo``. In that case you should do
+   ``save(sudo='user1')``
```
nejch, line 418: Could we maybe shorten the sentence a bit as it's inside an admonition block, so we can also mention why this is needed?

```suggestion
   When using ``sudo``, its usage is not remembered. If you use ``sudo`` to retrieve an object
   and then later use ``save()`` to modify the object, it will not use ``sudo``.  You should use
   ``save(sudo='user1')`` if you want to perform subsequent actions as the  user.
```

docs/api-usage.rst
```diff
@@ -411,6 +411,27 @@ user. For example:
...
+
+Updating with ``sudo``
+----------------------
+
+An example of how to ``get`` an object (using ``sudo``), modify the object, and
+then ``save`` the object (using ``sudo``)
```
nejch, line 424:
```suggestion
then ``save`` the object (using ``sudo``):
```

docs/api-usage.rst
```diff
@@ -411,6 +411,27 @@ user. For example:
...
+
+.. code-block:: python
+
+   group = gl.groups.get('example-group')
+   notification_setting = group.notificationsettings.get(sudo='user1')
+   notification_setting.level = gitlab.const.NOTIFICATION_LEVEL_GLOBAL
+   # Must use `sudo` again when doing the save.
+   notifcation_setting.save(sudo='user1')
```
nejch, line 432:
```suggestion
   notification_setting.save(sudo='user1')
```

nejch (changes requested):
Thanks @JohnVillalovos, this should help already even if we don't fully address it (this seems like a more general problem with our approach than the filter issue). Just a few nits :bow:

docs/api-usage.rst
JohnVillalovos: Done.

docs/api-usage.rst
JohnVillalovos: Done

JohnVillalovos:
> Thanks @JohnVillalovos, this should help already even if we don't fully address it (this seems like a more general problem with our approach than the filter issue). Just a few nits 🙇

Thanks @nejch! I have put your suggestions in place 🙂

nejch (approved):

PR #2949, opened 2024-08-08
chore: switch to ruff for linting

nejch:
## Changes

I've seen a lot of projects migrate already, and it's really crazy fast and avoids some of the cross-linter config we have in place atm.

We could also switch pylint to it later, but maybe wait for https://github.com/astral-sh/ruff/issues/970 first.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -31,13 +31,9 @@ jobs:
       - run: pip install --upgrade tox
       - name: Run commitizen (https://commitizen-tools.github.io/commitizen/)
         run: tox -e cz
-      - name: Run black code formatter (https://black.readthedocs.io/en/stable/)
-        run: tox -e black -- --check
-      - name: Run flake8 (https://flake8.pycqa.org/en/latest/)
-        run: tox -e flake8
+      - name: Run ruff linter and formatter (https://docs.astral.sh/ruff/)
+        run: tox -e ruff-check
       - name: Run mypy static typing checker (http://mypy-lang.org/)
         run: tox -e mypy
-      - name: Run isort import order checker (https://pycqa.github.io/isort/)
-        run: tox -e isort -- --check
       - name: Run pylint Python code static checker (https://github.com/PyCQA/pylint)
         run: tox -e pylint
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -2,23 +2,11 @@ default_language_version:
     python: python3
 
 repos:
-  - repo: https://github.com/psf/black
-    rev: 24.8.0
-    hooks:
-      - id: black
   - repo: https://github.com/commitizen-tools/commitizen
     rev: v3.28.0
     hooks:
       - id: commitizen
         stages: [commit-msg]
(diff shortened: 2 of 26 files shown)


JohnVillalovos:
I'll be honest, I don't want to.

1. Not written in Python
2. I have to install a new program on my system
3. Written by a company who are most likely losing money doing this. What happens if they go out of business?
4. Our code base isn't really that bad where the speed or lack of speed is a problem.

My initial thoughts on the matter.

nejch:
> I'll be honest, I don't want to.
> 
> 1. Not written in Python
> 2. I have to install a new program on my system
>
> And I did notice that Fedora 40 does have `ruff` available as an installable package.

Writing on fedora 40 here, no need for extra software as in the end it's still a python package, all pip installable. IMO this should never be installed outside of pypi. I don't see this as that much different from cryptography/numpy/etc that also have rust or C code packaged in the wheel.

> 3. Written by a company who are most likely losing money doing this. What happens if they go out of business?

It's newish but quite established it seems. With 30k stars already vs. 58k for black I'm actually not too worried about this, if they try to pull a Hashicorp I'd imagine someone else would pick it up because of the benefits it brings. But also if it's consistent with the existing rules which seems to be the case, we can always switch back.

> 4. Our code base isn't really that big where the speed or lack of speed is a problem.

True, I just liked the idea of simplifying the linting setup and with ruff we could easily enable more linting that would otherwise require more deps/plugins (e.g. I got some findings by enabling more https://docs.astral.sh/ruff/rules/ that would be extra flake8 plugins otherwise).

JohnVillalovos:
First off, if you really want this then go ahead and merge.

Myself I would prefer to stick with what we have.

- I would prefer to use Python based tools because if there is an issue it will be much more likely that I (and possibly others) would be able to look at the code base and figure out a fix if necessary. In the past I have had issues come up with our third party tools and proposed PRs to resolve those issues.
- The current tools we use are from the Python Software Foundation (black) and from the Python Code Quality Authority (isort, flake8) and by continuing to use these tools I feel it helps to promote those organizations work.
- To me ruff seems like it is of minimal benefit to this project. If it was dramatically better for python-gitlab then I would be more interested in changing. But I don't see that much benefit for our use case.
- And it still concerns me that ruff is developed by a venture capital funded company with no explanation (that I found) on how they intend to monetize their business. Granted the code is open-source and if they fail someone else would probably fork it and take over development.

Anyway, those are the reasons I would prefer to not switch to ruff. I realize that ruff has a lot of hype going for it at the moment and if we had a gigantic code base the speed improvements would be worth it. But on my ancient Intel i7-6700k (released in Q3 of 2015) system I can run `tox -e black,isort,flake8` in less than 2.2 seconds. (`black: OK (0.68=setup[0.06]+cmd[0.62] seconds)`) (`isort: OK (0.49=setup[0.01]+cmd[0.48] seconds)`) (`flake8: OK (0.89=setup[0.01]+cmd[0.88] seconds)`)

But as I said above if you really want it then go ahead and merge it

nejch:
Let's revisit it later then, we can wait and see if the ecosystem develops towards this, the same way that happened with black a few years ago.

PR #2953, opened 2024-08-12
test(cli): allow up to 30 seconds for a project export

JohnVillalovos:
Before we allowed a maximum of around 15 seconds for the project-export. Often times the CI was failing with this value.

Change it to a maximum of around 30 seconds.


nejch:
Noticed this as well, thanks for taking care of this @JohnVillalovos!

PR #2964, opened 2024-09-01
feat(api): add exclusive GET attrs for /groups/:id/members

Danil42Russia:
I'm not sure if this PR needs tests or documentation updates. If you think it is necessary - I will add it.

Taken from here: https://docs.gitlab.com/17.3/ee/api/members.html#add-a-member-to-a-group-or-project


nejch (approved):
Thanks @Danil42Russia for your contribution! I think this is fine, we don't need to test GitLab's behavior itself.

PR #2970, opened 2024-09-08
feat(api): add exclusive GET attrs for /projects/:id/members

dedlocc:
#2964 introduced support for passing a `username` to `/groups/:id/members`. The same should also be possible for `/projects/:id/members`.


nejch (approved):
Ah thanks @dedlocc. I missed that in the previous review.

If you find any more missing attributes that you think are needed feel free to open follow-up PRs :bow:

Issue #1357, opened 2021-03-03
Think about supporting the GraphQL API


max-wittig:
Not sure, if this would be worth a separate library, but doesn't look like Gitlab will depreciate the v4 API anytime soon.

So just something to keep in mind.

https://docs.gitlab.com/ee/api/graphql/reference/index.html

nejch:
That's funny last weekend I was hacking away locally on a branch to wrap https://github.com/graphql-python/gql with existing authentication/convenience stuff from `gitlab.Gitlab` 😁

IMO it would also make sense to provide a thin wrapper at least! Since `gql` has a DSL as well, not sure how really needs to be done on top of that.

max-wittig:
haha nice 👍  I just saw a depreciation warning somewhere in the docs (removed from v5 API) and checked out their GraphQL stuff again. Last time I checked they only had a project list endpoint.

irishismyname:
FWIW some of the existing REST APIs are unstable and marked for deprecation, notably [Vulnerabilities](https://docs.gitlab.com/ee/api/vulnerabilities.html) and [Vulnerability Findings](https://docs.gitlab.com/ee/api/vulnerability_findings.html). I'm pulling my hair out trying to develop a vulnerability report ATM because this python-gitlab doesn't support GraphQL (yet?) and the existing REST APIs are broken.

Also https://github.com/python-gitlab/python-gitlab/issues/2316#issuecomment-1874627875.

TimKnight-DWP:
In addition to whay @irishismyname posted above, they are changing how the CI_JOB_TOKEN works in GL 17.* and look to only be adding API control of "Allowlists" via the GraphQL API: https://gitlab.com/gitlab-org/gitlab/-/issues/435903

mitar:
I think it would be nice if this library abstracted out the REST vs. GraphQL difference and one could just use the library, not caring what is used internally.

TimKnight-DWP:
100% the end user of this library shouldn't need to worry about if it's GQL or REST, solely that they can perform a given action

irishismyname:
[...]

PR #2973, opened 2024-09-10
feat(client): add retry handling to GraphQL client

nejch:
## Changes

Related to https://github.com/python-gitlab/python-gitlab/issues/1357

### Documentation and testing

### DIFF
--- a/gitlab/client.py
+++ b/gitlab/client.py
@@ -26,6 +26,7 @@ from gitlab import _backends, utils
 
 try:
     import gql
+    import gql.transport.exceptions
     import graphql
     import httpx
 
@@ -746,7 +747,9 @@ class Gitlab:
             if 200 <= result.status_code < 300:
                 return result.response
 
-            if retry.handle_retry_on_status(result):
+            if retry.handle_retry_on_status(
+                result.status_code, result.headers, result.reason
+            ):
                 continue
 
             error_message = result.content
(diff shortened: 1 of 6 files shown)


gitlab/utils.py
```diff
@@ -100,33 +110,40 @@ def __init__(
         self.retry_transient_errors = retry_transient_errors
 
     def _retryable_status_code(
-        self,
-        result: requests_backend.RequestsResponse,
+        self, status_code: Optional[int], reason: str = ""
```
nejch, line 113: I dumbed this down from a `Response` object to individual arguments so we can feed it in from different kinds of clients more easily.

Coincidentally that simplifes the unit tests for retries quite a bit.

tests/unit/test_graphql.py
```diff
@@ -1,8 +1,15 @@
+import httpx
 import pytest
+import respx
```
nejch, line 3: Had to introduce respx here for httpx. But has a nice interface, I like it more than responses :+1:

gitlab/utils.py
```diff
@@ -100,33 +110,40 @@ def __init__(
...
             return True
 
         return False
 
-    def handle_retry_on_status(self, result: requests_backend.RequestsResponse) -> bool:
-        if not self._retryable_status_code(result):
+    def handle_retry_on_status(
+        self,
+        status_code: Optional[int],
+        headers: Optional[MutableMapping[str, str]] = None,
```
nejch, line 130: MutableMapping covers dict/CaseInsensitiveDict etc to keep mypy happy.

gitlab/client.py
```diff
@@ -1329,4 +1332,28 @@ def execute(
         self, request: Union[str, graphql.Source], *args: Any, **kwargs: Any
     ) -> Any:
         parsed_document = self._gql(request)
-        return self._client.execute(parsed_document, *args, **kwargs)
+        retry = utils.Retry(
+            max_retries=3, obey_rate_limit=True, retry_transient_errors=False
```
nejch, line 1336: As a follow-up, we could/should add this to the GraphQL client args. Just trying to keep the diff a bit smaller.

tests/unit/test_graphql.py
max-wittig: Mhmm but now we have two request test libraries in one project. What is the longterm plan?

That also might confuse contributors. Which response library do I use?

max-wittig:
I'm a bit questioning the decision to have two response libraries

tests/unit/test_graphql.py
nejch: @max-wittig currently httpx is only an optional dependency when selecting the `[graphql]` extra.

At some point we can switch to httpx completely with a major release, maybe when httpx hits 1.0.0 (but seems like they are already stable). `responses` isn't able to mock httpx which we use in the GraphQL client, so it's needed here.

Using httpx now lets us add an async GraphQL client very easily and avoid these requests in the future (similar to https://github.com/python-gitlab/python-gitlab/issues/1025, which is still the most upvoted issue, but kind of blocked by the requests/httpx support). If we stick to only requests, it would be another big breaking change on the GraphQL side.

So right now, if REST => `responses`, if GraphQL => `respx`. I could document this maybe.

JohnVillalovos (approved):
Looks good to me.

Thanks @nejch

PR #2977, opened 2024-09-12
fix(api): head requests for projectfilemanager

### LINKED ISSUE #2976: ProjectFileManager does not implement head method to meet API spec

Attempting to use a HEAD request to retrieve file metadata from a project repository ([Gitlab REST API docs](https://docs.gitlab.com/ee/api/repository_files.html#get-file-metadata-only)) fails with a 404

```python
dest_file_get_args = {
  'file_path': file_def['dest_filepath'],
  'ref': 'HEAD',
}
existing_file = project.files.head(**dest_file_get_args)
```

## Expected Behavior

Should return 200 and file metadata. The same request as a GET request returns 200, and they follow the same URL path and parameter requirements. Sending a GET request logs this URL path:

`
DEBUG:urllib3.connectionpool:https://gitlab.com:443 "GET /api/v4/projects/1/repository/files/somefile.txt?ref=HEAD HTTP/11" 200 418
`

## Actual Behavior

A HTTP HEAD request is sent with the following URL path:
`DEBUG:urllib3.connectionpool:https://gitlab.com:443 "HEAD /api/v4/projects/1/repository/files?file_path=somefile.txt&ref=HEAD HTTP/11" 404 0
`

## Specifications

  - python-gitlab version: ~~3.1.43~~ 4.9.0
  - Gitlab server version (or gitlab.com): [v16.7.4-ee](https://gitlab.com/gitlab-org/gitlab/-/tags/v16.7.4-ee)

holysoles:
## Changes

Closes #2976 

The `ProjectFileManager` class current has a `head` method, which throws a 404 error when used, as it doesn't have the same logic for URL path construction that its `get` method has. This is because it is inheriting from `GetMixin`, which inherits from `HeadMixin`. The benefit of having the HEAD request properly support is you can inspect and compare the file without needing to fully download it, as the actual file content is not returned.

This change:
- Removes the `ProjectFileManager` inheritance of `GetMixin`, since we're not making proper use of that class.
- adds a proper `head` method to `ProjectFileManager` that correctly constructs the API query
- Adds a unit test for the new `head` method based on mocked real-world headers

Note that because the HEAD request returns the requested data as headers and not a JSON object, we need to map the keys to the expected ones for `ProjectFile`. I did this in a dynamic way, but happy to re-implement it with a hardcoded map of expected values.

### Documentation and testing

### DIFF
--- a/gitlab/v4/objects/files.py
+++ b/gitlab/v4/objects/files.py
@@ -2,11 +2,11 @@ import base64
 from typing import (
     Any,
     Callable,
-    cast,
     Dict,
     Iterator,
     List,
     Optional,
+    Tuple,
     TYPE_CHECKING,
     Union,
 )
@@ -20,7 +20,6 @@ from gitlab.base import RESTManager, RESTObject
 from gitlab.mixins import (
     CreateMixin,
     DeleteMixin,
-    GetMixin,
     ObjectDeleteMixin,
     SaveMixin,
     UpdateMixin,
@@ -96,10 +95,11 @@ class ProjectFile(SaveMixin, ObjectDeleteMixin, RESTObject):
         self.manager.delete(file_path, branch, commit_message, **kwargs)
 
 
-class ProjectFileManager(GetMixin, CreateMixin, UpdateMixin, DeleteMixin, RESTManager):
+class ProjectFileManager(CreateMixin, UpdateMixin, DeleteMixin, RESTManager):
     _path = "/projects/{project_id}/repository/files"
     _obj_cls = ProjectFile
     _from_parent_attrs = {"project_id": "id"}
+    _optional_get_attrs: Tuple[str, ...] = ()
     _create_attrs = RequiredOptional(
         required=("file_path", "branch", "content", "commit_message"),
         optional=("encoding", "author_email", "author_name"),
@@ -112,11 +112,7 @@ class ProjectFileManager(GetMixin, CreateMixin, UpdateMixin, DeleteMixin, RESTMa
     @cli.register_custom_action(
         cls_names="ProjectFileManager", required=("file_path", "ref")
     )
-    # NOTE(jlvillal): Signature doesn't match UpdateMixin.update() so ignore
-    # type error
-    def get(  # type: ignore
-        self, file_path: str, ref: str, **kwargs: Any
-    ) -> ProjectFile:
+    def get(self, file_path: str, ref: str, **kwargs: Any) -> ProjectFile:
         """Retrieve a single file.
 
         Args:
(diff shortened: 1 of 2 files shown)


gitlab/v4/objects/files.py
```diff
@@ -114,9 +114,36 @@ class ProjectFileManager(GetMixin, CreateMixin, UpdateMixin, DeleteMixin, RESTMa
     )
     # NOTE(jlvillal): Signature doesn't match UpdateMixin.update() so ignore
     # type error
```
JohnVillalovos, line 116: I don't think this comment is needed anymore. And it was slightly wrong as it should have said `GetMixin.get()`

JohnVillalovos:
Also looks like it needs to be rebased against the current `main` branch. Thanks.

holysoles:
So looks like an existing functional test is failing because it expects the return value of headers to be a dict rather than a `ProjectFile`. Should I conform to that behavior or should I update the test based on me now constructing a ProjectFile from the metadata? I assume the former but I feel like the latter is more usable, but arguably a breaking change.

https://github.com/python-gitlab/python-gitlab/blob/main/tests/functional/api/test_repository.py#L43

nejch:
> So looks like an existing functional test is failing because it expects the return value of headers to be a dict rather than a `ProjectFile`. Should I conform to that behavior or should I update the test based on me now constructing a ProjectFile from the metadata? I assume the former but I feel like the latter is more usable, but arguably a breaking change.
> 
> https://github.com/python-gitlab/python-gitlab/blob/main/tests/functional/api/test_repository.py#L43

@holysoles our client's `http_head` method (and also `head()`) returns just the headers dict without really transforming them, so I think we should either keep it consistent with that, or change it everywhere. IMO it's best to leave this as-is for now, so it's a bit more generic, and discuss that for all head methods in a follow-up.

But I'm a bit confused our existing tests work as intended as opposed to your issue description, it seems like the issue there is actually your use of keyword args, but I'll have to check in more detail when I get home.

holysoles:
@nejch now that you say that, the keyword args do seem like the problem. I think my confusion initially stemmed from the official REST API docs indicate that there isnt a difference in how the API call should be made, but here, the user needs to pass `id=my_file_path` compared to `file_path=my_file_path`.

Given all this, does it make sense to **not** transform the headers dict and just return it, but still proceed with the other changes so the consistency of the parameter name can be improved?

nejch:
> @nejch now that you say that, the keyword args do seem like the problem. I think my confusion initially stemmed from the official REST API docs indicate that there isnt a difference in how the API call should be made, but here, the user needs to pass `id=my_file_path` compared to `file_path=my_file_path`.
> 
> Given all this, does it make sense to **not** transform the headers dict and just return it, but still proceed with the other changes so the consistency of the parameter name can be improved?

Yes, I think that makes sense. TBH I'm questioning whether we really need to even be overriding the `GetMixin` for the file manager because we technically already kind of cover this via the `_id_attr = "file_path"` and we could just document usage, but this is historical and not part of this PR.

For now IMO best to just align the `head()` override with the `get()` one and keep whatever is returned from `http_head`. They are HTTP headers in the end and not really one-to-one replacement for `ProjectFile` attributes so this could also be confusing in some cases otherwise. WDYT?

gitlab/v4/objects/files.py
```diff
@@ -112,11 +112,36 @@ class ProjectFileManager(GetMixin, CreateMixin, UpdateMixin, DeleteMixin, RESTMa
     @cli.register_custom_action(
         cls_names="ProjectFileManager", required=("file_path", "ref")
     )
-    # NOTE(jlvillal): Signature doesn't match UpdateMixin.update() so ignore
-    # type error
-    def get(  # type: ignore
-        self, file_path: str, ref: str, **kwargs: Any
-    ) -> ProjectFile:
+    def head(self, file_path: str, ref: str, **kwargs: Any) -> ProjectFile:
```
nejch, line 115: While we refactor this, maybe just another nit: let's maybe put `head` after `get`, will also keep the diff reasonable here?

gitlab/v4/objects/files.py
```diff
@@ -131,7 +127,37 @@ def get(  # type: ignore
...
+            file_path: Path of the file to retrieve
+            ref: Name of the branch, tag or commit
+            **kwargs: Extra options to send to the server (e.g. sudo)
+
+        Raises:
+            GitlabAuthenticationError: If authentication is not correct
+            GitlabGetError: If the file could not be retrieved
+
+        Returns:
+            The generated RESTObject
```
nejch, line 154: Maybe something like this would be more accurate, and then we're good to go:

```suggestion
            The response headers as a dictionary
```

nejch (approved):
Thanks again @holysoles for the PR and the rework, just had a small nit in the docstring that I applied directly so we can go ahead with this.

PR #2988, opened 2024-09-17
feat(api): Refactor branch creation to include default flag

mateusz-wozny:
## Changes
Add optional attribute for branch to specify if the branch is default branch of repository

## Use case
Automatic creation of merge requests to default branch


nejch:
Hi @mateusz-wozny, thanks for your contribution!

Are you sure this argument is supported by GitLab itself? I don't see it documented as a parameter for the create endpoint in https://docs.gitlab.com/ee/api/branches.html#create-repository-branch, and our integration tests are failing.

It seems to only be returned by the GET endpoints. But I'll take a look at the source code upstream, maybe it's something they've added very recently and not yet in the release images.

nejch:
As mentioned, from what I can tell this is not supported upstream:

https://gitlab.com/gitlab-org/gitlab/-/blob/master/lib/api/branches.rb?ref_type=heads#L170-184

@mateusz-wozny python-gitlab can only provide parameters that are also available in GitLab's API. So IMO you'd first need to contribute this functionality to GitLab itself. Let me know if anything's unclear here or if I missed something.

tests/functional/api/test_branches.py
```diff
@@ -7,11 +7,15 @@
 def test_branch_name_with_period(project):
     # Make sure we can create and get a branch name containing a period '.'
     branch_name = "my.branch.name"
-    branch = project.branches.create({"branch": branch_name, "ref": "main"})
+    branch = project.branches.create(
+        {"branch": branch_name, "ref": "main", "default": True}
```
mateusz-wozny, line 11: this doesn't work properly

tests/functional/api/test_branches.py
```diff
@@ -7,11 +7,15 @@
...
+    )
     assert branch.name == branch_name
+    assert branch.default is True
```
mateusz-wozny, line 14: but this works, the attribute is returned by gitlab

mateusz-wozny:
Hi @nejch , thanks for your feedback, you are right. I can't create new branch with `default` parameter, but still I can get this info from gitlab (for example if I created branch in UI, git etc), so my feature works correctly but I can't test this in right way

nejch:
> Hi @nejch , thanks for your feedback, you are right. I can't create new branch with `default` parameter, but still I can get this info from gitlab (for example if I created branch in UI, git etc), so my feature works correctly but I can't test this in right way

Thanks @mateusz-wozny. Is this pull request still necessary then, since we have all the required functionality in python-gitlab already, or should we close this?

mateusz-wozny:
@nejch No, it's not necessary, I have all required functionality in python-gitlab, thank you for your help!

PR #2989, opened 2024-09-17
fix: filter duplicates from previous page during pagination

### LINKED ISSUE #2979: Duplicate objects returned from list

At least several REST resources are returning duplicate objects.  I have noticed this on both projects and users.
This may be the expected behavior of GitLab itself, but perhaps this Python package which handles pagination could also handle deduplication based on `id`.

## Expected Behavior

I would expect no duplicate objects when using a `.list(get_all=True, iterator=True)` even if objects of that type are created while in the middle of all the pages.

## Actual Behavior

If calling `gl.projects.list(get_all=True, iterator=True)` and a project is created (or the same with users and likely all other object types as well), you'll get a duplicate object.

## end-user mitigation and thoughts

It would be nice if end users didn't have to dedupe themselves.

The below code is overkill but has info I was using while trying to understand the problem.

What I have found is that I do get that warning log about an exact match being returned.  I have never seen the `AssertionError` raised.  I also tracked the indices for information.  In every instance it's been at index `x99` and `x00` (right on a page boundary).
This makes sense as a new project or user is created we've already missed it and everything shifts by one index.

```
WARNING  Duplicate project id 31393 at index 1099 and 1100
WARNING  Duplicate project id 30028 at index 2099 and 2100
WARNING  Duplicate project id 22457 at index 7899 and 7900
WARNING  Duplicate user id 222 at index 10299 and 10300
```

If deduplication is implemented within python-gitlab itself it wouldn't need to keep track of all object ids, just the previous page's object ids, since this only occurs on page boundaries.

```python
def get_stuff(manager: CRUDMixin, **kwargs):
    things = []
    things_by_id = {}
    obj_type = manager.__class__.__name__.removesuffix("Manager").lower()
    for i, thing in enumerate(manager.list(iterator=True, **kwargs)):
        if thing.id in things_by_id:
            existing_idx, existing_thing = things_by_id.get(thing.id)
            if existing_thing == thing:
                logger.warning("Duplicate %s id %s at index %d and %d", obj_type, thing.id, existing_idx, i)
                continue
            else:
                p1 = Path(tempfile.gettempdir()) / f"{obj_type}_{existing_thing.id}_idx_{existing_idx}"
                p2 = Path(tempfile.gettempdir()) / f"{obj_type}_{thing.id}_idx_{i}"
                with p1.open("wt") as f:
                    print(json.dumps(existing_thing.attributes, indent=2, sort_keys=True), file=f)
                with p2.open("wt") as f:
                    print(json.dumps(thing.attributes, indent=2, sort_keys=True), file=f)
                raise AssertionError(
                    f"Duplicate {obj_type} id {thing.id} at index {existing_idx} and {i}; look at {str(p1)} and {str(p2)}"
                )
        things_by_id[thing.id] = (len(things), thing)
        things.append(thing)
        # TODO: this would be better done w/ rich or something
        if len(things) % 100 == 0:
            if len(things) % 200 == 0:
                click.secho("...", nl=False, fg="yellow", bold=True)
            else:
                click.secho("...", nl=False, fg="green", bold=True)
        if len(things) % 1000 == 0:
            click.secho(f"\n{len(things)} {obj_type}s", fg="blue")
    click.secho(f"\n{len(things)} {obj_type}s total", fg="blue", bold=True)
    return things
```

## Specifications

  - python-gitlab version: `python-gitlab==4.10.0`
  - API version you are using (v3/v4): `v4`
  - Gitlab server version (or gitlab.com): `16.11.6-ee`

ericfrederich:
fixes #2979

## Changes

Added a `_prev_page_objects` attribute to the GitlabList class to be able to filter out duplicates when items are created while pagination is happening.

Code could be much more concise if I didn't use the `utils.warn`:

```python
self._data: List[Dict[str, Any]] = [r for r in result.json() if r not in self._prev_page_objects]
...
self._prev_page_objects = list(self._data)
```

### Documentation and testing

Docs and tests skipped for now while discussion happens on #2979.

### DIFF
--- a/gitlab/client.py
+++ b/gitlab/client.py
@@ -1167,6 +1167,7 @@ class GitlabList:
         url: str,
         query_data: Dict[str, Any],
         get_next: bool = True,
+        dedupe: bool = True,
         **kwargs: Any,
     ) -> None:
         self._gl = gl
@@ -1174,6 +1175,7 @@ class GitlabList:
         # Preserve kwargs for subsequent queries
         self._kwargs = kwargs.copy()
 
+        self._dedupe = dedupe
         self._retrieved_object_ids: set[int] = set()
 
         self._query(url, query_data, **self._kwargs)
@@ -1207,17 +1209,20 @@ class GitlabList:
                 error_message="Failed to parse the server message"
             ) from e
 
-        duplicate_ids = set(o["id"] for o in self._data) & self._retrieved_object_ids
-        if duplicate_ids:
-            utils.warn(
-                message=(
-                    f"During pagination duplicate object(s) with id(s) "
-                    f"{duplicate_ids} returned from Gitlab and filtered"
-                ),
-                category=UserWarning,
+        if self._dedupe:
+            duplicate_ids = (
+                set(o["id"] for o in self._data) & self._retrieved_object_ids
             )
-        self._data = [o for o in self._data if o["id"] not in duplicate_ids]
-        self._retrieved_object_ids.update(o["id"] for o in self._data)
+            if duplicate_ids:
+                utils.warn(
+                    message=(
+                        f"During pagination duplicate object(s) with id(s) "
+                        f"{duplicate_ids} returned from Gitlab and filtered"
+                    ),
+                    category=UserWarning,
+                )
+            self._data = [o for o in self._data if o["id"] not in duplicate_ids]
+            self._retrieved_object_ids.update(o["id"] for o in self._data)
 
         self._current = 0
 
--- a/gitlab/client.py
+++ b/gitlab/client.py
@@ -1174,6 +1174,8 @@ class GitlabList:
         # Preserve kwargs for subsequent queries
         self._kwargs = kwargs.copy()
 
+        self._retrieved_object_ids: set[int] = set()
+
         self._query(url, query_data, **self._kwargs)
         self._get_next = get_next
 
@@ -1204,6 +1206,18 @@ class GitlabList:
             raise gitlab.exceptions.GitlabParsingError(
                 error_message="Failed to parse the server message"
             ) from e
+
+        duplicate_ids = set(o["id"] for o in self._data) & self._retrieved_object_ids
+        if duplicate_ids:
+            utils.warn(
+                message=(
+                    f"During pagination duplicate object(s) with id(s) "
+                    f"{duplicate_ids} returned from Gitlab and filtered"
+                ),
+                category=UserWarning,
+            )
+        self._data = [o for o in self._data if o["id"] not in duplicate_ids]
+        self._retrieved_object_ids.update(o["id"] for o in self._data)
 
         self._current = 0
 


gitlab/client.py
```diff
@@ -1174,6 +1174,8 @@ def __init__(
         # Preserve kwargs for subsequent queries
         self._kwargs = kwargs.copy()
 
+        self._prev_page_objects = []
```
nejch, line 1177: Are we sure it's enough to just compare against the previous page? Or is there a chance that an item from one of the earlier pages is repeated (e.g. by using a certain sort parameter that ends up returning an old item again)?

gitlab/client.py
```diff
@@ -1199,12 +1201,26 @@ def _query(
...
         try:
-            self._data: List[Dict[str, Any]] = result.json()
+            data: List[Dict[str, Any]] = result.json()
         except Exception as e:
             raise gitlab.exceptions.GitlabParsingError(
                 error_message="Failed to parse the server message"
             ) from e
 
+        self._data = []
+        for item in data:
```
nejch, line 1211: I think this should not be the default behavior, but conditional on an argument we can supply. For example something like `if remove_duplicates:`.

nejch, line 1211: Are we manually looping over items and adding them one by one for performance reasons here? Or could we just do a `list(OrderedDict.fromkeys(data))` and emit a single warning with the difference from the initial data vs. filtered?

nejch (changes requested):
Thanks for working on this @ericfrederich, if this is important to users we can can add it to the client.

However I'd add this as a `feat` rather than a fix as I think it's expected that offset pagination is not reliable, so we're providing an enhancement for getting around GitLab's limitation here, not fixing a bug in python-gitlab.

I just have a few additional questions :bow:

max-wittig:
Couldn't this one cause `OutOfMemory` errors as we need to store the dataset forever? What's the problem in handling this in client code?

nejch:
> Couldn't this one cause `OutOfMemory` errors as we need to store the dataset forever? What's the problem in handling this in client code?

I think this is why only the previous page is stored in this implementation (so max 100 entries). Not sure if this 100% ensures deduplication (see my comment above), but it's probably not as bad.

gitlab/client.py
ericfrederich: We could remove the loop and make the code more concise.  As far as complexity goes, doing a `set` difference still has a loop it's just abstracted away and done in C instead of Python.  I imagine the performance difference (loop is max 100 items) is negligible compared to the network API call.

I will change the code though and update this PR.

ericfrederich: If this were a feature offered an an option I have a hard time thinking anyone would opt to have the duplicates returned.

gitlab/client.py
ericfrederich: This duplication occurs when the sorting is done by ID and new objects are created because the new objects now exist on pages previously retrieved and the pagination requests have no context (i.e. no dedicated database curser for that query).  It's like in the UI looking at page 2 of something, then creating a new object in another tab, going back and refreshing page 2... everything shifted and stuff from page 1 (which you already retreived) is now on page 2.

I believe this duplication is independent of sort order but more pronounced with the default ID sorting because any new objects will result in duplication.

Let's think about any other kind of sorting... for example use alphabetical sorting.
In this case new objects which would appear on pages higher than your current page would not result in duplication, but objects which would appear on previous pages would result in duplication.
You're listing projects, you've got all the A's, B's, and are in the middle of listing projects starting with C's.  A new project called `Zoo` is created.  You will not see duplication, but if a new project called `Apples` is created the shift will happen and you'll get a duplicated project starting with C that you got on the previous page.

I do not believe it's necessary to store any extra items other than the ones from the previous page.

ericfrederich: I almost hate to mention it but there may be a case where the previous page is not enough, but it doesn't have to do with other sort parameters.

If between pages, more items are created on previous pages than the `per_page` number, then it would be possible for duplicates to occur.

The rarity of this occurring depends heavily on how the user code is implemented.

```python
# This code has higher possibility of 100 projects being created between pagination calls
for project in gl.projects.list(iterator=True):
  something_that_takes_a_long_time(project)

# This code has lower possibility of 100 projects being created between pagination calls
projects = gl.projects.list(get_all=True)
for project in projects:
  something_that_takes_a_long_time(project)
```

If we're worried about `OutOfMemory` errors as @max-wittig mentioned, the entire thing could be refactored to work off of sets of integers (objects IDs) instead of list of objects.

ericfrederich:
> Couldn't this one cause `OutOfMemory` errors as we need to store the dataset forever? What's the problem in handling this in client code?

Note: I've also made remarks above on possibly changing it to use a set of integers instead of objects which should help immensely with memory consumption.

I just wanted to mention here though that it would not not be forever, just during pagination.  The generator holding reference to the `GitlabList` with all this data can be garbage collected.

gitlab/client.py
ericfrederich: Actually, looks like we can't use `set()` or `OrderedDict` on these as here the objects are still just unhashable dicts.

I like option 2 as it's more concise.  Also perhaps CPython has optimized list comprehensions where option 1 is a generic for loop.

If we only keep a single page's previous objects we can choose between these two options.

If we decide to keep all objects, we should instead use a `set()` of `int`s.  So maybe let's decide that first.

## option 1

One loop; calculate dupes and filter within a single loop

```python
duplicates = []
self._data = []
for item in data:
    if item in self._prev_page_objects:
        duplicates.append(item)
        continue
    self._data.append(item)
```

## option 2

Two loops; one to calculate dupes, another to do the filtering.

```python
# loop once to calculate dupes
duplicates = [o for o in data if o in self._prev_page_objects]
# loop again to remove the dupes
self._data = [o for o in data if o not in dupes]
```

## single warning

In both cases we can defer the warning until all duplicates are detected and emit a single warning.

```python
if duplicates:
    utils.warn(
        message=(
            f"During pagination duplicate object with id(s) "
            f"{[item['id'] for item in duplicates]} returned from Gitlab and filtered"
        ),
        category=UserWarning,
    )
```

ericfrederich:
Sorry for ignoring this for 2 weeks, I was watching for activity on the "issue" and wasn't checking this "pr".

I see 3 open discussion points which need decisions (or just 2 decisions depending on the results of the others).

- retain last page of dicts or all pages of object ids.
- make deduplication optional? and if so what should default be?
- code style of for loop iteration (goes away if we decide to retain all object is's as they'd be ints and we could use sets) 

I'll refrain from changing code until we have consensus on those decision points.

I just wanted to apologize, summarize and communicate that I'm currently waiting for decisions.  We can use the dedicated conversations above for each of the points.

ericfrederich:
Updated the PR branch.

- It now holds all object ids in the case that more objects then `per_page` were added during pagination
- Single warning with all duplicated ids rather than a warning per item
- Deduplication is optional

gitlab/client.py
```diff
@@ -1205,6 +1209,21 @@ def _query(
                 error_message="Failed to parse the server message"
             ) from e
 
+        if self._dedupe:
+            duplicate_ids = (
+                set(o["id"] for o in self._data) & self._retrieved_object_ids
```
nejch, line 1214: Another issue, as you can see from all the failing tests: we're not actually guaranteed to have `id` attributes returned, some endpoints will return a different attribute as the unique identifier (we use `_id_attr` in our classes for this reason) and `http_list()` could potentially be used to return arbitrary paginated data.

This makes me think it would almost be better to do this in `ListMixin` based on the presence of `self._objc_cls._id_attr`, but not sure if it's too late at that stage to do it efficiently:

https://github.com/python-gitlab/python-gitlab/blob/541a7e3ec3f685eb7c841eeee3be0f1df3d09035/gitlab/mixins.py#L248-L250

nejch (changes requested):
Thanks again and sorry for the delay @ericfrederich.

I have some more concerns, see my comments. It's a bit tricky to get this right in a generic way it seems, due to the nature of what gets returned in paginated endpoints.

If feasible it would be great if we could test this somehow, based on the examples you outlined in the issue.

PR #2997, opened 2024-09-24
chore: update pylint to 3.3.1 and resolve issues

JohnVillalovos:
pylint 3.3.1 appears to have added "too-many-positional-arguments" check with a value of 5.

I don't disagree with this, but we have many functions which exceed this value. We might think about converting some of positional arguments over to keyword arguments in the future. But that is for another time.

For now disable the check across the project.


nejch (approved):
Makes sense for now and we can work on the rule(s) in a dedicated PR if needed :+1: Thanks!

PR #3002, opened 2024-09-30
fix(files): correctly raise GitlabGetError in get method

### LINKED ISSUE #2998: GitlabHttpError raised instead of GitlabGetError
`ProjectFileManager.get()` raises `GitlabHttpError` instead of `GitlabGetError` when trying to retrieve a file which does not exist.

## Expected Behavior
When calling `ProjectFileManager.get()`, it raises `GitlabGetError` if the given file does not exist.

## Actual Behavior
When calling `ProjectFileManager.get()`, it raises `GitlabHttpError` if the given file does not exist.

## Specifications

  - python-gitlab version: 4.12

nejch:
Closes #2998

### DIFF
--- a/gitlab/v4/objects/files.py
+++ b/gitlab/v4/objects/files.py
@@ -112,6 +112,7 @@ class ProjectFileManager(CreateMixin, UpdateMixin, DeleteMixin, RESTManager):
     @cli.register_custom_action(
         cls_names="ProjectFileManager", required=("file_path", "ref")
     )
+    @exc.on_http_error(exc.GitlabGetError)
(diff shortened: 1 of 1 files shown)


max-wittig:
@nejch Should we add a test for this somehow?

nejch:
@max-wittig I was also thinking about it initially, but not sure how we could do it explicitly while also covering other potential overrides. A simple test would only be for this specific case without adding more magic to collect all overrides.

Otherwise we'd have to go more in the direction of tests in https://github.com/python-gitlab/python-gitlab/tree/main/tests/unit/meta.

PR #3003, opened 2024-09-30
fix(ci): do not rely on GitLab.com runner arch variables

nejch:
See https://gitlab.com/python-gitlab/python-gitlab/-/jobs/7943404960 (aarch64 runner), seems like they run their ARM runners with an AMD64 helper binary/host.

Resulted in a failure on multi-arch build =>https://gitlab.com/python-gitlab/python-gitlab/-/jobs/7943404962

Also upstream docs:

> 	The OS/architecture of the GitLab Runner executable. Might not be the same as the environment of the executor.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -11,8 +11,8 @@ build-images:
     name: gcr.io/kaniko-project/executor:debug
     entrypoint: [""]
   script:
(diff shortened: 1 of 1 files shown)


JohnVillalovos (approved):
LGTM.

Thanks @nejch

nejch:
Thanks for the review @JohnVillalovos :bow:

nejch:
I'll try to release this now to see if it works and also to get https://github.com/python-gitlab/python-gitlab/issues/2998 back in.

Issue #2968, opened 2024-09-04
2024-10: Remove Python 3.8 as supported Python version


JohnVillalovos:
https://devguide.python.org/versions/

As of 2024-10 (October 2024), Python 3.8 will be End-of-Life.

We should remove it as a supported Python version at that point.

PR #3005, opened 2024-10-01
feat: remove support for Python 3.8, require 3.9 or higher

JohnVillalovos:
Python 3.8 is End-of-Life (EOL) as of 2024-10 as stated in https://devguide.python.org/versions/ and
https://peps.python.org/pep-0569/#lifespan

By dropping support for Python 3.8 and requiring Python 3.9 or higher it allows python-gitlab to take advantage of new features in Python 3.9, which are documented at:
https://docs.python.org/3/whatsnew/3.9.html

Closes: #2968

BREAKING CHANGE: As of python-gitlab 5.0.0, Python 3.8 is no longer supported. Python 3.9 or higher is required.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -26,8 +26,6 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         python:
-          - version: "3.8"
-            toxenv: py38,smoke
           - version: "3.9"
             toxenv: py39,smoke
           - version: "3.10"
--- a/README.rst
+++ b/README.rst
@@ -51,7 +51,7 @@ Features
 Installation
 ------------
 
-As of 4.0.0, ``python-gitlab`` is compatible with Python 3.8+.
+As of 5.0.0, ``python-gitlab`` is compatible with Python 3.9+.
 
 Use ``pip`` to install the latest stable version of ``python-gitlab``:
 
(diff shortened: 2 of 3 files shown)


nejch:
Thanks for this @JohnVillalovos. Should we take the opportunity here to add support for some syntax features in 3.9? https://docs.python.org/release/3.9.0/whatsnew/3.9.html. Specifically:

https://peps.python.org/pep-0585/
https://peps.python.org/pep-0584/ (if used at all and where it makes sense)

Maybe also https://peps.python.org/pep-0604/ if we add a future import.

JohnVillalovos:
> Thanks for this @JohnVillalovos. Should we take the opportunity here to add support for some syntax features in 3.9? https://docs.python.org/release/3.9.0/whatsnew/3.9.html. Specifically:
> 
> https://peps.python.org/pep-0585/ https://peps.python.org/pep-0584/ (if used at all and where it makes sense)
> 
> Maybe also https://peps.python.org/pep-0604/ if we add a future import.

Sure!  Once we drop the support for Python 3.8, we can use whatever we want in the future.

JohnVillalovos:
@nejch Please let me know if there is anything else I should do on this.

max-wittig (approved):

nejch (approved):
@JohnVillalovos sorry, for some reason I thought we were dependent on the 3.13 release that got postponed, but just realized this only drops support for the old one. I just did another PR earlier today to have a mini release before this breaking change :+1:

Issue #3004, opened 2024-10-01
Correctly raise GitlabHeadError in head method


acudovs:
## Description of the problem, including code/CLI snippet

GitlabHeadError should be raised in https://github.com/python-gitlab/python-gitlab/blob/main/gitlab/v4/objects/files.py#L140

## Expected Behavior

GitlabHeadError

## Actual Behavior

GitlabHttpError

## Specifications

  - python-gitlab version: v4.12.1

acudovs:
Here is also pydoc https://github.com/python-gitlab/python-gitlab/blob/main/gitlab/v4/objects/files.py#L153

PR #3006, opened 2024-10-01
fix: raise GitlabHeadError in `project.files.head()` method

JohnVillalovos:
When an error occurs, raise `GitlabHeadError` in
`project.files.head()` method.

Closes: #3004


nejch (approved):
Thanks @JohnVillalovos :bow: didn't even think this would be an issue

Issue #2832, opened 2024-03-28
Feature: add `pages` support in `project`


Karreg:
## Description of the problem, including code/CLI snippet

I was not able to determine how to access to [`/api/v4/projects/:id/pages`](https://docs.gitlab.com/ee/api/pages.html) with python-gitlab. I'm not sure to have searched well, but the [`Project`object](https://python-gitlab.readthedocs.io/en/stable/api/gitlab.v4.html#gitlab.v4.objects.Project) does not seem to provide such access.

Would it be possible to add it?

The use case behind is I want to provide a public script to add a badge linking to the deployed pages, which requires such API access.

## Specifications

  - python-gitlab version: 4.4.0
  - API version you are using (v3/v4): v4
  - Gitlab server version (or gitlab.com): 16.9.2

Thanks for you great API wrapper, it's a real improvement over standard GitLab API!

nejch:
@Karreg thanks for the report! This is available now in https://github.com/python-gitlab/python-gitlab/releases/tag/v4.13.0.

PR #3009, opened 2024-10-08
feat(api): add support for project Pages API

nejch:
## Changes

Implements https://docs.gitlab.com/ee/api/pages.html
Closes https://github.com/python-gitlab/python-gitlab/issues/2832

### Documentation and testing

### DIFF
--- a/docs/gl_objects/pagesdomains.rst
+++ b/docs/gl_objects/pagesdomains.rst
@@ -1,9 +1,38 @@
-#############
-Pages domains
-#############
+#######################
+Pages and Pages domains
+#######################
 
-Admin
-=====
+Project pages
+=============
+
+References
+----------
+
+* v4 API:
+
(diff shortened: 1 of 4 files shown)


docs/gl_objects/pagesdomains.rst
```diff
@@ -1,9 +1,38 @@
...
+Examples
+--------
+
+Get Pages settings for a project::
+
+    pages = project.pages.get()
+
+Update Pages settings for a project::
+
+    project.pages.update(new_data={'pages_https_only': True})
```
nejch, line 28: GitLab uses PATCH for this endpoint, and the attributes between get/patch are inconsistent (`force_https` returned from GET vs `pages_https_only` needed for PATCH).

So we can't use the SaveMixin here unless we do some magic mapping and only send the relevant data to the server as part of calling `save()`.

PR #3010, opened 2024-10-08
refactor: use more python3.9 syntax

nejch:
Follow-up to https://github.com/python-gitlab/python-gitlab/pull/3005

Mostly done with `pyupgrade --py39-plus` and some manual cleanup.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,5 +1,4 @@
 #!/usr/bin/env python3
-# -*- coding: utf-8 -*-
 #
 # python-gitlab documentation build configuration file, created by
 # sphinx-quickstart on Mon Dec  8 15:17:39 2014.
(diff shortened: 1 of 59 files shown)


JohnVillalovos:
Cool! I was thinking about this too 🙂  Well stop using `List[]` and instead use `list[]` style type annotations.

Thanks for doing this!

Once we (in reality you) get this done, we should do a follow-up adding the commit to `.git-blame-ignore-revs` which I'm sure you have already thought of.

nejch:
> Once we (in reality you) get this done, we should do a follow-up adding the commit to .git-blame-ignore-revs which I'm sure you have already thought of.

I was thinking of that as well, especially now that I see how big the diff is :sweat_smile: Maybe after this PR in case there's more rebasing etc.

I'll just need to tweak the type hint tests for this now.

nejch:
I'll rebase this after #3083 to avoid too many conflicts there, and that will likely fix the typing issues altogether.

nejch:
@JohnVillalovos this one's a chunky one now, but mostly auto-generated :sweat_smile:

JohnVillalovos (approved):
Thanks @nejch   

A lot of work there.

PR #3011, opened 2024-10-09
Make repr() of project approval rules consistent with merge request approval rules

holysoles:
## Changes

The representation attribute for `ProjectMergeRequestApprovalRule` is `name` (see [here](https://github.com/python-gitlab/python-gitlab/blob/main/gitlab/v4/objects/merge_request_approvals.py#L139)), which makes sense based on the API documentation. However, the API for `ProjectApprovalRule` is [nearly the same](https://docs.gitlab.com/ee/api/merge_request_approvals.html#create-project-level-rule), but we don't set the representation attribute to `name` for that class (see [here](https://github.com/python-gitlab/python-gitlab/blob/main/gitlab/v4/objects/merge_request_approvals.py#L55)).

This MR sets `_repr_attr` for the `ProjectApprovalRule` class , as well as adds a unit test for ensure we can list a project's approval rules, and call `repr()` on it.

### Documentation and testing

### DIFF
--- a/tests/unit/objects/test_project_merge_request_approvals.py
+++ b/tests/unit/objects/test_project_merge_request_approvals.py
@@ -25,6 +25,135 @@ updated_approval_rule_approvals_required = 1
 
 
 @pytest.fixture
+def resp_prj_approval_rules():
+    prj_ars_content = [
+        {
+            "id": approval_rule_id,
+            "name": approval_rule_name,
+            "rule_type": "regular",
+            "report_type": None,
+            "eligible_approvers": [
+                {
+                    "id": user_ids[0],
+                    "name": "John Doe",
+                    "username": "jdoe",
+                    "state": "active",
+                    "avatar_url": "https://www.gravatar.com/avatar/0?s=80&d=identicon",
+                    "web_url": "http://localhost/jdoe",
+                },
+                {
+                    "id": user_ids[1],
+                    "name": "Group Member 1",
+                    "username": "group_member_1",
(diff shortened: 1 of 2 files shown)


tests/unit/objects/test_project_merge_request_approvals.py
```diff
@@ -169,6 +298,14 @@ def test_project_approval_manager_update_method_post(project):
     assert approvals._update_method is UpdateMethod.POST
 
 
+def test_list_project_approval_rules(project, resp_prj_approval_rules):
+    approval_rules = project.approvalrules.list()
+    assert len(approval_rules) == 1
+    assert approval_rules[0].name == approval_rule_name
+    assert approval_rules[0].id == approval_rule_id
+    repr(approval_rules)  # ensure that `repr()` doesn't raise an exception
```
nejch, line 306: Question - if we're adding the test, shouldn't we be able to assert on the string returned by the repr here instead of having a comment?

nejch (commented):
Thanks a lot @holysoles looks great! Just a quick question.

tests/unit/objects/test_project_merge_request_approvals.py
holysoles: we certainly can, I had just mirrored the similar test being done in the function below this one. I'll update mine and happy to update the other if desired.

holysoles: updated

nejch (approved):

PR #3012, opened 2024-10-11
chore: add Python 3.13 as supported

JohnVillalovos:
Mark that Python 3.13 is supported.

Use Python 3.13 for the Mac and Windows tests.

Also remove the 'py38' tox environment. We no longer support Python 3.8.


JohnVillalovos:
@nejch I made this PR to only be about Python 3.13. Not enabling testing of Python 3.14 development code.

nejch (approved):
Thanks @JohnVillalovos, forgot we didn't already so this in the previous PR 😅

PR #3047, opened 2024-11-26
feat(const): add new Planner role to access levels

nejch:
See https://docs.gitlab.com/ee/api/members.html#roles

Context: https://about.gitlab.com/blog/2024/11/25/introducing-gitlabs-new-planner-role-for-agile-planning-teams/


matthias-reiff:
FYI, I added the following issue as the constant can not be imported easily currently: 
#3221

Issue #3039, opened 2024-11-14
Support creation of pull mirrors


nicoklaus15:
## Description of the problem, including code/CLI snippet

I'm at writing a script to automate the creation of GitLab Mirrors with several predefined project settings and so on. 

However, it seems that the API Endpoint [configure-pull-mirroring-for-a-project](https://docs.gitlab.com/ee/api/project_pull_mirroring.html#configure-pull-mirroring-for-a-project) is not to covered by python-gitlab.

The current implementation of [mirror_pull](https://python-gitlab.readthedocs.io/en/stable/api/gitlab.v4.html#gitlab.v4.objects.Project.mirror_pull) only covers the start of the pull process but not the configuration.

```python
data = {
    "enabled": True,
    "url": "https://github.com/python-gitlab/python-gitlab.git",
    "mirror_overwrites_diverged_branches": True
}

mirror = project.mirror_pull(data)

Exception has occurred: TypeError
Project.mirror_pull() takes 1 positional argument but 2 were given
  File "/workspaces/Development/tmp/external-mirrors/external-mirror.py", line 97, in create_project
    mirror = project.mirror_pull(data)
             ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspaces/Development/tmp/external-mirrors/external-mirror.py", line 122, in <module>
    if create_project(gitlab_mirror['destination'], gitlab_mirror['name'], gitlab_mirror['source']):
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: Project.mirror_pull() takes 1 positional argument but 2 were given
```

Maybe through the deprecation of the same API Endpoint in the past, the functionality was removed ?

https://docs.gitlab.com/ee/api/project_pull_mirroring.html#configure-pull-mirroring-for-a-project-deprecated

## Expected Behavior

Would configure the project to mirror "https://github.com/python-gitlab/python-gitlab.git".

```python
data = {
    "enabled": True,
    "url": "https://github.com/python-gitlab/python-gitlab.git",
    "mirror_overwrites_diverged_branches": True
}

mirror = project.mirror_pull(data)
```

## Actual Behavior

The additional arguments are not processed as the function only start the mirror pulling instead of configuration.

```
Exception has occurred: TypeError
[...]

PR #3056, opened 2024-12-04
feat(project): add support for pull mirror configuration

nicoklaus:
Add new pull mirroring api endpoint released in GitLab 17.6.

https://docs.gitlab.com/17.6/ee/api/project_pull_mirroring.html#configure-pull-mirroring-for-a-project

## Changes

Adds mirror_pull_configuration to projects.

Related to #3039

### Documentation and testing

### Additional Notes

Since the implementation of this API is slightly different from the standard that GitLab uses, there are small drawbacks to disabling pull mirrors.

There is no separate DELETE method, we use the PUT method in both cases to enable and disable pull mirroring. Since a URL is required for activation, this is now also required for deactivation, even though it is not actually needed for this.

I also noticed that the settings for mirror_regex are not deleted when deactivated, unlike the other settings. Therefore there is now an extra if condition for this. I'm not sure if this should be filed as a question / bug to GitLab ?


nejch:
Thanks a lot @nicoklaus, looks great in general!

I'm just thinking whether we should consolidate the GET, PUT and POST methods for `/pull/mirror` in a single manager+object so that we don't need the custom handling here, and that way people can just treat things like `.enabled` as another attribute to call `.save()` on. 

Just have to think about a clean way to do this, maybe `POST` would make more sense as a `.start()` method. I'll come back to you with some ideas ASAP :bow:

nicoklaus:
Hi @nejch, I had a little time and reworked the implementation based on your input. 🔨 would be pleased about another review 😄

docs/gl_objects/projects.rst
```diff
@@ -246,14 +246,6 @@ Get a list of users for the repository::
     # search for users
     users = p.users.list(search='pattern')
 
-Start the pull mirroring process (EE edition)::
-
-    project.mirror_pull()
-
-Get a project’s pull mirror details (EE edition)::
-
-    mirror_pull_details = project.mirror_pull_details()
```
nejch, line 255: Completely removing these 2 would be a breaking change, which we'd need to signal with a `BREAKING CHANGE` git trailer to trigger a new major version.

Alternatively, we can just keep it as a thin wrapper with our utility helper to add a deprecation warning. See this example:

https://github.com/python-gitlab/python-gitlab/blob/36d9b24ff27d8df514c1beebd0fff8ad000369b7/gitlab/v4/objects/snippets.py#L193-L200

Let me know if that works for you! That way we can remove all the deprecated methods together later.

nejch (changes requested):
> Hi @nejch, I had a little time and reworked the implementation based on your input. 🔨 would be pleased about another review 😄

Thanks a lot @nicoklaus, this is exactly what I had in mind. Looks great! Just a comment on the impact of removing the old code, let me know if you need some help :bow:

docs/gl_objects/projects.rst
nicoklaus: I have added the warning about the deprecation. As with the snippets example, I have removed the documentation so that new users do not use the obsolete functions. I hope this is correct, or should we keep the documentation until the major release?

nejch (approved):

PR #3074, opened 2025-01-02
feat(api): Narrow down return type of download methods using typing.overload

igorp-collabora:
## Changes

Currently the download methods such as `ProjectJob.artifacts` have return type set to `Optional[Union[bytes, Iterator[Any]]]` which means they return either `None` or `bytes` or `Iterator[Any]`.

However, the actual return type is determined by the passed `streamed` and `iterator` arguments. Using `@typing.overload` decorator it is possible to return a single type based on the passed arguments.

Add overloads in the following order to all download methods:

1. If `streamed=False` and `iterator=False` return `bytes`. This is the default argument values therefore it should be first as it will be used to lookup default arguments.
2. If `iterator=True` return `Iterator[Any]`. This can be combined with both `streamed=True` and `streamed=False`.
3. If `streamed=True` and `iterator=False` return `None`. In this case `action` argument can be set to a callable that accepts `bytes`.

This will eliminate unnecessary type checking for library users and also highlight some mistakes such as passing `action=` argument without setting `streamed=True`. (as `action` argument is never used without `streamed=True` `iterator=False`)

### DIFF
--- a/gitlab/mixins.py
+++ b/gitlab/mixins.py
@@ -6,7 +6,9 @@ from typing import (
     Dict,
     Iterator,
     List,
+    Literal,
     Optional,
+    overload,
     Tuple,
     Type,
     TYPE_CHECKING,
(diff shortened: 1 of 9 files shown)


igorp-collabora:
To test type hints you can create a following file:

```python
import typing
from gitlab.mixins import DownloadMixin


def test_mixin(mix: DownloadMixin) -> None:
    typing.reveal_type(
        mix.download(action=print, streamed=True)
    )
    typing.reveal_type(
        mix.download()
    )
    typing.reveal_type(
        mix.download(
            streamed=False,
            action=None,
            iterator=False,
        )
    )
    typing.reveal_type(
        mix.download(
            streamed=False,
            action=None,
            iterator=True,
        )
    )
```

Then you can run type checker on it like mypy and it will reveal the types:

```
$ mypy --follow-imports=silent --strict ./typing_test.py
typing_test.py:9: note: Revealed type is "None"
typing_test.py:12: note: Revealed type is "builtins.bytes"
typing_test.py:15: note: Revealed type is "builtins.bytes"
typing_test.py:22: note: Revealed type is "typing.Iterator[Any]"
```

igorp-collabora:
There is probably more room to add `typing.overload` to other methods. For example, for methods like `http_get` the return type depends on arguments passed.

JohnVillalovos:
Thanks. This looks cool. I'll try to review it within the next few days.

igorp-collabora:
Hmmm... Not sure why CI is failing. It does not seem like it is caused by these changes?

igorp-collabora:
I rebased against main branch.

JohnVillalovos (approved):
Thanks @igorp-collabora  Nice improvement for typing 👍

igorp-collabora:
The documentation now also shows different overloads the download functions can have: https://python-gitlab.readthedocs.io/en/latest/api/gitlab.v4.html#gitlab.v4.objects.ProjectJob.artifact

PR #3080, opened 2025-01-07
feat(api): Use protocols and overload to type hint mixins

### LINKED ISSUE #2062: Type hint on ListMixin.list is too strict

### LINKED ISSUE #2228: Some mixins are not properly type-hinted
E.g. for the `CreateMixin` the `.create(...)` method returns things like [`ProjectMergeRequest`](https://github.com/python-gitlab/python-gitlab/blob/181390a4e07e3c62b86ade11d9815d36440f5817/gitlab/v4/objects/merge_requests.py#L143) but not the `base.RESTObject` as specified [here](https://github.com/python-gitlab/python-gitlab/blob/181390a4e07e3c62b86ade11d9815d36440f5817/gitlab/mixins.py#L274).

---
`python-gitlab` version: 3.5.0

igorp-collabora:
## Changes

Currently mixins like ListMixin are type hinted to return base RESTObject instead of a specific class like `MergeRequest`.

The GetMixin and GetWithoutIdMixin solve this problem by defining a new `get` method for every defined class. However, this creates a lot of duplicated code.

`typing.Protocol` can be used to type hint that the mixed in method will return a class matching attribute `_obj_cls`. The type checker will lookup the mixed in class attribute and adjust the return type accordinly.

Delete `tests/unit/meta/test_ensure_type_hints.py` file as the `get` method is no required to be defined for every class.

Closes #2228 
Closes #2062


igorp-collabora:
This turned out to be a larger commit than I expected. I can probably split it by a mixin. For example, first commit for Get and GetWithoutId mixins, second one for ListMixin...

Also this is a pretty ad-hoc solution but requires the least amount of changes. Ideally making a `RESTManager` generic would be better as it would remove all `overload`. However, this MR can be a starting point to reworking the `RESTManager` hierarchy.

gitlab/client.py
```diff
@@ -397,10 +397,11 @@ def auth(self) -> None:
         The `user` attribute will hold a `gitlab.objects.CurrentUser` object on
         success.
         """
-        self.user = self._objects.CurrentUserManager(self).get()
+        user = self._objects.CurrentUserManager(self).get()
+        self.user = user
```
igorp-collabora, line 401: For some reason `self.user` will now be type evaluated as `Optional[User]`. I had to separate user to a new variable that can't be None.

tests/functional/conftest.py
```diff
@@ -104,12 +104,12 @@ def reset_gitlab(gl: gitlab.Gitlab) -> None:
             )
             continue
 
-        for deploy_token in group.deploytokens.list():
+        for group_deploy_token in group.deploytokens.list():
```
igorp-collabora, line 107: Because `list()` now returns a list containing a specific type it would case a type error here because the same variable is used between two different types.

gitlab/v4/objects/users.py
```diff
@@ -4,7 +4,9 @@
 https://docs.gitlab.com/ee/api/projects.html#list-projects-starred-by-a-user
 """
 
-from typing import Any, cast, Dict, List, Optional, Union
+from __future__ import annotations
```
igorp-collabora, line 7: For some reason without delayed annotations type checks failed to resolve types for `UserProject.list`. This means the quotes around the type annotation had to be removed or managers would not be initialized.

JohnVillalovos:
@igorp-collabora Thanks for this. Not ignoring your PRs. Just I haven't had time yet to fully review them. Hopefully this week or next week. Sorry for the delay.

Thanks for the PRs! They look interesting from the description 👍

igorp-collabora:
> Sorry for the delay.

No worries. I want to experiment with reworking RESTManager hierarchy meanwhile.

igorp-collabora:
The #3083 is a cleaner solution but more verbose. This one builds more work-around code on top of existing one.

PR #3087, opened 2025-01-17
feat(api): Narrow down return type of ProjectFileManager.raw using typing.overload

### LINKED ISSUE #1806: Type overloads needed for functions with streamed parameter

Most of the functions with a `streamed: bool` parameter currently return a `Union[..., requests.Response]` or `Optional[bytes]` type. In fact, the received return type can usually be inferred by type checkers based on the `streamed` parameter's value. For that, `typing.overload`s would have to be added to the function signatures.

## Specifications

  - python-gitlab version: 3.0.0
  - Gitlab server version (or gitlab.com): -

igorp-collabora:
This is equivalent to the changes in 44fd9dc1176a2c5529c45cc3186c0e775026175e but for `ProjectFileManager.raw` method that I must have missed in the original commit.

Closes #1806

### DIFF
--- a/gitlab/v4/objects/files.py
+++ b/gitlab/v4/objects/files.py
@@ -5,7 +5,9 @@ from typing import (
     Dict,
     Iterator,
     List,
+    Literal,
     Optional,
+    overload,
     Tuple,
     TYPE_CHECKING,
     Union,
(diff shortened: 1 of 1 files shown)


nejch (approved):
Thanks here as well @igorp-collabora, looks good and I just briefly checked locally with our test code :+1:

Issue #3091, opened 2025-01-21
CI: Functional CLI test keeps hanging on GitHub CI.


JohnVillalovos:
In the functional CLI job I keep seeing the test hanging/get stuck. 

The last test it finishes is typically:
`tests/functional/cli/test_cli.py::test_cli_does_not_print_token[subprocess] PASSED [ 22%]`

The test that should be next is:
`tests/functional/cli/test_cli_artifacts.py::test_cli_job_artifacts`

JohnVillalovos:
@nejch 

I think the `job_with_artifacts` fixture is failing. And it loops forever as it never gets job artifacts.

Debug PR here: https://github.com/python-gitlab/python-gitlab/pull/3092

JohnVillalovos:
The job fails with the following error:

```
Running with gitlab-runner 17.8.1 (a1fca00e) on python-gitlab-runner t1_p_wyG, system ID: r_6gTGeD1nqDhZ
Resolving secrets
Preparing the "shell" executor
Using Shell (bash) executor...
Preparing environment
su: user gitlab-runner does not exist
ERROR: Job failed: prepare environment: exit status 1. Check https://docs.gitlab.com/runner/shells/index.html#shell-profile-loading for more information
```

JohnVillalovos:
I have filed an issue about the missing user: https://gitlab.com/gitlab-org/gitlab-runner/-/issues/38503

nejch:
Thanks a lot @JohnVillalovos! Seems like they've completely reworked the runner helper image, which is causing a lot of issues including this one.

PR #3093, opened 2025-01-21
ci: use gitlab-runner:v17.7.1 for the CI

JohnVillalovos:
The `latest` gitlab-runner image does not have the `gitlab-runner` user and it causes our tests to fail.

Closes: #3091


tests/functional/cli/test_cli_artifacts.py
```diff
@@ -24,12 +24,18 @@
...
     project.files.create(data)
 
     jobs = None
     while not jobs:
         time.sleep(0.5)
         jobs = project.jobs.list(scope="success")
+        if time.time() - start_time > 60:
+            for job in jobs:
+                job.pprint()
+            assert None, "Fixture 'job_with_artifact' failed"
```
nejch, line 38: Are these leftover from debugging?

tests/functional/fixtures/docker-compose.yml
```diff
@@ -45,7 +45,7 @@ services:
       - gitlab-network
 
   gitlab-runner:
-    image: gitlab/gitlab-runner:latest
+    image: '${GITLAB_RUNNER_IMAGE}:${GITLAB_RUNNER_TAG}'
```
nejch, line 48: Wow that's a lot of config hehe, thanks. I guess we won't need this often but doesn't hurt :slightly_smiling_face:

nejch (commented):
Thanks @JohnVillalovos, looks good, just a question about the prints?

tests/functional/cli/test_cli_artifacts.py
JohnVillalovos: No, those are in case it fails we can get some debug info.

tests/functional/fixtures/docker-compose.yml
JohnVillalovos: Oh, I was copying how there is a `GITLAB_IMAGE` and `GITLAB_TAG` in the same file. Do you want to change it to just be a hard-coded version?

nejch: All good, I just didn't expect we'd need to customize this, but we can keep it the same way as here just in case!

tests/functional/cli/test_cli_artifacts.py
```diff
@@ -24,12 +24,24 @@
...
+        if time.time() - start_time > 60:
+            print("ERROR: job never succeeded")
+            for job in project.jobs.list():
+                job = project.jobs.get(job.id)
+                job.pprint()
+                if job.status == "failed":
+                    print()
+                    print("job log/trace:", job.trace())
+                print()
+            assert None, "Fixture 'job_with_artifact' failed"
```
nejch, line 44: What if we also have jobs that stuck running? Maybe we could just log them?

```suggestion
# import logging

        if time.time() - start_time < 60:
            continue

        logging.error("job never succeeded")
        for job in project.jobs.list():
            job = project.jobs.get(job.id)
            logging.debug(f"{job.status} job: {job.pprint()}")
            logging.debug(f"job log:\n{job.trace()}\n")

        pytest.fail("Fixture 'job_with_artifact' failed")
```

JohnVillalovos: Looks good. I went for `logging.info()` for the status. As I think if this fails we want to see this information.

nejch (approved):

PR #3095, opened 2025-01-22
Update groups.rst

rnelson01:
added docs example get all group members + access level

### Documentation and testing

### DIFF
--- a/docs/gl_objects/groups.rst
+++ b/docs/gl_objects/groups.rst
@@ -288,6 +288,15 @@ ancestor groups)::
 
     members = group.members_all.list(get_all=True)
 
+List all groups and all group member with access level::
+
+    groups = gl.groups.list(get_all=True)
+    for group in groups:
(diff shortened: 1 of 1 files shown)


docs/gl_objects/groups.rst
```diff
@@ -288,6 +288,15 @@ ancestor groups)::
 
     members = group.members_all.list(get_all=True)
 
+List all groups and all group member with access level::
```
nejch, line 291: This feels like it's covering a bit more than just the group members API and almost like a small cookbook example. WDYT about narrowing it down to just the API calls for members, or moving this example to a separate documentation page for longer examples?

nejch (changes requested):
Thanks a lot for your contribution @rnelson01!

The commit message will need to be changed to pass our CI, see our contribution guidelines if you need help. I just have one additional question as well.

PR #3097, opened 2025-01-22
fix(api): return the new commit when calling cherry_pick

### LINKED ISSUE #3096: ProjectCommit.cherry_pick should return the created commit

I noticed that `ProjectCommit.cherry_pick` returns `None` even though the API returns the created commit in the response. See: https://docs.gitlab.com/ee/api/commits.html#cherry-pick-a-commit

## Expected Behavior

Return the created commit.

## Actual Behavior

Returns `None`.

## Specifications

  - python-gitlab version: `5.3.1`
  - Gitlab server version (or gitlab.com): `gitlab.com`

mschoettle:
## Changes

Returns the response when calling `ProjectCommit.cherry_pick`.

### Documentation and testing

Fixes #3096

### DIFF
--- a/gitlab/v4/objects/commits.py
+++ b/gitlab/v4/objects/commits.py
@@ -48,7 +48,9 @@ class ProjectCommit(RESTObject):
 
     @cli.register_custom_action(cls_names="ProjectCommit", required=("branch",))
     @exc.on_http_error(exc.GitlabCherryPickError)
-    def cherry_pick(self, branch: str, **kwargs: Any) -> None:
+    def cherry_pick(
+        self, branch: str, **kwargs: Any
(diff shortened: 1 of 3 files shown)


nejch (changes requested):
Thanks a lot @mschoettle, makes sense! I think this must be very old code.

Would you mind adding a test for this, maybe just in the functional tests if it's easier than mocking? I know we didn't have one before, but just in case. Let me know if you need help.

mschoettle:
@nejch Thanks! I noticed yesterday that there were no tests for `cherry_pick` so I started adding a unit test and functional test based on the existing tests.

I am having an issue with running the functional tests in the codespace so I am relying on the pipeline for those.

PR #3110, opened 2025-01-29
chore: Remove trivial get methods in preparation for generic Get mixin

igorp-collabora:
Currently because the Get mixin is not generic every subclass has to implement its own `get` methods to type hint `get` method return beoynd the basic RESTObject.

The upcoming #3083 PR will change Get mixin to use generics. This means it will be able to type hint the corresponding `_obj_cls` and no `get` method would need to be redefined.

Because removing existing `get` methods modifies a lot of files split it in to a separate commit.

Also remove the `tests/unit/meta/test_ensure_type_hints.py` file because testing subclasses for `get` methods is no longer relevant.


gitlab/client.py
```diff
@@ -397,10 +397,10 @@ def auth(self) -> None:
         The `user` attribute will hold a `gitlab.objects.CurrentUser` object on
         success.
         """
-        self.user = self._objects.CurrentUserManager(self).get()
+        self.user = self._objects.CurrentUserManager(self).get()  # type: ignore[assignment]
 
         if hasattr(self.user, "web_url") and hasattr(self.user, "username"):
-            self._check_url(self.user.web_url, path=self.user.username)
+            self._check_url(self.user.web_url, path=self.user.username)  # type: ignore[union-attr]
```
igorp-collabora, line 403: This is the issue I described in this comment: https://github.com/python-gitlab/python-gitlab/pull/3083#discussion_r1932539880

I think it caused by `CurrentUserManager(self).get()` returning RESTObject so within this function's scope the `self.user` is still assumed to be None. However, once generic Get mixin gets merged the type ignore comments should no longer be needed. (and mypy will warn about unused ignore)

JohnVillalovos (approved):
WOW! Big PR, but relatively easy to review.

Thanks ❤️

nejch (approved):
Thanks a lot for splitting this out @igorp-collabora! Seeing all the removals isolated like that is super clear :+1:

PR #3111, opened 2025-01-30
feat(api): added support for tls client authentication against revers…

mgrechukh:
Certain reverse proxies may require TLS client pre-authentication in addition to further API authentication on application level. To facilitate CLI work in such environment, relevant config options added.


nejch (changes requested):
Thanks a lot for this contribution @mgrechukh!

Normally I'd say this is a perfect use for a custom session - but for the CLI, we'd really need this.

Would you be able to add a small test with a matcher to ensure the cert is passed?
https://github.com/getsentry/responses?tab=readme-ov-file#request-keyword-arguments-matcher

We should have examples from before in our tests for similar mocks, but let me know if it gets too difficult.

I'm also still thinking about the exact interface here. `requests` can take a tuple but also just a string to a combined cert file. `httpx` for example would only take the entire cert chain https://www.python-httpx.org/advanced/ssl/#client-side-certificates and not a tuple.

PR #3120, opened 2025-02-05
Reformat code with `skip_magic_trailing_comma = true`

JohnVillalovos:


nejch (approved):
Haha that's another chunky one @JohnVillalovos but I like the deletions :+1: I guess I should also add my other commit to the ignore revs.

max-wittig:
Magic trailing comma was the worst thing that black ever did.

PR #3121, opened 2025-02-06
feat(api)!: Make RESTObjectList typing generic

### LINKED ISSUE #2062: Type hint on ListMixin.list is too strict

igorp-collabora:
BREAKING CHANGE: Type narrowing of `list()` methods return can become redundant.

Currently the RESTObjectList type hints yielded objects as base RESTObject. However, the ListMixin is now generic and will return the RESTObject subclass based on the RESTManager typing.

Using `typing.Generic` it is possible to make RESTObjectList type hint a specific subclass of the RESTObject.

Iterating over `list()` call the ListMixin will now yield the same object class because both `list` and `RESTObjectList` will have the same type hinted subclass.

Closes #2062

### DIFF
--- a/gitlab/v4/objects/milestones.py
+++ b/gitlab/v4/objects/milestones.py
@@ -4,6 +4,7 @@ from gitlab import cli
 from gitlab import exceptions as exc
 from gitlab import types
 from gitlab.base import RESTObject, RESTObjectList
+from gitlab.client import GitlabList
 from gitlab.mixins import (
     CRUDMixin,
     ObjectDeleteMixin,
@@ -16,6 +17,7 @@ from gitlab.types import RequiredOptional
 from .issues import GroupIssue, GroupIssueManager, ProjectIssue, ProjectIssueManager
 from .merge_requests import (
     GroupMergeRequest,
+    GroupMergeRequestManager,
     ProjectMergeRequest,
     ProjectMergeRequestManager,
 )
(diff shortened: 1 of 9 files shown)


gitlab/v4/objects/milestones.py
```diff
@@ -53,14 +55,14 @@ def issues(self, **kwargs: Any) -> RESTObjectList:
         path = f"{self.manager.path}/{self.encoded_id}/issues"
         data_list = self.manager.gitlab.http_list(path, iterator=True, **kwargs)
         if TYPE_CHECKING:
-            assert isinstance(data_list, RESTObjectList)
+            assert isinstance(data_list, GitlabList)
```
igorp-collabora, line 58: Pretty sure there was some kind of incorrect assert here because `http_list` return type is `GitlabList | list[dict[str, Any]]`.

igorp-collabora:
I made this simple python file to check type hints:

```python
import typing
from gitlab.v4.objects import MergeRequestManager


def test_list(mang: MergeRequestManager) -> None:
    for m in mang.list():
        typing.reveal_type(m)
```

When run on main branch the revealed type will be:

```
typing_rest_list.py:7: note: Revealed type is "gitlab.base.RESTObject"
```

Using this branch the revealed type will be:

```
typing_rest_list.py:7: note: Revealed type is "gitlab.v4.objects.merge_requests.MergeRequest"
```

I believe this is because both `base.RESTObjectList[base.TObjCls] | list[base.TObjCls]` will yield `TObjCls` when iterated on. Not sure how much more value the overloads will bring but they can raise a deprecation warning when the `get_all` is not used.

gitlab/v4/objects/milestones.py
nejch: Makes sense. It was probably copy/pasted from one of the methods that returns full objects.

igorp-collabora:
I added the breaking changes footer to the commit message and an exclamation mark to the commit summary.

PR #3124, opened 2025-02-07
docs: Use get_all keyword arg instead of all in docstrings

igorp-collabora:
The `all` keyword had a lot of conflicts with other API parameters and was silently deprecated in favour of `get_all`. However, a lot of docstings were not updated to use the new keyword.


igorp-collabora:
I am also working on a second PR where `get_all` is used everywhere in the documentation examples.

nejch (approved):
Wow, nice catch, completely forgot about the docstrings. Thanks! :bow:

PR #3125, opened 2025-02-07
docs: Use list(get_all=True) in documentation examples

igorp-collabora:
The plain `list()` produces warnings to alert users that they might miss certain entities if pagination is not enabled using `get_all=True` to get a list of objects or `iterator=True` to have an iterator over all objects.

Use `get_all=False` where the list would me immediately indexed `[0]` and `iterator=True` where iteration is used.

This should help users encounter less warnings.

### DIFF
--- a/docs/api-usage-advanced.rst
+++ b/docs/api-usage-advanced.rst
@@ -34,7 +34,7 @@ properly closed when you exit a ``with`` block:
 .. code-block:: python
 
    with gitlab.Gitlab(host, token) as gl:
-       gl.projects.list()
+       gl.statistics.get()
 
 .. warning::
 
--- a/docs/gl_objects/deploy_tokens.rst
+++ b/docs/gl_objects/deploy_tokens.rst
@@ -29,7 +29,7 @@ Use the ``list()`` method to list all deploy tokens across the GitLab instance.
 ::
 
     # List deploy tokens
-    deploy_tokens = gl.deploytokens.list()
+    deploy_tokens = gl.deploytokens.list(get_all=True)
 
 Project deploy tokens
 =====================
@@ -52,7 +52,7 @@ Examples
 
 List the deploy tokens for a project::
 
-    deploy_tokens = project.deploytokens.list()
+    deploy_tokens = project.deploytokens.list(get_all=True)
 
 Get a deploy token for a project by id::
 
@@ -109,7 +109,7 @@ Examples
 
 List the deploy tokens for a group::
 
-    deploy_tokens = group.deploytokens.list()
+    deploy_tokens = group.deploytokens.list(get_all=True)
 
 Get a deploy token for a group by id::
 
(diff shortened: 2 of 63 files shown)


docs/api-usage-advanced.rst
```diff
@@ -34,7 +34,7 @@ properly closed when you exit a ``with`` block:
 .. code-block:: python
 
    with gitlab.Gitlab(host, token) as gl:
-       gl.projects.list()
+       gl.projects.list(get_all=True)
```
nejch, line 37: I believe we initially didn't include all of these as it would potentially lead users to paginate over all instance projects/users on gitlab.com and store it all in memory, which would be a pretty huge list. For any sub-resource this wouldn't be that much of an issue.

But for the sake of consistency I think we can go ahead with this.

docs/gl_objects/deploy_tokens.rst
```diff
@@ -24,12 +24,12 @@ Reference
 Examples
 --------
 
-Use the ``list()`` method to list all deploy tokens across the GitLab instance.
+Use the ``list(get_all=True)`` method to list all deploy tokens across the GitLab instance.
```
nejch, line 27: Maybe we could just add the argument in the example rather than in the description?

```suggestion
Use the ``list()`` method to list all deploy tokens across the GitLab instance.
```

docs/gl_objects/runners.rst
```diff
@@ -28,9 +28,9 @@ Reference
 Examples
 --------
 
-Use the ``runners.list()`` and ``runners_all.list()`` methods to list runners.
-``runners.list()`` - Get a list of specific runners available to the user
-``runners_all.list()``  - Get a list of all runners in the GitLab instance 
+Use the ``runners.list(get_all=True)`` and ``runners_all.list(get_all=True)`` methods to list runners.
+``runners.list(get_all=True)`` - Get a list of specific runners available to the user
+``runners_all.list(get_all=True)``  - Get a list of all runners in the GitLab instance
```
nejch, line 33: Same with this (and the bullet points just below) I'm not sure if we should have it in the description already or just in the examples. but no strong opinion.

nejch (commented):
Thank you @igorp-collabora! Just a question about adding these to the description already or if we should keep them for the code examples.

docs/api-usage-advanced.rst
igorp-collabora: This section is about using the `Gitlab` object as a context manager so it doesn't have to use the `.list`. I switched it to a `.statistics.get()` which shouldn't have any issues running on gitlab.com.

docs/gl_objects/deploy_tokens.rst
igorp-collabora: Done

docs/gl_objects/runners.rst
igorp-collabora: Done.

nejch (approved):

docs/api-usage-advanced.rst
nejch: Ah makes sense!

nejch:
Thanks again @igorp-collabora!

Issue #3103, opened 2025-01-27
Follow-up: add additional status check methods


nejch:
Thanks a lot @SachinKSingh28, this looks great already! I think we should maybe do the TODOs in a follow-up PR, as this is already a pretty big diff for reviews.

_Originally posted by @nejch in https://github.com/python-gitlab/python-gitlab/pull/3098#pullrequestreview-2574740937_
            
Tasks:

- [ ] Add documentation for project merge request status check manager.
- [ ] Add method for getting a single status check for project.
- [ ] Add method for setting the status check for an MR.

SachinKSingh28:
@nejch Added MR to implement two tasks and a discussion for the third one if you get a chance to take a look please.

https://github.com/python-gitlab/python-gitlab/discussions/3132
https://github.com/python-gitlab/python-gitlab/pull/3128

SachinKSingh28:
@nejch if you get a chance to take a look. I think https://github.com/python-gitlab/python-gitlab/pull/3128 is ready and that should complete the status check stuff.

PR #3128, opened 2025-02-08
fix: adds missing status check methods for merge requests

SachinKSingh28:
Implements promised changes in https://github.com/python-gitlab/python-gitlab/issues/3103

- [x] Add documentation for project merge request status check manager.
- [x] Add method for setting the status check for an MR.

### DIFF
--- a/docs/gl_objects/status_checks.rst
+++ b/docs/gl_objects/status_checks.rst
@@ -24,11 +24,11 @@ Examples
 
 List external status checks for a project::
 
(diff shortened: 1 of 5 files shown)


SachinKSingh28:
@nejch It doesn't seem like we have a Gitlab API to get a single external status check for a project, only list method is supported. is there a cleaner way this can be approached/implemented? I was actually thinking about raising with Gitlab.

Issue #2801, opened 2024-02-15
member_roles endpoint is not implemented


Nivalam:
The member_roles enpoint is not implemented in python-gitlab : https://docs.gitlab.com/16.8/ee/api/member_roles.html

We can't add a gitlab custom roles with the library

nejch:
Thanks for the report @Nivalam. Would be great if someone who uses GitLab Ultimate could contribute this so adding `help wanted` here  :bow:

Nivalam:
I would be happy to test if someone implement this

SachinKSingh28:
I think we can close this probably. May be @Nivalam can give it a try :)

nejch:
Awesome, thanks for your PR @SachinKSingh28!

PR #3133, opened 2025-02-13
feat: adds member role methods

SachinKSingh28:
Implements https://github.com/python-gitlab/python-gitlab/issues/2801

### DIFF
--- a/tests/functional/api/test_member_roles.py
+++ b/tests/functional/api/test_member_roles.py
@@ -1,0 +1,18 @@
+"""
+GitLab API:
+https://docs.gitlab.com/ee/api/member_roles.html
+"""
+
+
+def test_instance_member_role(gl):
+    member_role = gl.member_roles.create(
+        {
+            "name": "Custom webhook manager role",
+            "base_access_level": 20,
+            "description": "Custom reporter that can manage webhooks",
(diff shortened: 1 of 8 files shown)


SachinKSingh28:
Removing group member roles API tests because that API is deprecated from self managed so can only be tested with gitlab.com

nejch:
@SachinKSingh28 just quick note - it doesn't seem like group member roles are deprecated on self-managed, they are just behind a feature flag (meaning not production ready yet). We could potentially enable that feature flag for functional tests if you already had them written, but not necessary in this MR.

Unless of course I missed something upstream, feel free to provide a link if so!

tests/functional/api/test_member_roles.py
```diff
@@ -0,0 +1,18 @@
+"""
+GitLab API:
+https://docs.gitlab.com/ee/api/member_roles.html
+"""
+
+
+def test_instance_member_role(gl):
```
nejch, line 7: We could add a marker here as it's EE-only. But not a big issue.

```suggestion
@pytest.mark.gitlab_premium
def test_instance_member_role(gl):
```

nejch (approved):
Thanks a lot for this @SachinKSingh28! Very clean PR! :bow:

PR #3166, opened 2025-04-07
ci: disable testing of Python 3.14 development release

JohnVillalovos:
Issue: Dependencies we use do not yet support Python 3.14, and in fact fail. This in turn causes our CI to fail.

Until that changes disable testing of Python 3.14 development release in the CI.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -36,8 +36,9 @@ jobs:
             toxenv: py312,smoke
           - version: "3.13"
             toxenv: py313,smoke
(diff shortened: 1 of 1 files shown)


JohnVillalovos:
Actually this might not be necessary. It seems like even if the job fails we can still merge code. I can't see the settings but I suspect that the Python 3.14 job is not required to pass in order to allow a PR to be merged.

JohnVillalovos:
httpcore 1.0.8 has fixed the issue. This isn't needed.

PR #3175, opened 2025-04-15
feat(api): add support for avatar removal

### LINKED ISSUE #2933: Add support for removing avatars in the topics API.

If I want to remove the avatar from the topic, the following error will occur:
`gitlab.exceptions.GitlabUpdateError: 400: {'avatar': ['file format is not supported. Please try one of the following supported formats: image/png, image/jpeg, image/gif, image/bmp, image/tiff, image/vnd.microsoft.icon']}`

This is my code: 

```python
gitlab = Gitlab(private_token=access_token, url=host)

res = gitlab.topics.update(25, {
    'avatar': '',
})

```

## Expected Behavior

According to the [Topics API | GitLab](https://docs.gitlab.com/ee/api/topics.html), it supports removing the avatar during the update.

## Actual Behavior

Successfully updated, and the avatar has been removed.

## Specifications

  - python-gitlab version: 4.7
  - Gitlab server version (or gitlab.com): 16.11.5

rickbrouwer:
## Changes

When attempting to remove for example a [groups](https://docs.gitlab.com/api/groups/#remove-a-group-avatar), [projects](https://docs.gitlab.com/api/projects/#remove-a-project-avatar) or [topics](https://docs.gitlab.com/api/topics/#remove-a-topic-avatar) avatar by setting it to an empty string, the current implementation raises a validation error about unsupported file formats.

The solution is to add a special condition that checks if the value of a `FileAttribute` is an empty string. If so, it should remain in the `data` dictionary instead of being moved to the `files` dictionary.

Fixes #2933 

### Documentation and testing

- [X] [Unit tests](https://github.com/python-gitlab/python-gitlab/tree/main/tests/unit) and/or [functional tests](https://github.com/python-gitlab/python-gitlab/tree/main/tests/functional)

### DIFF
--- a/gitlab/utils.py
+++ b/gitlab/utils.py
@@ -188,6 +188,12 @@ def _transform_types(
 
         # if the type is FileAttribute we need to pass the data as file
         if isinstance(gitlab_attribute, types.FileAttribute) and transform_files:
+            # The GitLab API accepts mixed types
+            # (e.g. a file for avatar image or empty string for removing the avatar)
+            # So if string is empty, keep it in data dict
+            if isinstance(data[attr_name], str) and data[attr_name] == "":
+                continue
+
             key = gitlab_attribute.get_file_name(attr_name)
             files[attr_name] = (key, data.pop(attr_name))
             continue
--- a/docs/gl_objects/groups.rst
+++ b/docs/gl_objects/groups.rst
@@ -82,6 +82,11 @@ Set the avatar image for a group::
     group.avatar = open('path/to/file.png', 'rb')
     group.save()
 
+Remove the avatar image for a group::
+
+    group.avatar = ""
+    group.save()
+
 Remove a group::
 
     gl.groups.delete(group_id)
--- a/docs/gl_objects/projects.rst
+++ b/docs/gl_objects/projects.rst
@@ -109,6 +109,11 @@ Set the avatar image for a project::
     project.avatar = open('path/to/file.png', 'rb')
     project.save()
 
+Remove the avatar image for a project::
+
+    project.avatar = ""
+    project.save()
+
 Delete a project::
 
     gl.projects.delete(project_id)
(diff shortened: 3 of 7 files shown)


gitlab/utils.py
```diff
@@ -188,6 +188,10 @@ def _transform_types(
 
         # if the type is FileAttribute we need to pass the data as file
         if isinstance(gitlab_attribute, types.FileAttribute) and transform_files:
+            # If string is empty, keep it in data dict
+            if isinstance(data[attr_name], str) and data[attr_name] == "":
+                continue
```
amimas, line 193: Should we add a reference to GitLab's API for clarification since the API accepts mixed types (e.g. a file for avatar image or empty string for removing the avatar)?

rickbrouwer: That seems like a good idea to me! Do you have a proposal?

amimas: The API doc for creating/updating a project shows `avatar` attribute type is "Mixed".

https://docs.gitlab.com/api/projects/#edit-a-project

Unfortunately it doesn't add any explanation for it. Following sections in the doc shows adding/updating an avatar requires a file for the `avatar` attribute and an empty value for the same attribute to delete the avatar.

- https://docs.gitlab.com/api/projects/#upload-a-project-avatar
- https://docs.gitlab.com/api/projects/#remove-a-project-avatar

Haven't checked the other API endpoints but assuming they are all the same, maybe a summary of the above. Or, maybe there's a better suggestion.

rickbrouwer: Then maybe this is enough? WDYT?

```
# The GitLab API accepts mixed types (e.g. a file for avatar image or empty string for removing the avatar)
# So if string is empty, keep it in data dict
```

amimas: Yeah, that seems good to me.

rickbrouwer: Done 😄 Thanks for the feedback!

JohnVillalovos (approved):
Thanks @rickbrouwer 

Looks good.

PR #3191, opened 2025-05-12
feat(api): add iteration_id as boards create attribute

massimiliano96:
## Changes

Added support for iteration_id as an attribute when creating board lists.

### DIFF
--- a/gitlab/v4/objects/boards.py
+++ b/gitlab/v4/objects/boards.py
@@ -23,7 +23,7 @@ class GroupBoardListManager(CRUDMixin[GroupBoardList]):
     _obj_cls = GroupBoardList
     _from_parent_attrs = {"group_id": "group_id", "board_id": "id"}
     _create_attrs = RequiredOptional(
-        exclusive=("label_id", "assignee_id", "milestone_id")
+        exclusive=("label_id", "assignee_id", "milestone_id", "iteration_id")
     )
     _update_attrs = RequiredOptional(required=("position",))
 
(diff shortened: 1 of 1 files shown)


gitlab/v4/objects/boards.py
```diff
@@ -23,7 +23,7 @@ class GroupBoardListManager(CRUDMixin[GroupBoardList]):
     _obj_cls = GroupBoardList
     _from_parent_attrs = {"group_id": "group_id", "board_id": "id"}
     _create_attrs = RequiredOptional(
-        exclusive=("label_id", "assignee_id", "milestone_id")
+        exclusive=("label_id", "assignee_id", "milestone_id", "iteration_id")
```
JohnVillalovos, line 26: According to the docs I don't see that `iteration_id` is mutually exclusive. Can you confirm? Though for group they don't mention that any of them are mutually exclusive but project does.

https://docs.gitlab.com/api/group_boards/#new-group-issue-board-list

gitlab/v4/objects/boards.py
```diff
@@ -48,7 +48,7 @@ class ProjectBoardListManager(CRUDMixin[ProjectBoardList]):
     _obj_cls = ProjectBoardList
     _from_parent_attrs = {"project_id": "project_id", "board_id": "id"}
     _create_attrs = RequiredOptional(
-        exclusive=("label_id", "assignee_id", "milestone_id")
+        exclusive=("label_id", "assignee_id", "milestone_id", "iteration_id")
```
JohnVillalovos, line 51: According to the docs I don't see that `iteration_id` is mutually exclusive. Can you confirm?

https://docs.gitlab.com/api/boards/#create-a-board-list

gitlab/v4/objects/boards.py
massimiliano96: I made a test using a simple curl:

image

JohnVillalovos:
Thanks @massimiliano96

Issue #3210, opened 2025-06-07
python-semantic-release/upload-to-gh-release has been deprecated


JohnVillalovos:
Error seen in logfile:

```
18837 2025-06-04T02:38:35.1711497Z ##[warning]This action has been DEPRECATED. Please use the 'python-semantic-release/publish-action' instead.
```

Deprecated action:
https://github.com/python-semantic-release/upload-to-gh-release

New action:
https://github.com/python-semantic-release/publish-action

We are using it here:
https://github.com/python-gitlab/python-gitlab/blob/378a836bf5744ca6c9409dd60899e5d2f90b55be/.github/workflows/release.yml#L34-L38

PR #3211, opened 2025-06-07
build(release): use correct python-semantic-release/publish-action

JohnVillalovos:
The previous 'python-semantic-release/upload-to-gh-release' action has been deprecated.

Closes: #3210


nejch (approved):
Thanks @JohnVillalovos we might need to change some config if this fails again but LGTM!

PR #3215, opened 2025-06-09
chore: update to mypy 1.16.0 and resolve issues found

JohnVillalovos:
There was an error in the type-hints that was detected when updating to mypy 1.16.0

The error was:
  $ mypy
  gitlab/v4/cli.py:411: error: Argument 1 to "cls_to_gitlab_resource" has incompatible type "type[Any]"; expected "RESTObject"  [arg-type]
  Found 1 error in 1 file (checked 246 source files)


nejch (approved):
Thanks @JohnVillalovos for looking into this, I'm still catching up!

Issue #3220, opened 2025-06-24
feat(const): complete Planner Role


matthias-reiff:
## Description of the problem, including code/CLI snippet
`PLANNER ` access was added to gitlab.const.AccessLevel but is still mising as gitlab.const.PLANNER_ACCESS and in gitlab.const.__all__ 

## Expected Behavior
It should be possible to access `gitlab.const.PLANNER_ACCESS` similar to `gitlab.const.DEVELOPER_ACCESS`.

## Actual Behavior
Currently, PLANNER_ACCESS can only be imported via `gitlab.const.AccessLevel.PLANNER`

## Specifications

  - python-gitlab version: 6.0.0

matthias-reiff:
Fixed in https://github.com/python-gitlab/python-gitlab/pull/3221

PR #3221, opened 2025-06-24
feat: Publish const PLANNER_ACCESS

matthias-reiff:
From Issue #3220


nejch (approved):
Thanks @matthias-reiff! These string-based access constants are sort of soft-deprecated as we have the enums now, but makes sense for consistency.

PR #3239, opened 2025-08-06
feat(api) adding missing ProjectPackageManager list filters

nickbroon:
`package_version` was introduced in GitLab 16.6

See: https://docs.gitlab.com/api/packages/#for-a-project


nejch:
Thanks @nickbroon nice catch, I just retried the CLI test and setting auto-merge :+1:

PR #3242, opened 2025-08-07
feat(api): add missing ProjectJob list filters

nickbroon:
See: https://docs.gitlab.com/api/jobs/#list-project-jobs


nejch (approved):
Again, nice catch! Thanks @nickbroon :bow:

PR #3276, opened 2025-10-17
ci(stale): fix permission for stale action and allow manual run

JohnVillalovos:
As documented at: https://github.com/actions/stale it requires the
"actions: write" permission.

Also allow manually running the stale action.


JohnVillalovos:
@nejch I'm going to go ahead and merge this as it is a minor change in my opinion.

Issue #3209, opened 2025-06-04
Release job did not update CHANGELOG.md


JohnVillalovos:
When the 6.0.0 release occurred the release job did not update CHANGELOG.md

@nejch Any ideas on this?

KevinGimbel:
And why is it 6.0.0? I can't see any breaking changes.

JohnVillalovos:
> And why is it 6.0.0? I can't see any breaking changes.

```
commit 6eee494749ccc5387a0d3af7ce331cfb1e95308b
Author: Igor Ponomarev <igor.ponomarev@collabora.com>
Date:   Thu Feb 6 15:28:22 2025 +0000

    feat(api)!: ListMixin.list typing overload

    BREAKING CHANGE: The `http_list` method will no longer
...
```

nejch:
The breaking changes were mostly in typing, so if your code does not use type annotations you might not have noticed this @KevinGimbel. Otherwise, I've definitely seen CI break in our downstream projects.

@JohnVillalovos I haven't had time yet to look at this 🙇 I think I'll manually add this one and then possibly move to the node implementation of semantic-release as discussed. It's much more stable.

PR #3277, opened 2025-10-17
fix(semantic-release): enable CHANGELOG.md generation

JohnVillalovos:
Also update CHANGELOG.md with missing entries.

Exclude dependency updates from being added to the CHANGELOG.md file in the future.

Closes: #3209

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -101,6 +101,13 @@ version_variables = [
 ]
 commit_message = "chore: release v{version}"
 
+[tool.semantic_release.changelog]
+exclude_commit_patterns = [
+  '''chore\(deps\): .+''',  # Exclude dependency updates from the changelog
+]
+mode = "update"
+insertion_flag = "All versions below are listed in reverse chronological order."
+
 [tool.pylint.messages_control]
 max-line-length = 88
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -101,6 +101,13 @@ version_variables = [
 ]
 commit_message = "chore: release v{version}"
 
+[tool.semantic_release.changelog]
+exclude_commit_patterns = [
+  '''chore\(deps\): .+''',  # Exclude dependency updates from the changelog
```
nejch, line 106: Should we exclude all chores here, WDYT? Just so we can do small fixups after initial merges but before releases.

JohnVillalovos: @nejch Good question. I went with just `deps` as it seems like we do a lot of stuff using 'chores' which might be useful to know about. Maybe we (mainly me) should do a better job of picking 'fix' instead.

nejch: Ah ok. I deliberately choose chore when I want to omit it from changelogs (not that python-semantic-release has been good about omitting that recently, but the canonical node version does) :sweat_smile:  sure let's stick with this for now and if the changelogs are noisy we can adjust.

nejch (approved):

PR #3278, opened 2025-10-17
feat(api): add content_ref and dry_run_ref parameters to ProjectCiLintManager

### LINKED ISSUE #3260: Validate existing ci yml seems to be not implemented

It seems that _existing config validation_ is not implemented (at least I haven't found it in the source code), but the [documentation](https://python-gitlab.readthedocs.io/en/stable/gl_objects/ci_lint.html) doesn't say it explicitly.

While this curl call works:

```bash
$ curl --header "Private-Token: ${MY_CI_TOKEN}" https://${MY_GITLAB_HOST}/api/v4/projects/${MY_PROJECT_ID}/ci/lint?content_ref=${MYREF}
```

This code doesn't:

```python
from gitlab import Gitlab
g = Gitlab(my_gitlab_host, private_token=my_ci_token)
p = g.projects.get(my_project_id)
p.ci_lint.validate({"content_ref": myref})
```

## Expected Behavior

If I pass only `content_ref` instead of `content`, the same GET would be called as in the curl above according to the [GitLab documentation](https://docs.gitlab.com/api/lint/#validate-an-existing-cicd-configuration).

## Actual Behavior

It executes _new config validation_ as specified [here](https://docs.gitlab.com/api/lint/#validate-a-new-cicd-configuration), sends a POST call, then throws `AttributeError: Missing attributes: content`

## Specifications

  - python-gitlab version: v6.3.0
  - Gitlab server version (or gitlab.com): v14.7.6

JohnVillalovos:
Add support for the new GitLab API parameters for validating existing CI/CD configurations:
- content_ref: specify the branch, tag, or SHA to validate
- dry_run_ref: set branch/tag context for dry run simulations

The deprecated 'ref' parameter is kept for backward compatibility.

Also update documentation with examples showing how to validate CI configuration from specific branches and with dry run simulation.

Fixes #3260

### DIFF
--- a/docs/gl_objects/ci_lint.rst
+++ b/docs/gl_objects/ci_lint.rst
@@ -46,6 +46,18 @@ Lint a project's CI configuration::
     assert lint_result.valid is True  # Test that the .gitlab-ci.yml is valid
     print(lint_result.merged_yaml)    # Print the merged YAML file
 
+Lint a project's CI configuration from a specific branch or tag::
+
+    lint_result = project.ci_lint.get(content_ref="main")
+    assert lint_result.valid is True  # Test that the .gitlab-ci.yml is valid
+    print(lint_result.merged_yaml)    # Print the merged YAML file
+
+Lint a project's CI configuration with dry run simulation::
+
+    lint_result = project.ci_lint.get(dry_run=True, dry_run_ref="develop")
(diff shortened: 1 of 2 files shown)


nejch:
Thanks @JohnVillalovos for looking into this :bow:

PR #3279, opened 2025-10-18
fix(epics): use actual group_id for save/delete operations on nested epics

### LINKED ISSUE #3261: GroupEpic uses "iid" as ID attribute but should use "id"?

`objects.epics.GroupEpic` uses `iid` (group local ID) as ID attribute / `encoded_id` instead of the global `id` like, for example, `objects.issues.ProjectIssue` does.  This leads to an erroneous behavior when updating an epic, i.e., the wrong epic is being updated.

## Expected Behavior

- I retrieve an epic: `epic = top_level_group.epics.list(...)[0]`
- I manipulate the epic labels: `epic.labels.extend([...])` <-- epic from a sub group
- I save the epic: `epic.save()`
- The new labels are saved for _exactly_ this epic, identified by `epic.id`.

## Actual Behavior

- The labels are saved for the first epic with `epic.iid`, which is not necessarily the one that I intended to update.

## Discussion

The class `GroupEpic` explicitly sets `iid` as `_id_attr`.  When updating the epic, the request is sent to `epic.manager.path/{epic_iid}`, but since `epic.manager.path` is the path to the top level group, the epic with id 3 in this top level group is updated instead of the one from the nested group.

So maybe it's not just using `id` instead of `iid` but fixing the manager (path) so that is uses the path of the actual group that the epic belongs to and not the path of the group that was used to retrieve all epics.

## Specifications

  - python-gitlab version: 6.20 (but has been around since months or years and does still exist: https://github.com/python-gitlab/python-gitlab/blob/main/gitlab/v4/objects/epics.py#L26)
  - Gitlab server version (or gitlab.com): recent / not important

JohnVillalovos:
When an epic belonging to a subgroup is retrieved through a parent
group's epic listing, save() and delete() operations would fail because
they used the parent group's path instead of the epic's actual group_id.

Add an internal custom path hook for object save and delete operations, and let
UpdateMixin and DeleteMixin accept an explicit `_custom_path` keyword. GroupEpic
uses this hook to build mutation paths from the epic's server-provided group_id,
so save() and delete() target the epic's owning group.

Avoid using parent manager attrs as a fallback for the epic group path. Lazy
epics do not have a server-provided group_id, so raise a clear error instead of
silently using the parent group path.

Add unit coverage for custom update/delete paths, GroupEpic save/delete path
selection, and lazy epic path failures. Add a functional test for saving a
subgroup epic discovered through the parent group listing.

Assisted-by: OpenAI Codex (GPT-5)

Closes: #3261

### DIFF
--- a/gitlab/mixins.py
+++ b/gitlab/mixins.py
@@ -292,6 +292,8 @@ class UpdateMixin(base.RESTManager[base.TObjCls]):
         self,
         id: str | int | None = None,
         new_data: dict[str, Any] | None = None,
+        *,
+        _custom_path: str | None = None,
         **kwargs: Any,
     ) -> dict[str, Any]:
         """Update an object on the server.
@@ -299,6 +301,7 @@ class UpdateMixin(base.RESTManager[base.TObjCls]):
         Args:
             id: ID of the object to update (can be None if not required)
             new_data: the update data for the object
+            _custom_path: Optional custom path for special API endpoints
             **kwargs: Extra options to send to the server (e.g. sudo)
 
         Returns:
@@ -310,7 +313,9 @@ class UpdateMixin(base.RESTManager[base.TObjCls]):
         """
         new_data = new_data or {}
 
-        if id is None:
+        if _custom_path is not None:
+            path = _custom_path
+        elif id is None:
             path = self.path
         else:
             path = f"{self.path}/{utils.EncodedId(id)}"
@@ -357,18 +362,27 @@ class SetMixin(base.RESTManager[base.TObjCls]):
 
 class DeleteMixin(base.RESTManager[base.TObjCls]):
     @exc.on_http_error(exc.GitlabDeleteError)
-    def delete(self, id: str | int | None = None, **kwargs: Any) -> None:
+    def delete(
+        self,
+        id: str | int | None = None,
+        *,
+        _custom_path: str | None = None,
+        **kwargs: Any,
+    ) -> None:
         """Delete an object on the server.
 
         Args:
             id: ID of the object to delete
+            _custom_path: Optional custom path for special API endpoints
             **kwargs: Extra options to send to the server (e.g. sudo)
 
         Raises:
             GitlabAuthenticationError: If authentication is not correct
             GitlabDeleteError: If the server cannot perform the request
         """
-        if id is None:
+        if _custom_path is not None:
+            path = _custom_path
+        elif id is None:
             path = self.path
         else:
             path = f"{self.path}/{utils.EncodedId(id)}"
@@ -402,6 +416,12 @@ class SaveMixin(_RestObjectBase):
     _parent_attrs: dict[str, Any]
     _updated_attrs: dict[str, Any]
     manager: base.RESTManager[Any]
+
+    def _get_custom_path(self) -> str | None:
+        # NOTE(jlvillal): pylint will complain for the callers with an
+        # 'assignment-from-none' error, if we don't do this.
+        custom_path: str | None = None
+        return custom_path
 
     def _get_updated_data(self) -> dict[str, Any]:
         updated_data = {}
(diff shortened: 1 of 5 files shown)


JohnVillalovos:
FYI: This was tested and reported as working https://github.com/python-gitlab/python-gitlab/issues/3261#issuecomment-3506545729

gitlab/v4/objects/epics.py
```diff
@@ -29,6 +30,63 @@ class GroupEpic(ObjectDeleteMixin, SaveMixin, RESTObject):
     resourcelabelevents: GroupEpicResourceLabelEventManager
     notes: GroupEpicNoteManager
 
+    def _epic_path(self) -> str:
+        """Return the API path for this epic using its real group."""
+        if not hasattr(self, "group_id") or self.group_id is None:
```
nejch: Can a `group_id` actually be none for an epic, since it's always a group resource? Or can we go with a simple try/except here?

One other thought - will this work if we get an epic with `lazy=True` from an arbitrary group?

gitlab/v4/objects/epics.py
```diff
@@ -29,6 +30,63 @@ class GroupEpic(ObjectDeleteMixin, SaveMixin, RESTObject):
...
+
+        Raises:
+            GitlabAuthenticationError: If authentication is not correct
+            GitlabUpdateError: If the server cannot perform the request
+        """
+        # Use the epic's actual group_id to construct the correct path.
+        path = self._epic_path()
+
+        # Call SaveMixin.save() method
+        return super().save(_pg_custom_path=path, **kwargs)
```
nejch: Not sure if we should have some more generic `_get_custom_path` or `_resolve_path` helpers that would automatically get picked up by the mixins, as we already do this twice for delete/update. That way we don't need an override here (or maybe also in some other places), but I'm not sure if feasible or worth it.

gitlab/mixins.py
```diff
@@ -314,6 +314,8 @@ def update(
             path = self.path
         else:
             path = f"{self.path}/{utils.EncodedId(id)}"
+        if "_pg_custom_path" in kwargs:
```
nejch: nit: `pg_` always makes me think of postgres here, but maybe I've spent too much time in `gitlab.rb`.  Also depending on the answer in the other questions, would it make sense to have this as a keyword argument instead of popping it here?


```suggestion
        if "_custom_path" in kwargs:
```

tests/functional/api/test_epics.py
```diff
@@ -32,3 +37,42 @@ def test_epic_notes(epic):
...
+    subgroup_id = uuid.uuid4().hex
+    subgroup = gl.groups.create(
+        {
+            "name": f"subgroup-{subgroup_id}",
+            "path": f"sg-{subgroup_id}",
+            "parent_id": group.id,
+        }
+    )
+
+    nested_epic = subgroup.epics.create(
```
nejch, line 65: Could we have this one as a fixture that `yield`s the epic, to avoid the manual cleanup below?

Or I'm just confused by the try/finally block, maybe it was meant to test deletion even if previous things failed.

nejch (commented):
Thanks @JohnVillalovos, looks good and that's a tricky one. Not sure about lazily fetched epics - just have a few questions there.

tests/functional/api/test_epics.py
JohnVillalovos: Done.

gitlab/v4/objects/epics.py
JohnVillalovos: I added `_get_custom_path()` method.

gitlab/mixins.py
JohnVillalovos: Changed the code so no longer have this.

gitlab/v4/objects/epics.py
JohnVillalovos: @nejch Good catch. `lazy=True` would fail.

Will work on an update.

JohnVillalovos: Fixed.

PR #3284, opened 2025-10-20
chore(deps): upgrade coverage to 7.11.0 for Python > 3.9

JohnVillalovos:
For Python versions newer than 3.9 use 'coverage==7.11.0'

For 3.9 and older continue to use 'coverage==7.10.7'

### DIFF
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -1,7 +1,8 @@
 -r requirements.txt
 anyio==4.11.0
 build==1.3.0
-coverage==7.10.7
+coverage==7.10.7 ; python_version <= '3.9'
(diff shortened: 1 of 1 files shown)


nejch (approved):
Thanks @JohnVillalovos, LGTM. I'll merge this here now.

Though what do you think about dropping 3.9 already as it's EOL next week and we could align with that?

JohnVillalovos:
> Thanks @JohnVillalovos, LGTM. I'll merge this here now.
> 
> Though what do you think about dropping 3.9 already as it's EOL next week and we could align with that?

Great idea! I didn't even realize it was end of life.

PR #3300, opened 2025-11-20
feat(registry-protection): add support for registry protection rule deletion

sisp:
## Changes

I've added support for [deleting a registry protection rule](https://docs.gitlab.com/api/container_repository_protection_rules/#delete-a-container-repository-protection-rule). It seems this feature was missed in the initial implementation although [it was documented](https://python-gitlab.readthedocs.io/en/stable/gl_objects/protected_container_repositories.html#examples).

For this, I've added the `DeleteMixin`, and I've extended the functional API test as well as added a unit test.

### Documentation and testing

- [ ] ~~Documentation in the matching [docs section](https://github.com/python-gitlab/python-gitlab/tree/main/docs)~~ _The docs already exist._


tests/functional/api/test_registry.py
nejch, line 12: I guess we should enable this at some point haha, but don't want to break this PR for now in case something else is missing here.

nejch (approved):
Thanks a lot for this @sisp and sorry for the long delay 🙇

PR #3308, opened 2025-12-04
fix(utils): prevent negative sleep time in rate limit retry

AlexandreEXFO:
## Changes
Resolution of the error that sometimes occurs: "Error: sleep length must be non-negative"

### Documentation and testing

### DIFF
--- a/gitlab/utils.py
+++ b/gitlab/utils.py
@@ -133,7 +133,7 @@ class Retry:
             if "Retry-After" in headers:
                 wait_time = int(headers["Retry-After"])
             elif "RateLimit-Reset" in headers:
-                wait_time = int(headers["RateLimit-Reset"]) - time.time()
+                wait_time = max(0, int(headers["RateLimit-Reset"]) - time.time())
             self.cur_retries += 1
             time.sleep(wait_time)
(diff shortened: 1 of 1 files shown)


AlexandreEXFO:
@JohnVillalovos: Could you take a look at this change? Please

gitlab/utils.py
```diff
@@ -133,7 +133,7 @@ def handle_retry_on_status(
             if "Retry-After" in headers:
                 wait_time = int(headers["Retry-After"])
             elif "RateLimit-Reset" in headers:
-                wait_time = int(headers["RateLimit-Reset"]) - time.time()
+                wait_time = max(0, int(headers["RateLimit-Reset"]) - time.time())
```
JohnVillalovos, line 136: Shouldn't it be `min` ?

AlexandreEXFO: The maximum between the difference and 0 to ensure that the number is always positive (>=0).

max(0, -1) = 0
max(0, 1) = 1

JohnVillalovos: Duh. Sorry I just woke up 😀

PR #3313, opened 2025-12-18
feat: integrate pytest-gitlab

### LINKED ISSUE #3046: Incorporate pytest-gitlab plugin back into python-gitlab
Someone appears to have extracted our setup & fixtures for pytest and repackaged it all as a pytest plugin on pypi.org:

https://pypi.org/project/pytest-gitlab/

I was unaware of this but I have wanted something like this for years - i.e. expose our fixtures as a plugin to make them reusable in downstream projects. See https://gitlab.com/marge-org/marge-bot/-/merge_requests/479#note_2225265620 for a recent example use case.

I (and probably others @JohnVillalovos :sweat_smile: ) spent quite a few weekends adding and improving this pytest setup so IMO it would be nice to see the benefits of this shared here and also make it more discoverable for our users rather than have it a separate package that users might not trust if they can't recognize or see the sources.

I can't find the source for the project (seems to not be openly available though it reuses LGPL, purely by looking at the sdist). But I'll try to reach out to the author and see if we can incorporate it here. Otherwise the fixtures will almost definitely diverge anyway.

ptalbert:
Fixes: #3046

This takes the pytest-gitlab plugin from https://gitlab.com/thht_science_software/pytest-gitlab and puts it in gitlab/testing/. It is then exposed as a pytest plugin "gitlab" and the existing tests/ fixtures are removed.

### DIFF
--- a/gitlab/testing/docker/docker-compose.yml
+++ b/gitlab/testing/docker/docker-compose.yml
@@ -1,0 +1,52 @@
+version: '3.5'
+
+networks:
+  gitlab-network:
+    name: gitlab-network
+
+services:
+  gitlab:
+    image: '${GITLAB_IMAGE:-gitlab/gitlab-ee}:${GITLAB_TAG:-latest}'
+    container_name: 'gitlab-test'
+    hostname: 'gitlab.test'
+    privileged: true # Just in case https://gitlab.com/gitlab-org/omnibus-gitlab/-/issues/1350
+    environment:
+      GITLAB_ROOT_PASSWORD: 5iveL!fe
+      GITLAB_OMNIBUS_CONFIG: |
+        external_url 'http://127.0.0.1:8080'
+        registry['enable'] = false
+        nginx['redirect_http_to_https'] = false
+        nginx['listen_port'] = 80
+        nginx['listen_https'] = false
+        pages_external_url 'http://pages.gitlab.lxd'
+        gitlab_pages['enable'] = true
+        gitlab_pages['inplace_chroot'] = true
+        prometheus['enable'] = false
+        alertmanager['enable'] = false
+        node_exporter['enable'] = false
+        redis_exporter['enable'] = false
+        postgres_exporter['enable'] = false
+        pgbouncer_exporter['enable'] = false
+        gitlab_exporter['enable'] = false
+        letsencrypt['enable'] = false
+        gitlab_rails['initial_license_file'] = '/python-gitlab-ci.gitlab-license'
+        gitlab_rails['monitoring_whitelist'] = ['0.0.0.0/0']
+    entrypoint:
+      - /bin/sh
+      - -c
+      - ruby /create_license.rb && /assets/init-container
+    volumes:
+      - ${PWD}/tests/functional/fixtures/create_license.rb:/create_license.rb
+    ports:
+      - '8080:80'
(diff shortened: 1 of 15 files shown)


ptalbert:
All the existing tests pass when I run `tox`, however, I am struggling to get `api_func_v4` and `cli_func_v4` to run even on the original main branch. Still working on it...

gitlab/testing/docker/docker-compose.yml
```diff
@@ -36,15 +36,15 @@ services:
       - -c
       - ruby /create_license.rb && /assets/wrapper
     volumes:
-      - ${PWD}/tests/functional/fixtures/create_license.rb:/create_license.rb
+      - ./create_license.rb:/create_license.rb
```
nejch: Does this definitely work (i.e. it will be relative to the compose config file) regardless of where the tests are executed from? Just making sure, I know that different compose implementations have different behavior 🙇

We had some issues in the past when we tried to migrate to compose v2, so just wanted to check.

gitlab/testing/docker/docker-compose.yml
```diff
@@ -6,7 +6,7 @@ networks:
 
 services:
   gitlab:
-    image: '${GITLAB_IMAGE}:${GITLAB_TAG}'
+    image: '${GITLAB_IMAGE:-gitlab/gitlab-ee}:${GITLAB_TAG:-latest}'
```
nejch, line 9: Could we maybe extract the preparation changes (everything that can be applied without moving the plugin) into a separate PR that we merge first? Just so we can focus on integration aspects here. Let me know otherwise I can also try that on my end.

nejch (changes requested):
@ptalbert thank you so much for tackling this! I think some of this might need updating to match our current testing environment, since pytest-gitlab was last updated in 2024 from what I can see. It might also help fix the test failures, as we recently also updated that to make it work on GitLab 18.x.

To make that a bit more manageable, would it make sense to split this into smaller PRs - e.g. prepare the changes needed so that the final PR is mostly just copying files? That would make things easier to review as well I think 🙇 thanks!

gitlab/testing/docker/docker-compose.yml
ptalbert: I didn't realize there was so much drift between the fixtures in pytest-gitlab and python-gitlab fixtures.

To hopefully simplify things, I've redone this PR using the current python-gitlab fixtures as the starting point. It adds the pytest-gitlab plugin but stops short of having the existing tests use any of it. A follow up PR can switch the tests over.

Hopefully this makes more sense. Let me know.

gitlab/testing/docker/docker-compose.yml
ptalbert: I don't know why pytest-gitlab changed it. The latest version takes the python-gitlab approach.

ptalbert:
I suppose the codecov report is lighting up because this PR does not actually exercise the gitlab/testing/ code any more.

As I see it the options are:

1. exclude gitlab/testing/ dir from the check
2. add a separate set of tests for all the gitlab/testing/ fixtures
3. change over the existing tests/ to use the new pytest-gitlab plugin

Doing #1 as a temporary solution seems best if we do not want to do #3 yet. #2 would be great but that is a large undertaking.

PR #3330, opened 2026-01-20
fix: actually define project repr_attr

nickbroon:
While `_repr_attr=` is given to specify the attribute to use to repr the object, that object attribute itself is not actually defined, so tools like PyRight/MyPy complain that:

```
Type of "path_with_namespace" is unknown
```

### DIFF
--- a/gitlab/v4/objects/projects.py
+++ b/gitlab/v4/objects/projects.py
@@ -178,6 +178,8 @@ class Project(
     _repr_attr = "path_with_namespace"
     _upload_path = "/projects/{id}/uploads"
 
+    path_with_namespace: str
+
     access_tokens: ProjectAccessTokenManager
     accessrequests: ProjectAccessRequestManager
     additionalstatistics: ProjectAdditionalStatisticsManager


gitlab/v4/objects/projects.py
```diff
@@ -230,6 +230,7 @@ class Project(
     package_protection_rules: ProjectPackageProtectionRuleManager
     pages: ProjectPagesManager
     pagesdomains: ProjectPagesDomainManager
+    path_with_namespace: str
```
JohnVillalovos: Isn't this something that comes from the response from the GitLab server? We don't tend to type-hint those as it would be massive undertaking.

nickbroon: Yes, comes from the GitLab Server. And in general I agree that type hinting all object attributes is too large an undertaking to be worth the effort and constant maintenance as more as added. But in case of the attribute that are used for either _id_ or _repr_, which more fundamental I believe its worth explicitly defining these.

JohnVillalovos: Okay. That makes sense. Could you move it to around line 180 then? To me that makes more sense, as here it is in the middle of all the *Manager objects.

nickbroon: I just put it here as it appeared the whole list of attributes was sort alphabetically. But I'll move it to the start.

JohnVillalovos (approved):
Thanks @nickbroon 

LGTM

PR #3366, opened 2026-03-09
feat(api): add support for project feature flags and feature flag user lists

amimas:
## Changes

This change introduces support for managing Project Feature Flags and Feature Flag User Lists via the API and CLI.

It adds the `ProjectFeatureFlag` and `ProjectFeatureFlagUserList` objects and their corresponding managers, which are exposed on the `Project` object as `project.feature_flags` and `project.feature_flags_user_lists`.

To support the specific data formats required by these GitLab APIs, this change also introduces:

- `JsonAttribute`: A new attribute type to handle the strategies field, which is sent as a JSON object.
- `CommaSeparatedStringAttribute`: A new attribute type to handle the user_xids field, which is sent as a comma-separated string.

Additionally, a custom `save()` method has been implemented for `ProjectFeatureFlag` to correctly handle renaming, as the flag's name is also its unique identifier.

The implementation includes:

- API object and manager definitions.
- Functional API tests.
- Functional CLI tests.
- Unit tests for the new attribute types and the custom renaming logic.

### Documentation and testing

Please consider whether this PR needs documentation and tests. **This is not required**, but highly appreciated:

- [x] Documentation in the matching [docs section](https://github.com/python-gitlab/python-gitlab/tree/main/docs)
- [x] [Unit tests](https://github.com/python-gitlab/python-gitlab/tree/main/tests/unit) and/or [functional tests](https://github.com/python-gitlab/python-gitlab/tree/main/tests/functional)

**Documentation updates**:

- Added new documentation pages for Project Feature Flags and Feature Flag User Lists.
- Renamed the existing `features.rst` to `gitlab_features.rst` and updated its content to clearly distinguish that this API/feature is for admin-only GitLab Development Feature Flags - not for project-level feature flags.

## References:

Support for following API is being included in this PR
- GitLab Project Feature Flag API: https://docs.gitlab.com/api/feature_flags/
- GitLab Project Feature Flag User List API: https://docs.gitlab.com/api/feature_flag_user_lists/

Documentation changed/clarified related to following existing feature in python-gitlab
- API on Feature flag for GitLab Development: https://docs.gitlab.com/api/features/

closes #3350

PR #3367, opened 2026-03-10
docs(testing): document passing pytest options during local development

amimas:
## Changes

This improves the developer experience by making the testing workflow more discoverable and flexible.

- Updates `tox.ini` to use `posargs` with default values for the test environments. This allows developers to run the full test suite by default, or easily specify a single test file to run.

- Adds `description` fields to the unit and functional test environments in `tox.ini`. This makes instructions on how to run specific tests visible in the output of `tox list`.

- Updates `CONTRIBUTING.rst` with concrete examples for running specific unit, API functional, and CLI functional tests, aligning the documentation with the new `tox` configuration.

### Documentation and testing

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -111,9 +111,6 @@ You need to install ``tox`` (``pip3 install tox``) to run tests and lint checks 
    # run unit tests in one python environment only (useful for quick testing during development):
    tox -e py311
 
-   # run a specific unit test file:
-   tox -e py311 -- tests/unit/objects/test_projects.py
-
    # run unit and smoke tests in one python environment only
    tox -e py312,smoke
 
@@ -151,14 +148,8 @@ To run these tests:
    # run the CLI tests:
    tox -e cli_func_v4
 
-   # run a specific CLI functional test file:
-   tox -e cli_func_v4 -- tests/functional/cli/test_cli_v4.py
-
    # run the python API tests:
    tox -e api_func_v4
-
-   # run a specific API functional test file:
-   tox -e api_func_v4 -- tests/functional/api/test_projects.py
 
 When developing tests it can be a little frustrating to wait for GitLab to spin
 up every run. To prevent the containers from being cleaned up afterwards, pass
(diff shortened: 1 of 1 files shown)


amimas:
@JohnVillalovos  - Would you have some time to review this change? Basically the motivation for this came when I was working on #3366 . I wanted to be able to run only selected tests instead of everything while developing locally. Running entire test suite locally is time consuming. Plus, gitlab in docker in local machine requires is not quite stable dependent on available resource.

I'm not too familiar with `tox`. One thing I'm not sure is why I needed to add the `USER` environment variable. Please let me know if you have any other suggestions. I think ability to run specific tests would be helpful.

tox.ini
```diff
@@ -131,14 +137,20 @@ exclude_lines =
...
 
 [testenv:cli_func_v4]
-deps = -r{toxinidir}/requirements-docker.txt
+description =
+  Runs CLI functional tests. For a specific test: tox -e cli_func_v4 -- <path>
+deps = -r{toxinidir}/requirements-test.txt
+       -r{toxinidir}/requirements-docker.txt
 commands =
-  pytest --script-launch-mode=subprocess --cov --cov-report xml tests/functional/cli {posargs}
+  pytest --script-launch-mode=subprocess --cov --cov-report xml {posargs:tests/functional/cli}
```
JohnVillalovos: I don't think this needed to be changed.

tox.ini
```diff
@@ -131,14 +137,20 @@ exclude_lines =
...
 
 [testenv:api_func_v4]
-deps = -r{toxinidir}/requirements-docker.txt
+description =
+  Runs API functional tests. For a specific test: tox -e api_func_v4 -- <path>
+deps = -r{toxinidir}/requirements-test.txt
+       -r{toxinidir}/requirements-docker.txt
 commands =
-  pytest --cov --cov-report xml tests/functional/api {posargs}
+  pytest --cov --cov-report xml {posargs:tests/functional/api}
```
JohnVillalovos: I don't think this needs to be changed.

tox.ini
```diff
@@ -35,10 +36,13 @@ usedevelop = True
...
 isolated_build = True
 
+
+[testenv:py{314,313,312,311,310}]
+description = Runs unit tests. For a specific test: tox -e py311 -- <path>
 deps = -r{toxinidir}/requirements.txt
        -r{toxinidir}/requirements-test.txt
 commands =
-  pytest tests/unit {posargs}
+  pytest {posargs:tests/unit}
```
JohnVillalovos: I don't think this needs to be changed.

tox.ini
```diff
@@ -35,10 +36,13 @@ usedevelop = True
 install_command = pip install {opts} {packages} -e .
 isolated_build = True
 
+
+[testenv:py{314,313,312,311,310}]
```
JohnVillalovos: This is unneeded as far as I know. A person can already run `tox -e py313` without this.

CONTRIBUTING.rst
```diff
@@ -148,9 +151,15 @@ To run these tests:
    # run the CLI tests:
    tox -e cli_func_v4
 
+   # run a specific CLI functional test file:
+   tox -e cli_func_v4 -- tests/functional/cli/test_cli_v4.py
```
JohnVillalovos: The `-k` flag is the way to accomplish this. It would be good add to the documentation.

amimas:
Thank you for all the feedbacks. I'll take a look when I get a chance.

One thing I did notice with my current change is that when I run `tox -e api_func_v4 -- --keep-containers tests/functional/api/test_project_feature_flags.py`, even though I have the containers up and running from previous run, the test still pauses for a long time before it starts. Maybe attempting to bring up the containers?

CONTRIBUTING.rst
amimas: Oh!! So, does that mean none of the changes in this PR is needed? :)

amimas: I didn't know about the `-k` flag. I'll revert my changes and maybe document it for clarity.

PR #3379, opened 2026-03-26
Update and fix variable _create_attrs and _update_attrs

PhilipNelson5:
## Changes

1. make `value` an optional attribute for variable updates

    When updating a CI/CD variable that has been marked "Masked and hidden", you likely do not have the value when updating. In this case, the value is `None` which the api rejects (`400: {'value': ['is invalid']}`).

2. add `masked_and_hidden` the the list of optional attributes for creating variables

    When creating variables which you want to be hidden, you have to use the `masked_and_hidden` property.

    I don't think this was causing problems but for the sake completion and documentation I think it makes sense to include.

### Documentation and testing

- [x] ~~Documentation in the matching [docs section](https://github.com/python-gitlab/python-gitlab/tree/main/docs)~~

### DIFF
--- a/gitlab/v4/objects/variables.py
+++ b/gitlab/v4/objects/variables.py
@@ -27,10 +27,11 @@ class VariableManager(CRUDMixin[Variable]):
     _path = "/admin/ci/variables"
     _obj_cls = Variable
     _create_attrs = RequiredOptional(
-        required=("key", "value"), optional=("protected", "variable_type", "masked")
+        required=("key", "value"),
+        optional=("protected", "variable_type", "masked", "masked_and_hidden"),
     )
     _update_attrs = RequiredOptional(
-        required=("key", "value"), optional=("protected", "variable_type", "masked")
+        required=("key",), optional=("value", "protected", "variable_type", "masked")
     )
 
 
@@ -43,10 +44,11 @@ class GroupVariableManager(CRUDMixin[GroupVariable]):
     _obj_cls = GroupVariable
     _from_parent_attrs = {"group_id": "id"}
     _create_attrs = RequiredOptional(
-        required=("key", "value"), optional=("protected", "variable_type", "masked")
+        required=("key", "value"),
+        optional=("protected", "variable_type", "masked", "masked_and_hidden"),
     )
     _update_attrs = RequiredOptional(
-        required=("key", "value"), optional=("protected", "variable_type", "masked")
+        required=("key",), optional=("value", "protected", "variable_type", "masked")
     )
 
 
@@ -60,9 +62,15 @@ class ProjectVariableManager(CRUDMixin[ProjectVariable]):
     _from_parent_attrs = {"project_id": "id"}
     _create_attrs = RequiredOptional(
         required=("key", "value"),
-        optional=("protected", "variable_type", "masked", "environment_scope"),
+        optional=(
+            "protected",
+            "variable_type",
+            "masked",
+            "environment_scope",
+            "masked_and_hidden",
+        ),
     )
     _update_attrs = RequiredOptional(
-        required=("key", "value"),
-        optional=("protected", "variable_type", "masked", "environment_scope"),
+        required=("key",),
+        optional=("value", "protected", "variable_type", "masked", "environment_scope"),
     )
--- a/gitlab/v4/objects/variables.py
+++ b/gitlab/v4/objects/variables.py
@@ -27,8 +27,7 @@ class VariableManager(CRUDMixin[Variable]):
     _path = "/admin/ci/variables"
     _obj_cls = Variable
     _create_attrs = RequiredOptional(
-        required=("key", "value"),
-        optional=("protected", "variable_type", "masked", "masked_and_hidden"),
+        required=("key", "value"), optional=("protected", "variable_type", "masked")
     )
     _update_attrs = RequiredOptional(
         required=("key",), optional=("value", "protected", "variable_type", "masked")
--- a/gitlab/v4/objects/variables.py
+++ b/gitlab/v4/objects/variables.py
@@ -27,10 +27,19 @@ class VariableManager(CRUDMixin[Variable]):
     _path = "/admin/ci/variables"
     _obj_cls = Variable
     _create_attrs = RequiredOptional(
-        required=("key", "value"), optional=("protected", "variable_type", "masked")
+        required=("key", "value"),
+        optional=("description", "masked", "protected", "raw", "variable_type"),
     )
     _update_attrs = RequiredOptional(
-        required=("key",), optional=("value", "protected", "variable_type", "masked")
+        required=("key",),
+        optional=(
+            "description",
+            "masked",
+            "protected",
+            "raw",
+            "value",
+            "variable_type",
+        ),
     )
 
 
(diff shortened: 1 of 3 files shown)


PhilipNelson5:
I tried running the suite of functional tests but was unable to. I got the docker containers up but they were so unstable that even adding `gl = gitlab.Gitlab( ... retry_transient_errors=True)` couldn't get passed the `500` errors.

I was able to run the variable tests against a live gitlab instance to demonstrate to myself the tests failed before my changes and passed after.

gitlab/v4/objects/variables.py
```diff
@@ -27,10 +27,11 @@ class VariableManager(CRUDMixin[Variable]):
     _path = "/admin/ci/variables"
     _obj_cls = Variable
     _create_attrs = RequiredOptional(
-        required=("key", "value"), optional=("protected", "variable_type", "masked")
+        required=("key", "value"),
+        optional=("protected", "variable_type", "masked", "masked_and_hidden"),
```
JohnVillalovos: Looking at: https://docs.gitlab.com/api/instance_level_ci_variables/#create-instance-variable

I don't see `masked_and_hidden` I also see other variables that aren't here.

Image

gitlab/v4/objects/variables.py
```diff
@@ -27,10 +27,11 @@ class VariableManager(CRUDMixin[Variable]):
...
     )
     _update_attrs = RequiredOptional(
-        required=("key", "value"), optional=("protected", "variable_type", "masked")
+        required=("key",), optional=("value", "protected", "variable_type", "masked")
```
JohnVillalovos: Maybe add the other update variables?

Image

gitlab/v4/objects/variables.py
```diff
@@ -43,10 +44,11 @@ class GroupVariableManager(CRUDMixin[GroupVariable]):
     _obj_cls = GroupVariable
     _from_parent_attrs = {"group_id": "id"}
     _create_attrs = RequiredOptional(
-        required=("key", "value"), optional=("protected", "variable_type", "masked")
+        required=("key", "value"),
+        optional=("protected", "variable_type", "masked", "masked_and_hidden"),
```
JohnVillalovos: Maybe add the other variables? https://docs.gitlab.com/api/group_level_variables/#create-a-group-variable

Image

gitlab/v4/objects/variables.py
```diff
@@ -43,10 +44,11 @@ class GroupVariableManager(CRUDMixin[GroupVariable]):
...
     )
     _update_attrs = RequiredOptional(
-        required=("key", "value"), optional=("protected", "variable_type", "masked")
+        required=("key",), optional=("value", "protected", "variable_type", "masked")
```
JohnVillalovos: Maybe add the other "non-hash" type variables. hash variables aren't very well supported especially in the CLI.

https://docs.gitlab.com/api/group_level_variables/#update-a-group-variable

Image

gitlab/v4/objects/variables.py
```diff
@@ -60,9 +62,15 @@ class ProjectVariableManager(CRUDMixin[ProjectVariable]):
...
     _create_attrs = RequiredOptional(
         required=("key", "value"),
-        optional=("protected", "variable_type", "masked", "environment_scope"),
+        optional=(
+            "protected",
+            "variable_type",
+            "masked",
+            "environment_scope",
+            "masked_and_hidden",
+        ),
```
JohnVillalovos, line 98: Maybe add the other variables? https://docs.gitlab.com/api/project_level_variables/#create-a-variable

Image

gitlab/v4/objects/variables.py
```diff
@@ -60,9 +62,15 @@ class ProjectVariableManager(CRUDMixin[ProjectVariable]):
...
     )
     _update_attrs = RequiredOptional(
-        required=("key", "value"),
-        optional=("protected", "variable_type", "masked", "environment_scope"),
+        required=("key",),
+        optional=("value", "protected", "variable_type", "masked", "environment_scope"),
```
JohnVillalovos: Maybe add the other variables? https://docs.gitlab.com/api/project_level_variables/#update-a-variable

Image

JohnVillalovos (changes requested):

gitlab/v4/objects/variables.py
PhilipNelson5: oh yeah, that's inconsistent with what I experience interacting with the API and what is exercised in the unit tests I added.

How would you like this to proceed? I'm also happy to add the other missing variables.

JohnVillalovos: `masked_and_hidden` is available for group and project. Just not Admin according to their docs. Their docs could be wrong.

PhilipNelson5: my bad, you are correct, it is not an attribute on admin variables. removed in 4ea4cc04

gitlab/v4/objects/variables.py
PhilipNelson5: added missing optional attributes in d47e03f2

gitlab/v4/objects/variables.py
PhilipNelson5: added missing optional attributes in d47e03f2

gitlab/v4/objects/variables.py
PhilipNelson5: added missing optional attributes in d47e03f2

gitlab/v4/objects/variables.py
PhilipNelson5: added missing optional attributes in d47e03f2

gitlab/v4/objects/variables.py
PhilipNelson5: added missing optional attributes in d47e03f2

gitlab/v4/objects/variables.py
PhilipNelson5: added missing optional attributes in d47e03f2

gitlab/v4/objects/variables.py
```diff
@@ -43,10 +52,28 @@ class GroupVariableManager(CRUDMixin[GroupVariable]):
...
     _update_attrs = RequiredOptional(
-        required=("key", "value"), optional=("protected", "variable_type", "masked")
+        required=("key",),
+        optional=(
+            "description",
+            "environment_scope",
+            "masked",
+            "protected",
+            "raw",
+            "value",
```
JohnVillalovos, line 74: I believe `value` is required. The docs say so.

https://docs.gitlab.com/api/group_level_variables/#create-a-group-variable

JohnVillalovos (changes requested):

gitlab/v4/objects/variables.py
PhilipNelson5: This is where I think the docs are wrong. With the current implementation, you can't update the description of a `masked_and_hidden` variable because you can't know the `value`. The API accepts updating the description without providing the `value`.

PhilipNelson5: This is demonstrated in the functional test I added. When creating a `masked_and_hidden` variable, you don't even get the value back from the API during creation.

```py
def test_hidden_group_variables(group):
    variable = group.variables.create(
        {"key": "key1", "value": "secret_value", "masked_and_hidden": True}
    )
    assert variable.value is None
    assert variable.description is None
    assert variable in group.variables.list()

    variable.description = "new_description"
    variable.save()
    variable = group.variables.get(variable.key)
    assert variable.description == "new_description"
```

tests/functional/api/test_variables.py
```diff
@@ -43,3 +43,51 @@ def test_project_variables(project):
...
+        {"key": "key1", "value": "secret_value", "masked_and_hidden": True}
+    )
+    assert variable.value == "secret_value"
+    assert variable.description is None
+    assert variable in gl.variables.list()
+
+    variable.description = "new_description"
+    variable.save()
+    variable = gl.variables.get(variable.key)
+    assert variable.description == "new_description"
```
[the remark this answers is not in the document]
PhilipNelson5: Addressed in 5c0652de by removing the test. Instance variables do not support `masked_and_hidden`.

gitlab/v4/objects/variables.py
JohnVillalovos: Thanks!

JohnVillalovos (approved):
LGTM

PR #3400, opened 2026-05-19
fix(tests): update pformat/pprint assertions for Py3.15

traylenator:
## Changes

Python 3.15 changed the default behaviour of pprint.PrettyPrinter:
- default indent changed from 1 to 4
- default width changed from 80 to 88
- default compact=False layout now places opening braces/brackets on their own line with contents indented, similar to pretty-printed JSON

See: https://docs.python.org/3.15/library/pprint.html (gh-112632)


traylenator:
No longer required....

PR #3406, opened 2026-06-10
docs(ai_policy): add an AI attribution policy to CONTRIBUTING.rst

JohnVillalovos:
Require AI generated contributions to indicate so in commit messages.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -1,5 +1,6 @@
+############
 Contributing
-============
+############
 
 You can contribute to the project in multiple ways:
 
@@ -75,6 +76,9 @@ This creates a clearer project history, and automates our `Releases`_ and change
 * Bad:   ``Update documentation for projects``
 * Good:  ``docs(projects): update example for saving project attributes``
 
+Also see the `AI attribution section <ai_attribution_section_>`_ for
+information on attributing contributions from AI.
+
 Coding Style
 ------------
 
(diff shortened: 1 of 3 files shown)


bufferoverflow (approved):
@JohnVillalovos Nice, now you just need to add it to this commit as well ;-)

JohnVillalovos:
> @JohnVillalovos Nice, now you just need to add it to this commit as well ;-)

If you look at all my failures along the way it kind of gives away that I didn't use AI this time 😄

JohnVillalovos:
But we do use this policy over at LibreBooking https://github.com/LibreBooking/librebooking and I also do use AI quite often to help. If I had been smarter I would have used AI but I was lazy and thought it can't be that hard. My knowledge of RST was not as good as I thought it was. I should have just started up Claude...

max-wittig:
I'm not a big fan of this. All code is now written by LLMs and humans are just nudging it in the right direction. I would have agreed with adding this in 2025, but our work fundamentally changed in 2026 so I'm not sure this makes sense.

The model should not matter, but the quality of the code does.

CONTRIBUTING.rst
```diff
@@ -90,6 +94,30 @@ To format your code according to our guidelines before committing, run:
...
   tox -e black,isort
 
+.. _ai_attribution_section:
+
+AI Attribution
+--------------
+
+When an AI coding assistant contributed meaningfully to a commit (writing or
+significantly modifying code, suggesting the fix, generating tests, etc.),
+add an ``Assisted-by`` trailer to the commit footer::
```
nejch, line 104: I think this will pretty quickly get outdated, as agents are now autonomous enough that they're essentially doing the whole implementation. Likewise with models and harnesses, they change every month now and can be extended with MCPs, skills, plugins and other tools that I think muddy the water further.

WDYT about asking for the initial prompt that created the contribution, instead? As discussed offline as well, I think what's becoming more important now is the author's initial intent for a change, as that way we can better reason about whether something is well-intentioned or just slop to fill a GitHub punch card.

Though in that case, it might make more sense to include it in the original PR description as it could pollute the commit log much more than a simple trailer. We could even include it in the PR template, and have agents pick it up automatically.

nejch (commented):
Thanks @JohnVillalovos, I agree it'd be good to have something now that they've started, but would like something lightweight/generic to future-proof it. WDYT?

CONTRIBUTING.rst
JohnVillalovos: Thanks @nejch 

@nejch I'm not sure. As I know in my work I don't often do an initial prompt and the result of that is what I submit. For me there is usually a lot of iterating. But it could be useful to get the plan of what they were trying to do.

I still like would to have people declare if they used an AI or not. And adding the footer seems like an initial minimal ask of people.

One of my favorite quotes is, "Don't let perfect be the enemy of good". So my thought is get a start and then keep iterating on it as time goes by and we learn more and more.

PR #3412, opened 2026-07-05
fix(client): retry fetching server version after an unknown result

maretodoric:
## Changes

Fixes `Gitlab.version()` caching a failed lookup permanently. Previously, once a call to `/version` failed and `_server_version`/`_server_revision` were set to `"unknown"`, every subsequent call to `version()` would short-circuit on the `self._server_version is None` check and keep returning `("unknown","unknown")` forever - even if the GitLab server became reachable again.

`version()` now also retries the `/version` request when the cached value is `"unknown"`, not just when it's `None`.

### Documentation and testing

### DIFF
--- a/gitlab/client.py
+++ b/gitlab/client.py
@@ -404,7 +404,7 @@ class Gitlab:
             The server version and server revision.
                 ('unknown', 'unknown') if the server doesn't perform as expected.
         """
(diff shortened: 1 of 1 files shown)


JohnVillalovos (approved):
Thanks @maretodoric 

LGTM

Issue #3395, opened 2026-05-08
No new Container Images on registry.gitlab.com/python-gitlab/python-gitlab since v5.6.0


uluzox:
## Description of the problem, including code/CLI snippet
As per the [documentation](https://github.com/python-gitlab/python-gitlab#using-the-docker-images) docker images are pushed to the GitLab container registry. However, the [gitlab container registry of this project](https://gitlab.com/python-gitlab/python-gitlab/container_registry/592152)'s latest container image for registry.gitlab.com/python-gitlab/python-gitlab is version v5.6.0 while the py package's current version is [v8.3.0](https://github.com/python-gitlab/python-gitlab/releases/tag/v8.3.0).

## Expected Behavior
All releases produce a container image that is pushed to https://gitlab.com/python-gitlab/python-gitlab/container_registry/592152

## Actual Behavior
Since version v5.6.0 no further release was pushed as a container image to https://gitlab.com/python-gitlab/python-gitlab/container_registry/592152

TimKnight01:
Looks like the mirror to https://gitlab.com/python-gitlab/python-gitlab has been disabled, meaning that no new tags after 5.6.0 are in GitLab.

The .gitlab-ci.yml file hasn't changed, so as soon as the tags get mirrored to Gitlab then the images should get build

uluzox:
Can the team elaborate on why the mirror to gitlab was disabled?

TimKnight01:
@uluzox I cannot, I'm not a maintainer, all I can say from investigating is that is what has happened

JohnVillalovos:
One issue is that on GitLab we no longer have the Premium level. I believe that level has to be re-applied for each year for open source projects.

The Premium level is where "Pull" mirroring is available. That is how the sync between them is currently being done.

TimKnight01:
@JohnVillalovos yes I believe it does have to be re-applied for each year for OSS frustratingly

JohnVillalovos:
One issue is: https://gitlab.com/gitlab-org/gitlab/-/work_items/543742 as at the moment the LGPL v3 license is being detected as "Other" and thus not allowing to apply as an open-source project 😢 Can see that here: https://gitlab.com/python-gitlab/python-gitlab/-/blob/main/COPYING

JohnVillalovos:
I have proposed a PR to make our LGPGL v3 license be recognized by GitLab: #3420

JohnVillalovos:
[...]

PR #3420, opened 2026-07-14
fix(COPYING): Use the combined LGPGL and GPL license text.

JohnVillalovos:
Taken from https://www.gnu.org/licenses/lgpl-3.0.html in particular from https://www.gnu.org/licenses/lgpl+gpl-3.0.txt

There is a bug in the https://github.com/gl-gitaly/go-license-detector where it only detects LGPL if the license contains both LGPL and GPL text in it.

This bug causes the project to not be recognized as LGPL at GitLab.

Discussed in: https://gitlab.com/gitlab-org/gitlab/-/work_items/543742

Related: #3395

### DIFF
--- a/COPYING
+++ b/COPYING
@@ -1,7 +1,7 @@
                    GNU LESSER GENERAL PUBLIC LICENSE
                        Version 3, 29 June 2007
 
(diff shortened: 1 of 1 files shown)


JohnVillalovos:
I'm going to go ahead and merge this.

PR #3429, opened 2026-07-27
fix(docker): run container as non-root user

chernojagne:
## Changes

Add a non-root user (`app`, UID 1001) to the Dockerfile and set `USER 1001`
before the container entrypoint. Running as root in a container is a security
risk — it widens the blast radius if the process is compromised, and many
Kubernetes deployments enforce a non-zero `runAsUser` policy that would
reject a root container at admission.

The `RUN` instruction uses both Alpine-style (`addgroup -S` / `adduser -S`)
and glibc-style (`groupadd --system` / `useradd --system`) forms so the
image continues to build from either `python:3.x-alpine` or
`python:3.x-slim` base images.

### Documentation and testing

---

> This change was generated by [Autar](https://autar.ai) and reviewed by a
> human before submission.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,4 +16,8 @@ RUN pip install --no-cache-dir $(find dist -name *.whl) && \
     rm -rf dist/
 
 ENTRYPOINT ["gitlab"]
+
+# Run as a non-root user for container security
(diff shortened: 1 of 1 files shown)


chernojagne:
Closing this — we're revising how we approach these contributions and would rather withdraw it than leave it open unattended. Apologies for the noise, and thanks.

